Issue #7, opened 2017-12-09
Improve build process and reorganize code in a sane manner.


ospencer:
Our codebase still has remnants of the grand ol' SNEK project that needs to be cleaned up.

ospencer:
This was done in #20

PR #8, opened 2017-12-09
Migrate to jbuilder and Clean Up Lingering Snek References

### LINKED ISSUE #4: Remove Remnants of x86 Compile Target
There's still some code lying around from when Grain's ancestors targeted x86. We should clean this out.

belph:
This PR fixes #4 and (mostly) #7. As `jbuilder` is now the norm, the project is now configured to use that instead of our hand-rolled build system.


belph:
Note that this also removes our dependency on `wast2wasm`.

src/codegen.ml
```diff
@@ -176,12 +176,16 @@ let get_arity_func_type_idx arity env =
     Deque.size !(env.func_types) - 1
   | Some((i, _)) -> i
 
+(* Seems a little silly when named this way, but it makes a little more sense in the
+   context of this file *)
+let encode_string : string -> int list = Utf8.decode
```
ospencer: I don't think this is named sillyly. That's what it's doing. I don't think you need the comment. Or rather, have the comment explain why it's like this.

belph: You don't think
```
encode = decode
```
is silly?

src/runner.ml
```diff
@@ -155,24 +155,18 @@ let safe_run_process prog args =
...
+    | Wasm.Script.Quoted(_, s) -> get_encoded ((Wasm.Parse.string_to_module s).Wasm.Source.it) in
+  let ic = open_in asm_file in
+  let contents : Wasm.Script.definition = Wasm.Parse.string_to_module (input_all ic) in
+  close_in ic;
+  let oc = open_out_bin object_name in
+  output_string oc @@ get_encoded (contents.Wasm.Source.it);
+  close_out oc;
+  Right(""), (fun() -> ())
 
 let compile_assembly_to_binary asm debug outfile_name =
```
ospencer: Can you rename this?

src/codegen.ml
ospencer: I find it stupid, not silly. But it makes sense in this context. So I'd update the comment to explain why we're doing that, since this is something that looks just plain wrong/a bug.

PR #12, opened 2017-12-10
Clean up JS runtime

ospencer:
This PR moves the runtime into its own proper folder, and also cleans up the JS to be much more modern. The plan is to have the runtime consumable in two different ways. The first being a single script tag on the page, that has the runtime in it. The second being through NPM, if intergrating grain modules into a JS application.


ospencer:
Speaking of which, I registered the `@grain` organization on NPM.

runtime/src/core/tags.js
```diff
@@ -0,0 +1,52 @@
...
+import {
+  GRAIN_ERR_NOT_NUMBER_GENERIC,
+  GRAIN_ERR_NOT_BOOLEAN_GENERIC,
+  GRAIN_ERR_NOT_TUPLE_GENERIC,
+  GRAIN_ERR_NOT_LAMBDA_GENERIC,
+  GRAIN_ERR_NOT_STRING_GENERIC,
+  GRAIN_ERR_NOT_DOM_ELEMENT_GENERIC
+} from '../errors/error-codes';
+
+export const GRAIN_NUMBER_TAG_MASK = 0b0001;
```
belph: Maybe we should change these to a more OO-style?
Something like
```js
class GrainTag {
  constructor(code) {
    this.code = code;
  }
  // ...
}
```
(not just here, but other places as well...it's kind of obvious we ported a lot of this from C)

runtime/src/errors/error-codes.js
```diff
@@ -0,0 +1,24 @@
+export const GRAIN_ERR_NOT_NUMBER_COMP = 0;
```
belph: (this would be another place to maybe have some OO logic...especially w.r.t. formatting these as error messages)

script/public/javascripts/examples.js
```diff
@@ -0,0 +1 @@
+GrainRunner('examples/conditionals.wasm').then(res => console.log(res));
```
belph: Why `console.log`?

runtime/src/core/tags.js
ospencer: I agree that this can be better, but I feel like that's out of scope for what I was trying to accomplish with this. We should make an issue to tackle that.

script/public/javascripts/examples.js
ospencer: This is just an example file. `GrainRunner` returns a promise that resolves to whatever the Grain program returns. So if you want to see it, you need to print it or do something else with it, if you didn't already in your program.

belph:
@ospencer: Should `make test` run the JS tests as well?
Maybe something like this:
```make
RT_SRC=$(shell find runtime/src -name '*.js')
RT_DEPS=runtime/node_modules runtime/package.json $(RT_SRC)
runtime/node_modules:
    cd runtime && npm install
js-runtime: $(RT_DEPS)
    cd runtime && npm install
js-tests: $(RT_DEPS)
    make js-runtime
    cd runtime && npm test

tests: js-tests
```

script/public/javascripts/examples.js
belph: Oh, this is an example _for how to use the JS API_. Got it. I misunderstood.

runtime/src/core/tags.js
belph: Yes, that's what I meant; I'm not asking for you to change it for this PR.

PR #19, opened 2018-05-13
Update data types grammar

ospencer:
This PR updates the data types grammar to use the new syntax we discussed. It also updates the printing appropriately. 

I made the executive decision to force type variants to be capitalized. I found it far too confusing when reading a `match` and being unable to tell if a name was a variable or a type constructor without reading the rest of the file.


runtime/src/core/grain-module.js
```diff
@@ -80,6 +80,7 @@ export class GrainModule {
 }
 
 export async function readFile(path) {
+  const fs = require('fs');
```
belph: Oops.

src/grain-stdlib/lists.grlib
```diff
@@ -1,64 +1,64 @@
...
+data List<a> =
+  | Empty
+  | Cons(a, List<a>);
 
 let rec length = (lst) => {
   match (lst) {
-    | empty => 0
-    | cons(hd, tl : a List) => 1 + length(tl)
+    | Empty => 0
+    | Cons(hd, tl) => 1 + length(tl)
```
belph: Why drop the explicit type annotation here? Just for aesthetics? Just curious.

src/grain-stdlib/lists.grlib
```diff
@@ -1,64 +1,64 @@
...
   }
 };
 
 let rec sum = (lst) => {
-  match lst {
-    | empty => 0
-    | cons(hd, tl) => hd + sum(tl)
+  match (lst) {
```
belph: +1 for forced parens in `match`

src/grain-stdlib/lists.grlib
ospencer: Honestly, I thought I added it in when I was testing. It was breaking during parsing (after I updated the grammar) and I just removed it. I can add it back if you care. This goes to the larger point, though, of philosophy. I was leaning towards using type annotations as a debugging tool rather than for code, but obviously adding them in if it makes the code more readable. Here, I feel like it's clear that it's a list. It'll be even more clear once we have macros/syntax support for lists.

runtime/src/core/grain-module.js
ospencer: Not really your fault. This is specifically necessary since we can't require `fs` in the browser.

ospencer:
> This is the safer choice, since we can always enable them later, but would good syntax highlighting not address this?

Yes and no. There's no way to tell whether it's a binding or a constructor when looking at a single line, such as when highlighting with TextMate grammars. You could possibly do something with a language server, but you're going to have a hard time using any highlighter that isn't "aware", like a simple highlighter in a browser.

Issue #15, opened 2018-01-15
Support tail calls intra-module


ospencer:
This is an example transform that we believe is possible:

```grain
let rec is_even = lambda x:
  if x <= 1:
    x == 0
  else:
    is_odd(x - 1)
and is_odd = lambda x:
  if x <= 1:
    x == 1
  else:
    is_even(x - 1)
;;


let ie_x = ref false in
lee io_x = ref false in
let break_r = ref true in
let next_r = ref false in
let rec tc_is_even = lambda:
  if !ie_x <= 1:
    !ie_x == 0
  else:
    begin
      io_x := !ie_x - 1;
      break_r := false;
      next_r := tc_is_odd
    end
  in
and tc_is_odd = lambda:
  if !io_x <= 1:
    !io_x == 0
  else:
    begin
      ie_x := !io_x - 1;
      break_r := false;
      next_r := tc_is_even
    end
  in
let is_even = lambda x:
  begin
    ie_x := x;
    break_r := true;
    next_r := tc_is_even;
    let ret_v = ref false in
    while not !break_r do
      ret_v := !next_r()
    end;
    !ret_v;
  end
  in
let is_odd = lambda x:
  begin
    io_x := x;
    break_r := true;
    next_r := tc_is_odd;
    let ret_v = ref false in
    while not !break_r do
      ret_v := !next_r()
    end;
    !ret_v;
  end
;;
```

PR #25, opened 2018-07-08
Add while loops, boxes, and box assignments.

ospencer:
This removes 10 TODOs in the tests! 🚀🙌 
🌾🌾🌾🌾🌾🌾🌾🌾🌾🌾🌾🌾🌾🌾🌾

This moves us towards completing #15.

### DIFF
--- a/src/parsing/parser.dyp
+++ b/src/parsing/parser.dyp
@@ -129,7 +129,7 @@ let make_program statements body =
 
 }
 
-%relation pe<pt<pp<pb<pc
+%relation pe<pt<pp<pb<pc<pa
 
 %token <int> NUM
 %token <string> ID
@@ -144,6 +144,7 @@ let make_program statements body =
 %token COLON COLONCOLON
 
 %token ADD1 SUB1 ISBOOL ISNUM ISTUPLE
+%token BOX UNBOX
 %token PLUS MINUS TIMES
 %token TRUE FALSE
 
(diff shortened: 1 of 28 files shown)


src/parsing/parser.dyp
```diff
@@ -167,11 +168,16 @@ prim1 :
...
   | SUB1 { Sub1 }
   | NOT { Not }
+  | BOX { Box }
+  | UNBOX { Unbox }
   | ISBOOL { IsBool }
   | ISNUM { IsNum }
   | ISTUPLE { IsTuple }
 
-binop_expr:
+expr :
```
ospencer: I made this alias because it's weird to say that lots of things take a `binop_expr`, when what you really want to say is that they take any `expr`.

src/typed/typecore.ml
```diff
@@ -509,6 +529,17 @@ and type_expect_ ?in_function ?(recarg=Rejected) env sexp ty_expected_explained
...
       exp_env = env
     }
+  | PExpWhile(scond, sbody) ->
+    let cond = type_expect env scond (mk_expected ~explanation:While_loop_conditional Builtin_types.type_bool) in
+    let body = type_expect env sbody ty_expected_explained in
+    re {
+      exp_desc = TExpWhile(cond, body);
+      exp_loc = loc;
+      exp_extra = [];
+      (* While loops don't evaluate to anything *)
```
belph: I'm not a huge fan of this, given our `if`/`else` semantics, but I don't have a good alternative. I'd argue that `while` loops should evaluate to something, but it's hard to say what they should be in the case of `while (false) { ... }`...

ospencer: I thought the same thing and wanted them to evaluate to something too. I think I disagree about it not being similar to `if`/`else`, since it's the same deal as having a one-sided `if`. A couple of options, though:

1) We could also/otherwise have `do {} while` loops, which would evaluate to something:
```grain
let i = box(5);
do {
  i := unbox(i) - 1;
} while (unbox(i) > 0)
# Evaluates to box(0)
```

2) Some languages have this kinda thing:
```grain
while (someBool) {
  # Do some stuff
} else {
  # Some other stuff
}
```

I think I'd prefer option 1, since I also don't really like one-sided `if`s (which I don't know if we even have them implemented right now, though we do have a parsing rule for them) because the semantics change based on whether or not you've got that `else` clause (which I think people could possibly get behind, especially if we had a helpful message when they did it, but I feel like I rather avoid it if we could, but I'll think on it more). So instead of something like one-sided `if`, I might be more in favor of calling it `when` or something similar like other languages do (but also maybe something different, because I feel like it's confusing when you also have async built into your language).

PR #37, opened 2018-08-02
Docker support

nickbreaton:
Hello all. I was struggling a couple days ago to get grain running, especially since I'm unfamiliar with the ocaml ecosystem. I really wanted to just jump in and see what it could do, however this proved challenging.

To help others get up and running quickly, I've made a docker image that contains the compiler and CLI runtime. To be clear, this is intended to help people *use* the compiler and runtime, not for development of the compiler or runtime.

If you'd like to go ahead and try it out, I have it up on dockerhub building off of my fork. Just be sure you have docker on your system and run,

```sh
docker run --rm -it nickbreaton/grain
```

Once inside the shell, you will have `grainc` and `grain` available to you to start playing around. I have also installed `vim` to make it easy to create and modify `.gr` files within the running container. Since the image is based on the node image, `node` and `npm` are also available.

I understand if you don't want to include this in the main repository, and if not, I will just continue to use it for personal use, however if you do, this PR should contain all that is needed. I did have to make a small number of modifications to the grain CLI, but I will explain those in comments.

I would also be more than happy to setup an organization on dockerhub for grain-lang, and create a continuous deployment system. That way the above build command will look like,

```
docker run --rm -it grain-lang/grain
```

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -1,4 +1,4 @@
-#!node --harmony
+#!/usr/bin/env node
 
 let program = require('commander');
 let compile = require('./compile.js');
@@ -16,7 +16,7 @@ program
 program.on('--help', () => {
   console.log('\n\n  File can be either a Grain (.gr) or WebAssembly (.wasm) file.\n');
 });
-  
+
 program.parse(process.argv);
 
 let wasmFile;
--- a/cli/bin/locator-path.js
+++ b/cli/bin/locator-path.js
@@ -1,0 +1,6 @@
+const path = require('path');
+
+const relativePath = '_build/install/default/lib/grain/stdlib';
+const absolutePath = path.resolve(__dirname, '../../', relativePath);
+
+module.exports = absolutePath;
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,5 @@
 # dockerignore
+.git
 *.dockerfile
 docker-compose.yml
 Dockerfile
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,34 @@
+# dockerignore
+*.dockerfile
+docker-compose.yml
+Dockerfile
+
+# gitignore
+main
+*.install
+output/*.o
+output/*.run
+output/*.s
+output/*.dSYM
+script/node_modules
+.installed-pkgs
+*.sw*
+*~
+*.o
+*.log
+_build/*
+.installed-pkgs
+notes.org
+**/node_modules/**/*
+*.wasm
+*.wasm.wast
+*.wasm.modsig
+.tern-port
+setup.data
+setup.ml
+myocamlbuild.ml
+.merlin
+runtime/dist
+script/public/javascripts/grain-runtime.js
+script/public/javascripts/grain-runtime-browser.js
+/*.gr
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,4 +46,4 @@ RUN npm install && \
 
 # SETUP FOR END USER
 WORKDIR /root
-ENTRYPOINT ["/bin/bash"]
+CMD ["bash"]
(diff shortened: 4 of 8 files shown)


cli/bin/locator-path.js
```diff
@@ -0,0 +1,6 @@
+const path = require('path');
+
+const relativePath = '_build/install/default/lib/grain/stdlib';
+const absolutePath = path.resolve(__dirname, '../../', relativePath);
+
+module.exports = absolutePath;
```
nickbreaton: I was having some issues with the `grain` command not being able to find `_build/install/default/lib/grain/stdlib` based off of where it was executed from. I created this file to ensure we are always pointing towards an absolute path for `stdlib`.

 I then required this file from both `run.js` and `compile.js`.

cli/bin/grain.js
```diff
@@ -1,4 +1,4 @@
-#!node --harmony
+#!/usr/bin/env node
```
nickbreaton: So two issues were occurring here on Ubuntu. 

First off, `#!node` didn't see to work on its own, I added the `#!/usr/bin/env node` as I believe this is the most common way to handle shebangs for node.

Secondly, the `--harmony` was not playing nicely with the above command, and I did some digging, and there does not seem to be a standard for if/how shebangs support flags. However, I didn't really see the use case for it at all. It seems to run fine on node versions `8`, `9`, and `10`.

Please let me know if this is a problem.

cli/bin/locator-path.js
belph: Yeah, the `grain` CLI tool on `master` has some bitrot. A somewhat more "legitimate" fix was implemented last night in e1578e6 (on the `gc` branch), but wiring up `grain` to `GRAIN_ROOT` is still a bit of a work in progress...

cli/bin/grain.js
belph: cc: @ospencer 
I have to test whether this works for me later this evening, but I see no issues (in fact, I would have probably complained had I noticed `node --harmony` originally....but I digress.)

belph:
I have to give this a more thorough review later, but I appreciate you fixing the CLI. I'm thinking it makes the most sense to combine this with the parallel fixes in e1578e6. Perhaps we can merge this into `gc` and do a merge of `gc` into `master`; but again, I need to look this over first.

cli/bin/grain.js
ospencer: Oh, yep. That's definitely one of those testing things that got merged. 🙂 

This should definitely be fine, but we should make sure that we have a note in the README about having `node` >= 8.

nickbreaton: @ospencer I can add that note on #38 if you'd like.

ospencer: Though actually now that I think about it, we don't need the shebang at all, since this script is always invoked through the node tooling. I remember adding that in specifically because I was trying to do something with some harmony feature (which never ended up happening).

nickbreaton: @ospencer I think we still do need it. I just removed it, and it was not happy:

```sh
root@ad15e8007768:~# grain
/usr/local/bin/grain: line 1: syntax error near unexpected token `('
/usr/local/bin/grain: line 1: `let program = require('commander');'
```

ospencer: Alright! I'm down with it.

belph:
This looks good to me, but I would like to incorporate the CLI fixes from #42 onto this branch before merging. Once that's merged into `master`, one of us can rebase this PR onto the new `HEAD` and fix the conflicts.

nickbreaton:
Hey guys. I'm sorry I never got around to wrapping this up. I still don't have time so I'm going to close this. If anyone is going to pick it up and has a question about what I had done, please feel free to reach out.

PR #49, opened 2018-12-01
Implement tail call optimization

### LINKED ISSUE #15: Support tail calls intra-module

ospencer:
Fixes #15

### DIFF
--- a/src/middle_end/analyze_tail_calls.ml
+++ b/src/middle_end/analyze_tail_calls.ml
@@ -32,6 +32,7 @@ let push_tail_call analysis =
 let push_tail_recursive analysis =
   analysis := TailRecursive(true)::!analysis
 
+(* Indicate whether or not this expression contails a tail call that needs optimization *)
 let rec analyze_comp_expression ({comp_desc = desc; comp_analyses = analyses}) =
   match desc with
   | CIf(_, t, f) ->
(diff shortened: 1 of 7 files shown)


ospencer:
I agree that directly-recursive functions can be further optimized, but the call count goes from `n` to `n + 1` (you get one additional call to begin the iteration, but each tail call is replaced by exactly one iteree call).

PR #51, opened 2018-12-01
Move the Ident hash to the Ident module

ospencer:


src/typed/ident.ml
```diff
@@ -110,6 +110,13 @@ let print ppf i =
   | -1 -> fprintf ppf "%s#" i.name
   | n -> fprintf ppf "%s/%i%s" i.name n (if global i then "g" else "")
 
+type _t = t
```
ospencer: If you could tell me a way I could do this without having to do this, I'd be delighted.

belph: Nope, I think this is idiomatic.

PR #53, opened 2018-12-06
Clean up test suite, enable Travis builds

ospencer:
I didn't delete `wasm_runner`, but if you think we should I will. A small bug was uncovered when going through the tests, so there's also a small change to address that.


.travis.yml
```diff
@@ -0,0 +1,11 @@
...
+node_js:
+  - "10"
+sudo: required
+install: echo "Begin install"
+script: bash -ex .travis-opam.sh
+env:
+  - OCAML_VERSION=4.05 GRAIN_ROOT=~/build/grain-lang/grain/_build/install/default
+os:
+  - linux
+
```
belph: :grimacing:

src/runner.ml
```diff
@@ -104,8 +105,27 @@ let extract_wasm {cstate_desc} =
   | Compiled({asm}) -> asm
   | _ -> raise (Invalid_argument "Expected WASM State")
 
-let run_output cstate =
-  Wasm_runner.run_wasm (extract_wasm cstate)
+let read_stream cstream =
+  let buf = Bytes.create 2048 in
+  let i = ref 0 in
+  Stream.iter (fun c ->
+    (if !i >= 2048 then failwith "Program output exceeds 2048 characters");
```
belph: Would be good to have a comment describing that this is an arbitrary limit

cli/bin/grain.js
```diff
@@ -1,31 +1,35 @@
...
 
 program
   .version('Grain compiler 0.0.0\nGrain cli 0.0.0', '-v, --version')
+  .description('Compile and run Grain programs. 🌾')
   .arguments('<file>')
-  .action((givenFile) => {
-    file = givenFile || '';
+  .option('-w, --wasm', 'run a wasm file')
+  .option('-p, --print-output', 'print the output of the program')
+  .option('-g, --graceful', 'return a 0 exit code if the program errors')
```
belph: Oof. I understand why you put this, but it's kind of gross

cli/bin/grain.js
```diff
@@ -1,31 +1,35 @@
...
 program.parse(process.argv);
 
-let wasmFile;
-if (file.endsWith('gr')) {
-  wasmFile = compile(file);
-} else if (file.endsWith('wasm')) {
-  wasmFile = file;
-} else {
-  program.help();
+if (typeof givenFile === 'undefined') {
```
belph: Comment? (I see what you're doing, but some code clarity would be nice)

src/runner.ml
```diff
@@ -6,6 +6,7 @@ open Printf
 open OUnit2
```
belph: Really, I'd love if this wasn't in `src` at all. Is that possible?

PR #58, opened 2019-05-20
Switch from jbuilder to dune

ospencer:


belph:
LGTM, but I see you dropped the `jbuild_version` lines. Is there a Dune equivalent that we can put instead?

ospencer:
It doesn’t seem like there is: https://dune.readthedocs.io/en/latest/dune-files.html#jbuild-version

belph:
There is, but you've already taken care of it.
```
(lang dune 1.5)
```
is the equivalent line, it seems.

PR #60, opened 2019-09-08
New import system

ospencer:
# New import system
Excited to finally get this in place. I'll drop the commit that fixes the asm.js warnings once #59 is merged.

## Importing all values from a module
Any value may be excluded.
### New syntax! 🎉
```grain
import * from 'path-to-module'
import * except {values, to, Exclude} from 'path-to-module'
```

## Importing some values from a module
All imported values are aliasable.
### New syntax! 🎉 
```grain
import {value1, value2, Constructor1} from 'path-to-module'
import {x as y, X as Y} from 'path-to-module'
```

### New syntax! 🎉 
```grain
import ModuleName from 'path-to-module'
```

### New syntax! 🎉 
Values in modules may be used with dot accessors:
```grain
ModuleName.foo
ModuleName.bar(5)
ModuleName.Cons(x, y)
print(ModuleName.baz)
```

### New syntax! 🎉 
```grain
import List, {length} from 'lists'
```

## Other important notes/changes
- Grain libraries are now regular `.gr` files, and the `.grlib` filetype has been removed.
- Printing data structures does not respect aliases, but I'm not sure that we'd actually it want to (and it'd also be fairly difficult to implement).


cli/bin/run.js
```diff
@@ -1,13 +1,16 @@
...
 
 let runtime = require('../../runtime/dist/grain-runtime.js');
 let grainRoot = execSync('grain-root').toString('utf8').trim();
 
-module.exports = async function run(path, options) {
+module.exports = async function run(filename, options) {
   try {
-    let locator = runtime.defaultFileLocator(options.includeDirs.concat(`${grainRoot}/lib/grain/stdlib`));
+    let base_path = path.dirname(filename)
+    let includeDirs = [base_path].concat(options.includeDirs, `${grainRoot}/lib/grain/stdlib`)
```
belph: Is this functionally the same as the original line? Methinks not, unless either `concat` is smarter than I think or you changed the type of `includeDirs`

src/typed/env.mli
```diff
@@ -96,16 +97,15 @@ val add_local_type: Path.t -> type_declaration -> t -> t
 val add_item: signature_item -> t -> t
 val add_signature: signature -> t -> t
 
-(* Remember the name of the current compilation unit. *)
-val set_unit_name: string -> unit
-val get_unit_name: unit -> string
+(* Remember the current compilation unit. *)
+val set_unit: string * string -> unit
```
belph: @ospencer Can you explain this one?

cli/bin/run.js
ospencer: It's the same. `concat` takes any number of arrays or values.

Would it read easier if it were
```js
let includeDirs = [base_path, ...options.includeDirs, `${grainRoot}/lib/grain/stdlib`]
```
?

src/typed/env.mli
ospencer: The name alone isn't enough to determine if we're in the same module, since modules could have the same name. This makes it filepath * name instead.

cli/bin/run.js
belph: @ospencer For me, I think that would be more readable, yeah.

src/typed/env.mli
belph: Can you maybe add a comment to this explaining what the two tuple elements are?

ospencer:
@belph updated.

PR #64, opened 2019-10-30
Implement records

ospencer:
# Records
This PR adds record functionality to Grain.

### New syntax 🎉 
```grain
data Foo = {
  foo: Number,
  bar: List<String>
}
```

### New syntax 🎉
```grain
{
  foo: 5,
  bar: Cons("hi", Empty)
}
```

We also support puns.
### New syntax 🎉
```grain
let foo = 5;
let bar = Empty;
{ foo, bar }
```

### New syntax 🎉
```grain
let x = { foo: 5, bar: Empty };
x.foo
```
```grain
funcThatReturnsAFoo().bar
```
```grain
anyNonBinopExpr.bar
```

### New syntax 🎉
```grain
let x = { foo: 5, bar: 6 };
let { foo, bar } = x;
let { bar, _ } = x;
let { bar: baz, _ } = x;
```
```grain
let x = { foo: 5, bar: 6 };
match (x) {
  | { foo, bar} => foo + bar
};
match (x) {
  | { foo, _ } => foo
};
```

## Breaking changes
The program
```grain
{ x }
```
no longer parses as a block with the single expression `x`, but rather as a record with punned field `x`.


ospencer:
@belph PR updated to include record destructuring.

PR #72, opened 2019-12-05
Support deeply-nested pattern matching

ospencer:
There was a bug in our pattern-matching implementation that caused nested patterns to be treated as catch-alls. For example,
```grain
match (...) {
  | Cons(a, Cons(b, Empty)) => ...
  ...
}
```
was treated as though it were
```grain
match (...) {
  | Cons(a, b) => ...
  ...
}
```

This PR resolves this, and adds extensive tests around pattern matching.

### DIFF
--- a/src/middle_end/matchcomp.ml
+++ b/src/middle_end/matchcomp.ml
@@ -18,25 +18,29 @@ type decision_tree =
       to the action number (i.e. switch branch) to be taken *)
   | Fail
   (** Represents a failed match. *)
-  | Switch of (int * decision_tree) list
+  | Switch of (int * decision_tree) list * decision_tree option
   (** Multi-way test for an occurence. The list is a
       non-empty list of pairs of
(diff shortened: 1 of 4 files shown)


src/middle_end/matchcomp.ml
```diff
@@ -257,20 +280,9 @@ let matrix_head_constructors mtx =
...
-    | hd :: _ when pattern_always_matches hd.mb_pat -> [hd]
-    | hd :: tl -> hd :: (help tl) in
-  help match_branches
-
 let make_matrix branches =
-  List.map (fun ({mb_pat} as mb) ->
-      match mb_pat.pat_desc with
-      | TPatTuple(patts) -> patts, mb
-      | _ -> ([mb_pat], mb)) branches
+  (* Form matrix and assign each branch a number *)
```
ospencer: This was done to make debugging much easier. The switch case numbers in the compiled output now map 1:1 onto the branches in the original program.

PR #82, opened 2020-03-17
Improve array stdlib

ospencer:
The `Array` stdlib now implements these fine functions:
```
length
make
init
get
set
append
concat
copy
forEach
forEachi
toList
fromList
```

TODO: Documentation (mainly) along with some proper error handling, and tests. Some code is a little jank, just because we don't have something like a `failwith` function for impossible cases, so there's just some awkward expressions to allow things to typecheck.

Next on my todo list is to implement `assert` and `failwith`, since that's better than having bad code.


ospencer:
@belph FYI.

PR #83, opened 2020-03-17
Remove stale example files & build script

phated:
As per discussion on the stream, this removes the stale examples and script for generating website examples.


ospencer:
Thanks for getting this cleaned up!

PR #86, opened 2020-03-18
Remove grain-root & get stdlib/cli/runtime working together

phated:
This PR does a few things. I'll outline below with reasons for "why".

## Moves grain-stdlib to root and rename it to stdlib

Having the grain standard library in `src/` seems weird when it is implemented in grain itself. By moving to the root of the project, contributors can find it easier and you can even publish it as an npm package (so I also added a package.json file, but more on that later).

## Makes the CLI depend on `@grain/runtime` and `@grain/stdlib`

I've added these two projects to the dependencies list of `@grain/cli` because it used to be doing a require on a relative path to load the runtime. This won't work when you publish it.

Currently, this is using the `file:` protocol to keep things simple for the tests. The goal is to remove that and have something else do the linking in development.

## Removes grain-root and defaults to no stdlib

The removal of `grain-root` really needed to happen because I was unable to move `grainc` around on my filesystem successfully, even when moving them together. This made it almost impossible to include the `grainc` binary in the `@grain/cli` package. Now that I've removed it, I am able to put the `grainc` binary in the `cli/bin/` directory and everything works great!

After doing a ton of digging, it seemed that `grain-root` was only used to auto-detect where the grain stdlib was. However, if you control the CLI through `@grain/cli` and you publish the stdlib as a separate npm package, you can just pass the location of the the stdlib to `grainc`. With that in mind, combined two `grainc` options into one—instead of `--no-stdlib` and `--grain-root`, you now have `--stdlib` to specify the standard library. I don't think it's hotswappable quite yet, but that could happen in the future.

Anyway, hope that all makes sense. Cheers! 🎉


ospencer:
This largely looks really good! I think I'm good with `GRAIN_STDLIB` or `--stdlib` being required, though it'd be really nice if there was some reasonable default. If there's some kind of proper install, that could plop the stdlib somewhere and set up a PATH variable, that could work. I suppose that there's a lot of options since we don't yet distribute binaries, and maybe this is just fine for now.

ospencer:
Looks like the tests aren't finding the stdlib though.

PR #87, opened 2020-03-18
Add assertions and failing

ospencer:
This PR implements:
```
assert someCondition
```
and
```
fail "with some explanation"
```

Huzzah 🎉


ospencer:
@belph FYI.

PR #88, opened 2020-03-19
Add proper stdlib tests

ospencer:
There's now a `stdlib` folder in `test` with tests implemented (in Grain!) for the current standard libraries.


ospencer:
@belph FYI.

PR #91, opened 2020-03-21
Implement new Box utilities

ospencer:
This PR adds a shorthand syntax for unboxing: `^box`. It also adds four new utilities for `Box<Number>`:

```grain
box += 6
box -= 6
box *= 6
box /= 6
```

These should make working with the most common uses for boxes much nicer.


ospencer:
@belph FYI.

Issue #89, opened 2020-03-19
Complete the test suite for the list stdlib


ospencer:
There are some list stdlib functions that aren't tested. You can find the tests in `test/stdlib/lists.test.gr`.

PR #93, opened 2020-03-22
Add tests for list append and for_all

TyrelDenison:


ospencer:
Closes #89

ospencer:
Thanks for the first contrib! Looking forward to many more in the future 😄

PR #96, opened 2020-03-25
Ban nonfunctions from letrec

ospencer:
There's currently an issue with creating recursive data types using `let rec`, and there isn't _currently_ an easy fix. For the time being, we'll disallow non-functions on the right side of `let rec`.


ospencer:
@belph FYI.

PR #98, opened 2020-03-28
Implement generic hash function

ospencer:
This PR adds an AssemblyScript implementation of the MurmurHash3 hashing algorithm over the current Grain data types. It's nice because it's probably exactly as fast as if we'd hand-written the wasm since most things are a 1:1 mapping.

This adds a bit of fragmentation, since the constant tags/masks are now duplicated within the stdlib module. I think this is okay, as we've chatted about rewriting most of the runtime in AssemblyScript anyway.

This does, however, add a new build step to the stdlib (npm i && npm run build). 

The `foreign` syntax is now a bit different:

```
import foreign wasm foo : Number -> Number from 'some/path'
export foreign wasm foo : Number -> Number from 'some/path'
```

This was done to allow foreign imports from a path rather than from just a name.


src/codegen/compcore.ml
```diff
@@ -1015,11 +1017,11 @@ let compile_prim2 (env : codegen_env) p2 arg1 arg2 : Wasm.Ast.instr' Concatlist.
     ] @
     (* Convert remainder result into modulo result *)
     compiled_arg1 +@ [
-      Ast.Const(encoded_const_int32 31);
+      Ast.Const(const_int32 31);
       Ast.Binary(Values.I32 Ast.IntOp.ShrU);
     ] @
     compiled_arg2 +@ [
-      Ast.Const(encoded_const_int32 31);
+      Ast.Const(const_int32 31);
```
ospencer: 🤫

stdlib/hashUtils.ts
```diff
@@ -0,0 +1,176 @@
...
+  GRAIN_ARRAY_HEAP_TAG,
+} from './ascutils/tags.ts'
+
+import {
+  GRAIN_TRUE,
+  GRAIN_FALSE,
+  GRAIN_VOID,
+} from './ascutils/primitives.ts'
+
+const seed: u32 = 0xe444
```
ospencer: 🌾

src/codegen/compcore.ml
belph: :eyes:

runtime/src/core/grain-module.js
```diff
@@ -59,6 +59,10 @@ export class GrainModule {
     return this.instantiated.exports;
   }
 
+  get isGrainModule() {
+    return !!this.exports["GRAIN$MAIN"]
```
belph: Is this really the JS-idiomatic way to do to-bool?

ospencer: Yes, actually. https://github.com/airbnb/javascript#coercion--booleans

JS is the wild west.

torch2424:
@ospencer Hello again! Awesome to see y'all are using AssemblyScript! 

Was wondering if you and @belph would like an invite to the AssemblyScript slack? It's just where a lot of the core team / community chats and asks general questions in things.

I just need y'alls email if you are interested! Thanks! :smile:

ospencer:
@torch2424 Hey Aaron! Definitely would love to hang and chat. I'll DM you on Twitter. :)

PR #102, opened 2020-04-03
Update README and contributor docs; fix gotchas

ospencer:
When I was updating the contributor docs and was trying to outline gotchas, I felt that was kinda dumb instead of just fixing them, so I did.

Two new things happen now:
1. Post `yarn compiler:install` we now clean the stdlib, so the stdlib will get recompiled will with the new installed compiler.
2. The tests now run with `dune runtest --force`, which always runs the test binary, not just if the dependencies changed.


ospencer:
@belph FYI.

PR #105, opened 2020-04-07
Fix ADT printing

ospencer:
We were double-tagging constructor tags and specially tagging non-constant constructors, and our match/allocation implementations don't depend on that. We also didn't use separate numbers for constant/tuple constructors which caused printing to be dependent on the order in which the type variants were defined, i.e.
```grain
export data Option<a> = None | Some(a)

print(Some(9))
print(None)
```
correctly prints
```
Some(9)
None
```

but
```
export data Option<a> = Some(a) | None

print(Some(9))
print(None)
```
casually prints
```
None
Some(0)
```

Nothing is wrong with the data in memory (so pattern matching and whatnot still works properly) but the printing is busted. Also, if you define more than one tuple constructor or more than one constant constructor, that doesn't work at all 😬

But now it's fixed!

### DIFF
--- a/compiler/src/middle_end/linearize.ml
+++ b/compiler/src/middle_end/linearize.ml
@@ -10,8 +10,8 @@ module MatchCompiler = Matchcomp.MatchTreeCompiler
 let compile_constructor_tag =
   let open Grain_typed.Types in
   function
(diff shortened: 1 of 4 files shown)


ospencer:
@belph FYI.

PR #107, opened 2020-04-10
Heap allocate should call malloc

ospencer:
`toString`, `stringSlice`, and `stringAppend` haven't been working since we had real malloc. This actually calls malloc for those methods.


ospencer:
(tests are coming in a later PR)

PR #108, opened 2020-04-11
Use GitHub actions

ospencer:
Seems like GitHub actions has much better cache handling than Travis. On a cache hit, the build and tests run in under three minutes. This'll also make it easier for us to have nightly builds that push to GitHub.


compiler/test/stdlib/lists.test.gr
```diff
@@ -24,8 +24,6 @@ assert append([], [4]) == [4]
 
 # List.contains
 
-assert contains(true, list) == false
-assert contains(false, list) == false
```
ospencer, line 28: Since I merged the `==` change, these should be type errors now. That breaks on the GitHub actions CI, but it didn't on Travis... odd.

PR #122, opened 2020-05-23
Remove janestreet/core from project

phated:
Whille working on `js_of_ocaml` stuff, I found that janestreet's core module was being included but not actually used. So I removed it and things seem to compile 🤷 

I think this and #121 are going to conflict but that's just the esy.lock directory which can be deleted and recreated.


phated:
Instead of just removing `core`, I think Batteries (and maybe extlib?) should be replaced with `core_kernel` which janestreet guarantees to be cross-platform compatible and works with JSOO. It also seems that they have an Int64 implementation there that can replace `Stdint`, which uses C-bindings and can't be JSOO'd

phated:
Not sure what's up with CI, I have it running locally.

ospencer:
I re-ran the CI a couple of times and it does the same thing. I would guess would be that it's some change in `compiler/src/utils/files.ml`.

When I try to build this branch locally I get this error:
```
info building @opam/grain_utils@path:./grain_utils.opam@d41d8cd9
error: build failed with exit code: 1
  build log:
    # esy-build-package: building: @opam/grain_utils@path:./grain_utils.opam
    # esy-build-package: pwd: /Users/oscar/.esy/3/b/opam__s__grain__utils-b8aae6a5
    # esy-build-package: running: 'dune' 'build' '-p' 'grain_utils' '-j' '4'
    File "src/utils/dune", line 6, characters 3-12:
    6 |    batteries)
           ^^^^^^^^^
    Error: Library "batteries" not found.
```

Which definitely doesn't make sense since that line doesn't read that any more, but deleting the esy cache at `compiler/_esy` doesn't make this go away. Is there some other cache I should be trying to remove?

ospencer:
I just deleted my whole `~/.esy` directory and that seems to do the trick.

ospencer:
It looks like I'm able to reproduce the CI error if I run the tests twice. Does that happen for you?

phated:
Yeah, I had to blow away `~/.esy`, `compiler/esy.lock` and `compiler/_esy`.  I think that might be an issue with us using `path:` instead of `link:` for the resolutions, but I'm not exactly sure.

Anyway, once having a clean build, I can run the tests repeatedly.
Screen Shot 2020-05-25 at 1 22 22 PM

ospencer:
Does anything different happen if you run `yarn compiler:install` before running the tests?

phated:
Why do we run that? Maybe we need to remove `@install` from the dune build in `esy.json` because I thought that did the install part

ospencer:
`yarn compiler:install` "installs" the compiler so you can run `grainc` from the command line, versus `@install` which I believe builds things that are installable. You still have to run `dune install` (which is what `yarn compiler:install` does) to gain access to `grainc`.

We can run the tests without installing, since all the compilation happens through Grain as a library rather than the executable.

ospencer:
It theoretically shouldn't make a difference whether this test fails or not; I was mostly asking as a sanity check.

phated:
I'm going to close this because @ospencer implemented new esy fixes into master, and I'm not sure it is worth the effort currently to try pulling all the native deps from the project because it will end up using `binaryen.ml` anyway.

PR #137, opened 2020-05-27
Make compiler subpackage linkable

ospencer:
This makes the development `grainc` available via `yarn workspace @grain/compiler link`, which I think is a good solution until we decide if we want to bundle the compiler with the CLI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
           yarn setup
       - name: esy install
         run: |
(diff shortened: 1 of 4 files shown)


ospencer:
This also means that `yarn compiler:install` goes away—the built version of the compiler is always the one on your path.

ospencer:
For legacy `opam` folks, you can uninstall your version installed through `opam` with `yarn workspace @grain/compiler run esy dune uninstall`.

Issue #94, opened 2020-03-22
Add new list methods


ospencer:
Let's add methods for these:
```grain
some # checks that at least one item in the list is true for a given condition
nth # gets the nth item in the list
filter
reject # opposite of filter
count # counts how many elements satisfy a given condition
hd # gets the first item in the list
tl # returns the list without the first item
flatten # takes a list of lists and combines them into a single list
insert # takes a list, item, index. inserts the item at the index-th position
rotate # takes a list, count. item at count-th position becomes the first item. should work for negative counts as well
uniq
part # takes a list and and a number. returns a tuple containing the first n items and the rest of the list
```

ospencer:
Removed `every` since that rename has already been done.

designsbysm:
It looks like several of these functions should be included with Array.

ospencer:
@JacobGinsparg is currently taking this on, and I think it's almost complete?

ospencer:
I removed `sort` and `sortBy` from this in favor of #125. 

I also added a new method, `part`, since it seems like it'd be useful for `rotate`.

PR #141, opened 2020-05-31
List functions

JacobGinsparg:
Adding functions to stdlib.List for #94


stdlib/lists.gr
```diff
@@ -87,3 +91,81 @@ let forEachi = (f, l) => {
...
+  match (l) {
+    | [] => []
+    | [hd, ...tl] => if (!f(hd)) [hd, ...reject(f, tl)] else reject(f, tl)
+  }
+}
+
+let hd = (l) => {
+  match (l) {
+    | [] => fail 'hd'
+    | [hd, ...tl] => hd
```
ospencer, line 112: It's probably better to just ignore a pattern if we won't use it:
```suggestion
    | [hd, ..._] => hd
```

stdlib/lists.gr
```diff
@@ -87,3 +91,81 @@ let forEachi = (f, l) => {
...
+  }
+}
+
+let tl = (l) => {
+  match (l) {
+    | [] => fail 'tl'
+    | [hd, ...tl] => tl
```
ospencer, line 119:
```suggestion
    | [_, ...tl] => tl
```

stdlib/lists.gr
```diff
@@ -87,3 +91,81 @@ let forEachi = (f, l) => {
...
+  }
+}
+
+let rec nth = (n, l) => {
+  if (n == 0) hd(l) else nth(n - 1, tl(l))
+}
```
ospencer, line 125: This should probably `fail 'nth'` if `n` is greater than the length of the list rather than `fail 'hd'`. I don't think that `nth` needs to support negative indexes, so we should `fail` in that case too. Once exceptions are merged we can change it to `raise InvalidArgument`.

stdlib/lists.gr
```diff
@@ -87,3 +91,81 @@ let forEachi = (f, l) => {
...
+
+let rec flatten = (l) => {
+  match (l) {
+    | [] => []
+    | [hd, ...tl] => append(hd, flatten(tl))
+  }
+}
+
+let rec insert = (x, i, l) => {
```
ospencer, line 134: This should fail if i < 0.

stdlib/lists.gr
```diff
@@ -87,3 +91,81 @@ let forEachi = (f, l) => {
...
+
+let part = (n, l) => {
+  let rec help = (l1, l2, n) => {
+    match (l2) {
+      | [] => if (n > 0) fail 'part' else (reverse(l1), l2)
+      | [hd, ...tl] => if (n > 0) help([hd, ...l1], tl, n - 1) else (reverse(l1), l2)
+    }
+  }
+  help([], l, n)
+}
```
ospencer, line 156: I'm not sure that I understand this. Why does reverse?

stdlib/lists.gr
```diff
@@ -87,3 +91,81 @@ let forEachi = (f, l) => {
...
+
+let rotate = (n, l) => {
+  let (beginning, end) = if (n >= 0) part(n, l) else part(length(l) + n, l)
+  append(end, beginning)
```
ospencer, line 160: 😍

PR #150, opened 2020-06-24
Updated Grain Guide Link in README

torch2424:
https://grain-lang.org/guide 404'd for me :smile: 

And I think it should be: https://grain-lang.org/docs/ ? :thinking: 

Awesome project by the way! Excited to start playing with this :smile:


ospencer:
You're absolutely right; thanks for catching this!

PR #156, opened 2020-06-27
OCamlFormat & ReasonML proof-of-concept

phated:
With Dune 2, they included [automatic formatting](https://dune.readthedocs.io/en/stable/formatting.html) by default, for both OCaml **and** ReasonML syntax. This auto-formatting assumes that `ocamlformat` and/or `reason` are installed depending on what extensions it sees in the project.

For this PR, I wanted to set up `ocamlformat` to have a consistent formatter, and to test out the dune command. I also tested running `esy x refmt --parse ml grainc/grainc.ml > grainc/grainc.re` in the project and committed the results of that.

If y'all feel that Reason syntax could increase contributions (I prefer the C-style syntax), I can finish using refmt to translate the files. We could also include some esy commands to convert files, if some people prefer to write primarily in OCaml.

What do ya think?


phated:
I swear that grain's CI has it out for me...

ospencer:
👏🏽👏🏽👏🏽👏🏽👏🏽

I've been wanting ocamlformat to be set up for _ages_; I wish I had known it were that easy!

I'm pro-switching over to Reason—I do think it would help with getting more folks to contribute. I'm a fan of the OCaml syntax, but I understand that _most_ people these days are riding the Reason train. Reason also feels pretty appropriate still it feels a lot closer to Grain, and that'll probably make it easier further down the line when we start thinking about Whole Grain.

@belph any thoughts?

belph:
I think that Reason is a much better fit for Grain, just by nature of it being targeted towards a similar group of developers. I'm happy for us to make the switch 🙂

phated:
Superseded by #160

PR #158, opened 2020-06-27
Add Option to stdlib

phated:
This adds the `Option` data type to pervasives, and additionally adds an `option.gr` module to the standard library that shamelessly reimplements APIs I've liked from OCaml, BuckleScript's Belt library, and Rust.

I also added `peek` because I think it is useful once there's currying and you can chain things with `|>` or similar.

One glaring omission from this implementation is converting to/from Result because that doesn't exist yet.

### DIFF
--- a/stdlib/option.gr
+++ b/stdlib/option.gr
@@ -1,0 +1,122 @@
+# Standard library for list functionality
+
+export *
+
+let isSome = (opt) => {
+  match (opt) {
+    | Some(_) => true
+    | None => false
+  }
+}
+
+let isNone = (opt) => {
+  match (opt) {
+    | None => true
+    | Some(_) => false
+  }
+}
+
+let contains = (val, opt) => {
+  match (opt) {
+    | Some(x) => x == val
+    | None => false
+  }
+}
+
+let expect = (msg, opt) => {
+  match (opt) {
+    | Some(x) => x
+    | None => fail msg
+  }
+}
+
+let unwrap = (opt) => {
+  expect('Could not unwrap None value', opt)
+}
+
+let unwrapWithDefault = (default, opt) => {
+  match (opt) {
+    | Some(x) => x
+    | None => default
+  }
+}
+
+let map = (fn, opt) => {
+  match (opt) {
+    | Some(x) => Some(fn(x))
+    | None => None
+  }
+}
+
+let mapWithDefault = (default, fn, opt) => {
+  match (opt) {
+    | Some(x) => fn(x)
+    | None => default
+  }
+}
+
+let flatMap = (fn, opt) => {
+  match (opt) {
+    | Some(x) => fn(x)
+    | None => None
+  }
+}
+
+let filter = (pred, opt) => {
+  match (opt) {
+    | Some(x) =>
+      if (pred(x)) {
+        Some(x)
+      } else {
+        None
+      }
+    | None => None
+  }
+}
+
+let zip = (optA, optB) => {
+  match ((optA, optB)) {
+    | (Some(a), Some(b)) => Some((a, b))
+    | _ => None
+  }
+}
+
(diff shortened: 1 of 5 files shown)


stdlib/option.gr
```diff
@@ -0,0 +1,129 @@
...
+    | None => None
+  }
+}
+
+let zip = (optA, optB) => {
+  match ((optA, optB)) {
+    | (Some(a), Some(b)) => Some((a, b))
+    | (Some(_), None) => None
+    | (None, Some(_)) => None
+    | (None, None) => None
```
belph, line 83:
```suggestion
    | (Some(a), Some(b)) => Some((a, b))
    | _ => None
```

stdlib/option.gr
```diff
@@ -0,0 +1,129 @@
...
+  }
+}
+
+let zipWith = (fn, optA, optB) => {
+  match ((optA, optB)) {
+    | (Some(a), Some(b)) => Some(fn(a, b))
+    | (Some(_), None) => None
+    | (None, Some(_)) => None
+    | (None, None) => None
```
belph, line 92:
```suggestion
    | (Some(a), Some(b)) => Some(fn(a, b))
    | _ => None
```

stdlib/option.gr
```diff
@@ -0,0 +1,129 @@
...
+  }
+}
+
+# TODO: Would be cool to have a recursive version
+let flatten = (opt) => {
+  match (opt) {
+    | Some(Some(x)) => Some(x)
+    | Some(None) => None
+    | None => None
```
belph, line 101:
```suggestion
    | Some(Some(x)) => Some(x)
    | _ => None
```

stdlib/option.gr
```diff
@@ -0,0 +1,129 @@
...
+}
+
+let mapWithDefault = (default, fn, opt) => {
+  match (opt) {
+    | Some(x) => fn(x)
+    | None => default
+  }
+}
+
+# TODO: Do you like `andThen` or should it be `flatMap` or both?
```
ospencer, line 58: I would lean towards `flatMap`, since the "then" part of `andThen` feels like it could imply that something async is happening.

stdlib/option.gr
```diff
@@ -0,0 +1,129 @@
...
+let zipWith = (fn, optA, optB) => {
+  match ((optA, optB)) {
+    | (Some(a), Some(b)) => Some(fn(a, b))
+    | (Some(_), None) => None
+    | (None, Some(_)) => None
+    | (None, None) => None
+  }
+}
+
+# TODO: Would be cool to have a recursive version
```
ospencer, line 96: I don't think the typechecker would let us write one without a custom type to match against :/

The type would have to look like this:
```grain
data RecOpt<a> = RecOptCons(Option<a>, RecOpt<a>) | RecOptNone
```

You could maybe get this to work with higher kinded types, but I'm not sure.

stdlib/option.gr
belph: +1. Plus, `flatMap` is nice since Options are monads/functors

stdlib/option.gr
belph: Yeah, I'm not sure how calls to a recursive version would type-check in the first place...

phated:
@belph when I make your suggested changes, I get this warning:
> Warning 9: this pattern-matching is fragile.
It will remain exhaustive when constructors are added to type pervasives.Option.

How should I avoid that?

phated:
I made the changes and rebased. We should figure out how to disable that warning for the stdlib maybe?

ospencer:
Yeah, I'll make an issue.

ospencer:
I wrote a thing to set the compiler's default warnings, and opened an issue so we can have them be configurable. (I made FragileMatch default to off, and left everything else as on for now.)

PR #160, opened 2020-06-28
Convert to ReasonML

phated:
Based on the conversation in #156, I decided to supersede that PR with just converting everything to ReasonML and not pulling in ocamlformat at all.

This is just the results of running refmt on the project and doesn't actually change anything to more idiomatic Reason patterns (if that's even a thing).

Something else to consider: the JS code could also be written in Reason and compiled with Bucklescript (so we can have nice npm library interop). What do you think?


compiler/esy.json
```diff
@@ -11,7 +11,8 @@
     "compile": "dune build @all @install",
     "clean": "rm -rf #{self.root}/_esy",
     "install-compiler": "dune install",
-    "test": "dune runtest --force"
+    "test": "dune runtest --force",
+    "format": "esy dune build @fmt --auto-promote"
```
ospencer, line 15: Do you think `format` should just be moved into the regular build command? I want a way to ensure that formatting always happens. Maybe it'd be a good idea too to have the CI fail if things aren't formatted.

ospencer:
> Something else to consider: the JS code could also be written in Reason and compiled with Bucklescript (so we can have nice npm library interop).

If there's a codemod then I'd be into this, but we're slowly trying to remove all of the JS code. Probably wouldn't be worth it to have someone take the time to rewrite it. But if it's free then 👌🏽

phated:
> If there's a codemod

You can't go from unsound to sound :wink: so there's not a tool for that.

compiler/esy.json
phated: Yeah, there's a bunch of build and CI stuff to solve. I'm going to send another PR that tries to fix some of the stuff that has been popping up in the discord.

phated:
Let me know if you need me to rebase this now that other changes have landed.

ospencer:
Yeah, that'd be great.

PR #161, opened 2020-06-28
Copy the grainc binary to the CLI directory

phated:
This adds another step to the "compiler:build" command that copies the generated binary of `grainc.exe` into the `cli/bin` directory and then the execSync calls use the absolute path to that binary.

This should avoid people having to install the grainc binary. There might be a possible issue with that absolute path on Windows, but we'll need to get something in CI to check it.

This is also phase 1 in getting a publishable CLI that people can just use.

### DIFF
--- a/compiler/esy.json
+++ b/compiler/esy.json
@@ -4,13 +4,15 @@
   "esy": {
     "buildEnv": {
       "DUNE_BUILD_DIR": "#{self.target_dir}",
-      "GRAIN_STDLIB": "#{self.root}/../stdlib"
+      "GRAIN_STDLIB": "#{self.root / '..' / 'stdlib'}",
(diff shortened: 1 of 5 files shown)


compiler/esy.json
```diff
@@ -4,13 +4,15 @@
   "esy": {
     "buildEnv": {
       "DUNE_BUILD_DIR": "#{self.target_dir}",
-      "GRAIN_STDLIB": "#{self.root}/../stdlib"
+      "GRAIN_STDLIB": "#{self.root / '..' / 'stdlib'}",
+      "GRAINC_BIN_PATH": "#{self.target_dir / 'default' / 'grainc' / 'grainc.exe'}",
+      "CLI_BIN_DIR": "#{self.root / '..' / 'cli' / 'bin'}"
```
phated, line 9: I switched this stuff to esy's path expansion to hopefully help with Windows support of the commands.

PR #162, opened 2020-06-28
Use del-cli to clean with globs

phated:
This uses a node module to have cross-platform cleaning of the wasm files in the stdlib. It uses globs and supports negative globbing for `stdlib-external`

### DIFF
--- a/stdlib/package.json
+++ b/stdlib/package.json
@@ -32,11 +32,12 @@
   },
   "homepage": "https://grain-lang.org",
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


ospencer:
I wish I knew Weylin's GitHub handle to ping them on this. I'll have to remember in the morning.

phated:
@ospencer we still need to solve Windows support for the build script, but I'm hitting a wall for tonight, so I can look tomorrow.

I also want to get Windows set up in the GH Actions CI so we can make sure we don't break things.

PR #164, opened 2020-06-28
Clean up heap top

ospencer:
Wanted to get this out of the way before starting work on moving to Binaryen :)

### DIFF
--- a/compiler/src/codegen/compcore.ml
+++ b/compiler/src/codegen/compcore.ml
@@ -12,8 +12,6 @@ let source_it (x : 'a Source.phrase) : 'a = Source.(x.it)
 
 (** Environment *)
 type codegen_env = {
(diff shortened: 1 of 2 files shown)


compiler/src/codegen/compcore.ml
```diff
@@ -1999,7 +1989,6 @@ let compile_exports env {functions; imports; exports; num_globals} =
           };
           add_dummy_loc {
             Ast.name=encode_string (Ident.name table_size);
-            (* We add one here because of heap top *)
```
ospencer, line 2002: Oddly enough, changing this misaligns the global table, even with the removal of heap_adjust.

ospencer: I also tried just changing this to `compute_table_size` but that's not correct either.

ospencer:
@belph told me this was 👍

PR #165, opened 2020-06-28
Remove unused variables

ospencer:
These seem to be left over from GC implementation.


phated:
Yeah, I saw this too. Wasn't sure if they were placeholders for future things.

ospencer:
FYI @belph

PR #166, opened 2020-06-28
Adds a build script to run assemblyscript

phated:
This adds a cross-platform build script that generates all wasm files based on assemblyscript files in `stdlib-external`.

I also fixed a globbing issue with my clean script.

### DIFF
--- a/stdlib/build.mjs
+++ b/stdlib/build.mjs
@@ -13,7 +13,7 @@ const asc = require('assemblyscript/cli/asc');
 
 const compileFile = promisify(asc.main);
 
-// Binaryen is ready
+// AssemblyScript compiler is ready
 await asc.ready;
 
 const filenames = await globby('stdlib-external/**/*.ts');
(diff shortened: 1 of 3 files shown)


stdlib/build.mjs
```diff
@@ -0,0 +1,29 @@
...
+import globby from 'globby';
+import replaceExt from 'replace-ext';
+
+// Making asc interop with ESM
+const require = createRequire(import.meta.url);
+const asc = require('assemblyscript/cli/asc');
+
+const compileFile = promisify(asc.main);
+
+// Binaryen is ready
```
ospencer, line 16: I feel like this will confuse people since this is AssemblyScript's Binaryen and people might think this has something to do with ours.

```suggestion
// AssemblyScript compiler is ready
```

phated: Yeah, makes sense. I literally stole that line from their implementation 😛

PR #172, opened 2020-06-30
Utilize dune to generate opam files

phated:
The other day @ospencer pointed out that Dune can do the generation of OPAM files, so I went down that rabbit hole today.

There's still some duplication (but a lot less) and I actually caught some issues with the current OPAM files (incorrect `doc:` value, wrong license metadata).

I think deferring to Dune for this generation can help us streamline other areas of the build tooling and catch potential problems.

There is one caveat to this right now: when you change something in dune-project that alters the opam files, the esy cache will be invalidated during the run and `yarn compiler:build` will fail; running it a 2nd time will then succeed because the opam files don't change again. I believe this will be solvable by fixing some issues with linking our opam files with esy and I hope to work on that next.


ospencer:
Yeah that makes sense. Not a huge deal though, since the opam files are checked in. Ideally though the build command would work on the first try 🙂 

I imagine Dune doesn't have a separate command to check/generate the opam files before we run the build command?

phated:
> I imagine Dune doesn't have a separate command to check/generate the opam files before we run the build command?

@ospencer it doesn't seem so. They are still building out this functionality and only added `dune describe opam-files` 5 days ago 😛

PR #174, opened 2020-06-30
Clean stdlib for test

phated:
It turns out that if you copy the `grainc.exe` binary to a different location, the CMIs generated from each binary are different. This causes issues if you run `yarn test` then try to use the grain CLI to compile other code.

I'm not exactly sure the reasons for that, but I've heard OCaml has portability issues.

Anyway, this PR ensures that stdlib is cleaned before and after testing.


ospencer:
You know what? This might not actually be the result of moving `grainc`. Since the tests compile the stdlib, we shouldn't need to clean. I realized that I introduced a bug back when I moved the `sys` stdlibs into a subdirectory.

This is the bit of code that needs to be updated: https://github.com/grain-lang/grain/blob/master/compiler/test/test.re#L26

Up to you if you want to take a stab at that or if you want me to do it.

phated:
@ospencer Yeah, there was a problem with that where the stdlib wasn't getting updated before the tests so I thought we needed to clean before we ran them. However, if we don't clean **after** we test, the CMIs for the stdlib are incorrect for use with the `grain` command, so I left the post-clean in.

ospencer:
I opened an issue to track the underlying bug. This is a good measure in the meanwhile 👍

PR #176, opened 2020-07-01
Proxy the root scripts to packages

phated:
This changes the scripts from all being defined as `package:command` to being proxied through `yarn workspace` into the package that defines the build script. It also moves the `esy` commands into the package.json of the compiler so we can utilize them from the CI.

A really neat trick is that you can use `yarn run` to navigate into packages and select a command because they all spawn `yarn run` commands. Here's an example of me navigating from the root into the compiler package where I can select a new command:
Screen Shot 2020-06-30 at 5 27 58 PM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,9 +33,9 @@ jobs:
           yarn
           yarn setup
 
-      - name: Esy install
+      - name: Esy setup
         run: |
-          yarn compiler install
+          yarn compiler setup
 
       - name: Esy cache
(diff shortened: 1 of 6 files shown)


phated:
An additional benefit is that we can now navigate to the `compiler/` directory and use `yarn build` and `yarn test` to run scripts on just the compiler. I was having a frustrating time trying to navigate while running tests & `grain` against a test file.

phated:
I'm going to squash this one into master because it has a merged commit

PR #179, opened 2020-07-01
Small opam/dune cleanup

phated:
I've been diving into OPAM & Dune to understand them better and I found some issues with the way some our stuff was configured.

The primary things being that ppx_deriving isn't a true preprocessor and it doesn't need to be included in the libraries stanza because deriving plugins that need the API or runtime will pull it in themselves (this is what ppx_deriving_yojson does). I aslo noticed that we were bringing in all of ppx_sexp_conv when the runtime-lib should suffice (though this required explicitly referencing the compiler-libs because that's where Misc exists).


phated:
These changes drop the size of the grainc binary from 23mb to 16mb on my computer.

PR #180, opened 2020-07-01
Add OCaml LSP w/ setup script

phated:
We have all sorts of issues with the reason-vscode extension and development on the OCaml Platform extension seems to be moving quickly, so I added some quick-and-dirty scripts to get contributors set up with using it.


phated:
Hopefully the detection gets solved so we don't need this script. Ref https://github.com/ocamllabs/vscode-ocaml-platform/issues/245

PR #184, opened 2020-07-02
Implement a Map module for stdlib

phated:
This is a base implementation of Map (aka HashMap) for the stdlib. It was heavily influenced by [BuckleScript's Belt.HashMap](https://github.com/BuckleScript/bucklescript/blob/3b317c4784e5e4b4bbe47893241aed8a873a9a76/jscomp/others/belt_HashMap.ml).

There's still a ton of work to do but I wanted an initial review because I've never implemented something like this before. 🙏

### DIFF
--- a/compiler/test/stdlib/map.test.gr
+++ b/compiler/test/stdlib/map.test.gr
@@ -1,0 +1,59 @@
+import Map from 'map'
+
+# With Number keys
+let nums = Map.make();
+
+assert Map.set(1, "🌾", nums) == void;
+assert Map.set(2, "🐑", nums) == void;
+assert Map.set(3, "🧱", nums) == void;
+
+assert Map.get(1, nums) == Some("🌾");
+assert Map.get(2, nums) == Some("🐑");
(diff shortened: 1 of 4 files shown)


phated:
@ospencer Hmm, that's the way Belt does it but I thought collection-last was better for currying. Maybe not for this data structure?

The last test was supposed to be for overwriting data, but I didn't create a new Map, should I adjust it?

ospencer:
Whoops, I completely missed that last test.

As for the argument order, I guess I'm just used to the OCaml implementation. Haskell and Elm both put the map last. I guess the other order made sense to me because I've never been in a situation where I've worked with more than one hash table, but I have done the simple `let add = Hashtbl.add tbl` to just do `add key val` in places.

I see arguments for both ways, but it's probably more consistent to put the map last.

phated:
> I have done the simple `let add = Hashtbl.add tbl` to just do add key val in places.

I think that's an argument for placeholder partial application 😉

PR #188, opened 2020-07-02
stdlib: Add Array.contains

phated:
I need this for writing tests for my additions to Map.

### DIFF
--- a/compiler/test/stdlib/arrays.test.gr
+++ b/compiler/test/stdlib/arrays.test.gr
@@ -116,3 +116,10 @@ let square = (n, i) => n * numbers[i]
 
 assert Array.mapi(square, [>]) == [>]
 assert Array.mapi(square, numbers) == [> 4, 9, 16]
+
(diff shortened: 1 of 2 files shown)


compiler/test/stdlib/arrays.test.gr
```diff
@@ -116,3 +116,10 @@ let square = (n, i) => n * numbers[i]
 
 assert Array.mapi(square, [>]) == [>]
 assert Array.mapi(square, numbers) == [> 4, 9, 16]
+
+# Array.contains
+
+let arr = [> 1, 2, 3]
+
+assert Array.contains(3, arr)
+assert Array.contains(4, arr) == false
```
ospencer, line 125:
```suggestion
assert !Array.contains(4, arr)
```

phated: I prefer the explict false check, but seems fine.

PR #191, opened 2020-07-03
stdlib: Add more methods to Map

phated:
This adds a bunch of methods to the Map stdlib, and also improves the original implementation by resizing it once the amount of nodes exceeds double the size of the buckets.

There are still improvements to be made but I will open issues for some right now!

### DIFF
--- a/stdlib/map.gr
+++ b/stdlib/map.gr
@@ -1,5 +1,6 @@
 # Standard library for Map (aka HashMap) functionality
 
+import List from 'lists'
 import Array from 'arrays'
 import { hash } from 'hash'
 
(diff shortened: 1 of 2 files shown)


stdlib/map.gr
```diff
@@ -51,24 +101,29 @@ export let set = (key, value, map) => {
...
+    | Some(node) => {
+      if (replaceInBucket(key, value, node)) {
+        buckets[idx] := Some({ key: box(key), value: box(value), next: box(bucket) });
         map.size += 1;
       };
     }
-  };
-  # TODO: Need to check and possibly resize the buckets array
+  }
+  # Resize if there are more than 2x the amount of nodes as buckets
```
ospencer, line 114: Oh this is interesting. I figured resizing would happen sooner!

stdlib/map.gr
```diff
@@ -81,10 +136,160 @@ export let get = (key, map) => {
...
+  map
+}
+
+export let toArray = (map) => {
+  Array.fromList(toList(map))
+}
+
+export let fromArray = (array) => {
+  let map = make();
+  # TODO: Fix this when Array.forEach is collection-last
```
ospencer, line 283: The tests will fail when we do this!
```suggestion
```

stdlib/map.gr
phated: Hah, it was just taken from BuckleScript but they do `lsl` so I added a comment to explain what I learned 😛

PR #194, opened 2020-07-05
Emit WebAssembly code using Binaryen

ospencer:
# Binaryen as Grain's compiler backend

Previously, Grain used the [official WebAssembly specification](https://github.com/WebAssembly/spec) as our compile target. Since it's written in OCaml, it was a quick way for us to get up and running and start building `.wasm` files.

The wasm specification was never intended to be used as a compiler backend, and thus working with it could be difficult at times, especially when it comes to debugging.

From the Binaryen README:

> Binaryen is a compiler and toolchain infrastructure library for WebAssembly, written in C++. It aims to make compiling to WebAssembly easy, fast, and effective.

With Binaryen, debugging is a breeze. Binayren's validation errors made it _significantly_ easier for me to get everything working than if we were still compiling to lists of wasm instructions. In fact, once validation passed, it turned out that I only made one mistake in the actual code generation, and that was an easy fix.

Binaryen will make our backend far easier to maintain, and implementing long-awaited features (like sourcemaps) becomes fairly trivial. It also gives us access to WebAssembly features before they're finalized, so we can test them out and be ready to go on day one.

This _does_ mean that I rewrote the entire compiler backend, and this was possible due to a new project I've been working on.

# Introducing Binaryen.ml 🎉 🚀 

I wrote [a full suite of OCaml bindings for Binaryen's C API](https://github.com/grain-lang/binaryen.ml), and it's cleverly named _Binaryen.ml_. It's available on opam, and since it's a standalone package, other compilers written in OCaml will be able to use it too!

I'm excited that this has all come together and that it has also helped us contribute something back to the WebAssembly community 😄


compiler/src/codegen/compcore.re
```diff
@@ -2624,173 +3699,152 @@ let compile_exports = (env, {functions, imports, exports, num_globals}) => {
...
-        Ast.gtype:
-          [@implicit_arity] Types.GlobalType(Types.I32Type, Types.Immutable),
-        Ast.value:
-          add_dummy_loc([
-            add_dummy_loc @@
-            Ast.Const(const_int32(compute_table_size(env, prog))),
-          ]),
-      }),
-    ],
+let compile_elems = (wasm_mod, env, prog) =>
```
phated, line 3753: Does this do nothing?

compiler/src/codegen/compcore.re
```diff
@@ -360,384 +362,467 @@ let init_codegen_env = () => {
...
-let wrap_float64 = n => add_dummy_loc(Values.F64Value.to_value(n));
+let wrap_int32 = n => Literal.int32(n);
+let wrap_int64 = n => Literal.int64(n);
+let wrap_float32 = n => Literal.float32(n);
+let wrap_float64 = n => Literal.float64(n);
 
-let grain_number_max = 0x3fffffff;
-let grain_number_min = (-0x3fffffff); /* 0xC0000001 */
+let grain_number_max = 1073741823;
+let grain_number_min = (-1073741823); /* 0xC0000001 */
```
ospencer, line 382: Note to self to change these back (I guess refmt didn't like them)

compiler/src/codegen/compcore.re
ospencer: Huh, yeah, the diff makes this look really weird. But yes, this doesn't do anything (at least for now). Previously this function did what `compile_tables` does now, and the old `compile_tables` didn't do anything. There's an issue though to clean up this file, so I'll probably take care of this then.

compiler/src/codegen/compcore.rei
```diff
@@ -2,6 +2,7 @@ open Grain_typed;
 open Grain_middle_end;
 open Mashtree;
 open Wasm;
```
belph, line 4: We can probably drop this import now? In fact, do we need the original dependency at all?

compiler/src/codegen/compcore.re
```diff
@@ -2,14 +2,12 @@ open Grain_typed;
 open Mashtree;
 open Value_tags;
 open Wasm;
```
belph, line 4: Same as above; do we need this import?

compiler/src/codegen/compcore.re
```diff
@@ -2,14 +2,12 @@ open Grain_typed;
...
+open Binaryen;
 open Concatlist; /* NOTE: This import shadows (@) and introduces (@+) and (+@) */
 
 /* [TODO] Should probably be a config variable */
 let memory_tracing_enabled = false;
 
-let add_dummy_loc = (x: 'a): Source.phrase('a) => Source.(x @@ no_region);
```
belph, line 10: 🙏 🙏 🙏 🙏

compiler/src/codegen/compcore.re
belph: +1

compiler/src/codegen/compcore.re
```diff
@@ -1403,864 +1815,1490 @@ let buf_to_ints = (buf: Buffer.t): list(int64) => {
...
+          compiled_num_elts(),
+          Expression.const(wasm_mod, const_int32(1)),
+        ),
+      ),
+      set_loop_counter(Expression.const(wasm_mod, const_int32(0))),
+      Expression.block(
+        wasm_mod,
+        outer_label,
+        [
+          Expression.loop(
```
belph, line 2471: It'd be nice if we could somehow avoid this massive pyramid, but nbd

compiler/src/codegen/compcore.re
```diff
@@ -2889,52 +3941,42 @@ let compile_wasm_module = (~env=?, ~name=?, prog) => {
...
+  let () = ignore @@ compile_functions(wasm_mod, env, prog);
+  let () = ignore @@ compile_imports(wasm_mod, env, prog);
+  let () = ignore @@ compile_exports(wasm_mod, env, prog);
+  let () = ignore @@ compile_globals(wasm_mod, env, prog);
+  let () = ignore @@ compile_tables(wasm_mod, env, prog);
+  let () = ignore @@ compile_elems(wasm_mod, env, prog);
+  
+  validate_module(~name?, wasm_mod);
+  // TODO: Enable Binaryen optimizations
+  // Module.optimize(wasm_mod);
```
belph, line 3956: Story behind this? Should be uncommented or better documented why it remains commented

compiler/src/codegen/compcore.rei
ospencer: Meant to leave a comment about this.

We can't (just yet) because we still depend on the [Utf8 module](https://github.com/grain-lang/wasm-spec/blob/master/interpreter/binary/utf8.ml) that it provides for our strings (and it turns out there's no other module out there that does this exactly). I figured we'd have a discussion about breaking that functionality out into its own module, or potentially using WebAssembly's data section for strings (though tbd if this is possible right now).

Additionally, we currently write custom sections using the `Wasm` module, but it looks like we sprinkle a little ✨ Grain magic ✨ when we do that, so I wanted to touch that in another PR (rather than trying to change that out in this one).

compiler/src/codegen/compcore.re
ospencer: Doesn't work and needs investigation as to why, but it's probably a problem with the bindings. We have complete parity without the optimizations (we didn't optimize our WASM output before), but we will want to turn this on, so I left it with a TODO. I planned to open an issue for it after this was merged, but I can do it beforehand.

PR #198, opened 2020-07-08
Remove wasm spec dependency

ospencer:


compiler/src/utils/wasm_utils.re
```diff
@@ -259,6 +259,20 @@ let serialize_abi_version = ({major, minor, patch}) => {
...
+  }, ints);
+  Buffer.contents(buf);
+}
+
+let utf8_decode = (str) => {
+  List.init(String.length(str), (i) => {
+    Uchar.to_int(Uchar.of_char(String.get(str, i)));
+  });
+}
+
```
ospencer, line 275: Turns out we can accomplish this all from the standard library now. Huzzah! 🎉

compiler/src/utils/wasm_utils.re
```diff
@@ -520,14 +534,6 @@ let () =
...
     | [@implicit_arity] MalformedSectionType(tag, None) =>
       Some(Printf.sprintf("Malformed WASM section tag: %d", tag))
-    | [@implicit_arity] Wasm.Script.Syntax(src, msg) =>
-      Some(
-        Printf.sprintf(
-          "WebAssembly Parse Error at %s: %s",
-          Wasm.Source.string_of_region(src),
-          msg,
-        ),
-      )
```
ospencer, line 530: We can't raise this error anymore, since we don't parse wasm files using the wasm spec.

compiler/test/test_wasm_utils.re
```diff
@@ -80,18 +70,18 @@ let test_leb128_i32 = ctxt => {
 };
 
 let test_get_wasm_sections = ctxt => {
-  ignore(compile_wasm_file("testmod"));
```
ospencer, line 83: I just went ahead and checked in a compiled version of this file (compiled via wat2wasm) and it sits next to the source.

compiler/src/utils/wasm_utils.re
```diff
@@ -259,6 +259,20 @@ let serialize_abi_version = ({major, minor, patch}) => {
   bytes;
 };
 
+let utf8_encode = (ints) => {
+  let buf = Buffer.create(14);
```
phated, line 263: It's always 14?

compiler/src/utils/wasm_utils.re
phated: Is that for the better or worse?

compiler/test/test_wasm_utils.re
```diff
@@ -80,18 +70,18 @@ let test_leb128_i32 = ctxt => {
...
-      {sec_type: Import, offset: 28, size: 25},
-      {sec_type: Function, offset: 59, size: 2},
-      {sec_type: Export, offset: 67, size: 17},
-      {sec_type: Code, offset: 90, size: 12},
+      {sec_type: Type, offset: 10, size: 8},
+      {sec_type: Import, offset: 20, size: 25},
+      {sec_type: Function, offset: 47, size: 2},
+      {sec_type: Export, offset: 51, size: 17},
+      {sec_type: Code, offset: 70, size: 8},
+      {sec_type: Custom("name"), offset: 85, size: 28},
```
phated, line 84: Just curious why these offsets changed

compiler/src/utils/wasm_utils.re
ospencer: It's a good thing! Pre-Binaryen, one of the ways we validated modules was by creating the bytes and then parsing it back to make sure it was correct. Now we just assume it's correct since it was produced via Binaryen :P

compiler/src/utils/wasm_utils.re
ospencer: It's just an initial guess for how many characters you think there will be in the buffer, and then it resizes accordingly. Sorta like hash tables.

phated: Does the buffer ever shrink though? I thought it only grew

compiler/test/test_wasm_utils.re
ospencer: Just a different compiler. The old one was compiled via (a pretty old version of) the wasm spec, versus this one was compiled via wat2wasm. I figured we should check in the one compiled with wat2wasm since we won't really have easy access to generating the `.wasm` file again via the spec if we need to, but someone's probably got [the wabt suite](https://github.com/WebAssembly/wabt) on their machine (which includes wat2wasm).

compiler/src/utils/wasm_utils.re
ospencer: That's true. It only grows unless you call `Buffer.reset` on it, and then it shrinks back to the initial size you told it to be (and throws away everything in your buffer). That size parameter though is for the amount of space it allocates to store buffer contents, and that's separate from how much stuff is actually in the buffer. I imagine it's got some sort of internal state that it keeps track of that.

Here's a quick test from the REPL:
```ocaml
# let buf = Buffer.create 14;;
val buf : Buffer.t = <abstr>
# Buffer.add_string buf "hello";;
- : unit = ()
# Buffer.contents buf;;
- : string = "hello"
```

Issue #202, opened 2020-07-09
Lang: Revisiting Box


phated:
Inspired by writing our Map stdlib and [this twitter thread](https://twitter.com/BlaineBublitz/status/1281317794064166912) I was thinking that it might be nice to just have a `mut` keyword instead of using the Box construct.

Thoughts?

ospencer:
I've actually thought about this quite a bit. I actually imagined this as `let mutable` but that might be a little verbose. I'm fully on board with this idea. I think it makes a ton of sense for Grain with its "functional yet flexible" mentality.

We could theoretically get rid of boxes since exported mutable globals seem to be on by default now in WebAssembly.

This change though is largely a matter of the technical stuff, and making sure our typechecking will all still be sound. I believe it would be, but we'd have to make sure. We'd also need to be careful with some of our optimizations, but that could be worked out.

I'd be down to prototype this and see how we feel about it.

s0kil:
Grain is getting Rusty with the `let mut`, haha

phated:
This was completed in #225 🎉

PR #204, opened 2020-07-11
`let mut` prototype

ospencer:
# `let mut` prototype ✨ 

This PR lets us see what Grain would look like with the addition of `let mut`. I thought about this some time ago, and sparked by #202, I took the plunge and prototyped it out. I felt that we could play around with this and see how we liked it, and then make a decision on if we should move forward with it. If we decide to move forward with it, I'll clean up this PR and we can merge it.

So far, I love it! It feels very much like a welcome addition to Grain and makes doing anything imperative feel a lot less clunky.

## Notes/thoughts

This does not replace boxes. This is largely because we don't yet have mutable record fields. Once we have that, though, boxes probably won't really have much of a real use case. I do see some potential cases where you might want to pass boxes to functions (say, in third-party libraries) that wouldn't be able to see your custom record type (but that library could just provide that type to you 🙃), so maybe we'll keep them forever.

`let rec mut` is not supported, and its use seems like it be rare enough that we don't really need to support it. If you really needed this functionality, you could just declare your function with `let rec` and create a new binding to it with `let mut`

Importing a mutable value from another module and updating it properly updates the value in both places (spoiler: it's the same reference in memory)

I somehow borked toplevel mutable destructuring, but it works if you do it within a block or a function body, i.e. `let mut (x, y, z) = (5, 'foo', false)`—`x`, `y`, and `z` are all mutable. ~I'll get that fixed and I'll push a change for it.~ This is now fixed.

The infix operators `+=`, `-=`, `*=`, and `/=` no longer work with boxes, and instead now work on mutable values. There's now also `%=`.

If you try to mutate a function parameter, you get the same `Error: The identifier x was not declared mutable` that you'd get elsewhere. This implies that you could somehow declare a parameter as mutable. That's up for debate, but in the meanwhile this error message could be improved.

## Some programs

```gr
let mut count = 0
count = 4
count += 6
count /= 2
```

---

`test2.gr`
```gr
export let mut x = 5

export let xFromTest2 = () => x
```

`test.gr`
```gr
import {x, xFromTest2} from './test2'

print(x)
print(xFromTest2())

x = 2

print(x)
print(xFromTest2())
```

---

```gr
let foo = () => {
  let mut (x, y, z) = ('foo', 3, true)
  print(x)
  print(z = false)
}
```

---

```gr
let mut count = 5
let inc = () => {
  count += 1
}
inc()
inc()
inc()
print(count)
```

## Technical details

I initially tried to implement this by directly mutating locals (this is why you'll see `MSet` etc. in the code). This worked great, but completely falls flat when it comes to closures because of how WebAssembly handles functions. I instead switched mutables to be largely sugar on top of boxes. `let mut x = 5` automatically allocates a new box for `x`. Whenever `x` is accessed, we just unbox `x`. This is possible since `value_description`s now have a `val_mutable` property. The extra benefit to this is that it extends to imported values. Since they can have `val_mutable` set to `true`, imported mutable values can be mutated. We may want to put some guardrails on this though, since it's not obvious what's mutable from an import statement. I think that could be addressed in a later PR though.

We could potentially optimize mutables that are never exported or closed over into mutations of its local.

We could also theoretically implement this via globals, and every `let mut` would have a global value. I think this approach still gets hairy with closures, and managing the imports feels like it would be difficult.

Lastly, I wouldn't bother looking through the code too much. There's a lot that needs to be cleaned up here and some bits left over from previous approaches. If you'd like me to talk through some of the specifics though, I'd be happy to.


phated:
This feature is 🔥!!!

I love how the updates to the Grain code look. It definitely makes using mutable values seem more natural. Oh, and destructuring into `let mut` is SICK!

I'm definitely going to try writing some Grain using this feature to see how it feels.

How much will it take to have `mut` properties on records? I used those pretty heavily in my Map implementation.

ospencer:
@phated I actually don't think it would be terribly hard to add mutable record fields. All of the hard stuff was figured out while I was doing this, and it would be a fairly similar approach. From a technical standpoint, it should actually be much easier than this was, since there doesn't need to be any fancy box magic.

ospencer:
I went ahead and implemented that as well. Here's an example program:
```gr
export data Ref<a> = {
  mut contents: a
}

let ref = {contents: 'foo'}

print(ref)

ref.contents = 'bar'

print(ref)
```

phated:
I started converting my `map.gr` implementation to use mutable record fields and I ran into this error:
```
[wasm-validator error in function func_90] unexpected false: names in Binaryen IR must be unique - IR generators must ensure that, on 
record_set
[wasm-validator error in function func_93] unexpected false: names in Binaryen IR must be unique - IR generators must ensure that, on 
record_set
/Users/phated/github/grain/cli/bin/grainc.exe: internal error, uncaught exception:
                                               WASM Runner Exception in stdlib/map.gr: WARNING: Invalid module
```

phated:
I committed on https://github.com/grain-lang/grain/tree/let-mutable-stdlib for debugging purposes.

phated:
@ospencer can you rebase this on master whenever you get a chance?

ospencer:
@phated this has been rebased.

phated:
I found another case that breaks but I'm not sure why. Here is a small reproduction:

```
import Map from 'map';

let iter = (acc, key, value) => [(key, value), ...acc];

let m = Map.fromList([('a', 1), ('b', 2)]);

# Works
let works = Map.reduce(iter, [], m);
print(works);

# Breaks when returned
let broken_when_returned = (map) => {
  Map.reduce(iter, [], map)
}
let doesNotWork = broken_when_returned(m);
print(doesNotWork);
```

PR #206, opened 2020-07-11
Fix browser runtime

ospencer:
Looks like I broke the runtime in the browser when I added the WASI stuff, but this fixes that. I also updated the `defaultURLLocator`, since we haven't updated that since we started doing separate compilation.

### DIFF
--- a/runtime/src/core/grain-module.js
+++ b/runtime/src/core/grain-module.js
@@ -1,17 +1,27 @@
 import { GrainError } from '../errors/errors';
-import { grainToJSVal } from '../utils/utils';
 
 import { WASI } from "@wasmer/wasi/lib/index.cjs";
-import wasiBindings from "@wasmer/wasi/lib/bindings/node";
+import { WasmFs } from "@wasmer/wasmfs";
+
+const wasmFs = new WasmFs();
+
+let bindings
+
+if (__RUNTIME_BROWSER) {
+  bindings = {
+    ...wasiBindings.default,
+    fs: wasmFs.fs
+  }
+} else {
+  bindings = wasiBindings.default
+}
 
 export const wasi = new WASI({
-  args: process.argv,
-  env: process.env,
-  bindings: {
-    ...wasiBindings
-  },
+  args: __RUNTIME_BROWSER ? [] : process.argv,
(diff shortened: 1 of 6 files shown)


phated:
I wonder if we can add puppeteer in CI to verify this.

ospencer:
We could! Though I figure we'd want to do that once we had a better bundle for the stdlib. Could do it sooner though.

runtime/src/core/grain-module.js
```diff
@@ -1,17 +1,27 @@
 import { GrainError } from '../errors/errors';
-import { grainToJSVal } from '../utils/utils';
 
 import { WASI } from "@wasmer/wasi/lib/index.cjs";
-import wasiBindings from "@wasmer/wasi/lib/bindings/node";
+import { WasmFs } from "@wasmer/wasmfs";
+
+const wasmFs = new WasmFs();
```
phated, line 6: Should this be constructed only is `__RUNTIME_BROWSER`

runtime/src/utils/locators.js
```diff
@@ -7,15 +7,20 @@ function normalizeSlash(s) {
...
   return async (raw) => {
-    if (base === null) {
-      return null;
-    }
     let module = raw.replace(/^GRAIN\$MODULE\$/, '');
-    return readURL(base + "/" + module + ".wasm");
+    for (const base of bases) {
+      let fullpath = base + "/" + module + ".wasm";
+      try {
+        return await readURL(fullpath);
```
phated, line 18: Not an async function

ospencer: Awkward.

ospencer: Oh yes it is: https://github.com/grain-lang/grain/blob/master/runtime/src/core/grain-module.js#L189

phated: Sorry, I meant the function that you are using the `await` inside, which shouldn't work if it isn't `async`, right?

ospencer: There's an `async` on line 13. A little hard to see.

phated: Ah, didn't realize this function returned another function.

Issue #182, opened 2020-07-01
Lang: Trailing comma in balanced-syntax data structures


phated:
I'm working on defining some record types and just noticed that they don't support trailing commas.

Edit: As mentioned below, this should be supported in all balanced-syntax data structures:

- [x] Records
- [x] Lists
- [x] Arrays
- [x] ADT (inside the variant constructor)
- [x] Tuples

ospencer:
We should also support trailing commas for all balanced-syntax data structures (lists, arrays, records, ADTs, tuples).

I believe we currently only support a trailing comma for single-item tuples, i.e. `(0,)`. But we may have also gotten rid of single element tuples.

bmakuh:
Is it cool if I work on tackling this?

ospencer:
@bmakuh Definitely! You might find this helpful: https://github.com/grain-lang/grain/blob/master/docs/contributor/compiler_walkthrough.md

And as always, reach out on Discord if you run into any trouble 🙂

phated:
Is the last thing here the ADTs? I've updated the list in the original issue.

ospencer:
@phated I think so, but we'll probably need a separate issue for declarations. Or we could just use this one.

phated:
I think we can just use this issue and update the list at the top.

bmakuh:
Yeah, I'm working on ADTs now, just running into some snags

bmakuh:
Right now I have this working:
```
data Menu = Pizza(Topping, Dough,)
```
But this is still throwing a syntax error:
```
let item = Pizza(Peppers, WholeWheat,)
```
I cannot seem to figure out where in parser.dyp the latter is parsed.

ospencer:
Should be here: https://github.com/grain-lang/grain/blob/746c9bdab027b145b931129592e4b312b2b61f09/compiler/src/parsing/parser.dyp#L402-L405

bmakuh:
Wait, `Pizza` is considered a function here? 🤯

ospencer:
I'm not sure about all functional languages, but they're not treated as functions in OCaml/Reason at least. It's kinda neat having it this way since you can do something like `List.map(Some, [1, 2, 3])` in Grain (but that doesn't work in OCaml).

ospencer:
(And that's why it might feel kinda weird 😄 )

bmakuh:
Wow. That's really interesting. Thanks for the explanation!

PR #207, opened 2020-07-12
lang: Support trailing commas in records

phated:
Ref #182

I only implemented the Record syntax because it is a pain-point for me, but this should also give guidance on how someone could add it to the other balanced-syntax data structures mentioned there.


phated:
Done!

PR #209, opened 2020-07-12
Unicode escape sequences for strings

JohnnyRidout:
Unicode escape sequences for strings.

Examples: for letter "a"
Octal \141
Hex \x61
Unicode \u{61}


stdlib/stdlib-external/string.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+## Combining and normalization
+
+Some graphemes are encoded by multiple code points. Some graphemes can be
+encoded by either a single code point or multiple code points. For example,
+__ñ__ is U+00F1, but it can also be encoded by combining __n__ U+0006E and tilde
+ __◌̃__ U+0303.
+
+ Normalization is the process by which combining code points are reduced to
```
belph, line 49: We should open an issue to add a `normalize()` function to the standard library if we are going to discuss Unicode normalization in the docs.

stdlib/stdlib-external/string.md
```diff
@@ -0,0 +1,56 @@
...
+ either single code points or put into canonical order. This is necessary for
+ searching and comparing strings. Strings in Grain are not normalized.
+
+ ## JavaScript
+
+ JavaScript strings are not encoded in UTF-8.
+
```
belph, line 56: I don't think we need to mention this.
```suggestion
```

stdlib/stdlib-external/string.md
ospencer: Meant to mention this as well, but I can open the issue.

stdlib/stdlib-external/string.md
ospencer: I think this is important to mention in the contributor docs, especially as we expand our JS FFI. Though I think we should expand on it and maybe point out https://github.com/grain-lang/grain/blob/master/runtime/src/utils/utils.js#L76 specifically.

belph: We certainly can mention it in the contributor docs, but I should note that I wouldn't expect it to impact many users, since the Grain FFI layer translates directly from our UTF-8 strings into native ones with whatever encoding JS uses (I forget off the top of my head).

JohnnyRidout: JS uses UCS-2 with unvalidated surrogate pairs. A DOMString is the same but with a replacement character for invalid surrogate pairs. A USVString is the same but with validated surrogate pairs so it's UTF-16 restricted to the Basic Multilingual Plane. A CSSOMString is the same as a DOMString. The encoder converts to USVString.

ospencer: @belph These are contributor docs. I agree that we shouldn't mention this on the website or anything.

@JohnnyRidout Could you add that to this section?

JohnnyRidout: @ospencer,   I've added it and moved the document to docs/contributor.

I can add `\u0000` style escapes in another pull request next weekend.

@belph I can update the user docs with information about the escape sequences next weekend.

Does that cover everything?

ospencer: Yes, this is awesome! Thanks for your hard work on this!!

PR #215, opened 2020-07-13
Utilize the OCaml stdlib when possible

phated:
The OCaml stdlib has come a **long** way since this project was started, so I propose we drop the variety of external utility libs that were in use.

This work began when I dropped janestreet/core in an attempt to use js_of_ocaml on this project, but I've expanded it to remove `extlib` (polluted the globals and stopped us from using OCaml's Option stdlib) and `batteries` (native dependencies that might cause Windows or JS compilation to be difficult).

@ospencer also spent some time to remove the Stdint dependency, which was the other big native dependency.

I started adding some of the [Reason Native](https://github.com/facebookexperimental/reason-native) projects because they can help us clean up some code and they have a goal of being cross-platform & compile to JS while having minimal/no dependencies. 

In addition to general cleanup, this PR also fixes a bug that BatPathGen wouldn't work on paths that began with `.` so we couldn't use `./something` or `../something` in Grain imports. I believe there is still an issue with absolute paths but I didn't dig into that (I'll open an issue later because it is late).


compiler/src/codegen/compcore.re
```diff
@@ -1795,19 +1793,16 @@ let heap_check_memory = (wasm_mod, env, num_words: int) => {
...
-    out_ints := [bytes_to_int(byte_buf, i * 8), ...out_ints^];
-  };
-  List.rev @@ List.map(Stdint.Uint64.to_int64, out_ints^);
+  let num_ints = num_bytes / 8;
+  let num_ints = num_bytes mod 8 == 0 ? num_ints : num_ints + 1;
+  let total_bytes = num_ints * 8;
+
+  let bytes = Buffer.to_bytes(buf);
+  let bytes = Bytes.extend(bytes, 0, total_bytes - num_bytes);
+  // Clear out those unitialized bytes
```
ospencer, line 1802: This was my bad:
```suggestion
  // Clear out those uninitialized bytes
```

compiler/src/parsing/well_formedness.re
```diff
@@ -153,7 +156,7 @@ let types_have_correct_case = (errs, super) => {
 let modules_have_correct_case = (errs, super) => {
   let check_uppercase = (loc, s) => {
     let first_char = s.[0];
-    if (first_char != BatChar.uppercase(first_char)) {
+    if (Char_utils.is_lowercase_letter(first_char)) {
```
ospencer, line 159: Same deal here.

compiler/src/codegen/compcore.re
```diff
@@ -332,7 +333,7 @@ let runtime_function_imports =
         mimp_type:
           [@implicit_arity]
           MFuncImport(
-            BatList.init(Runtime_errors.max_arity + 1, _ => I32Type),
+            List.init(Runtime_errors.max_arity + 1, _ => I32Type),
```
belph, line 336: 😍 Very happy to see this is in the stdlib now

compiler/src/codegen/compcore.re
```diff
@@ -378,7 +378,7 @@ let grain_number_min = (-0x3fffffff); // 0xC0000001
 
 /** Constant compilation */
 
-let rec compile_const = c: Literal.t => {
+let rec compile_const = (c): Literal.t => {
```
belph, line 381: Is there a prettifier somewhere that we could use to catch things like this and make it part of our CI? Now that I say that out loud, I feel like I am forgetting a thread somewhere about this...

compiler/src/utils/files.re
```diff
@@ -1,49 +1,88 @@
+// This file is converting strings into Fp.t and then back into Strings
+// TODO: We should consider switching to type safe Fp.t where ever filepaths are used
```
belph, line 2: @phated Do you mind opening an issue for this?

compiler/src/codegen/compcore.re
phated: Yeah, we have a `esy format` but I haven't proxied that to the yarn commands yet or set up CI because I think the Reason formatter will update a bunch and I was waiting until some of the bigger PRs landed to avoid conflicts.

It might be time though because the exceptions PR still needs to be converted to Reason AFAIK

compiler/src/parsing/well_formedness.re
phated: What's the same?

compiler/src/utils/files.re
phated: Will do! Opening a few this morning

compiler/src/codegen/compcore.re
phated: I enjoyed my usage in `String_utils.explode`, very handy method!

compiler/src/parsing/well_formedness.re
ospencer: That this logic should be "isn't uppercase" rather than "is lowercase."

phated: Is that just because special characters should throw here?

phated: Updated.

phated:
I squashed this into master because the number of commits was pretty intense.

Issue #218, opened 2020-07-13
CI: Refmt linting


phated:
We need to get the autoformatting stuff figured out and add a formatting check in CI.

I'd advise against trying to format/autopromote + push in CI because it doesn't work on forks (I struggled a long time with that for gulp), but we should at least warn someone if the formatting is wrong.

belph:
Maybe a git hook which warns on commits would be appropriate?

phated:
Generally, I'm against commit hooks because they add significant lag to git commands.

belph:
Valid point. I'm just trying to think of what could be the lowest-friction way to make sure it runs on everything. 🤔

phated:
I actually have my VSCode editor set up to do "format on save", which is why some of my commits have formatting changes. OCaml LSP does a pretty great job with that.

belph:
Totally, I think we should recommend people use that or a similar tool. We definitely need a way to run a check on code as it comes in, though, since we will be having more and more contributors as time goes on.

phated:
Oh yeah, I already figured out how to add a linting phase to CI that will fail if code wasn't formatted

phated:
Closing this with #220 landed. We can open another issue once we figure out how it can be improved.

PR #220, opened 2020-07-13
ReasonML formatting

phated:
This formats the Reason code and then adds a format checking phase to our CI.

As you can probably tell, Reason has some issues with block comments and we should probably switch them to line comments (which it seems to handle much more elegantly).

Not sure if this is enough to satisfy all of #218 - Thoughts?

I'll push up a sample of the lint failing once it passes just as an example.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,11 @@ jobs:
         run: |
           yarn compiler test
 
+      # Formatting lint last because building is more important
+      - name: Run formatting lint
+        run: |
+          yarn compiler check-format
+
       # This is to test the CLI is working
       - name: Log Grain version
         run: |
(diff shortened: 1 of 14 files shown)


phated:
I pushed some incorrectly formatted code to another branch. This build: https://github.com/grain-lang/grain/actions/runs/167901512 should fail shortly with a formatting problem.

ospencer:
Do you think we could make the linting a separate CI job? That'd make it more obvious that the linting was the reason it failed.

phated:
@ospencer then we'd have 2 builds racing to build the compiler. Unfortunately, everything needs to be built before esy can execute the dune commands.

ospencer:
That's fair. Thoughts on making the format command just be the default build command?

phated:
I also tried that but the command exits with non-zero if it formats (even with auto-promotion) so I didn't want to do that because building the compiler shouldn't fail for just formatting problems.

PR #222, opened 2020-07-14
Switch stdlib boxes to let-mut

phated:
This also adds `let mut` tests in test-end-to-end.

and LET me just say `let mut` is a dream to work with!


compiler/test/input/recursive-equal.gr
```diff
@@ -1,21 +1,24 @@
...
 assert !(z == y)
 
-y := Some(y)
+y = Some(y)
 assert z == y
 
-export data R = Atom | Recurisve(Box<Opt<R>>)
+# These fail with `let mut` because we don't know they shouldn't be unwrapped
+# We'd probably need something like `mut Opt<R>`
+export data R = Atom | Recursive(Box<Opt<R>>)
```
ospencer, line 16: Might as well drop this comment here, since you pointed out that this can be accomplished with boxes/mutable record fields. _We_ know that `let mut` is a box under the hood, but for the average user they should think of `let mut` as similar to, say, JavaScript `let`.

```suggestion
export data R = Atom | Recursive(Box<Opt<R>>)
```

compiler/test/input/recursive-equal.gr
```diff
@@ -25,22 +28,23 @@ u := Some(w)
 
 assert x == w
 
+# These succeed with mutable record props
```
ospencer, line 31: Excellent to know. 😂 

Maybe it would be good to duplicate this test with both a box version and a mutable record field version so both code paths are exercised.

compiler/test/test_end_to_end.re
```diff
@@ -664,6 +664,30 @@ let box_tests = [
   t("box_division2", "let b = box(76); b := unbox(b) / 19; ^b", "4"),
 ];
 
+let let_mut_tests = [
```
ospencer, line 667: Woah! I wasn't expecting you to write these!

compiler/test/test_end_to_end.re
```diff
@@ -664,6 +664,30 @@ let box_tests = [
...
+  t("let-mut3_2", "let mut b = box(4); ^b", "4"),
+  t("let-mut4", "let mut b = 4;b = 3;b", "3"),
+  t("let-mut5", "let mut b = 4;b = b - 1;b", "3"),
+  tfile("counter-mut", "counter-mut", "1\n2\n3\nvoid"),
+  te(
+    "test_mut_typing",
+    "let mut a = false; a + 4",
+    "expression has type Bool but",
+  ),
+  /* Operations on Box<Number> */
```
ospencer, line 680:
```suggestion
  /* Operations on mutable `Number`s */
```

compiler/test/test_end_to_end.re
```diff
@@ -664,6 +664,30 @@ let box_tests = [
...
+  t("let-mut_addition1", "let mut b = 4; b = b + 19", "23"),
+  t("let-mut_addition2", "let mut b = 4; b = b + 19; b", "23"),
+  t("let-mut_subtraction1", "let mut b = 4; b = b - 19", "-15"),
+  t("let-mut_subtraction2", "let mut b = 4; b = b - 19; b", "-15"),
+  t("let-mut_multiplication1", "let mut b = 4; b = b * 19", "76"),
+  t("let-mut_multiplication2", "let mut b = 4; b = b * 19; b", "76"),
+  t("let-mut_division1", "let mut b = 76; b = b / 19", "4"),
+  t("let-mut_division2", "let mut b = 76; b = b / 19; b", "4"),
```
ospencer, line 688: I re-implemented `+=`, etc., to work with mutable values rather than boxes. May be good to exercise those here too?

compiler/test/input/recursive-equal.gr
phated: Alright. I'm going to actually create 2 separate files so I won't accept this right now.

compiler/test/input/recursive-equal.gr
phated: 👍 see above

compiler/test/test_end_to_end.re
phated: 😆 I just grepped for box and copy-pasted or replaced everything

phated:
@ospencer Updated and squashed. Did you want me or you to merge this into your branch?

ospencer:
I'll merge it!

PR #224, opened 2020-07-15
cli: Add compile & run subcommands (closes #223)

### LINKED ISSUE #223: CLI: Add ability to compile only
Currently, you have to use `grainc` directly to *just* compile files without running them. Since grainc is copied into the CLI now, we can expose a command for compiling files.

phated:
Adds some subcommands to make things easier to do with just the `grain` CLI

By moving people to use one CLI, any underlying changes we make to structure shouldn't affect the majority of users, like when I started copying `grainc.exe` into the CLI directory.

### DIFF
--- a/cli/bin/compile.js
+++ b/cli/bin/compile.js
@@ -12,7 +12,7 @@ module.exports = (file, options) => {
     if (options.graceful) {
       process.exit()
     } else {
-      process.exit(-1)
+      process.exit(1)
     }
   }
 }
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -34,7 +34,6 @@ program
     process.exit(0)
   })
   .description('Compile and run Grain programs. 🌾')
-  .option('-w, --wasm', 'run a wasm file')
   .option('-p, --print-output', 'print the output of the program')
   .option('-g, --graceful', 'return a 0 exit code if the program errors')
   .option('-I, --include-dirs <dirs>', 'include directories the runtime should find wasm modules', list, [])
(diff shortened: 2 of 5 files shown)


ospencer:
So if I'm reading this correctly, it looks like `grain foo.gr` and `grain run foo.gr` do the same thing?

phated:
No, run doesn't compile first so it really only takes a wasm file. I was conflicted about that though because it feels like you should be using grain files with our CLI

ospencer:
Ah, so `run` does the same thing as `--wasm`. I suppose we could remove `--wasm` then.

It's useful if you just don't want to recompile something. That's also how the tests work; the tests generate the wasm file and then invoke the runtime with `-w`.

Also fun fact: if you have a wasm file that doesn't require any imports, you can just run them from the Grain CLI, i.e. if you compile a Rust program to a wasm file, it'll work with the Grain CLI.

phated:
I'm good with removing the `--wasm` flag and just telling people to use `grain run` to exec wasm files. I'll change the tests too.

cli/bin/compile.js
```diff
@@ -12,7 +12,7 @@ module.exports = (file, options) => {
     if (options.graceful) {
       process.exit()
     } else {
-      process.exit(-1)
+      process.exit(1)
```
phated, line 15: Btw, my new ZSH setup (it reports the error code of the last command) helped me identify this issue because using `-1` actually caused a 255 error code instead of 1 like we were trying to do 😛

ospencer: Yeah—I think this one was my fault 😛

phated:
@ospencer Updated this with removal of the --wasm flag

PR #228, opened 2020-07-21
Write Grain CMI using Binaryen

ospencer:
~This is waiting on https://github.com/grain-lang/binaryen.ml/pull/18 to land, but then it's all set.~

This should be ready to go!


phated:
I can't wait to try this! You know I love those negative lines of code!

compiler/src/codegen/compcore.re
```diff
@@ -3984,6 +3984,13 @@ let compile_wasm_module = (~env=?, ~name=?, prog) => {
   let () = ignore @@ compile_tables(wasm_mod, env, prog);
   let () = ignore @@ compile_elems(wasm_mod, env, prog);
 
+  let serialized_cmi = Cmi_format.serialize_cmi(prog.signature);
+  Module.add_custom_section(
+    wasm_mod,
+    "cmi",
```
phated, line 3990: We might want to consider renaming this to `grain_cmi` or something in case someone starts putting their CMIs in a custom section named `cmi` 😛

compiler/src/codegen/compcore.re
```diff
@@ -3984,6 +3984,13 @@ let compile_wasm_module = (~env=?, ~name=?, prog) => {
...
+    Bytes.to_string(serialized_cmi),
```
phated, line 3991: Any particular reason this API takes a string instead of bytes? The JS API takes a uint8array.

compiler/src/codegen/compcore.re
ospencer: Yeah, that's fair. Though at least we sprinkle some [Grain magic](https://github.com/grain-lang/grain/blob/master/compiler/src/utils/wasm_utils.re#L33) to know that it's a Grain CMI. But we could open an issue for that, yeah.

compiler/src/codegen/compcore.re
ospencer: Just because when I wrote it, I figured that it would always be a string. The [C API called it `char *`](https://github.com/WebAssembly/binaryen/blob/master/src/binaryen-c.h#L2322) so my brain just went straight to strings. I would be down to change that though.

Issue #231, opened 2020-07-27
Use full filepath for persistent structures


ospencer:
We currently use unresolved paths for persistent structures, which is problematic if modules have the same name but exist in different places (e.g. the standard library versus a local file).

Relevant code here:
https://github.com/grain-lang/grain/blob/dd23a5f0b27be2b5289ae8f1aa2e029e79f5399a/compiler/src/typed/env.re#L688-L689

PR #232, opened 2020-07-27
[Stdlib] Standardize library names

ospencer:
[BREAKING]

This PR renames a few libraries for consistency:

`lists.gr` -> `list.gr`
`arrays.gr` -> `array.gr`
`stdint64.gr` -> `int64.gr`

`int64.gr` was previously named `stdint64.gr` because of a previously unidentified bug with module loading. That bug has now been identified in #231. Until #231 is fixed, I had to rename a few unit tests to get the tests to pass.


ospencer:
I'll also update the docs before I merge this.

PR #239, opened 2020-08-02
feat: add CLI option for no_gc

willemneal:
Prevent reference counting with `no_gc` flag.


phated:
@willemneal heads up, we needed to do some work around this to debug some memory management issues we were having so we pushed it up to this branch. I'm going to mark this as "ready for review" and we will likely land it shortly.

phated (approved):
Quick once-over on this and it looks solid. @ospencer are we ready to merge this?

ospencer (approved):
I think it's a solid first pass!

PR #256, opened 2020-08-13
lang: Support trailing comma in list literals

phated:
This adds support for parsing list literals with a trailing comma, given they aren't empty and don't end with a spread.

I also added tests for the list syntax because they didn't seem to exist. I also discovered that lists are heterogeneous?!!?

There is an additional commit here that cleans up trailing whitespace in the parser file because I'm sick of dodging it in my PRs 🤣 

Ref #182

### DIFF
--- a/compiler/src/parsing/parser.dyp
+++ b/compiler/src/parsing/parser.dyp
@@ -412,15 +412,15 @@ ampamp_op :
   | AMPAMP { "&&" }
 pipepipe_op :
   | PIPEPIPE { "||" }
-pluseq_op :  
+pluseq_op :
   | PLUS EQUAL { "+" }
-dasheq_op :  
+dasheq_op :
   | DASH EQUAL { "-" }
-stareq_op :  
+stareq_op :
   | STAR EQUAL { "*" }
-slasheq_op :  
+slasheq_op :
   | SLASH EQUAL { "/" }
-percenteq_op :  
+percenteq_op :
   | PERCENT EQUAL { "%" }
 
 infix_op :
@@ -520,12 +520,12 @@ stmt_expr :
   | FAIL expr { Exp.apply ~loc:(symbol_rloc dyp) (mkid_expr dyp ["fail"]) [$2] }
 
 assign_binop_op :
-  | pluseq_op 
-  | dasheq_op 
-  | stareq_op 
+  | pluseq_op
+  | dasheq_op
+  | stareq_op
   | slasheq_op
   | percenteq_op
-  
+
 assign_expr :
   | binop_expr(<pl) eols? GETS eols? expr { no_array_access $1; Exp.box_assign ~loc:(symbol_rloc dyp) $1 $5 }
   | id_expr equal expr { Exp.assign ~loc:(symbol_rloc dyp) $1 $3 }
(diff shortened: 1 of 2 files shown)


phated:
Oh, I just realized that this makes the comma optional between an element and spread. I'll need to fix that before landing this.

compiler/src/parsing/parser.dyp
```diff
@@ -507,9 +507,13 @@ match_branches :
...
   | MATCH lparen expr rparen lbrace match_branches rbrace { Exp.match_ ~loc:(symbol_rloc dyp) $3 $6 }
 
+list_expr_ending :
+  | comma? { None }
+  | comma ELLIPSIS expr { Some $3 }
+
 list_expr :
   | lbrack rbrack { Exp.list ~loc:(symbol_rloc dyp) [] None }
-  | lbrack expr [comma expr {$2}]* [comma ELLIPSIS expr {$3}]? rbrack { Exp.list ~loc:(symbol_rloc dyp) ($2::$3) $4 }
+  | lbrack expr [comma expr {$2}]* list_expr_ending rbrack { Exp.list ~loc:(symbol_rloc dyp) ($2::$3) $4 }
```
phated, line 516: @ospencer is there a way to improve this? I had to hoist the Some/None to the new rule because comma's value is unit.

ospencer: This is probably fine, though I might rename `list_expr_ending` to something like `list_spread_opt` or `list_spread`.

compiler/src/parsing/parser.dyp
```diff
@@ -507,9 +507,13 @@ match_branches :
 match_expr :
   | MATCH lparen expr rparen lbrace match_branches rbrace { Exp.match_ ~loc:(symbol_rloc dyp) $3 $6 }
 
+list_expr_ending :
+  | comma? { None }
+  | comma ELLIPSIS expr { Some $3 }
```
ospencer, line 512: Maybe trailing commas after a spread as well? Might not make sense syntactically (just yet) until we support multiple spreads.
```suggestion
  | comma ELLIPSIS expr comma? { Some $3 }
```

compiler/test/test_end_to_end.re
```diff
@@ -337,6 +337,28 @@ let tuple_tests = [
...
+    "let a = [3, 4]; [1, 2 ...a]",
+    "Error: Syntax error",
+  ),
+  // trailing commas
+  t("list1_trailing", "[1, 2, 3,]", "[1, 2, 3]"),
+  t("list1_trailing_space", "[1, 2, 3, ]", "[1, 2, 3]"),
+  te("invalid_empty_trailing", "[,]", "Error: Syntax error"),
+  te(
+    "invalid_list_spread_trailing",
+    "let a = [3, 4]; [1, 2, ...a,]",
```
ospencer, line 357: And then this would pass.

compiler/src/parsing/parser.dyp
phated: I don't think it is a good idea to allow a comma at the end of a spread until multi-spread or spreads in non-terminal positions are supported.

I imagine a situation where someone rights code like `[a, b, ...rest,]` and then another dev comes to the code and tries to add `[a, b, ...rest, c]` and it doesn't work.

compiler/src/parsing/parser.dyp
phated: I don't like naming this anything with `spread`, I actually considered the naming for a long time and it's not actually just a spread pattern but instead how the list expression "ends". Ideally, `comma ELLIPSIS expr` will be extracted into a `list_spread` rule when they are supported in more positions.

PR #258, opened 2020-08-13
Add result type

### LINKED ISSUE #244: Stdlib: Result
Result is similar to our Option stdlib, but it has 2 generic type parameters, one for the data and one for the error.

The type would be defined like:
```
data Result<data, error> =
    | Ok(data)
    | Error(error);
```

The stdlib module would contain many of the same utilities as Option.

c0d3d:
Closes #244 

~~This isn't complete yet. Please review API and give thoughts. Also I did `Err` instead of `Error` can switch if asked.~~

### DIFF
--- a/compiler/test/stdlib/option.test.gr
+++ b/compiler/test/stdlib/option.test.gr
@@ -124,3 +124,7 @@ assert c == 4
 let mut d = '🌾'
 assert Option.peek((x) => fail 'Should not be called', None) == None
 assert d == '🌾'
+
+# Option.toResult
+assert Option.toResult(Some(1), "Nope") == Ok(1)
+assert Option.toResult(None, "Yes!") == Err("Yes!")
--- a/compiler/test/stdlib/result.test.gr
+++ b/compiler/test/stdlib/result.test.gr
@@ -1,0 +1,95 @@
+import Result from 'result';
+
+# isOk
+assert Result.isOk(Ok(1)) == true
+assert Result.isOk(Err("Ooops")) == false
+
+# isErr
+assert Result.isErr(Ok(1)) == false
+assert Result.isErr(Err("Ooops")) == true
+
+# toOption
+assert Result.toOption(Ok(1)) == Some(1)
+assert Result.toOption(Err("Ooops")) == None
+
+# map
+assert Result.map((a) => "some other type", Ok(1)) == Ok("some other type")
+assert Result.map((v) => fail "Shouldn't run", Err("Ooops")) == Err("Ooops")
+
+# mapErr
+assert Result.mapErr((x) => fail "Shouldn't run", Ok(1)) == Ok(1)
+assert Result.mapErr(incr, Err(1)) == Err(2)
+
+# flatMap
+assert Result.flatMap((a) => Ok("some other type"), Ok(1)) == Ok("some other type")
+assert Result.flatMap((a) => Err("some other type"), Ok(1)) == Err("some other type")
+assert Result.flatMap((x) => fail "Shouldn't run", Err("Ooops")) == Err("Ooops")
+
+# flatMapErr
+assert Result.flatMapErr((x) => fail "Shouldn't run", Ok(1)) == Ok(1)
+assert Result.flatMapErr((x) => Err(incr(x)), Err(1)) == Err(2)
+assert Result.flatMapErr((x) => Ok(incr(x)), Err(1)) == Ok(2)
+
+# mapWithDefault
+assert Result.mapWithDefault(toString, "Nope!", Ok(1)) == "1"
+assert Result.mapWithDefault(toString, "Nope!", Err(1)) == "Nope!"
+
+# mapWithDefaultFn
+assert Result.mapWithDefaultFn(toString, (x) => fail "Shouldn't run", Ok(1)) == "1"
+assert Result.mapWithDefaultFn((x) => fail "Shouldn't run", (x) => "7", Err("Awesome")) == "7"
+
+# or
+assert Result.or(Ok(1), Ok(2)) == Ok(1)
+assert Result.or(Ok(1), Err(2)) == Ok(1)
+assert Result.or(Err(1), Ok(2)) == Ok(2)
+assert Result.or(Err(1), Err(2)) == Err(2)
+
+# and
+assert Result.and(Ok(1), Ok(2)) == Ok(2)
+assert Result.and(Ok(1), Err(2)) == Err(2)
+assert Result.and(Err(1), Ok(2)) == Err(1)
+assert Result.and(Err(1), Err(2)) == Err(1)
+
+let mut a = 1
+let testMut = (f) => {
+    a = 1
+    f()
+}
+
+# peek
+testMut(() => {
+    Result.peek(
+       (v) => a = v,
+       (v) => fail 'Shouldn\'t run',
+       Ok(2)
+    )
+    assert a == 2
+})
+testMut(() => {
+    Result.peek(
+       (v) => fail 'Shouldn\'t run',
+       (v) => a = v,
+       Err(2)
+    )
+    assert a == 2
+})
+
+# peekOk
+testMut(() => {
+    Result.peekOk((v) => a = v, Ok(2))
+    assert a == 2
+})
+testMut(() => {
+    Result.peekOk((v) => a = v, Err(2))
+    assert a == 1
+})
+
+# peekErr
+testMut(() => {
+    Result.peekErr((v) => a = v, Ok(2))
+    assert a == 1
+})
+testMut(() => {
+    Result.peekErr((v) => a = v, Err(2))
+    assert a == 2
+})
--- a/stdlib/pervasives.gr
+++ b/stdlib/pervasives.gr
@@ -46,3 +46,9 @@ export data List<a> = [] | [...](a, List<a>)
 
 # Maybe some data, maybe not!
 data Option<a> = Some(a) | None
+
+# Maybe some data, maybe an error!
+data Result<t, e> = Ok(t) | Err(e)
+
+# Identity function
+let identity = (x) => x
--- a/stdlib/result.gr
+++ b/stdlib/result.gr
@@ -1,0 +1,87 @@
+export let isOk = (v) => {
+    match (v) {
+        | Ok(_)  => true
+        | _      => false
+    }
+};
+
+export let isErr = (v) => !isOk(v);
+
+export let toOption = (v) => {
+    match (v) {
+        | Ok(x)  => Some(x)
+        | _      => None
+    }
+}
+
+export let flatMap = (fn, v) => {
+    match (v) {
+        | Ok(x)  => fn(x)
+        | _      => v
+    }
+}
+
+export let flatMapErr = (fn, v) => {
+    match (v) {
+        | Err(e) => fn(e)
+        | _      => v
+    }
+}
+
+export let map = (fn, v) => {
+    match (v) {
+        | Ok(x) => Ok(fn(x))
+        | _     => v
+    }
+}
+
+export let mapErr = (fn, v) => {
+    match (v) {
+        | Err(x) => Err(fn(x))
+        | _      => v
+    }
+}
+
+export let mapWithDefault = (fn, def, v) => {
+    match (v) {
+        | Ok(a) => fn(a)
+        | _     => def
+    }
+}
+
+export let mapWithDefaultFn = (fnOk, fnErr, v) => {
+    match (v) {
+        | Ok(a)  => fnOk(a)
+        | Err(a) => fnErr(a)
+    }
+}
+
+export let or = (r1, r2) => {
+    match (r1) {
+        | Ok(x) => r1
+        | _     => r2
+    }
+}
+
+export let and = (r1, r2) => {
+    match (r1) {
+        | Ok(_) => r2
+        | Err(_) => r1
+    }
+}
+
+export let peek = (fnOk, fnErr, r) => {
+    match (r) {
+        | Ok(x) => ignore(fnOk(x))
+        | Err(x) => ignore(fnErr(x))
+    }
+    r
+}
+
+export let peekOk = (fn, r) => {
+    peek(fn, identity, r)
+}
+
+export let peekErr = (fn, r) => {
+    peek(identity, fn, r)
+}
--- a/compiler/test/test_end_to_end.re
+++ b/compiler/test/test_end_to_end.re
@@ -627,6 +627,7 @@ let stdlib_tests = [
   tlib("strings.test"),
   tlib("sys.file.test"),
   tlib("map.test"),
+  tlib("result.test"),
   tlib(~returns="", ~code=5, "sys.process.test"),
 ];
 
--- a/stdlib/option.gr
+++ b/stdlib/option.gr
@@ -120,3 +120,10 @@ let peek = (fn, option) => {
   sideEffect(fn, option)
   option
 }
+
+let toResult = (option, err) => {
+  match (option) {
+    | Some(a) => Ok(a)
+    | None    => Err(err)
+  }
+}


stdlib/result.gr
```diff
@@ -0,0 +1,63 @@
...
+export let mapOk = (v, fn) => {
+    match (v) {
+        | Ok(x)  => Ok(fn(x))
+        | Err(e) => Err(e)
+    }
+}
+
+export let mapErr = (v, fn) => {
+    match (v) {
+        | Err(e) => Err(fn(v))
```
phated, line 26: Should just the `e` be passed to `fn`? Maybe that's a typechecker error?

stdlib/pervasives.gr
```diff
@@ -46,3 +46,9 @@ export data List<a> = [] | [...](a, List<a>)
 
 # Maybe some data, maybe not!
 data Option<a> = Some(a) | None
+
+# Maybe some data, maybe an error!
+data Result<t, e> = Ok(t) | Err(e)
```
phated, line 51: 👍 to `Err`, it looks good in the code!

stdlib/pervasives.gr
```diff
@@ -46,3 +46,9 @@ export data List<a> = [] | [...](a, List<a>)
...
+
+# Identity function
+let id = (x) => x
```
phated, line 54: Woot! I've been thinking we need an identity

stdlib/result.gr
```diff
@@ -0,0 +1,63 @@
...
+        | _      => v
+    }
+}
+
+export let map = (v, fnOk, fnErr) => {
```
phated, line 31: I think we chose `flatMap` naming for this type of function. I think we can also have a `map` that rewraps in Ok/Err

stdlib/result.gr
```diff
@@ -0,0 +1,63 @@
...
+    match (v) {
+        | Ok(x)  => fnOk(x)
+        | Err(x) => fnErr(x)
+    }
+}
+
+export let okOr = (v, default) => {
```
phated, line 38: In `Option`, this is `unwrapWithDefault` so maybe we should name this the same or `okWithDefault`?

stdlib/result.gr
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export let or = (r1, r2) => {
+    match (r1) {
+        | Ok(x) => r1
+        | _     => r2
+    }
+}
+
+let runForSideEffects = (fn) => (v) => {
```
phated, line 52: I didn't realize we had manually currying (but of course we do 🤦)

stdlib/result.gr
c0d3d: Good catch, there is a type checker error right now.

compiler/test/stdlib/result.test.gr
```diff
@@ -0,0 +1,99 @@
...
+assert Result.isErr(Ok(1)) == false
+assert Result.isErr(Err("Ooops")) == true
+
+# toOption
+assert Result.toOption(Ok(1)) == Some(1)
+assert Result.toOption(Err("Ooops")) == None
+
+# mapOk
+assert Result.mapOk(Ok(1), a => "some other type") == Ok("some other type")
+assert Result.mapOk(Err("Ooops"), incr) == Err("Ooops")
```
phated, line 17:
```suggestion
assert Result.mapOk(Err("Ooops"), (v) => fail 'Shouldn\'t run') == Err("Ooops")
```

compiler/test/stdlib/result.test.gr
```diff
@@ -0,0 +1,99 @@
...
+
+# mapErr
+assert Result.mapErr(Ok(1), incr) == Ok(1)
```
phated, line 20:
```suggestion
assert Result.mapErr(Ok(1), (v) => fail 'Shouldn\'t run') == Ok(1)
```

compiler/test/stdlib/result.test.gr
```diff
@@ -0,0 +1,99 @@
...
+assert Result.mapErr(Err(1), incr) == Err(2)
+
+# map
+assert Result.map(Ok(1), (x) => toString(x), (x) => 7) == Ok("1")
```
phated, line 24:
```suggestion
assert Result.map(Ok(1), (x) => toString(x), (v) => fail 'Shouldn\'t run') == Ok("1")
```

compiler/test/stdlib/result.test.gr
```diff
@@ -0,0 +1,99 @@
...
+assert Result.map(Err("awesome"), toString, (x) => 7) == Err(7)
```
phated, line 25:
```suggestion
assert Result.map(Err("awesome"), (v) => fail 'Shouldn\'t run', (x) => 7) == Err(7)
```

compiler/test/stdlib/result.test.gr
```diff
@@ -0,0 +1,99 @@
...
+
+# flatMapOk
+assert Result.flatMapOk(Ok(1), a => Ok("some other type")) == Ok("some other type")
+assert Result.flatMapOk(Ok(1), a => Err("some other type")) == Err("some other type")
+assert Result.flatMapOk(Err("Ooops"), (x) => Ok(incr(x))) == Err("Ooops")
```
phated, line 30:
```suggestion
assert Result.flatMapOk(Err("Ooops"), (v) => fail 'Shouldn\'t run') == Err("Ooops")
```

compiler/test/stdlib/result.test.gr
```diff
@@ -0,0 +1,99 @@
...
+
+# flatMapErr
+assert Result.flatMapErr(Ok(1), (x) => Ok(incr(x))) == Ok(1)
```
phated, line 33:
```suggestion
assert Result.flatMapErr(Ok(1), (v) => fail 'Shouldn\'t run') == Ok(1)
```

compiler/test/stdlib/result.test.gr
```diff
@@ -0,0 +1,99 @@
...
+assert Result.or(Ok(1), Ok(2)) == Ok(1)
+assert Result.or(Ok(1), Err(2)) == Ok(1)
+assert Result.or(Err(1), Ok(2)) == Ok(2)
+assert Result.or(Err(1), Err(2)) == Err(2)
+
+# and
+assert Result.and(Ok(1), Ok(2)) == Ok(2)
+assert Result.and(Ok(1), Err(2)) == Err(2)
+assert Result.and(Err(1), Ok(2)) == Err(1)
+assert Result.and(Err(1), Err(2)) == Err(1)
```
phated, line 55: These are awesome, and I'm going to get them added for Option too!

compiler/test/stdlib/result.test.gr
phated: I think we should make sure to use failures on anything that shouldn't be called to make sure some refactoring isn't done that hoists the function to always be called then throws away the result.

compiler/test/stdlib/result.test.gr
```diff
@@ -0,0 +1,99 @@
...
+assert Result.flatMapOk(Ok(1), a => Err("some other type")) == Err("some other type")
+assert Result.flatMapOk(Err("Ooops"), (x) => Ok(incr(x))) == Err("Ooops")
+
+# flatMapErr
+assert Result.flatMapErr(Ok(1), (x) => Ok(incr(x))) == Ok(1)
+assert Result.flatMapErr(Err(1), (x) => Err(incr(x))) == Err(2)
+assert Result.flatMapErr(Err(1), (x) => Ok(incr(x))) == Ok(2)
+
+# flatMap
+assert Result.map(Ok(1), (x) => toString(x), (x) => 7) == Ok("1")
```
phated, line 38:
```suggestion
assert Result.flatMap(Ok(1), (x) => Ok(toString(x)), (v) => fail 'Shouldn\'t run') == Ok("1")
```

compiler/test/stdlib/result.test.gr
```diff
@@ -0,0 +1,99 @@
...
+assert Result.map(Err("awesome"), toString, (x) => 7) == Err(7)
```
phated, line 39:
```suggestion
assert Result.flatMap(Err("awesome"), (v) => fail 'Shouldn\'t run', (x) => Err(7)) == Err(7)
```

stdlib/result.gr
```diff
@@ -0,0 +1,94 @@
...
+        | _      => v
+    }
+}
+
+export let flatMap = (v, fnOk, fnErr) => {
+    match (v) {
+        | Ok(x)  => fnOk(x)
+        | Err(x) => fnErr(x)
+    }
+}
```
phated, line 36: Can this fool the type system if I pass 2 methods that neither return a Result type? cc @ospencer

c0d3d: I wouldn't call this fooling it.
The answer should be yes, since `fnOk` and `fnErr` should both be constrained to return the same type.
Either way, they need to return the same type, and it won't be a problem.

ospencer: That's the intended behavior, right? Since this is `flatMap` rather than `map`? The typechecker should just have to make sure `fnOk` and `fnErr` return the same type.

phated: Isn't `flatMap` usually expected to return the same type as the value? Like `bind` in the OCaml stdlib?

c0d3d: Perhaps I mixed them up. Currently, `map` re-wraps the output automatically.

phated: You didn't mix them up, the implementation is correct, but the type system will cause (what should be) invalid usage to work.

```
Result.flatMap(Ok(1), (o) => "test", (e) => "test_err")
```

Should be invalid because `flatMap` functions should always return a Result type.

ospencer: If we don't want to wrap the output, I suppose this could just be solved with type annotations.

c0d3d: I guess I don't really see the problem. Why is it necessary to arbitrarily constrain user's functions?

compiler/test/stdlib/result.test.gr
ospencer: Can also use double quoted strings for these:
```suggestion
assert Result.mapOk(Err("Ooops"), (v) => fail "Shouldn't run") == Err("Ooops")
```

stdlib/result.gr
ospencer: I feel like it's fine. If we required you to return a result, it would just be the same as `map`. @phated were you thinking of this as a combination of `flatten` and `map`? I think that might be the confusion. I would expect a function that takes a `Result<Result<a, e> e>` and called `flatMap` to produce a `Result<b, f>`.

This could potentially be named `reduce` or something like that, and `flatMap` would do this ☝️

ospencer: Rust's `Result` `flatten`, for example: https://doc.rust-lang.org/beta/core/result/enum.Result.html#method.flatten

ospencer: Yeah, Rust calls this function `map_or_else`: https://doc.rust-lang.org/beta/core/result/enum.Result.html#method.map_or_else

phated: @ospencer no, our `flatMap` should be similar to `and_then` in Rust (which is also the same in Elm). We talked about that in my Option PR

This is the OCaml type for [`bind`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Result.html#VALbind):
```ocaml
val bind : ('a, 'e) result -> ('a -> ('b, 'e) result) -> ('b, 'e) result
```

And Rust's [`and_then`](https://doc.rust-lang.org/beta/core/result/enum.Result.html#method.and_then):
```rust
pub fn and_then<U, F: FnOnce(T) -> Result<U, E>>(self, op: F) -> Result<U, E>
```

And Belt's [`flatMap`](https://rescript-lang.org/apis/latest/belt/result#flatmap):
```reason
let flatMap: (t<'a, 'c>, 'a => t<'b, 'c>) => t<'b, 'c>
```

phated: Now that I'm looking at it, `flatMapOk` if the implementation that every other language use for their version of `flatMap`. Should we just rename that and delete this?

chriskrycho: The normal definitions, where `f` is the type wrapping the generics `a` and `b`, e.g. `Maybe(a)` or `Maybe(b)`:

- **`map`**: `f a -> (a -> b) -> f b` where `f` is the type wrapping the inner type
- **`flatMap`** (or `bind`)`: `f a -> (a -> f b) -> f b`

You'll see this exact same definition (with appropriate variation in syntax) across Haskell, Elm, OCaml, Rust, Swift, F♯… and this isn't arbitrary! It's because `map` and `flatMap` are well-defined interfaces (in category theory, the interfaces are named *functor* and *monad* respectively, and they really are just extremely-useful interfaces).

An example of how conflating these things gets weird: JavaScript's `Promise` chains and `await`, which are so difficult to type correctly that the TypeScript team is *still* working on better ways to plumb them through that type system, and it has more special cases than nearly any other part of the type system I can think of. That is so in part because `Promise` combines the `map` and `flatMap` interface—exactly the confusion 

Additionally, I'd argue that you *want* your core types to be well-defined type-wise (rather than relying on inference for them) because the inference for everything *else* needs to fall out correctly from these.

So for example with `Result`, `map` and `flatMap` would normally be constrained like this:

```ts
export let map = <t, u, e>(
  result: Result<t, e>,
  fn: (t: t) => u
): Result<u, e> => {
  match (result) {
    | Ok(v) => Ok(fn(v))
    | Err(e) => Err(e)
  }
}

export let flatMap = <t, u, e>(
  result: Result<t, e>,
  fn: (t: t) => Result<u, e>
): Result<u, e> => {
  match (result) {
    | Ok(v) => fn(v)
    | Err(e) => Err(e)
  }
}
```

This is because each of these defaults to one specific "path" through a set of variants for any given type with variants, e.g. `Option.map`/`Maybe.map` applies the function to the `Some`/`Just` case (for obvious reasons), `Result.map` to the `Ok` case, `Either` to the `Right` case, etc. Covering *both* scenarios with a function of that name is not only a bit unusual, it actually violates the interface! However, having the *additional* interfaces (e.g. as Rust supplies with names like `map_or_else`) is totally fine and *doesn't* violate those standard interfaces.

ospencer: @chriskrycho that's helpful and is in line with what @phated said.

Thus, I believe the API should look like this:

```
map : (a -> b, Result<a, e>) -> Result<b, e>
# formerly mapOk
# map(fn, r) is Ok(fn(a)) if r is Ok(a) and is r otherwise

mapErr : (e -> f, Result<a, e>) -> Result<a, f>
# mapErr(fn, r) is Err(fn(e)) if r is Err(e) and is r otherwise

mapWithDefault : (a -> b, b, Result<a, e>) -> b
# new function
# mapWithDefault(fn, default, r) is fn(a) if r is Ok(a) and is default otherwise

mapWithDefaultFn : (a -> b, e -> b, Result<a, e>) -> b
# formerly flatMap
# mapWithDefaultFn(fnOk, fnErr, r) is fnOk(a) if r is Ok(a) and fnErr(e) if r is Err(e)


flatMap : (a -> Result<b, e>, Result<a, e>) -> Result<b, e>
# formerly flatMapOk
# flatMap(fn, r) is fn(a) if r is Ok(a) and is r otherwise

flatMapErr : (e -> Result<a, f>, Result<a, e>) -> Result<a, f>
# flatMapErr(fn, r) is fn(e) if r is Err(e) and is r otherwise

# There is no flatMapWithDefault or flatMapWithDefaultFn, as these are both easily accomplished with mapWithDefault and mapWithDefaultFn, but it would not hurt to provide them.
```

This makes our API match that of OCaml/Rust/Elm/Belt & friends, while using (what I believe to be) the most clear names.

One thing to note: I've swapped the argument orders to put the data last. 🙂 

@c0d3d, I know that's quite a bit to ask for; if you'd like, I can make those changes.

chriskrycho: Looks perfect and I’m always a big fan of data-last!

compiler/test/stdlib/option.test.gr
```diff
@@ -124,3 +124,7 @@ assert c == 4
 let mut d = '🌾'
 assert Option.peek((x) => fail 'Should not be called', None) == None
 assert d == '🌾'
+
+# Option.toResult
+assert Option.toResult(Some(1), "Nope") == Ok(1)
```
phated, line 129: This didn't get switched to data-last, but I'll get this merged and follow up with a PR that changes it and adds the Option methods to mirror this PRs awesome additions (like `or` and `and`)

PR #259, opened 2020-08-15
Include program comments in program parsetree

ospencer:
This PR adds program comments to the parsetree. This will allow build tools that look at parsed programs to also read comments, for things like documentation or formatting.


phated:
While writing this, we ran into bugs when we made changes to `String_utils.slice`, so I'd like to try to set up inline testing for our utility methods before we call this ready

compiler/test/test_end_to_end.re
```diff
@@ -1283,47 +1284,83 @@ let string_tests = {
     tparse(
       "string_parse_dqs1",
       "\"foo\"",
-      {statements: [str("foo")], prog_loc: Location.dummy_loc},
+      {
```
belph, line 1287: @ospencer Maybe add a test case?

belph (changes requested):
Great stuff! LGTM; just one small request. Also, we should think about block comments soon.

ospencer:
@belph I dunno if you remember, but the plan is to just swap our comment syntax to C-style.

ospencer:
Made https://github.com/grain-lang/grain/issues/275 to track.

compiler/test/test_end_to_end.re
phated: Good call here. I am trying to write these now and it seems like a bunch of comments are being added to this block.

Investigating.

phated: I added a test for this in a separate block (see individual commit).

phated:
@ospencer @belph Alright, I believe I've added enough tests and fixed an issue I came across now. I'll re-request reviews from you both, so please review when you have a chance.

PR #268, opened 2020-08-26
Source Map Support

ospencer:
Source map support is here! 🎉 

Wasm source maps are currently supported in Chrome and Firefox. I'm going to look for some options for Node until there's support, like using `node --inspect`, but that's not important for this initial pass.

Here's what it looks like:

image

(The example program does `assert 1 == 2` on line 5 in `hello.gr`, and `assert` is defined in `pervasives.gr` on line 30.)

We can do some things to make it nicer, like using the original source names in the wasm output—but this is a great start.

Made #273 to track the above ☝️

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ _build/
 notes.org
 node_modules/
 *.wasm
+*.wasm.map
 *.wasm.wast
 *.wasm.modsig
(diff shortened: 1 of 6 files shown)


belph (approved):
🚀

PR #280, opened 2020-09-06
Add string concatenation to Pervasives

### LINKED ISSUE #279: Support string concatenation in Pervasives

### LINKED ISSUE #282: Add runtime wasm file
We want to start offloading much of our JS runtime into wasm functions, as well as support some usage for stdlib functions.

ospencer:
This adds string concatenation to Pervasives, which allows the user to concatenate strings without having to import/depend on the String library.

```grain
'Hello' ++ ', world'
```

As I mentioned in Discord, I'm not 100% sold on this syntax since with `let mut` we might want to have the `++` and `--` operators for use with numbers, and there's some potential ambiguous syntax if string concatenation also uses `++`.

We _also_ don't quite have the ability to use `^` like OCaml does, since we use that syntax for unboxing. (I also mentioned in Discord that I'm more or less okay with removing/changing the syntax for unboxing, since it's a lot less common now that we have `let mut`.)

Open to ideas for new syntax or another path forward; also fine just merging this and worrying about it when we actually want to support `++` and `--` for use with numbers.

Closes #279.
Closes #282.

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -33,7 +33,7 @@ let gensym_label = s => {
 };
 
 /* Number of swap variables to allocate */
-let swap_slots_i32 = [|Type.int32, Type.int32|];
+let swap_slots_i32 = [|Type.int32, Type.int32, Type.int32|];
 let swap_slots_i64 = [|Type.int64|];
 let swap_i32_offset = 0;
 let swap_i64_offset = Array.length(swap_slots_i32);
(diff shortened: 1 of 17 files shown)


compiler/src/codegen/compcore.re
```diff
@@ -33,7 +33,7 @@ let gensym_label = s => {
 };
 
 /* Number of swap variables to allocate */
-let swap_slots_i32 = [|Type.int32, Type.int32|];
+let swap_slots_i32 = [|Type.int32, Type.int32, Type.int32|];
```
ospencer, line 36: This seems fine since Binaryen will optimize away the unused slots.

phated (approved):
💯 💯 💯 💯 Love this!

I'm actually not a fan of `++` or `--` for incr/decr on mutable values because it is awkward for usage on mutable record properties, so I'm happy to use `++` for string concatenation.

belph (changes requested):
This is great! 

Before merging, though, I would to discuss why this was added as a mashtree implementation as opposed to a stdlib function. In my numbers branch, I am moving more in favor of cleaning up compcore.re such that it delegates out anything more complex than malloc (and sometimes even that) to appropriate stdlib functions. My understanding is that Binaryen should have our back in terms of inlining where appropriate (maybe that's wrong, though?).

@ospencer How would you feel about making this a function instead (even if there is hard-coded syntax, it the `compile_prim2` implementation could call the function)?

ospencer:
Summary from chat outside GH: this was removed as a stdlib function and built into the compiler to avoid Pervasives depending on another library/wasm file, which in the browser results in an extra network call or natively an extra disk read for *every* Grain program, regardless of your use of the `++` function. We plan to start offloading much of the runtime from JS to wasm anyway, so we figured we could start that effort now to avoid bloating `compcore.re` even more. 

#282 to track. This PR adds a set of runtime imports and imports the concat function from there.

@belph ready for you to take another look.

belph (approved):
Looks great. Thanks for the change!

PR #281, opened 2020-09-07
Exceptions and minimal support for open types

ospencer:
This PR adds the ability to make exceptions in Grain and perform pattern matching on them. It does not, however, allow the user to raise exceptions (that will come in a later PR) or catch exceptions (also in a later PR). It also adds some barebones support for open types to make exceptions work, and with some additional effort we could support open types in general.

Exceptions can be made like so:
```
exception Foo
exception Bar(String)
export exception Baz
```


compiler/src/typed/includemod.re
```diff
@@ -131,12 +134,19 @@ let type_declarations =
...
 
-/*let extension_constructors ~loc env ~mark cxt subst id ext1 ext2 =
-  let mark = mark_positive mark in
-  let ext2 = Subst.extension_constructor subst ext2 in
-  if Includecore.extension_constructors ~loc env ~mark id ext1 ext2
-  then ()
-  else raise(Error[cxt, env, Extension_constructors(id, ext1, ext2)])*/
+let extension_constructors = (~loc, env, ~mark, cxt, subst, id, ext1, ext2) => {
+  let mark = mark_positive(mark);
+  // let ext2 = Subst.extension_constructor(subst, ext2);
```
ospencer, line 139: @belph I spent a few hours looking into why this substitution doesn't work and continuously came up blank. It only matters because we do [this check](https://github.com/grain-lang/grain/blob/5b0739ccbfce28bcfcf18f09c26c4542f4801046/compiler/src/typed/typemod.re#L895) after typechecking. I'm not really sure what it's for, as Grain doesn't allow separate interface files so the inferred signature should always match the actual signature. Unless it's some sort of sanity check? But then, the error messages are pretty cryptic.

Regardless, this doesn't work and I'm not really sure if it needs to.

belph: Hmm, okay. I think it at least merits a TODO, if not an investigation issue, though. One of these days, we really need to properly fix up the typing module.

compiler/src/typed/types.re
```diff
@@ -71,10 +71,13 @@ and abbrev_memo =
   // Abbreviations can be found after this indirection
   | TMemLink(ref(abbrev_memo));
 
+/*true if a constant false if a block*/
```
belph, line 74: ADT might be better, idk

belph (changes requested):
Great stuff! Let's chat about a couple of little things, but this looks mostly ready to go.

compiler/src/typed/types.re
ospencer: Yeah, I had that thought, but didn't know if it was worth it. Do you want me to change it?

belph (approved):

compiler/src/parsing/ast_mapper.re
```diff
@@ -167,6 +168,29 @@ module D = {
   };
 };
 
+module Exc = {
+  let map = (sub, {ptyexn_constructor: ext, ptyexn_loc: loc}) => {
+    open Except;
+    let cloc = sub.location(sub, loc);
+    let {pext_name: n, pext_kind: k, pext_loc: loc} = ext;
+    let name = map_loc(sub, n);
+    let loc = sub.location(sub, loc);
```
phated, line 177: @ospencer I know this has already been merged, but why does `cloc` and `loc` use the same `sub.location(sub, loc);` call?

ospencer: @phated `loc` is shadowed here on line 175, but maybe that's not clear 😛

PR #286, opened 2020-09-09
Support pattern matching with guards

### LINKED ISSUE #283: Compiler: Support guards when pattern matching

ospencer:
Closes #283.

This PR adds the ability to "guard" a match pattern with a condition, like so:

```grain
data Example = Foo(Number) | Bar

match (Foo(6)) {
  | Foo(n) when n == 2 => print('it was two')
  | Foo(n) when n == 6 => print('it was six')
  | Foo(_) => print('it was not two or six')
  | Bar => print('it was bar')
}
```

### DIFF
--- a/compiler/src/middle_end/linearize.re
+++ b/compiler/src/middle_end/linearize.re
@@ -531,10 +531,7 @@ let rec transl_imm =
         transl_imm,
         exp_ans,
       );
-    (
-      Imm.id(~loc, ~env, tmp),
-      (exp_setup @ setup) @ [BLet(tmp, ans)],
-    );
+    (Imm.id(~loc, ~env, tmp), (exp_setup @ setup) @ [BLet(tmp, ans)]);
   | TExpConstruct(_) => failwith("NYI: transl_imm: Construct")
   }
 
--- a/compiler/src/middle_end/linearize.re
+++ b/compiler/src/middle_end/linearize.re
@@ -63,15 +63,41 @@ let lookup_symbol = (mod_, mod_decl, name, original_name) => {
   };
 };
 
-type anf_bind =
-  | BSeq(comp_expression)
-  | BLet(Ident.t, comp_expression)
-  | BLetRec(list((Ident.t, comp_expression)))
-  | BLetExport(rec_flag, list((Ident.t, comp_expression)));
-
 type either('a, 'b) =
   | Left('a)
   | Right('b);
+
+let convert_binds = anf_binds => {
+  let void_comp = Comp.imm(Imm.const(Const_void));
+  let void = AExp.comp(void_comp);
+  let (last_bind, top_binds) =
+    switch (anf_binds) {
+    | [bind, ...rest] => (bind, rest)
+    | _ => (BSeq(void_comp), [])
+    };
+  let ans =
+    switch (last_bind) {
+    | BSeq(exp) => AExp.comp(exp)
+    | [@implicit_arity] BLet(name, exp) =>
+      AExp.let_(Nonrecursive, [(name, exp)], void)
+    | BLetRec(names) => AExp.let_(Recursive, names, void)
+    | [@implicit_arity] BLetExport(rf, binds) =>
+      AExp.let_(~glob=Global, rf, binds, void)
+    };
+  List.fold_left(
+    (body, bind) =>
+      switch (bind) {
+      | BSeq(exp) => AExp.seq(exp, body)
+      | [@implicit_arity] BLet(name, exp) =>
+        AExp.let_(Nonrecursive, [(name, exp)], body)
+      | BLetRec(names) => AExp.let_(Recursive, names, body)
+      | [@implicit_arity] BLetExport(rf, binds) =>
+        AExp.let_(~glob=Global, rf, binds, body)
+      },
+    ans,
+    top_binds,
+  );
+};
 
 let extract_bindings = (mut_flag, pat, cexpr) => {
   let get_imm =
@@ -506,15 +532,12 @@ let rec transl_imm =
       MatchCompiler.compile_result(
         Matchcomp.convert_match_branches(branches),
         transl_anf_expression,
+        transl_imm,
         exp_ans,
       );
     (
       Imm.id(~loc, ~env, tmp),
-      (
-        exp_setup
-        @ List.map(((n, e)) => [@implicit_arity] BLet(n, e), setup)
-      )
-      @ [[@implicit_arity] BLet(tmp, ans)],
+      (exp_setup @ setup) @ [[@implicit_arity] BLet(tmp, ans)],
     );
   | TExpConstruct(_) => failwith("NYI: transl_imm: Construct")
   }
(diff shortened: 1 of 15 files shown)


compiler/src/middle_end/matchcomp.re
```diff
@@ -23,6 +22,13 @@ type decision_tree =
     Leaf(
       int,
     )
+  | /** Represents a guarded branch. The left tree corresponds to a successful
+      guard check, and the right tree corresponds to a failed guard check. */
+    Guard(
+      match_branch,
+      decision_tree,
+      decision_tree,
+    )
```
ospencer, line 31: This was simpler, but potentially this could/should be `pattern, expression, decision_tree, decision_tree` to guarantee that the `match_branch` actually has a guard, but I feel it's a non-issue since it's contained within this module.

compiler/src/middle_end/matchcomp.re
```diff
@@ -238,26 +244,59 @@ module MatchTreeCompiler = {
...
     switch (tree) {
     | Leaf(i) => (Comp.imm(Imm.const(Const_int(i))), [])
+    | Guard(mb, true_tree, false_tree) =>
+      let guard =
+        switch (mb.mb_guard) {
+        | Some(guard) => guard
+        | None =>
+          failwith(
+            "Impossible: (compile_tree_help): guarded match branch contained no guard",
+          )
```
ospencer, line 272: See my earlier point... though I honestly feel this could just be `assert(false)`.

compiler/src/middle_end/linearize.re
```diff
@@ -63,16 +63,42 @@ let lookup_symbol = (mod_, mod_decl, name, original_name) => {
...
   | Left('a)
   | Right('b);
 
+let convert_binds = anf_binds => {
+  let void_comp = Comp.imm(Imm.const(Const_void));
+  let void = AExp.comp(void_comp);
+  let (last_bind, top_binds) =
+    switch (anf_binds) {
+    | [bind, ...rest] => (bind, rest)
+    | _ => (BSeq(void_comp), [])
```
phated, line 76: Is there any need to match on `[bind]` and use an empty list for the top_binds?

compiler/src/middle_end/linearize.re
```diff
@@ -63,16 +63,42 @@ let lookup_symbol = (mod_, mod_decl, name, original_name) => {
...
+    };
+  let ans =
+    switch (last_bind) {
+    | BSeq(exp) => AExp.comp(exp)
+    | [@implicit_arity] BLet(name, exp) =>
+      AExp.let_(Nonrecursive, [(name, exp)], void)
```
phated, line 82:
```suggestion
    | BLet(name, exp) => AExp.let_(Nonrecursive, [(name, exp)], void)
```

phated: Reason doesn't need implicit_arity afaik. I know we haven't removed all of them from the codebase, but I think we should avoid it in new code.

compiler/src/middle_end/linearize.re
```diff
@@ -63,16 +63,42 @@ let lookup_symbol = (mod_, mod_decl, name, original_name) => {
...
+    | BLetRec(names) => AExp.let_(Recursive, names, void)
+    | [@implicit_arity] BLetExport(rf, binds) =>
+      AExp.let_(~glob=Global, rf, binds, void)
```
phated, line 85:
```suggestion
    | BLetExport(rf, binds) => AExp.let_(~glob=Global, rf, binds, void)
```

compiler/src/middle_end/linearize.re
```diff
@@ -63,16 +63,42 @@ let lookup_symbol = (mod_, mod_decl, name, original_name) => {
...
+    };
+  List.fold_left(
+    (body, bind) =>
+      switch (bind) {
+      | BSeq(exp) => AExp.seq(exp, body)
+      | [@implicit_arity] BLet(name, exp) =>
+        AExp.let_(Nonrecursive, [(name, exp)], body)
```
phated, line 92:
```suggestion
      | BLet(name, exp) => AExp.let_(Nonrecursive, [(name, exp)], body)
```

compiler/src/middle_end/linearize.re
```diff
@@ -63,16 +63,42 @@ let lookup_symbol = (mod_, mod_decl, name, original_name) => {
...
+      | BLetRec(names) => AExp.let_(Recursive, names, body)
+      | [@implicit_arity] BLetExport(rf, binds) =>
+        AExp.let_(~glob=Global, rf, binds, body)
```
phated, line 95:
```suggestion
      | BLetExport(rf, binds) => AExp.let_(~glob=Global, rf, binds, body)
```

compiler/src/middle_end/linearize.re
```diff
@@ -506,15 +532,12 @@ let rec transl_imm =
...
         exp_ans,
       );
     (
       Imm.id(~loc, ~env, tmp),
-      (
-        exp_setup
-        @ List.map(((n, e)) => [@implicit_arity] BLet(n, e), setup)
-      )
-      @ [[@implicit_arity] BLet(tmp, ans)],
+      (exp_setup @ setup) @ [[@implicit_arity] BLet(tmp, ans)],
```
phated, line 540:
```suggestion
      (exp_setup @ setup) @ [BLet(tmp, ans)],
```

phated: This needs a refmt but github won't let me alter the lines around it.

compiler/src/middle_end/matchcomp.re
phated: I am generally against `assert(false)`-type things. At least a failure message like "Impossible" might get someone to open an issue with "well actually!" so we can fix it.

compiler/src/typed/parmatch.re
```diff
@@ -1972,7 +1961,7 @@ let check_unused = (pred, casel) =>
     let rec do_rec = pref =>
       fun
       | [] => ()
-      | [{mb_pat: q, /*c_guard;*/ mb_body}, ...rem] => {
+      | [{mb_pat: q, mb_guard, mb_body}, ...rem] => {
```
phated, line 1964: It's so cool when some piece of commented code gets uncommented to add a killer feature! 🎉

phated (changes requested):
@ospencer This is sooooooo cool! However, I think we need to improve the compiler warning when a match isn't fully guarded. I tried this:

```
let a = Some(true);
match (a) {
  | Some(inner) when inner == true => print(a)
  | None => print(a)
}
```

And got this error:

```
File "test.gr", line 2, character 0-line 5, character 1:
Warning 8: this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
#$Some0(_)
(However, some guarded clause may match this value.)
Some(true)
```

Specifically, the `#$Some0(_)` seems wrong to show end users.

compiler/src/middle_end/linearize.re
ospencer: Yeah, it just feels weird to have code that doesn't match the surrounding code.

compiler/src/middle_end/linearize.re
ospencer: The `[bind, ...rest]` case should handle that. 🙂

ospencer:
@phated re: warnings, that wasn't introduced with this PR, and is just a general bug with printing of constructors for errors/warnings. I realized we didn't have an issue to track it though, so I made #291.

compiler/src/middle_end/linearize.re
phated: Ha, well I'll be. For some reason my brain thought that only matched lists with 2+ elements.

phated (approved):
I didn't try this out extensively, but my only test surfaced the exhaustiveness printing issue that is now fixed.

Amazing work!

PR #292, opened 2020-09-09
Fix printing of constructors for exhaustiveness checking

### LINKED ISSUE #291: Compiler: Fix pattern matching exhaustiveness checking errors
When pattern matching, if a match is not exhaustive, printing of constructors is pretty broken:
```grain
File "test.gr", line 2, character 0-line 4, character 1:
Warning 8: this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
#$Foo0()
```

ospencer:
Funny enough, I tried to show examples of this working, but there's another bug that I fixed in #286 that's preventing me from doing so in all cases, but once that's merged everything should be all set. Based on the code though, I'm sure you can probably see how this fixes the issue we were having. 😛 

Closes #291.


phated (approved):
You are a gentleman and a scholar! 💯

PR #294, opened 2020-09-09
Add operator precedence document

### LINKED ISSUE #287: Compiler: Establish Grain's operator precedence (and ensure we match that precedence)
We should create a table and then check it into source control. We should also probably add a section to the docs on it.

ospencer:
Closes #287. I'm going to open a few issues as well.

@phated added `isnt`.


belph (approved):
LGTM

phated:
Sorry for the delay here. @ospencer should we also add the string concat operator to the table?

ospencer:
I literally told myself to do that like five times and I didn't. Yes we should.

PR #296, opened 2020-09-10
Operator precedence adherence

### LINKED ISSUE #285: Compiler: Priority of `==` should be higher
`a == 2 && b == 3` parses as `equal(equal(and(2, b), 3))` but should instead parse as `and(equal(a, 2), equal(b, 3))`.

### LINKED ISSUE #295: Fix operator precedence according to our doc
The guidelines were introduced in #294.

ospencer:
Closes #295.
Closes #285.

This makes all of the infix operators adhere to the document. I _think_ everything else is mostly fine, but we can deal with them on a case-by-case basis.


stdlib/sys/file.gr
```diff
@@ -298,7 +298,7 @@ let fdStats = (fd) => {
 
   let mut rightsInheritingList = []
   List.forEachi((flag, i) => {
-    if (rightsInheriting && Int64.lsl(1L, i) > 0L) {
+    if ((rightsInheriting && Int64.lsl(1L, i)) > 0L) {
```
ospencer, line 301: (`&&` is shadowed here to be `&` for Int64s.)

This also makes me think that even if they aren't defined, we should still parse `&` & friends so they can be defined in a context like this.

ospencer: https://github.com/grain-lang/grain/issues/297

belph: half-lol @ "Good First Issue"

belph (approved):
Looks to be in order with the new document

phated:
This is a very cool PR. I like how the precedence table matches the relations in the parser. After reading both, I know understand the binop stuff quite a lot more! Thanks for putting all that together @ospencer!

PR #317, opened 2020-09-16
Add Docker Build

belph:
This PR adds a Dockerfile to build the Grain compiler and CLI. This is primarily aimed at Windows users who would like to try out or contribute to Grain while we come up with a more elegant Windows solution. Additionally, this PR will add a new workflow which publishes builds of the Grain compiler corresponding to each Grain Pull Request and push to `master`.

To build, run the following:
```bash
$ docker build .
```


ospencer (approved):
LGTM, though I can't really say much without trying it out (and I'll give it a go sometime later)

phated:
This needs a lockfile update.

phated:
Oh, and the opam files were changed but not the dune-project?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,30 @@
+name: Publish Docker image (HEAD)
+on:
+  push:
+    branches: [master]
+    tags: ['v*.*.*']
+  pull_request:
```
phated, line 6:
```suggestion

```

phated: We shouldn't run this on PR because it requires secrets. They won't be available to external contributors and this CI run will fail which will be a negative experience for those folks.

compiler/src/typed/env.re
```diff
@@ -2036,8 +2036,9 @@ let add_components =
             Tbl.add(m, x, acc);
           } else {
             assert(
+              // TODO: Shouldn't this be a `try`?
```
phated, line 2039: @ospencer @belph what's going on with this code?

compiler/src/typed/env.re
```diff
@@ -2229,8 +2230,9 @@ let open_signature_of_initially_opened_module =
     (~loc=Location.dummy_loc, root, env) => {
   let load_path = Grain_utils.Config.include_dirs^;
   let filter_modules = m =>
+    // TODO: Shouldn't this be a `try`?
```
phated, line 2233: And this?

phated (changes requested):
@belph Amazing work on this! I pushed some cleanup directly to this branch to fix some weirdness around some of the changes you needed for the docker build to work.

I only really want 1 change (to not run on PR), but I also ran into 2 questions related to the new refmt formatting.

compiler/src/typed/env.re
belph: Looks like it came from this commit (which you made): https://github.com/grain-lang/grain/pull/317/commits/c3133d25a797bfae252e0f8d6a6107f5e35a90ab

What led to you needing to make the change?

belph:
@phated Thanks for taking a look. I understand some of these changes you made, but not all of them. For example, I understand that we should keep the `.opam` files in sync with the `dune-project` file, but why was the fix to roll back the `.opam` files instead of rolling forward the `dune-project` one? Additionally, I don't understand what the problems were with the workflow file that was there; can you elaborate? Finally, how did you fix it so we don't need the additional calls to `esy install`? I want to understand what went wrong there.

compiler/src/typed/env.re
phated: Yes, I added this comment because I needed you and @ospencer to look at the code right below it. When you upgraded the Reason to latest master, it removed the parens surrounding the annotated type and forced me to look at this code to figure out what was going on.

As you'll notice, this code is using a `switch` statement with a catch-all in the first match, so nothing else will fallthrough. Further, you'll see there is a pattern match that is trying to match an exception, which I'm pretty sure only works with a `try` statement, and not a `switch` statement.

So I wanted you both to look at it and figure out if there are other problems in this code, and I added the TODO as a point of reference in this PR.

phated:
Cool, cool. I'll explain below.

> I understand that we should keep the .opam files in sync with the dune-project file, but why was the fix to roll back the .opam files instead of rolling forward the dune-project one?

The published version of Reason on opam is 3.6.0, we should keep the range to include the proper published version and not some theoretical next release. The only time we should stray from this is if something isn't published on opam, then we can reference it with `>= 0.0.0`. We can then use esy resolutions to do the "patching" we need to get the proper dependencies to build the project.

> Additionally, I don't understand what the problems were with the workflow file that was there; can you elaborate?

The workflow file was a combination of examples from docker-push-build v1 and v2. This was brought to my attention by warnings in the Action runs that said some properties were invalid. I started doing research and found that using the v2 (still in beta) would require a lot more changes and additional build steps to log into container registries and stuff. Instead of doing that, I just removed any of the stuff you got from v2 examples until we upgrade it.

>  Finally, how did you fix it so we don't need the additional calls to esy install?

Esy needs the build tree to be immutable. When we generate our opam files with dune, that's not the case (if there is a change). So when you made manual changes to the opam files, dune tried to regenerate them every time you would run an esy command because they were different. If you need to make a change to the opam dependencies, you need to change the `dune-project` file, run `esy build` once (which will fail) and then commit the opam files. Once the files match the `dune-project`, they won't be regenerated and the tree will be immutable again. Further information on this issue is at https://github.com/esy/esy/issues/1104

compiler/src/typed/env.re
ospencer: You can catch exceptions with `switch`/`match` statements like this. In this case since there's a wildcard branch, it's just acting like a regular `try`. It's a bit different though since with a try, you'd have to write it like this to be equivalent:

```re
try (IdTbl.find_name(...); true) { _ => false }
```

ospencer: or actually
```re
try (ignore(IdTbl.find_name(...)); true) { _ => false }
```

belph: Yeah, what he said :)
So, @phated, are we good here?

belph:
@phated Sorry for the delay in responding to this:
> We can then use esy resolutions to do the "patching" we need to get the proper dependencies to build the project.
I think that's why I was a bit confused; one thing this PR adds is an esy resoluton to 3.6.1...

> docker-push-build
Okay, I understand. Will the version you pushed still tag the images as appropriate, similar to what the code I had checked in was doing?

> opam files
Got it, understood.

compiler/src/typed/env.re
phated: I think we should just remove the TODOs

phated, line 2039:
```suggestion

```

phated:
> I think that's why I was a bit confused; one thing this PR adds is an esy resoluton to 3.6.1...

But `>= 3.6.0` includes 3.6.1 😛 

> Okay, I understand. Will the version you pushed still tag the images as appropriate, similar to what the code I had checked in was doing?

I'm not sure, but I think that's what `tag_with_ref: true` is doing. You can look at the artifact that was pushed when I got the build actually working.

compiler/src/typed/env.re
belph: Done.

belph:
Okay, it looks like the image tagging works still. @phated Are we good to merge now?

phated (approved):

phated:
:shipit:

Issue #310, opened 2020-09-14
Stdlib: Implement List.drop and List.dropWhile


ospencer:
`List.drop` should take a number `n` and a list `l`, and return `l` with the first `n` elements removed. If `n` is greater than `List.length(l)`, return an empty list.

`List.dropWhile` takes a function instead of `n`.  The function takes one argument, an element of the list. `List.dropWhile` removes elements of the list until the given function returns `false`.

tatchi:
Hey, I would be happy to give it a shot (if ok for you ofc)

ospencer:
@tatchi Of course! Thanks a ton. Let us know in Discord if there's anything we can do to help out 😄

ospencer:
Completed in #318

Issue #284, opened 2020-09-08
Compiler: Support constant patterns in pattern matching


ospencer:

phated:
What is a constant pattern?

ospencer:
It's being able to do like 
```
match (...) { 
  | 5 => print('five') 
  | 6 => print('six') 
  | _ => print('not five or six') 
}
```

It's sorta syntactic sugar over `| a when a = 5 => ...` (though I suppose you could implement it directly without it being sugar).

PR #318, opened 2020-09-17
Stdlib: add drop and dropWhile

tatchi:
Implement #310

### DIFF
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -195,3 +195,25 @@ let unique = (list) => {
   }
   iter(list, [])
 }
+
+let rec drop = (n, list) => {
+  if (n < 0) {
+    fail 'number of items to drop cannot be a negative number'
+  } else {
+    match ((n, list)) {
+      | (_,[]) => []
+      | (n, _) when n == 0 => list
+      | (n, [first, ...rest]) => drop(n - 1, rest)
+    }
+  }
+}
+
+let rec dropWhile = (fn, list) => {
+  match (list) {
+    | [] => list
+    | [first, ...rest] => if (fn(first)) dropWhile(fn, rest) else list
+  }
+}
+
+
+
--- a/compiler/test/stdlib/list.test.gr
+++ b/compiler/test/stdlib/list.test.gr
@@ -127,3 +127,19 @@ assert rotate(-2, list) == [2, 3, 1]
 assert unique([]) == []
 assert unique(list) == list
 assert unique([1, 1, 1, 1]) == [1]
+
+# List.drop
+
+assert drop(4, []) == []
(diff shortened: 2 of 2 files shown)


stdlib/list.gr
```diff
@@ -195,3 +195,25 @@ let unique = (list) => {
...
   iter(list, [])
 }
+
+let rec drop = (n, list) => {
+  if (n < 0) {
+    fail 'number of items to drop cannot be a negative number'
+  } else {
+    match ((n, list)) {
+      | (_,[]) => []
+      | (n, _) when n == 0 => list
```
tatchi, line 205: Looks like we can't pattern match on number so I used a guard here.

stdlib/list.gr
```diff
@@ -195,3 +195,25 @@ let unique = (list) => {
...
+    }
+  }
+}
+
+let rec dropWhile = (fn, list) => {
+  match (list) {
+    | [] => list
+    | [first, ...rest] => if (fn(first)) dropWhile(fn, rest) else list
+  }
+}
```
tatchi, line 216: I'm not sure this one is tail-recursive but from what I saw, others from this file like `count` are not. I don't know at which point it matters?

stdlib/list.gr
belph: @ospencer Is this a bug or a TODO?

stdlib/list.gr
belph: Hmm, I think we want to move towards making both this and `count` tail-recursive. @ospencer should we make a separate issue for this, or do we want to include making this/these tail-recursive inside of this PR?

belph (commented):
This is great stuff, thank you! I've requested a review from @ospencer so that we can address the comments you've left, but I have no issues merging this once those are ironed out.

stdlib/list.gr
ospencer: We already have an issue: https://github.com/grain-lang/grain/issues/284

stdlib/list.gr
ospencer: Maybe it's really early for me, but is this not tail-recursive? The call to `dropWhile` is in the tail position of the true branch of a tail position `if`.

`count` does need to be changed though. Made https://github.com/grain-lang/grain/issues/319.

ospencer (approved):
Thanks @tatchi!

stdlib/list.gr
belph: Oh, yes, lol. I missed that; you're right.

tatchi: Thanks for the review, I was indeed not sure for the `dropWhile` :)

Issue #319, opened 2020-09-17
Stdlib: List.count should be tail-recursive


ospencer:

tatchi:
I can pick this one up if you agree :)

ospencer:
@tatchi That'd be great, thanks!

ospencer:
Completed in #320

PR #320, opened 2020-09-17
Stdlib: make sum, length and count tail-recursive

tatchi:
Implement #319 and makes also `length` and `sum` tail-recursive


ospencer (approved):
Looks great! Thanks @tatchi! 😄

PR #322, opened 2020-09-18
Fix parsing of negative numbers

### LINKED ISSUE #314: Compiler: Fix parsing of 1-1
This should parse as `minus(1, 1)` but instead fails to parse (probably because it's parsing `-1` as negative 1).

ospencer:
Fixes #314.

### DIFF
--- a/compiler/test/test_end_to_end.re
+++ b/compiler/test/test_end_to_end.re
@@ -163,6 +163,11 @@ let basic_functionality_tests = [
   t("binop1", "2 + 2", "4"),
   t("binop2", "2 - 2", "0"),
   t("binop2.1", "2-2", "0"),
+  t("binop2.2", "2 -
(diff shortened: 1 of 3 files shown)


compiler/test/test_end_to_end.re
```diff
@@ -162,6 +162,7 @@ let basic_functionality_tests = [
   t("complex2", "print(2 + 3)", "5\nvoid"),
   t("binop1", "2 + 2", "4"),
   t("binop2", "2 - 2", "0"),
+  t("binop2.1", "2-2", "0"),
```
belph, line 165: Please add tests for the cases with newlines which we discussed on Discord (and a reminder about opening that issue).

belph (changes requested):
LGTM; just one comment

belph (approved):

PR #327, opened 2020-09-21
Literal rational literals

ospencer:
This PR bakes in support for rational literals instead of only creating them via division. This will make it possible to do pattern matching on rational constants (as well as be slightly more performant than having to perform the division/factoring at runtime).

Requires #326.

### DIFF
--- a/compiler/src/typed/parmatch.re
+++ b/compiler/src/typed/parmatch.re
@@ -1757,8 +1757,8 @@ let untype_constant =
   | Const_number(Const_number_rational(n, d)) =>
     Parsetree.PConstNumber(
       Parsetree.PConstNumberRational(
-        Int64.to_string(n),
-        Int64.to_string(d),
+        Int32.to_string(n),
+        Int32.to_string(d),
       ),
     )
   | Const_string(s) => Parsetree.PConstString(s)
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -2107,6 +2107,10 @@ let allocate_int64 = (wasm_mod, env, i) => {
   call_new_int64(wasm_mod, env, [i]);
 };
 
+let allocate_rational = (wasm_mod, env, n, d) => {
+  call_new_rational(wasm_mod, env, [n, d]);
+};
+
 /* Store an int64 */
 let allocate_int64_imm = (wasm_mod, env, i) => {
   let get_swap64 = () => get_swap(~ty=Type.int64, wasm_mod, env, 0);
@@ -2900,6 +2904,13 @@ let compile_allocation = (wasm_mod, env, alloc_type) =>
       wasm_mod,
       env,
       Expression.const(wasm_mod, Literal.float64(i)),
+    )
+  | MRational(n, d) =>
+    allocate_rational(
+      wasm_mod,
+      env,
+      Expression.const(wasm_mod, Literal.int32(n)),
+      Expression.const(wasm_mod, Literal.int32(d)),
     )
   };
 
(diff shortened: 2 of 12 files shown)


ospencer:
Just realized that I didn't add parsing rules for negative rationals. Whoops. Though this _also_ made me realize that we have no tests for negative rationals! Will resolve after some sleep.

compiler/src/typed/parmatch.re
```diff
@@ -1739,8 +1739,8 @@ let untype_constant =
   | Const_number(Const_number_rational(n, d)) =>
     Parsetree.PConstNumber(
       Parsetree.PConstNumberRational(
-        Int64.to_string(n),
-        Int64.to_string(d),
+        Int32.to_string(n),
```
belph, line 1742: We should probably open an issue to expand this to i64s.

belph (commented):
I'm not very enthusiastic about this. I can see why we would need it to make pattern-matching work, but, if it were not for that, I would really want it to be something that is produced by the optimization passes, rather than at parse time. @ospencer , do you think that's possible, or does pattern matching predicate the need for this approach?

ospencer:
@belph It's _possible_ (though then the parsing code gets a little gross). I think the real question is this: is `1/3` a rational literal or is it `1` divided by `3`? If it's a rational literal, then this implementation would be a more accurate representation of the program. If it's division, then producing it during optimization would be correct. The major consideration is what happens during shadowing:

If `/` is shadowed, `1` divided by `3` would be different than the rational literal.
If it's division, that means that `1/3` is not a constant, and therefore you cannot pattern-match on it. If we let you pattern-match on it anyway, that means that in the context of pattern matching it _is_ a rational literal, but that's a little awkward since if `/` is shadowed, you'd get different results while matching within the same scope. If we respected the shadowing and actually performed the division for a match, we're now special-case allowing pattern matching on a compound expression.

It'd be sorta weird to have first-class support for rationals and not allow pattern matching on them, and I think it would be confusing to let actual division be a valid match pattern when you can't do that for anything else (+ having to document/explain that this one case is division and not a rational literal).

We can support something somewhere in there, but I thought we wanted actual rational literals + pattern matching on rational literals.

belph:
First, let me say that we almost certainly want pattern matching on rational literals; it will be confusing to users if they cannot match those but they can match any other number which gets printed by Grain. 

> I think the real question is this: is 1/3 a rational literal or is it 1 divided by 3?

In my mental model, it was the latter. That being said, the precedent from Racket is it being the former. However, I do find it somewhat counterintuitive that we have the following, due to Grain operators (unlike Racket's) being infix:
```grain
# Works:
let (+) = (-)
assert (2) + (2) == 0
assert 2+2 == 0

# Does not work:
let (/) = (*)
assert (2) / (2) == 4
assert 2/2 == 4 # <-- boom
```

So I'm not sure what the correct solution is here. We need to sacrifice one of these two competing objectives: the ability to shadow properly or the ability to pattern match...

ospencer:
I'm actually okay with this:
```grain
# Does not work:
let (/) = (*)
assert (2) / (2) == 4
assert 2/2 == 4 # <-- boom
```
as long as `<int literal> / <int literal>` is well-defined as a rational literal, which we can make clear with documentation.

It should be pretty rare that `/` is ever shadowed and still used with the `Number` type (you'd be asking for trouble anyway).
More commonly you'd see something like this:
```grain
let (/) = Int64.div
assert (2L) / (2L) == 1L
assert 2L/2L == 1L
```

which would work as expected.

belph:
I agree that shadowing `/` is asking for trouble, but I am still worried about the inconsistency between `/` and the other binops...I suppose it makes sense, but it still feels a little unsettling to me.

That being said, in the absence of a suitable alternative, I am going to go ahead and approve this PR. The mention inconsistencies are definitely something for us to think about, though.

belph (approved):

PR #331, opened 2020-09-23
Implement pattern matching on constants

### LINKED ISSUE #284: Compiler: Support constant patterns in pattern matching

ospencer:
This PR adds the ability to do pattern matching on constant patterns:
```grain
match (('foo', 5, true)) {
  | ('bar', _, false) => false
  | ('foo', 5, _) => true
  | _ => false
}
```

Ended up going with the desugaring into guards (after typechecking) approach since I believe it is more performant than checking constants on the fly (though this seems to only be true for matches with many cases with constant patterns—I believe it's faster to check them on the fly if there are few cases with constant patterns).

Tests are light since they're just exercising already-tested guarded branches.

Closes #284.

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -64,6 +64,7 @@ let float32_to_number_ident =
   Ident.create_persistent("coerceFloat32ToNumber");
 let float64_to_number_ident =
   Ident.create_persistent("coerceFloat64ToNumber");
+let equal_ident = Ident.create_persistent("equal");
 /* Variants used for tracing */
 let incref_adt_ident = Ident.create_persistent("incRefADT");
 let incref_array_ident = Ident.create_persistent("incRefArray");
@@ -423,6 +424,14 @@ let runtime_function_imports =
         mimp_mod: stdlib_external_runtime_mod,
         mimp_name: float64_to_number_ident,
         mimp_type: [@implicit_arity] MFuncImport([I32Type], [I32Type]),
+        mimp_kind: MImportWasm,
+        mimp_setup: MSetupNone,
+      },
+      {
+        mimp_mod: stdlib_external_runtime_mod,
+        mimp_name: equal_ident,
+        mimp_type:
+          [@implicit_arity] MFuncImport([I32Type, I32Type], [I32Type]),
         mimp_kind: MImportWasm,
         mimp_setup: MSetupNone,
       },
@@ -892,6 +901,13 @@ let call_float64_to_number = (wasm_mod, env, args) =>
   Expression.call(
     wasm_mod,
     get_imported_name(stdlib_external_runtime_mod, float64_to_number_ident),
+    args,
+    Type.int32,
+  );
+let call_equal = (wasm_mod, env, args) =>
+  Expression.call(
+    wasm_mod,
+    get_imported_name(stdlib_external_runtime_mod, equal_ident),
     args,
     Type.int32,
   );
@@ -2846,7 +2862,8 @@ let compile_prim2 = (wasm_mod, env: codegen_env, p2, arg1, arg2): Expression.t =
   | LessEq
   | Int64Lte =>
     failwith("Unreachable case; should never get here: LessEq/Int64Lte")
-  | Eq =>
+  | Eq => call_equal(wasm_mod, env, [compiled_arg1(), compiled_arg2()])
+  | Is =>
     // Physical equality check
     encode_bool(
       wasm_mod,
--- a/compiler/src/middle_end/matchcomp.re
+++ b/compiler/src/middle_end/matchcomp.re
@@ -616,19 +616,19 @@ let rec default_matrix = mtx => {
   );
 };
 
-let rec compile_matrix = (branches, mtx) =>
+let rec compile_matrix = mtx =>
   switch (mtx) {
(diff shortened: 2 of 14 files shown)


compiler/src/middle_end/matchcomp.re
```diff
@@ -616,19 +616,19 @@ let rec default_matrix = mtx => {
   );
 };
 
-let rec compile_matrix = (branches, mtx) =>
+let rec compile_matrix = mtx =>
```
ospencer, line 619: I cleaned this up so we don't have to pass the branches through. That was originally necessary because we assigned branch numbers along the way, but some time ago we switched to assigning each branch its own number before compiling.

ospencer: Otherwise, this function hasn't changed.

compiler/src/middle_end/matchcomp.re
```diff
@@ -705,18 +696,123 @@ let rec compile_matrix = (branches, mtx) =>
...
+      | Const_int32(_) => Builtin_types.type_int32
+      | Const_int64(_) => Builtin_types.type_int64
+      | Const_float32(_) => Builtin_types.type_float32
+      | Const_float64(_) => Builtin_types.type_float64
+      | Const_string(_) => Builtin_types.type_string
+      };
+    };
+    let add_constant_guard = (id, c, loc) => {
+      let vd = make_vd(id, type_constant(c));
+      let ty = type_constant(c);
```
phated, line 734:
```suggestion
      let ty = type_constant(c);
      let vd = make_vd(id, ty);
```

compiler/test/test_end_to_end.re
```diff
@@ -1053,6 +1054,17 @@ let match_tests = [
...
+  t(
+    "constant_match_1",
+    "match (1/3) { | 5 => false | 2/6 => true | _ => false }",
+    "true",
+  ),
+  t(
+    "constant_match_2",
+    "match (('foo', 5, false)) { | ('bar', 5, false) => false | ('foo', _, true) => false | ('foo', _, false) => true | _ => false }",
+    "true",
+  ),
```
phated, line 1067: I'd like to add some tests that match constants with additional guards added.

ospencer: Good call.

compiler/src/codegen/compcore.re
```diff
@@ -2846,7 +2862,8 @@ let compile_prim2 = (wasm_mod, env: codegen_env, p2, arg1, arg2): Expression.t =
   | LessEq
   | Int64Lte =>
     failwith("Unreachable case; should never get here: LessEq/Int64Lte")
-  | Eq =>
+  | Eq => call_equal(wasm_mod, env, [compiled_arg1(), compiled_arg2()])
```
peblair, line 2865: This does something? I thought pervasives exports the function binding, not the primitive.

compiler/src/middle_end/matchcomp.re
peblair: 🙏

compiler/src/typed/parmatch.re
```diff
@@ -1761,9 +1761,13 @@ let untype_constant =
...
       ),
     )
+  | Const_int32(i) => Parsetree.PConstInt32(Int32.to_string(i))
+  | Const_int64(i) => Parsetree.PConstInt64(Int64.to_string(i))
+  | Const_float32(f) => Parsetree.PConstFloat32(Float.to_string(f))
+  | Const_float64(f) => Parsetree.PConstFloat64(Float.to_string(f))
   | Const_string(s) => Parsetree.PConstString(s)
   | Const_bool(b) => Parsetree.PConstBool(b)
-  | _ => failwith("NYI untype_constant");
+  | Const_void => Parsetree.PConstVoid;
```
peblair, line 1770: 🚀

compiler/src/middle_end/matchcomp.re
```diff
@@ -705,18 +696,123 @@ let rec compile_matrix = (branches, mtx) =>
           "Internal error: convert_match_branches: no non-wildcard column found but not all patterns always match (should be impossible)",
         )
       };
-    let {tree, _} as result = compile_matrix(branches, rotated);
-    {...result, tree: [@implicit_arity] Swap(i, tree)};
+    let result = compile_matrix(rotated);
+    Swap(i, result);
+  };
+
+let prepare_match_branches = branches => {
```
peblair, line 703: So all this function does is massage constant patterns into being guards? Either way, might merit a comment

peblair (approved):
Looks good! Do take a look at my comment before merging, though, please.

compiler/src/codegen/compcore.re
ospencer: You're correct, but we can't do like `call("==", arg1, arg2)` because if someone shadowed `==`, pattern matching would break 😂 

So doing it this way guarantees that we always call the real equals function rather than what `==` is bound to.

ospencer: (Because the desugaring does `Primitive2(Eq, generated_var_pat, constant)`.)

compiler/src/middle_end/matchcomp.re
ospencer: Yup. I can add a comment.

PR #332, opened 2020-09-23
Update docker working directory and default command

ospencer:
This makes it so if you `docker run` without any command, you're in a shell where `grain` works rather than the node repl. It also sets the default directory to the Grain repo rather than just `/`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,5 +39,5 @@ RUN yarn workspace @grain/compiler esy compile
 RUN yarn workspace @grain/compiler esy copy-compiler
 
 # Set up container environment
-WORKDIR /grain
+WORKDIR /
 CMD [ "/bin/bash" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,4 +39,5 @@ RUN yarn workspace @grain/compiler esy compile
 RUN yarn workspace @grain/compiler esy copy-compiler
 
 # Set up container environment
-WORKDIR /
+WORKDIR /grain
+CMD [ "/bin/bash" ]


Dockerfile
```diff
@@ -39,4 +39,5 @@ RUN yarn workspace @grain/compiler esy compile
 RUN yarn workspace @grain/compiler esy copy-compiler
 
 # Set up container environment
-WORKDIR /
+WORKDIR /grain
```
phated, line 42: Why do we want the working dir to be the grain project itself if we want the default command to be executing grain?

belph:
I'm confused by this. `grain` should be on the PATH inside of the image, so it shouldn't matter what directory you're in. Furthermore, we've checked in the past that you can do `docker run -ti grainlang/grain:pr-317-merge grain -h`, which means that the `grain` command works from the existing directory.

belph:
Clarification: I'm confused why we want to change the working directory

ospencer:
Yes, the working directory doesn't matter. That said, the repo being the working directory is more useful to contributors than randomly being at the root of the container.

With these changes, if you decide to `docker run -it grainlang/grain`, you get a `bash` shell in the directory of the repo where you can contribute immediately.

(I'm writing docs on how to use the docker container, and found it really odd that running the above command drops you into a `node` repl in the root directory.)

phated:
Thinking about it some more, I don't actually think this docker image is useful for contributors right now. What we should be doing if we want this to help contributors is linking the grain directory into the docker image instead of copying it.

ospencer:
I think you're right, but I see two different Dockerfiles for those cases: one where you download the image so you can just use the project immediately, and another where you can just `docker build .` and run all of your build commands from within the container.

With what we have now, this is at least a start where you can work from within the container. But it shouldn't be too hard to set up the other file (and this one can inherit from the other one).

belph:
I really don't think we should encourage people to work from within the container, since it's a totally separate environment. Instead, I think we should primarily use it for the `docker build .` command on Windows, with contributions being done on the host machine. As such, I don't think there's any utility in directing the pwd to be the project root (since people shouldn't be playing in there anyway). If you wanted to use Grain in a docker environment, you could use the image to have the compiler available, but that's pretty much the only practical use outside of building for the Docker image.

ospencer:
Sure. I can change it back, but then we'll need someone to make the appropriate updates to actually use this with `docker build` pretty soon.

phated (approved):
👍 I'm good with this now. It definitely seems weird to stick someone into the node repl.

peblair (approved):

PR #340, opened 2020-09-27
Provide type-specific operator methods

ospencer:


phated (approved):
👍 Looks good to me, but I'm still a noob with the number stuff. Would be good to get @peblair's review too.

ospencer:
I'm going to merge to get the bug fixes in. If there's anything that needs to be changed, we can do them in a follow-up PR 👍

PR #342, opened 2020-10-01
Update Map type

ospencer:
This makes the Map type referenceable as `Map.Map<k, v>` (tbd on if it should be `Map.T`) instead of `Map.Container<Map.Bucket<k, v>>`.

Side note: I would love syntax for optional types, like `Array<Bucket<k, v>?>` or `Array<?Bucket<k, v>>` (and also shorthands for list and maybe array types)!

### DIFF
--- a/stdlib/map.gr
+++ b/stdlib/map.gr
@@ -4,16 +4,16 @@ import List from 'list'
 import Array from 'array'
 import { hash } from 'hash'
 
-data Container<a> = {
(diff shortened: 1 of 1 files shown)


phated (approved):
I'm conflicted on shorthand for Option in the type signature. Maybe we should open an issue to get feedback from the community?

Anyway, this is good to go!

PR #344, opened 2020-10-03
Add Map.filter and Map.reject (fixes #193)

### LINKED ISSUE #193: Stdlib: Add Map.filter & Map.reject
It would be helpful if you could filter/reject a Map in-place. I imagine the API to be something like:

```
let map = Map.fromList([('foo', 1), (bar', 2)]);

Map.filter((key, val) => val == 1, map);
```

I'm unsure if it should use a Boolean or an Option to determine the filters. @ospencer @belph which do you prefer?

ohana54:
Hi, this is my first contribution so do let me know if anything's wrong.

I'd like to say the contributing guide was great and everything worked right from the start.
The [understandable] weaker points are stack traces / error messages and test assertions that don't give much info why something failed.

### DIFF
--- a/stdlib/map.gr
+++ b/stdlib/map.gr
@@ -279,6 +279,20 @@ export let fromArray = (array) => {
   map
 }
 
+export let filter = (predicate, map) => {
+  let filteredMap = make();
+  forEach((key, value) => {
+    if (predicate(key, value)) {
+      set(key, value, filteredMap);
+    }
+  }, map);
+  filteredMap
+}
+
+export let reject = (predicate, map) => {
+  filter((key, value) => !predicate(key, value), map)
+}
+
 # TODO: Should return a Record type instead of a Tuple
 # Waiting on https://github.com/grain-lang/grain/issues/190
 export let getInternalStats = (map) => {
--- a/stdlib/map.gr
+++ b/stdlib/map.gr
@@ -280,11 +280,15 @@ export let fromArray = (array) => {
 }
 
 export let filter = (predicate, map) => {
-  forEach((key, value) => {
+  let keysToRemove = reduce((list, key, value) => 
     if (!predicate(key, value)) {
-      remove(key, map);
-    }
-  }, map)
+      [key, ...list]
+    } else {
+      list
+    }, [], map);
+  List.forEach((key) => {
+    remove(key, map);
+  }, keysToRemove);
 }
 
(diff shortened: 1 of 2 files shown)


stdlib/map.gr
```diff
@@ -279,6 +279,20 @@ export let fromArray = (array) => {
...
 }
 
+export let filter = (predicate, map) => {
+  let filteredMap = make();
+  forEach((key, value) => {
+    if (predicate(key, value)) {
+      set(key, value, filteredMap);
+    }
+  }, map);
+  filteredMap
```
ospencer, line 289:
```suggestion
  forEach((key, value) => {
    if (!predicate(key, value)) {
      remove(key, map);
    }
  }, map)
```

ospencer (changes requested):
Sweet, thanks for taking this on! 😄 

This looks awesome, though these two methods should filter/reject in place (though I could definitely see having APIs to do exactly this). Normally this is exactly how you'd want to do it, though, and your code is perfect 👌 

Also, thanks for the feedback! We do have source map support when running in the browser, but haven't yet figured out a way to get it working with Node, unfortunately. I've been noodling on some ideas, but nothing concrete yet (other than trying to add wasm source map support directly to Node, and that sounds... hard).

ohana54:
Thanks, I should've asked whether it should be an immutable implementation or not :)
I automatically accepted the changes which obviously failed the tests, fixing it now.

ospencer (approved):
Looks great! Will merge as soon as the CI passes :)

ohana54:
Thanks, 2 notes about the implementation:
1. Removing elements inside a `forEach` caused the program to hang, I don't know if it's considered a limitation of the API or a bug that should be fixed.
2. In the tests, when I tried creating the `filterTestMap` using the following function the program hanged:
```
let makeFilterTestMap = () => {
  let map = Map.make();
  Map.set(Grain, "g", map);
  Map.set(Sheep, "s", map);
  Map.set(Brick, "b", map);
  map
}
```
Any idea why?

ospencer:
Also, thanks for matching the style of the tests. Made https://github.com/grain-lang/grain/issues/349 to clean up :)

ohana54:
I'll go ahead and tackle #349  :)

ospencer:
Oh, interesting. I'll make an issue for us to look into that. Thanks for letting me know!

PR #345, opened 2020-10-03
Slightly better node source map support

ospencer:
This makes it so if you run Grain via `node --inspect`, you can see a properly source-mapped call stack in the Chrome Node Debugger:

image

You can even set breakpoints and they work:

image

This doesn't fix the issue with the mappings not showing in the actual stack trace, though, since those are generated by Node and not the Chrome debugger. 😞 

The mappings still work properly if you're running directly in Chrome, though.

### DIFF
--- a/runtime/package.json
+++ b/runtime/package.json
@@ -47,6 +47,7 @@
     "@wasmer/wasi": "^0.11.2",
     "@wasmer/wasmfs": "^0.11.2",
     "fast-text-encoding": "^1.0.0",
-    "treeify": "^1.1.0"
+    "treeify": "^1.1.0",
+    "wasm-sourcemap": "^1.0.0"
   }
 }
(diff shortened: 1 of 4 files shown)


ohana54:
I don't know enough about the subject, but maybe you can hook into the node stack trace in a similar manner to https://www.npmjs.com/package/source-map-support
We use it and it works great with a bundled Typescript cli stack traces.

ospencer:
I've looked at this a bit, and couldn't really find much of anything to help. I think that Node itself needs better support for it unfortunately :/

peblair (approved):
@phated should maybe weigh in as well, but this looks good to me.

ospencer:
We can def make more changes if we realize another way to move forward!

phated:
Fwiw, I think this is a solid solution for now. 👍 Thanks for digging deeper on this @ospencer!

PR #351, opened 2020-10-03
Clean up usage of `== false` and `== true` (fixes #349)

### LINKED ISSUE #349: Refactor map tests to use negation (rather than == false)
In https://github.com/grain-lang/grain/blob/master/compiler/test/stdlib/map.test.gr, the tests should `assert !thing` rather than `assert thing == false`.

ohana54:


ospencer (approved):
Sweet, thank you!

ohana54:
Seems like the CI failure is on a sporadic network issue, although I think I don't have permissions to re-run it.

phated:
fwiw, I don't like this syntax. I opened #386 so we can add a proper assertion library.

PR #353, opened 2020-10-03
Add `List.find`, `List.findIndex`

ohana54:
Note I'm not familiar with how to catch exceptions, so I don't know how to test the failure cases.


ospencer (approved):
Looks great! We don't have exception handling just yet (pending https://github.com/WebAssembly/exception-handling) so no need to test that just yet.

PR #354, opened 2020-10-03
Add `List.flatMap` (fixes #352)

### LINKED ISSUE #352: [Stdlib] Add List.flatMap
Here's `flatMap` signature:

```grain
flatMap : (a -> List<b>, List<a>) -> List<b>
```

ohana54:


compiler/test/stdlib/list.test.gr
```diff
@@ -41,6 +41,11 @@ let multiply = (n, i) => n * i
 assert mapi(multiply, numbers) == [0, 3, 8]
 assert mapi(multiply, []) == []
 
+# List.flatMap
+
+assert flatMap(x => [x + 1], list) == [2, 3, 4]
```
ospencer, line 46: Would maybe be good to also test having multiple items in the list:
```suggestion
assert flatMap(x => [x + 1], list) == [2, 3, 4]
assert flatMap(x => [x, x + 1], list) == [1, 2, 2, 3, 3, 4]
```

ospencer (changes requested):
Looks great! Just one idea about an additional test.

compiler/test/stdlib/list.test.gr
ohana54: No idea how I missed this case 🤦

ospencer (approved):
Nice! 🎉

PR #356, opened 2020-10-03
Add isnt operator

### LINKED ISSUE #338: [Lang] Add `isnt` operator
We have `==` and `!=` for structural equality, but we only have `is` for physical equality. The `isnt` operator should be the negation of `is`.

For this change, we'll need to add a token for `isnt` to the lexer and the parser.

https://github.com/grain-lang/grain/blob/4c03b34094e500fb90f7ed81f19fcfddf4c112ce/compiler/src/parsing/lexer.mll#L169-L171
https://github.com/grain-lang/grain/blob/4c03b34094e500fb90f7ed81f19fcfddf4c112ce/compiler/src/parsing/parser.dyp#L175

We'll also need to add it as an infix operator:
https://github.com/grain-lang/grain/blob/4c03b34094e500fb90f7ed81f19fcfddf4c112ce/compiler/src/parsing/parser.dyp#L421-L426
https://github.com/grain-lang/grain/blob/4c03b34094e500fb90f7ed81f19fcfddf4c112ce/compiler/src/parsing/parser.dyp#L450

And then finally as a binary operator:
https://github.com/grain-lang/grain/blob/4c03b34094e500fb90f7ed81f19fcfddf4c112ce/compiler/src/parsing/parser.dyp#L274-L276

Code-wise, all that needs to be done is define a function similar to `!=` in Pervasives:
https://github.com/grain-lang/grain/blob/4c03b34094e500fb90f7ed81f19fcfddf4c112ce/stdlib/pervasives.gr#L57-L61

fa7ad:
should function similar to != operator

Fixes #338


ospencer (changes requested):
Excellent! Would you mind adding a test for this, around here-ish? https://github.com/grain-lang/grain/blob/ecca7798889c357aed4205f01a6687e37a042e1c/compiler/test/test_end_to_end.re#L209

fa7ad:
Updated with some basic tests

ospencer (approved):
Awesome work on this! I know @phated will be pretty hyped 😄

PR #366, opened 2020-10-04
Add array fill and fillRange methods

### LINKED ISSUE #361: [Stdlib] Implement Array.fill and Array.fillRange
Here's the type for `fill`:
```grain
fill : (a, Array<a>) -> Void
```

`fill` should replace every item in the array with the given argument.

And here's the type for `fillRange`:
```grain
fillRange : (a, Number, Number, Array<a>) -> Void
```

`fillRange` should replace items in the array starting from the start index up to (but not including) the end index. If negative indices are supplied, they should be treated as offsets from the end of the array.

clovis1122:
Add fill and fillRange methods. Closes #361

### DIFF
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -94,9 +94,9 @@ export let mapi = (fn, array) => {
 
 export let fill = (value, array) => {
   let length = length(array)
-  init(length, (index) => {
+  forEachi((_, index) => {
     array[index] := value
-  })
+  }, array)
   void
 }
 
@@ -111,11 +111,11 @@ export let fillRange = (value, start, stop, array) => {
   if (startIndex > stopIndex) {
     ignore(fail 'The start index cannot be higher than the stop index')
   }
-  init(length, (index) => {
+  forEachi((_, index) => {
     if (index >= startIndex && index < stopIndex) {
       ignore(array[index] := value)
     }
-  })
+  }, array)
   void
 }
 
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -111,11 +111,12 @@ export let fillRange = (value, start, stop, array) => {
   if (startIndex > stopIndex) {
     ignore(fail 'The start index cannot be higher than the stop index')
   }
-  forEachi((_, index) => {
-    if (index >= startIndex && index < stopIndex) {
-      ignore(array[index] := value)
-    }
-  }, array)
+
+  let mut index = startIndex
+  while (index < stopIndex && index < length) {
+    ignore(array[index] := value)
+    index += 1
+  }
   void
 }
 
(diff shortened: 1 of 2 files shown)


stdlib/array.gr
```diff
@@ -92,6 +92,33 @@ export let mapi = (fn, array) => {
...
+  }
+  if (startIndex > stopIndex) {
+    ignore(fail 'The start index cannot be higher than the stop index')
+  }
+  init(length, (index) => {
+    if (index >= startIndex && index < stopIndex) {
+      ignore(array[index] := value)
+    }
+  })
+  void
```
ospencer, line 119: Similarly, we'll want to avoid using `init` here. We could use `forEach` here as well, but it'd be a tad more efficient to use a while loop to only iterate the range 😄

ospencer (changes requested):
Looks pretty good! We'll want to change it a bit to make it more efficient, but otherwise looks great!

stdlib/array.gr
```diff
@@ -92,6 +92,33 @@ export let mapi = (fn, array) => {
   })
 }
 
+export let fill = (value, array) => {
+  let length = length(array)
+  init(length, (index) => {
+    array[index] := value
+  })
+  void
```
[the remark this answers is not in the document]
clovis1122: Good feedback, fixed!

stdlib/array.gr
```diff
@@ -92,6 +92,33 @@ export let mapi = (fn, array) => {
   })
 }
 
+export let fill = (value, array) => {
+  let length = length(array)
+  forEachi((_, index) => {
+    array[index] := value
+  }, array)
+  void
```
ospencer, line 100: We can drop the void now!
```suggestion
```

stdlib/array.gr
```diff
@@ -92,6 +92,33 @@ export let mapi = (fn, array) => {
...
+  }
+  if (startIndex > stopIndex) {
+    ignore(fail 'The start index cannot be higher than the stop index')
+  }
+  forEachi((_, index) => {
+    if (index >= startIndex && index < stopIndex) {
+      ignore(array[index] := value)
+    }
+  }, array)
+  void
```
ospencer, line 119: Here we can use a while loop so we don't have to iterate the whole thing 👌🏽
```suggestion
  let mut index = startIndex
  while (index < stopIndex) {
    array[index] := value
    index += 1
  }
```

stdlib/array.gr
clovis1122: Good feedback, fixed!

ospencer (approved):
👏 👏 👏

stdlib/array.gr
clovis1122: Oh, actually the tests failed when I removed void, haha.

ospencer: Oh, dang.

stdlib/array.gr
clovis1122: Yup! Done, just added an extra check for index > length

PR #368, opened 2020-10-04
fix empty array as input in Array.find function

### LINKED ISSUE #357: Implement Array.find and Array.findIndex
Should be similar to #353.

lvaniscak:
I forgot to check if input array is not empty. Closes  #357


ospencer (approved):
Great catch! Thanks for fixing this :)

PR #369, opened 2020-10-04
Implement Array.count

### LINKED ISSUE #359: [Stdlib] Implement Array.count
Should be similar to List.count.

MScheibel:
Implemented Array.count function. Closes #359


ospencer (changes requested):
Looks great! I have one small ask though: could you re-assign the array in your tests so it doesn't depend on the previous test? Should be ready to go after that 🙂

ospencer (approved):
Thanks!

PR #370, opened 2020-10-04
Implement List.product and Array.product

### LINKED ISSUE #362: [Stdlib] Implement List.product and Array.product
Here's the type for `List.product`:

```grain
product : (List<a>, List<b>) -> List<(a, b)>
```

`List.product` should take two lists, and return a new list of tuples of all of the combinations of elements from the two lists.

Similarly, here's the type for `Array.product`:

```grain
product : (Array<a>, Array<b>) -> Array<(a, b)>
```

Dony477:
Implemented product function on array and list. Closes #362

### DIFF
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -252,8 +252,8 @@ let product = (a, b) => {
   let mut list = []
   forEach((aItem) => {
     forEach((bItem) => {
-     list = append(list, [(aItem, bItem)])
+     list = [(aItem, bItem), ...list]
     },b)
   }, a)
-  list
+  reverse(list)
 }
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -247,3 +247,13 @@ let findIndex = (fn, list) => {
   }
   findItemIndex(list, 0)
 }
+
+let product = (a, b) => {
+  let mut list = []
+  forEach((aItem) => {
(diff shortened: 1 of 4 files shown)


stdlib/list.gr
```diff
@@ -246,4 +246,14 @@ let findIndex = (fn, list) => {
...
   findItemIndex(list, 0)
-}
\ No newline at end of file
+}
+
+let product = (a, b) => {
+  let mut list = []
+  forEach((aItem) => {
+    forEach((bItem) => {
+     list = append(list, [(aItem, bItem)])
```
ospencer, line 255: Here we could instead do `list = [(aItem, bItem), ...list]` and call `reverse` at the end for a bit more efficiency 🙂

ospencer (changes requested):
Awesome work! I left one note, but other than that, this is fantastic 👏🏽

stdlib/list.gr
Dony477: @ospencer  good point 👍

ospencer (approved):
Sweet, thank you!

Dony477:
@ospencer there were some conflicts but i resolved them :)

ospencer (commented):
Thanks for fixing that! 😄

PR #371, opened 2020-10-04
Add Queue to stdlib

### LINKED ISSUE #363: [Stdlib] Queue
We should add a new standard library called `queue.gr`. It should be immutable. There's a neat algorithm that you can find here: https://github.com/asif31iqbal/ImmutableQueue

We should start by implementing the methods described in that project's README. For the `Stack` implementation, we should just use a `List`.

miguelcarvalho13:
Closes #363 

Adds a Queue library to the `stdlib`.

### DIFF
--- a/compiler/test/stdlib/queue.test.gr
+++ b/compiler/test/stdlib/queue.test.gr
@@ -6,8 +6,8 @@ let sampleQueue = Queue.enQueue(3, Queue.enQueue(2, Queue.enQueue(1, empty)))
 
 # Queue.isEmpty
 
-assert Queue.isEmpty(empty) == true
-assert Queue.isEmpty(sampleQueue) == false
+assert Queue.isEmpty(empty)
+assert !Queue.isEmpty(sampleQueue)
 
 # Queue.head
 
--- a/compiler/test/stdlib/queue.test.gr
+++ b/compiler/test/stdlib/queue.test.gr
@@ -2,7 +2,7 @@ import Queue from 'queue'
 
 let empty = Queue.make()
 # 1 <- 2 <- 3
-let sampleQueue = Queue.enQueue(3, Queue.enQueue(2, Queue.enQueue(1, empty)))
+let sampleQueue = Queue.enqueue(3, Queue.enqueue(2, Queue.enqueue(1, empty)))
 
 # Queue.isEmpty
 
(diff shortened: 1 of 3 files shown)


ospencer (commented):
This looks fantastic!!

How would you feel about having the Empty queue just be `{ forwards: [], backwards: [] }`? The `dequeue` pattern match made me realize it could be consolidated a bit, looking something like this:

```
match (queue) {
  | { forwards: [], backwards: [] } => queue
  | { forwards: [head, ...ftail], backwards } => { forwards: ftail, backwards }
  ...
}
```

miguelcarvalho13:
Oh...yeah! I think that's a good idea and it will make the code cleaner!
I'll try to work on this

miguelcarvalho13:
@ospencer I did the changes! Seems way better than before 😁

compiler/test/stdlib/queue.test.gr
```diff
@@ -0,0 +1,28 @@
+import Queue from 'queue'
+
+let empty = Queue.make()
+# 1 <- 2 <- 3
+let sampleQueue = Queue.enQueue(3, Queue.enQueue(2, Queue.enQueue(1, empty)))
+
+# Queue.isEmpty
+
+assert Queue.isEmpty(empty) == true
+assert Queue.isEmpty(sampleQueue) == false
```
ospencer, line 10:
```suggestion
assert Queue.isEmpty(empty)
assert !Queue.isEmpty(sampleQueue)
```

compiler/test/stdlib/queue.test.gr
```diff
@@ -0,0 +1,28 @@
...
+# Queue.enQueue
+
+assert Queue.head(Queue.enQueue(1, empty)) == Some(1)
+assert Queue.head(Queue.enQueue(4, sampleQueue)) == Some(1)
+
+# Queue.deQueue
+
+assert Queue.isEmpty(Queue.deQueue(empty)) == true
+assert Queue.isEmpty(Queue.deQueue(Queue.enQueue(1, empty))) == true
+assert Queue.isEmpty(Queue.deQueue(Queue.deQueue(Queue.deQueue(sampleQueue)))) == true
```
ospencer, line 26:
```suggestion
assert Queue.isEmpty(Queue.deQueue(empty))
assert Queue.isEmpty(Queue.deQueue(Queue.enQueue(1, empty)))
assert Queue.isEmpty(Queue.deQueue(Queue.deQueue(Queue.deQueue(sampleQueue))))
```

ospencer (changes requested):
Realized we could tidy up the tests a little bit!

And one last thing (sorry this one is annoying 🙈) could we rename `enQueue` and `deQueue` to `enqueue` and `dequeue`? I realized the Java implementation used the camel case names, but I believe these are more common since they're both just one word.

But otherwise this looks great and we'll be ready to merge after!

compiler/test/stdlib/queue.test.gr
miguelcarvalho13: Done ✅

compiler/test/stdlib/queue.test.gr
miguelcarvalho13: Done ✅

miguelcarvalho13:
> Realized we could tidy up the tests a little bit!
> 
> And one last thing (sorry this one is annoying 🙈) could we rename `enQueue` and `deQueue` to `enqueue` and `dequeue`? I realized the Java implementation used the camel case names, but I believe these are more common since they're both just one word.
> 
> But otherwise this looks great and we'll be ready to merge after!

Hahaha That's fine and it does make sense!
I did the changes let me know if there's something still missing 😄

ospencer (approved):
Perfect, thanks so much! :D

PR #372, opened 2020-10-05
Use full filenames for persistent structures

### LINKED ISSUE #231: Use full filepath for persistent structures

ospencer:
Fixes #231.

### DIFF
--- a/compiler/src/typed/env.re
+++ b/compiler/src/typed/env.re
@@ -736,17 +736,17 @@ let check_consistency = ps =>
 /* Reading persistent structures from .cmi files */
 
 let save_pers_struct = (crc, ps) => {
-  let modname = ps.ps_name;
-  Hashtbl.add(persistent_structures, modname, Some(ps));
+  let filename = ps.ps_filename;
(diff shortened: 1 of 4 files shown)


phated (approved):
So awesome!! This has been on the backlog for awhile now. One thing that I'd like see is going back through the tests and naming the first test to be the colliding module name (e.g. `"list"` and `"array"` tests), but I suppose that cleanup can be opened as another issue—maybe for Hacktoberfest?

ospencer:
I thought about this a bit... I did add an explicit test, so I don't know that we need to do that. We totally could, though.

ospencer:
@phated let me know what you think, but I'll merge this for now.

PR #378, opened 2020-10-09
Add couple of new functions to the list module that will help fix #125

### LINKED ISSUE #125: Implement Array.sort and List.sort
These functions should take the array/list and a comparator function as arguments and produce a sorted array/list.

Open questions:
- should the array version be a sort in place?
- should there be two array versions, a `sort` and `sort!` where the latter sorts in place?

tmphey:
Hey guys! 

Here is some initial work on the #125 - new methods for `List` module:
- `List.take(n, list)` - takes `n` first items from the list. 
- `List.takeWhile(fn, list)` - takes elements from the list until `fn` returns true.
- `List.sub(start, len, list)` - creates a sublist for `list` starting from `start` with `len` elements

### DIFF
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -277,6 +277,6 @@ let product = (a, b) => {
   reverse(list)
 }
 
-let sub = (start, lenght, list) => {
-  take(lenght, drop(start, list))
+let sub = (start, length, list) => {
+  take(length, drop(start, list))
 }
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -231,6 +231,25 @@ let rec dropWhile = (fn, list) => {
   }
 }
 
+let rec take = (n, list) => {
+  if (n < 0) {
+    fail 'number of items to take cannot be a negative number'
+  } else {
+    match ((n, list)) {
+      | (_, []) => list
+      | (n, _) when n == 0 => []
+      | (n, [first, ...rest]) => [first, ...take(n-1, rest)]
+    }
+  }
+}
+
+let rec takeWhile = (p, list) => {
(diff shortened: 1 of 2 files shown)


stdlib/list.gr
```diff
@@ -231,6 +231,25 @@ let rec dropWhile = (fn, list) => {
   }
 }
 
+let rec take = (n, list) => {
+  if (n < 0) {
+    fail 'number of items to take cannot be a negative number'
+  } else {
+    match ((n, list)) {
+      | (_, []) => list
+      | (n, _) when n == 0 => []
```
ospencer, line 240: Could match on `(0, _)` here but I think this is more obvious/clear.

stdlib/list.gr
```diff
@@ -257,3 +276,7 @@ let product = (a, b) => {
   }, a)
   reverse(list)
 }
+
+let sub = (start, lenght, list) => {
+  take(lenght, drop(start, list))
```
ospencer, line 281:
```suggestion
let sub = (start, length, list) => {
  take(length, drop(start, list))
```

tmphey:
> This looks awesome! 😄 

Thanks! Do you want me to write docs for these methods?

stdlib/list.gr
tmphey: Yeah, you absolutely right. The syntax is not that familiar for me still

ospencer:
You definitely don't have to ask twice 😛 

Here's the issue for it: https://github.com/grain-lang/grain-lang.org/issues/121

Thanks so much! Let me know if you need help finding your way around the docs :)

PR #383, opened 2020-10-13
Fix cross-module type annotations

ospencer:
This PR enables (and fixes) the ability to reference types from a module in a type annotation, like so:

```
import Map from 'map'

let map : Map.T = Map.make()
```


phated:
I'd still like to have a voice chat before calling this good. I just want to understand everything going on.

phated (approved):
👍 Thanks for discussing this with me. I think this can land in current state but we should really consider updates to our module system (for ease of use and to make potential debugging a lot easier).

ospencer:
Agreed!

PR #385, opened 2020-10-14
Fix imported ADT typechecking

### LINKED ISSUE #315: Compiler: Typechecking of imported type variables is broken
Take this data type for example:
```grain
export data Foo<a> = Cons(a, Foo<a>) | Nothing
```

When imported into another module, the typechecker does not properly resolve/link/register the type variables, allowing a program like 
```grain
import Other from './other'
Other.Cons(true, Other.Cons(4, Other.Nothing))
```
to compile.

This is _not_ a problem if the data type is used locally.

ospencer:
Fixes #315.

In addition to linking type variables for imported values, we now also do it for imported types (which includes variants and records).

Fixing this made some of the Result tests fail, and after reviewing the code it made perfect sense. I updated the implementations to have the behavior the tests expected. 😄

### DIFF
--- a/compiler/src/typed/translsig.re
+++ b/compiler/src/typed/translsig.re
@@ -32,8 +32,6 @@ let rec collect_type_vars = typ =>
   };
 
 let link_type_vars = ty => {
-  reset_type_variables();
-  collect_type_vars(ty);
   let rec link_types = texpr => {
     let desc =
       switch (texpr.desc) {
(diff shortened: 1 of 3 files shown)


phated (approved):
I am beyond happy that you were able to get this fixed! 🎊

ospencer:
Me too!!

PR #389, opened 2020-10-15
Comment syntax update

### LINKED ISSUE #275: Lang: Use C-style comment syntax
`// line comment`
`/* block comment */`

ospencer:
Closes #275. 

We still support shebangs. They're not currently restricted to just the first line of the file, but we can update that later.

### DIFF
--- a/compiler/src/parsing/parser.dyp
+++ b/compiler/src/parsing/parser.dyp
@@ -27,8 +27,20 @@ let debug_print_state () =
   when_debug !last_state_printer
 
 let line_comment source loc =
-  let content = String_utils.slice ~first:1 source |> String.trim in
+  let content = String_utils.slice ~first:2 source |> String.trim in
   Line { cmt_content=content; cmt_source=source; cmt_loc=loc }
+
+let shebang_comment source loc =
+  let content = String_utils.slice ~first:2 source |> String.trim in
+  Shebang { cmt_content=content; cmt_source=source; cmt_loc=loc }
+
+let block_comment source loc =
+  let content = String_utils.slice ~first:2 ~last:(-2) source |> String.trim in
+  Block { cmt_content=content; cmt_source=source; cmt_loc=loc }
+
+let doc_comment source loc =
+  let content = String_utils.slice ~first:3 ~last:(-2) source |> String.trim in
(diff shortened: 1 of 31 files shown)


compiler/test/test_end_to_end.re
```diff
@@ -1717,10 +1718,10 @@ let comment_tests = {
     ),
     te(
       "comment_line_numbers_1",
-      "#comment\n#comment\n5 + 5L",
+      "//comment\n//comment\n5 + 5L",
       "line 3, characters 4-6",
     ),
-    t("comment_lone_#", "#\nlet x = 10\nx", "10"),
+    t("comment_lone_//", "//\nlet x = 10\nx", "10"),
```
peblair, line 1724: Explicit block comment and shebang parser tests would be prudent here.

compiler/src/parsing/parser.dyp
```diff
@@ -27,9 +27,21 @@ let debug_print_state () =
...
+  Shebang { cmt_content=content; cmt_source=source; cmt_loc=loc }
+
+let block_comment source loc =
+  let content = String_utils.slice ~first:2 ~last:(-2) source |> String.trim in
+  Block { cmt_content=content; cmt_source=source; cmt_loc=loc }
+
+let doc_comment source loc =
+  let content = String_utils.slice ~first:3 ~last:(-2) source |> String.trim in
+  Doc { cmt_content=content; cmt_source=source; cmt_loc=loc }
+
```
peblair, line 44: Where are these used?

peblair (changes requested):
LGTM; just asking for tests, plus one question

compiler/src/parsing/parser.dyp
ospencer: They're used in the lexer.

peblair: @ospencer Ah, I see. I think it's a little confusing, since `line_comment` and `shebang_comment` are redefined inside of the lexer, but the others are not (so I think there are two parallel scopes going on?). Do you think there is a way to make it a little more clear what name refers to what?

peblair (approved):

PR #390, opened 2020-10-15
Added an LSP mode for use with vscode lsp implementation.

ng-marcus:
(works with https://github.com/ng-marcus/lsp-grain which isn't a published extension yet so you need to run it in debug mode)

I've modified the cli and compiler to support an lsp mode where the cli passes through stdin as the content to compile,  and reports back the errors (it any) on stdout in a form of

startline startchat endline endchar
Error message 
over multiple lines

The assumption is the compiler stops after the first error so there's only one to report.

I'm not really sure what's happening in location.re (report exception) so I just copied it and sprintfed the result, I assume it's to do with nested exceptions.   It works for now, but I'm not confident in the code.

Your feedback on approach (modifying the compiler in this way, method of reporting back) would be appreciated.

### DIFF
--- a/cli/bin/lsp.js
+++ b/cli/bin/lsp.js
@@ -9,7 +9,7 @@ const grainc = path.join(__dirname, 'grainc.exe');
 
 module.exports = (file, options) => {
   try {
-    execSync(`${grainc} --stdlib=${options.stdlib} ${options.cflags ? options.cflags : ''} --lsp ${file}`, { stdio: 'inherit' });
+    execSync(`${grainc} --stdlib=${options.stdlib} ${options.cflags ? options.cflags : ''} -g --lsp ${file}`, { stdio: 'inherit' });
     process.exit()
   } catch (e) {
     //  console.log(e); //.stdout.toString());
--- a/cli/bin/lsp.js
+++ b/cli/bin/lsp.js
@@ -12,7 +12,6 @@ module.exports = (file, options) => {
     execSync(`${grainc} --stdlib=${options.stdlib} ${options.cflags ? options.cflags : ''} -g --lsp ${file}`, { stdio: 'inherit' });
     process.exit()
   } catch (e) {
-    //  console.log(e); //.stdout.toString());
     console.log(e);
     if (options.graceful) {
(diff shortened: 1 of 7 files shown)


peblair (approved):
This is awesome! I definitely want @ospencer to look at it as well, but this looks great.

ng-marcus:
I was thinking about reporting back in json too, as that would support multiple errors easily in future.  I'll do that!

ng-marcus:
Changed the reporting format to JSON.    Only handles a single error at the moment, but we can return an array of them once we get multiple.

One other thing I'm not happy with - I pass in in the filename (for returning in the reporting) but although I'm using stdin, the compiler is checking for the existence of the file.    That might not be there for a file not yet saved, although I've yet to see that in vscode

cli/bin/lsp.js
```diff
@@ -0,0 +1,23 @@
...
+// call the compiler, passing stdio through so the compiler gets the source code on stdin
+// and we get the compiler output in stdout
+// we still take the file name so we have it available
+
+module.exports = (file, options) => {
+  try {
+    execSync(`${grainc} --stdlib=${options.stdlib} ${options.cflags ? options.cflags : ''} -g --lsp ${file}`, { stdio: 'inherit' });
+    process.exit()
+  } catch (e) {
+    //  console.log(e); //.stdout.toString());
```
ospencer, line 15:
```suggestion
```

ospencer:
This looks great! I like the JSON response 👍 

How would you feel about merging `grain-lsp` with `vscode-grain`? I like the file structure of `reason-language-server` where the language server is the top-level project, and editor extensions are all kept in a sub-directory. (And then you only need to worry about installing one extension, and it just comes with the language server.)

ospencer:
But as for just this change, it's ready to go!

ng-marcus:
Thanks for the merge!   I think merging grain-lsp and vscode-grain is the way to go.     100% agreed on one install to get syntax highlighting and language server.  I'll take a look at that next

phated:
🤯 🤯 🤯 🤯

PR #391, opened 2020-10-15
remove some unneeded dependencies

phated:
As per the discussion in discord, this is just cleaning up some deps in preparation of other cleanup I'm going to do.


ospencer (approved):
Good as long as the CI passes!

PR #393, opened 2020-10-17
Don't re-raise exceptions when printing compiler errors in LSP mode

ng-marcus:
Stop re-raising lex/parse/compile errors in LSP mode to stop them causing the LSP server to report them on trace output


ospencer (approved):
Seems reasonable!

PR #394, opened 2020-10-17
[BREAKING] Enum/Record syntax update

### LINKED ISSUE #379: [Lang] Data declaration syntax update
# Data Declaration Syntax Update

New syntax for declaring variant types:
```
enum Foo<a> {
  Bar(String),
  Baz(List<a>, Number),
  Quux
}

export enum Foo<a> { ... }
```

New syntax for declaring record types:
```
record Foo<a> {
  bar: String,
  baz: (List<a>, Number),
  quux: Bool
}
```

New syntax for pattern matching:
```
match (foo) {
  Bar(a) => a,
  Baz([hd, ...tl], 5) => {
    ...
  },
  Quux => 'quux'
}
```

## Why we're making these changes

We want Grain to be a very straightforward language, and we realized that our `data` keyword was starting to get a little overloaded. Between defining variant types, record types, and in the future creating abstract types and type aliases, we realized that it would be better for users to have explicit keywords for these actions. This way, it's immediately clear to both new and experienced Grain users what's happening with the types when reading a Grain program. 🎉

ospencer:
Closes #379.

### DIFF
--- a/compiler/src/parsing/lexer.mll
+++ b/compiler/src/parsing/lexer.mll
@@ -160,7 +160,8 @@ rule token = parse
   | "from" { FROM }
   | "->" { ARROW }
   | "=>" { THICKARROW }
-  | "data" { DATA }
+  | "enum" { ENUM }
+  | "record" { RECORD }
   | "|" { PIPE }
   | "let" { LET }
   | "mut" { MUT }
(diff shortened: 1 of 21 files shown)


phated (approved):
It's awesome that this was just parser/lexer changes and all the behavior is the same!

Also, I'm loving the ergonomics of this syntax change while reading through the PR. 😍

PR #395, opened 2020-10-20
Switch to using esy exclusively over combination of esy/opam files

phated:
This PR does a few things and I'll outline why below:

1. It removes the opam files and disables the generation of these files.
I've removed these because using dune generation [causes esy to break](https://github.com/esy/esy/issues/1104) due to the tree not being immutable.
Also, we started relying on some [ReasonNative](https://github.com/facebookexperimental/reason-native) projects that haven't been published to opam yet, which means we are unable to publish these packages to opam currently.
Ideally, we would be able to add opam generation to esy for which I've opened a [feature request](https://github.com/esy/esy/issues/1130).
2. Moves all our dependencies to the esy.json file and updates some dependencies, including depending on a published version of our `binaryen.ml` project published to npm. It also relies less on resolutions, except in a couple of ReasonNative cases where our 4.11.0 OCaml dependency breaks their published versions.
3. Moves the `dune build` command to a proper `build` stanza instead of the `compile` script. I believe this will give us better caching with esy and hopefully reduce rebuilds of binaryen.
4. Streamlines the build commands to use just the recommended `esy` command to install and build a project.
5. Removes the cmake requirement in our docs because the published `binaryen.ml` project depeneds on `esy-python` and `esy-cmake`, which means they no longer need to exist on a users system before building our compiler.
6. Some general cleanup.

A couple things that still need to be done:
1. The dockerfile needs to be updated for these patterns.
2. I'm planning to further streamline some of our build commands.


compiler/grainc/dune
```diff
@@ -3,4 +3,4 @@
  (public_name grainc)
  (package grain)
  (modules grainc)
- (libraries grain dune-build-info))
+ (libraries grain binaryen.native dune-build-info))
```
ospencer, line 6: Is this necessary?

ospencer (commented):
Nice! I kinda wanted to do this for some time. I agree that this is the right change for now, especially since we weren't publishing to opam anyway. Not that we really _should_ publish, but it could be useful. I imagine we'd just reference a git tag or something. Which we could still do (maybe with a few modifications).

I digress.

This looks great! 🎉

compiler/grainc/dune
phated: It's not strictly required because we have binaryen.ml use the native lib by default, but I like being explicit because we ship the js lib with it now too.

ospencer: I meant like, is this necessary to specify for `grainc`?

ospencer: (Explained, and I get it now 👍)

ospencer (approved):

PR #398, opened 2020-10-21
Fix the dockerfile after my esy changes

phated:
Was able to get some really nice cleanup in here too!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,9 +5,6 @@ LABEL description="Grain CLI and Runtime"
 LABEL vcs-url="https://github.com/grain-lang/grain"
 LABEL maintainer="philip@grain-lang.org"
 
-RUN apt-get update
-RUN apt-get install -y cmake=3.13.4-1
-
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -6,7 +6,6 @@ LABEL vcs-url="https://github.com/grain-lang/grain"
 LABEL maintainer="philip@grain-lang.org"
 
 RUN apt-get update
-RUN apt-get install -y cmake=3.13.4-1
```
ospencer, line 9: Probably don't need the line above anymore either huh?

ospencer (commented):
This is sooooooo much better!

Dockerfile
phated: Probably not. I wasn't sure if that updated like gcc or something. Do you think I should remove it?

ospencer: Yeah, let's remove it.

phated: Amended and force-pushed.

ospencer (approved):
Woohoo!

PR #399, opened 2020-10-22
Bump Grain version to 0.2.0

phated:
I believe we are at a point where we should bump the language's version again. We've introduced a lot of features (`let mut`, number system) and breaking changes (`enum`, `record`, comment syntax).

Bumping the version will indicate to anyone coming back from a break that things have changed and their code might not compile.

@ospencer can you merge this and tag a release? I think we might want to editorialize some release notes with the highlights too.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grain/cli",
-  "version": "0.1.0",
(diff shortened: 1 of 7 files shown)


ospencer (approved):
🎉 Loads of goodies here!

PR #405, opened 2020-10-24
Remove unnecessary wasm exports

ospencer:
We were exporting everything that was global in a module rather than just things that are explicitly exported. This wasn't a huge deal, since the compiler won't let you import them, but this is the first step to some wasm cleanup.

### DIFF
--- a/compiler/src/middle_end/anf_helper.re
+++ b/compiler/src/middle_end/anf_helper.re
@@ -88,8 +88,8 @@ module AExp = {
     anf_env: or_default_env(env),
     anf_analyses: ref([]),
   };
-  let let_ = (~loc=?, ~env=?, ~glob=Nonglobal, rec_flag, binds, body) =>
-    mk(~loc?, ~env?, AELet(glob, rec_flag, binds, body));
+  let let_ = (~loc=?, ~env=?, ~global=Nonglobal, rec_flag, binds, body) =>
+    mk(~loc?, ~env?, AELet(global, rec_flag, binds, body));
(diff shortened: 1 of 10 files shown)


compiler/src/middle_end/anf_helper.re
```diff
@@ -95,13 +95,17 @@ module AExp = {
...
     imp_use_id: use_id,
     imp_desc: d,
     imp_shape: s,
+    imp_exported: e,
     imp_analyses: ref([]),
   };
-  let grain_value = (a, md, name, s) => mk(a, GrainValue(md, name), s);
-  let wasm_func = (a, md, name, s) => mk(a, WasmFunction(md, name), s);
-  let js_func = (a, md, name, s) => mk(a, JSFunction(md, name), s);
+  let grain_value = (~glob=Nonglobal, a, md, name, s) =>
```
phated, line 105: Since `glob` is an actual programming term, can we use a different name for these?

compiler/src/middle_end/linearize.re
```diff
@@ -1016,12 +1016,18 @@ let rec transl_anf_statement =
       (None, []);
     };
   | TTopException(_, ext) => (Some(linearize_exception(env, ext)), [])
-  | TTopForeign(desc) =>
+  | TTopForeign(exported, desc) =>
     let arity = Ctype.arity(desc.tvd_desc.ctyp_type);
+    let glob =
```
phated, line 1021: Rename please

phated (changes requested):
Looks good, I just don't like naming them `glob` when that's a term already.

phated (approved):

PR #407, opened 2020-10-27
Add support for loading Grain modules directly from an ArrayBuffer

z0w0:
This allows Grain's browser runtime to be executed with an `ArrayBuffer`
of compiled Web Assembly code, which is extremely useful for bundling Grain apps (e.g. Webpack) rather than depending
on the WASM being hosted somewhere for `readURL`.


ospencer (approved):
This looks great, thanks!

Issue #39, opened 2018-08-02
JavaScript FFI


ospencer:
Users should be able to reference JavaScript values (most notably, functions) via an FFI.

The syntax should be similar to the [wasm FFI syntax](https://github.com/grain-lang/grain/blob/b050f22c307c9238543110f0f2088939a9023b1d/src/grain-stdlib/pervasives.grlib#L2) and allow the user to reference a certain module and value from that module. The user must provide a type annotation.

For now, Grain will just assume that the type annotation is correct. Maybe in the future it could do some type of checking (maybe just for JavaScript primitives).

The user also must pass those values into the Grain runtime in the form of a plain JavaScript object, with keys as module names and values which are also plain JavaScript objects, with keys as value names, i.e.
```js
{
  myLibrary: {
    myFunction () { ... }
  }
}
```

Suggested syntax:
`foreign js myLibrary myFunction : String -> Number`

PR #409, opened 2020-10-28
Add JS FFI import support

z0w0:
This adds a `import foreign js funcName : Args -> Result from 'module'` syntax for
importing and calling JS functions directly from Grain per #39. Any modules must be declared
when instantiating the runtime using a new `.addJSImports` method like so:

```js
runner.addJSImports({
  moduleName: {
    someFunc: () => { console.log('this can do anything like a plain old JS function'); }
  }
});
```

The runtime then wraps those JS function imports so that Grain argument values are converted into JS argument values and JS return values are converted to Grain return values. Under the hood, JS imports are outputted just like WASM imports in the compiled WASM, except JS imports implicitly live under the "js/" module name prefix. This module name prefix is then populated using the aforementioned `.addJSImports` method.

At the moment there are no pre-defined JS imports within the runtime, but over at [grain-loader](https://github.com/z0w0/grain-loader) I've added support for passing in JS imports when calling a compiled Grain program, for e.g.

```
import foreign js alert : String -> Void from 'window'

alert("Hello world");
```

Can be compiled and used in a Webpack bundle like so:

```js
const hello = require('./hello.gr');

hello({
  window: {
    alert: window.alert
  }
});
```

We can add some pre-defined JS imports to the runtime here in the future and possibly allow additional ones to be passed into `grain run` since it's running on Node.js.


ospencer:
Hey Zack, awesome that you took this on! The code is good and the API is exactly what we wanted. We hadn't done this yet since we don't have well-defined semantics for the conversion functions and they're currently only used internally. Since this exposes them (albeit indirectly) to users, we have to hammer down what all of the semantics are supposed to be and get them documented.

If you're willing to work with us to get that done, we can get this over the line 🙏

z0w0:
No problem, I'm happy to improve and document the conversion functions as part of this PR. It's definitely a blocker at the moment anyway since anything other than primitive types will throw a runtime error right now. Let me know how you would like to proceed with standardising the semantics.

ospencer:
@z0w0 Sorry we haven't gotten back to this yet! I wanted to land #414 and get some reference types in so we can use abstract types with this. I've been a little bit stuck on an annoying bug with it for a bit. Hopefully I'll have some updates for you soon!

ospencer:
@z0w0 we swapped our default branch from master -> main, so I got this rebased and updated in #549. I just need to figure out how to get you commit access on that branch 😅 

Closing this since the histories are unrelated.

PR #410, opened 2020-10-31
Graindoc

phated:
This is my rough draft of implementing a `graindoc` command line tool.

Currently, it only handles docblocks that are on the line preceding a `let` or a type declaration (like `enum` or `record`). There are a lot more type printing things to fix and this should also be able to output to a file.

I'm not sure what to do about multiple `let` bindings on the same line - e.g. `let a = 1, b = 2`.

Also looking for any other tips to make this cleaner.


compiler/src/typed/printtyp.re
```diff
@@ -978,32 +978,7 @@ let extension_only_constructor = (id, ppf, ext) => {
...
-    };
-
-  Osig_value(vd);
-};
-
-let value_description = (id, ppf, decl) =>
-  Oprint.out_sig_item^(ppf, tree_of_value_description(id, decl));
-
-/* Print a value declaration */
-
```
phated, line 1004: This code was just duplicated.

compiler/src/typed/oprint.re
```diff
@@ -631,22 +631,42 @@ and print_out_sig_item = ppf =>
...
+      //   | Orec_not => "type nonrec"
+      //   | Orec_first => "type"
+      //   | Orec_next => "and"
+      //   };
+      let kwd =
+        switch (td.otype_type) {
+        | Otyp_record(_) => "record"
+        | Otyp_variant(_, _, _, _) => "enum"
+        // TODO: What is variant if sum is enum?
+        | Otyp_sum(_) => "enum"
```
phated, line 645: Why are our `enum`s the `Otyp_sum` and what are `Otyp_variant`s then?

compiler/src/typed/oprint.re
```diff
@@ -631,22 +631,42 @@ and print_out_sig_item = ppf =>
...
-      },
-      ppf,
-      td,
-    )
+  | Osig_type(td, rs) => {
+      // let kwd = switch (rs) {
+      //   | Orec_not => "type nonrec"
+      //   | Orec_first => "type"
+      //   | Orec_next => "and"
+      //   };
```
phated, line 639: Do we want to use these at all?

compiler/src/typed/oprint.re
```diff
@@ -631,22 +631,42 @@ and print_out_sig_item = ppf =>
...
+        | Otyp_constr(_, _) => "constr"
+        | Otyp_manifest(_, _) => "manifest"
+        | Otyp_object(_, _) => "object"
+        | Otyp_stuff(_) => "stuff"
+        | Otyp_tuple(_) => "tuple"
+        | Otyp_var(_, _) => "var"
+        | Otyp_poly(_, _) => "poly"
+        | Otyp_module(_, _, _) => "module"
+        | Otyp_attribute(_, _) => "attribute"
+        };
```
phated, line 661: I don't know which of these we want to print

phated:
I'm currently using this file to test locally:
```
/**
  Some test doc comments
*/
export let a = 1;

/**
  Adds `x` and `y` together.
*/
export let add = (x, y) => x + y;

/**
  Tracks the math operation you would like to perform.
*/
export enum Math<x, y> {
  Add(x, y),
  Subtract(x, y),
}
```

PR #413, opened 2020-10-31
Switch the setup command to prepare

phated:
Yarn supports the [`prepare` script](https://classic.yarnpkg.com/en/docs/package-json#toc-scripts), which is a lifecycle script that is run after `yarn` or `yarn install` is run.

By renaming our `setup` to `prepare`, it will reduce the initial setup to 2 commands, instead of 3.


ospencer (approved):
I'm a big fan of this!

PR #416, opened 2020-11-07
First support for LSP lenses

ng-marcus:
Updated the LSP mode of the compiler to be more generic, and it now returns error and lenses in the same JSON.

Will return multiple errors if supported in the future.

Refactored the code to capture errors properly using fprintf and a buffer and removed the error printing code from Location.re where it didn't really belong by creating a separate lsp library to contain this code.


compiler/dune-project
```diff
@@ -115,3 +115,18 @@
     (fp (>= 0.0.0))
     (fs (>= 0.0.0))
     (cmdliner (>= 1.0.2))))
+
+
+
```
ospencer, line 120:
```suggestion

```

compiler/grainc/grainc.re
```diff
@@ -57,12 +56,31 @@ let compile_string = name => {
 
   try(
     ignore(
-      Compile.compile_string(~hook=stop_after_typed, ~name, program_str^),
+      {
+        let compileState =
+          Compile.compile_string(~hook=stop_after_typed, ~name, program_str^);
+
+        switch (compileState.cstate_desc) {
```
ospencer, line 63: Tiny thing here wrt casing:
```suggestion
        let compile_state =
          Compile.compile_string(~hook=stop_after_typed, ~name, program_str^);

        switch (compile_state.cstate_desc) {
```

compiler/src/lsp/lenses.re
```diff
@@ -0,0 +1,76 @@
...
+};
+
+let lens_to_yojson = (l: lens_t): Yojson.t =>
+  `Assoc([("line", `Int(l.line)), ("signature", `String(l.signature))]);
+
+let lens_list_to_yojson = (lenses: list(lens_t)): Yojson.t =>
+  `List(List.map(l => lens_to_yojson(l), lenses));
+
+let lenses_to_json_string = (l: list(lens_t)): string =>
+  Yojson.to_string(lens_list_to_yojson(l));
```
ospencer, line 15: Could you just `[@deriving yojson]` the `lens_t` here?

compiler/src/lsp/output.re
```diff
@@ -0,0 +1,55 @@
+type lsp_error = {
+  file: string,
+  line: int,
+  startchar: int,
+  endline: int,
+  endchar: int,
+  lsp_message: string,
+};
+
+let lsp_error_to_yojson = (e: lsp_error): Yojson.t =>
```
ospencer, line 10: Sort of the same deal here, could we just derive it?

ng-marcus:
Thanks for the comments, I'll get on them!

compiler/src/lsp/lenses.re
ng-marcus: Now I know, yes I can :-). And a much better solution!

compiler/grainc/grainc.re
ng-marcus: :-D.  Absolutely.  I laugh because my team can spot my code from my use of snake case where they all use camel, so they've obviously succeeded in their brainwashing! ;)

compiler/src/lsp/output.re
ng-marcus: That's done.  I'm not 100% confident with yojson stuff, not sure if I have to go from safe to basic to print it?

compiler/src/lsp/dune
```diff
@@ -0,0 +1,8 @@
+(library
+ (name lsp)
+ (public_name lsp)
+ (synopsis "LSP modules for Grain compiler")
+ (libraries compiler-libs.common dyp grain_utils grain_typed utf8.lib str
+   ppx_deriving_yojson.runtime ppx_sexp_conv.runtime-lib sexplib yojson)
```
phated, line 6: I think we need to trim this down. I don't believe all these will be be needed for this package (especially `dyp`).

compiler/grainc/dune
```diff
@@ -3,4 +3,4 @@
  (public_name grainc)
  (package grain)
  (modules grainc)
- (libraries grain binaryen.native dune-build-info))
+ (libraries grain lsp binaryen.native dune-build-info))
```
phated, line 6: Since we are splitting this out now, I propose we rename it to `diagnostics` because I foresee us building a lot of the tooling into this.

ng-marcus: Agreed, I'll do that.

compiler/src/lsp/dune
ng-marcus: Good point, I was lazy and just copied the existing project.  I'll trim it down to what we need

ng-marcus: Trimmed down.

compiler/grainc/dune
ng-marcus: Renamed to Grain_diagnosics to avoid any potential conflicts.

ospencer (approved):
Looks great! I'll have @phated take another look at the dependencies stuff, but otherwise I think we're good to go.

compiler/src/diagnostics/lenses.re
```diff
@@ -0,0 +1,77 @@
...
+      (statement: Typedtree.toplevel_stmt) => {
+        let d = statement.ttop_desc;
+        switch (d) {
+        | TTopExpr(expr) =>
+          let buf = Buffer.create(64);
+          let ppf = Format.formatter_of_buffer(buf);
+          let t: Types.type_expr = expr.exp_type;
+          Printtyp.type_expr(ppf, t);
+          Format.pp_print_flush(ppf, ());
+          let signt = Buffer.contents(buf);
```
phated, line 30: I _believe_ this can be cleaned up with some of the utilities in `Format`, but I don't have it on-hand. Let's land it with this because it is working and we can always do a cleanup pass when I find those utils.

phated (approved):
This is so awesome! Let's land it!

PR #418, opened 2020-11-10
fix: ensure Dockerfile works with `yarn prepare` command

phated:
When switching our setup command to `yarn prepare`, I broke the docker container creation. This was due to the caching strategy we were trying to employ, which didn't allow for the stdlib or runtime scripts to be executed immediately after `yarn install` (which is when `yarn prepare` is guaranteed to run).

Since this container is mostly consumed from container registries, I've just reduced the dockerfile to have no caching. We should consider a better strategy for our lifecycle scripts going forward, but I think that will require some changes to the compiler itself.


ospencer (approved):
LGTM!

PR #421, opened 2020-11-11
feat: add update method to map stdlib

### LINKED ISSUE #192: Stdlib: Add Map.update method
A common method I see for Map is `update` which is an in-place update of the value. I imagine the API to be something like:

```
let map = Map.fromList([('foo', 1), (bar', 2)]);

Map.update('foo', (val) => val + 1, map);
```

phated:
Closes #192 

This implements the `Map.update` method that sets, updates, or removes the value at a given key in a Map data structure. The updater function receives an `Option` of the existing (or not) value and must return an `Option` indicating whether to set or remove the value.


ospencer (approved):
This is spicy. I love it!

PR #422, opened 2020-11-11
refactor: extract header & trailer from parser.dyp

phated:
This extracts the header and trailer code from the `parser.dyp` file and puts it into a `parser_extra.re` file. It also converts the OCaml code to ReasonML and documents why they are externalized.

I only had to change the `parse_program` method to include dypgen's `program` value in the function signature and then it is partially applied at the bottom of the `parser.dyp` file.


phated:
FWIW, I could be convinced to change the filename to `parser_header` now that everything is included into the header

ospencer (changes requested):
Looks great! I would prefer `parser_header` 🙂

phated:
Rename complete!

ospencer (approved):
Woohoo!

Issue #297, opened 2020-09-10
Compiler: Parse infix bitwise operators


ospencer:
Even if they aren't defined yet, we should still parse `&`, `|`, `^` (tbd), `<<`, `>>`, and `>>>` so they can be used.

bmakuh:
@ospencer you cool if I take this one?

bmakuh:
Also, what's the tbd on `^`?

ospencer:
@bmakuh Absolutely!

ospencer:
The tbd on `^` is that we currently use it as the unboxing operator. We have `let mut` now, so it's a little less necessary that we have syntax for that operator.

Now that I think about it though, we can totally unambiguously parse when it's an infix operator versus when it's a prefix operator.

ospencer:
If you feel like giving that a shot, definitely give it a go. It may be a little annoying depending on your experience with parser generators (and more specifically dypgen), but we could also pair on it if it gives you a headache!

ospencer:
Here's where we keep all of the operator precedences: https://github.com/grain-lang/grain/blob/master/docs/contributor/operator_precedence.md

bmakuh:
@ospencer sorry just finally getting around to this. For these operators, is this what you were intending?
| sigil | operator |
| --- | --- |
|`&` | AND |
| `\|` | OR |
| `^` | XOR |
| `<<` | left shift |
| `>>` | right shift |
| `>>>` | ? |

bmakuh:
Also, do we want a NOT operator like `~`?

ospencer:
Yup, `>>` is the arithmetic (signed) right shift operator, and `>>>` is the logical (unsigned) right shift operator.

And yeah, we don't want `~` right now since we might be using `~` for something else 🙂

bmakuh:
Ok, thanks. And is there any danger of ambiguity between an infix bitwise pipe operator and other uses of the pipe? I guess we're not using it for `match` anymore, but it came to mind after you mentioned the unboxing operator

ospencer:
There shouldn't be! It's only used for `||` now (and pipeline in the future).

phated:
This was completed by #425 🎉

Issue #183, opened 2020-07-01
Lang: Syntax error with with unbox operator


phated:
This program doesn't parse:

```
data Container = {
    size: Box<Number>
}

let make = () => {
    { size: box(1) }
}

let getSize= (m) => {
  let size = ^m.size;
  size
}

let a = make();
print(getSize(a));
```

It gives this error:

```
$ grain test.gr 
File "test.gr", line 1, characters 0-0:
Error: The Grain compiler had trouble parsing your program.
```

PR #425, opened 2020-11-12
Add support for bitwise operators

bmakuh:
Addresses #297

## Todo
- [x] rebase to bring in unboxing operator removal
- [x] tests


ospencer (commented):
Yikes, I checked out your branch to see what error you got and it's definitely one of our least helpful errors. That error just means that there's an ambiguity in the grammar.

In this case, it's because `^` is both an infix _and_ a prefix operator (I knew it would be `^` that bit us!). The parser can't tell which you mean whenever you write `^`. This also made me realize that unboxing is currently a function bound to the identifier `^`. To actually get both to work, we'd need to make it something statement-like which would be a bit out of scope for this PR (and would also be a breaking change), so I think the move is to just leave `^` out for now.

Also, it looks like you just didn't add the new tokens to the lexer:
https://github.com/grain-lang/grain/blob/master/compiler/src/parsing/lexer.mll#L194-L196

Forgetting that one would just give you a regular syntax error, though 🙂

ospencer:
But I'm really glad to see this come through! 😄

phated:
I also previously ran into issues with this operator at https://github.com/grain-lang/grain/issues/183 - I believe we chatted in the past about just dropping that symbol as the unbox operator. It's also not really needed much anymore since almost all mutable data can be handled with `let mut`.

I'll try to land a PR tonight that removes that operator as `unbox` so we can keep it for bitwise operations in this PR!

bmakuh:
Thanks for the tips @ospencer and thanks for the unboxing PR @phated! Learning a lot trying to make this work 😅

bmakuh:
Are we going to have ambiguity in situations like these for right shift operators?
```
mut next: Option<Bucket<k, v>>
```
```
mut buckets: Array<Option<Bucket<k, v>>>
```

ospencer:
Nope, we shouldn't have to worry about that! Speaking of which, though, we should probably stick to having only one token each for the carets... We would totally run into some issues if there was a token for `>>>` versus just matching a token for a single `>` three times.

phated:
@bmakuh feel free to rebase on master now that I landed the removal of `^` as unbox 🍻

bmakuh:
@ospencer this should be ready for review now.

compiler/src/parsing/parser.dyp
```diff
@@ -127,11 +128,17 @@ binop_expr :
...
+  | binop_expr(<=p90) llcaret_op eols?   binop_expr(<p100) { Exp.apply ~loc:(symbol_rloc dyp) (mkid_expr dyp [$2]) [$1; $4] } p100
   | binop_expr(<=p90) rcaret_op eols?    binop_expr(<p90) { Exp.apply ~loc:(symbol_rloc dyp) (mkid_expr dyp [$2]) [$1; $4] } p90
+  | binop_expr(<=p90) rrcaret_op eols?   binop_expr(<p100) { Exp.apply ~loc:(symbol_rloc dyp) (mkid_expr dyp [$2]) [$1; $4] } p100
+  | binop_expr(<=p90) rrrcaret_op eols?  binop_expr(<p100) { Exp.apply ~loc:(symbol_rloc dyp) (mkid_expr dyp [$2]) [$1; $4] } p100
   | binop_expr(<=p90) lesseq_op eols?    binop_expr(<p90) { Exp.apply ~loc:(symbol_rloc dyp) (mkid_expr dyp [$2]) [$1; $4] } p90
   | binop_expr(<=p90) greatereq_op eols? binop_expr(<p90) { Exp.apply ~loc:(symbol_rloc dyp) (mkid_expr dyp [$2]) [$1; $4] } p90
+  | binop_expr(<=p40) amp_op eols?       binop_expr(<p70) { Exp.apply ~loc:(symbol_rloc dyp) (mkid_expr dyp [$2]) [$1; $4] } p70
   | binop_expr(<=p40) ampamp_op eols?    binop_expr(<p40) { Exp.apply ~loc:(symbol_rloc dyp) (mkid_expr dyp [$2]) [$1; $4] } p40
+  | binop_expr(<=p30) pipe_op eols?      binop_expr(<p50) { Exp.apply ~loc:(symbol_rloc dyp) (mkid_expr dyp [$2]) [$1; $4] } p50
```
ospencer, line 141: It looks like for these some of the priorities for the operands are off. They should just match the priority of the operation (that's on the far right side).

For example, `binop_expr(<=p90) caret_op eols?     binop_expr(<p60)` should be `binop_expr(<=p60) caret_op eols?     binop_expr(<p60)`.

I don't think we really need tests to make sure all of these priorities hold as long as all of the numbers match up. Was there a reason why you needed to make them something different?

ospencer (commented):
Woohoo! I left one comment about the priorities, but otherwise it looks great! We might want to keep the original names of the bit-shifting operators in Pervasives in addition to the new syntax, since otherwise it'd be a breaking change. I'm going to double check with the core team though.

Excellent work!

compiler/src/parsing/parser.dyp
bmakuh: Oops, it was a mistake.

ospencer:
@bmakuh We chatted, and we're just going to have the breaking change. I dropped your fix commit (and thank you for changing it so quickly!) and we're pretty much good to merge.

ospencer (approved):

PR #426, opened 2020-11-13
fix!: remove ^ as unbox operator (fixes #183)

### LINKED ISSUE #183: Lang: Syntax error with with unbox operator

phated:
We've had a few scenarios where the `^` being used for unboxing has caused problems, specifically around ambiguity (see #183 & #425) in the parsetree.

In this PR, I remove the `^` operator as the unbox operator but left the token in the lexer/parser for use as a bitwise operator in #425. I also removed the operator from Pervasives and cleaned up the tests.

### DIFF
--- a/compiler/src/parsing/lexer.mll
+++ b/compiler/src/parsing/lexer.mll
@@ -193,6 +193,7 @@ rule token = parse
   | "]" { RBRACK }
   | "<" { LCARET }
   | ">" { RCARET }
+  | "^" { CARET }
   | "++" { PLUSPLUS }
   | "+" { PLUS }
   | "-" { DASH }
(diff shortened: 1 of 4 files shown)


bmakuh (approved):

ospencer (approved):
LGTM. We'll also need to update the docs. Maybe let's make an issue for that?

phated:
Docs! https://github.com/grain-lang/grain-lang.org/pull/139

PR #428, opened 2020-11-17
fix: Only backpatch closures for let-bindings

ospencer:


phated (approved):
Looks solid to me! Glad to have this working 🎊

PR #430, opened 2020-11-25
fix: replace shebang trick with flagged-respawn to support Windows

phated:
The hashbang as it is currently written doesn't work on Windows, because npm creates wrappers that are informed by the hashhang. When it sees `/usr/bin/env node`, it just replaces the call with something like `%PATH%\node.exe` but if you try to specify `/bin/sh` it tries to do `%PATH%\sh.exe` which crashes.

@ospencer I know you put this in place, so do you believe this simplification will work?

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -1,5 +1,4 @@
-#!/bin/sh
-// 2>/dev/null; exec /usr/bin/env node --experimental-wasm-bigint "$0" "$@"
+#!/usr/bin/env node --experimental-wasm-bigint
 // https://github.com/grain-lang/grain/issues/114
 
 const { execSync } = require('child_process');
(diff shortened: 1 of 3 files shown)


phated:
With these test failures, it seems like it won't work ☹️

ospencer:
Yeah the tldr on why the shebang is like that is so you can pass flags. We should check if we even need the flags anymore with newer versions of Node.

phated:
It's still needed in the v14 release stream. Do you know that I've already solved that problem though? https://github.com/gulpjs/flagged-respawn

ospencer:
See, that should be the top answer on stack overflow.

phated:
🤣 Creators rarely hang out on SO.

phated:
Implemented flagged-respawn and tests seem to be working locally.

ospencer (approved):

phated:
Another cool thing that comes from this is that users can now pass other experimental wasm flags to the grain compiler which will be added to the respawn!

PR #434, opened 2020-11-25
fix: temporarily disable inline tests for Windows

phated:
Inline tests are currently broken on Windows until https://github.com/janestreet/ppx_inline_test/issues/23 is fixed (hopefully in 0.15 🤞)

I couldn't come up with a good way to enable them on other platforms, so let's just turn them off temporarily.


ospencer (approved):
What inline tests do we even have?

phated:
Tests around our string utils. I added them when working on some of those.

PR #435, opened 2020-11-25
feat!: upgrade binaryen to 0.4.0 & update test invocation to work with it

phated:
This bumps Binaryen.ml to 0.4.0 and fixes the test invocation so the tests will run on Windows.

I've marked this as a breaking change since technically we lose support for things like `x86_32` operating systems by vendoring binaryen.


phated:
Something seems to be very wrong with the tests.

phated:
It seems that `esy b dune runtest` broke the tests on Mac and Linux. I removed it to see if this works in CI (it works locally again). However, that is likely needed for Windows, so I need to investigate further.

compiler/esy.json
```diff
@@ -15,12 +15,12 @@
   "scripts": {
     "copy-compiler": "cp #{$GRAINC_BIN_PATH} #{$CLI_BIN_DIR}",
     "clean": "rm -rf #{self.root}/_esy",
-    "test": "dune runtest --force",
+    "test": "dune runtest",
```
ospencer, line 18: Any way we can keep `--force`? It's useful for when you change the runtime and want to re-run the tests.

phated: I didn't know what this did 🙈 Re-added

ospencer (approved):
Nice!

PR #439, opened 2020-11-28
chore: update Reason to use published 3.6.2

phated:
Reason v3.6.2 was finally published to opam, so we no longer need a resolution! This was the release that had ocaml 4.11.0 support.


ospencer (approved):
Sweet!

phated:
A really cool thing is that it will parse `type t<'k, 'v>;` too. Unfortunately, the pretty printer will turn it back into parens until v4 😢

PR #440, opened 2020-11-30
fix: set test sharding to 1 in attempt to avoid test race condition

phated:
I was poking around oUnit2 and I found it does parallel tests by default, so I added the `-help` flag to our test binary and it output this:
```
  -shards i                       Number of shards to use as worker (threads or processes). (default: 2)
```

I believe this means that oUnit is running our tests in parallel in 2 processes and that **might** be causing an issue where 1 process is trying to create a stdlib wasm file, like `pervasives.gr.wasm`, at the same time another process is trying to open that file, which would result in the input channel being closed immediately and `End_of_file` being surfaced which fails the test with `Cmi_format.Error`. It also _kind of_ explains why only 1 tests fails with that error because no other test going forward would be recreating that file, so they all could just read from it. Thoughts?

I haven't had a chance to run this change over and over on my system yet, but hoping to find time tomorrow.

### DIFF
--- a/compiler/test/dune
+++ b/compiler/test/dune
@@ -11,4 +11,4 @@
   (source_tree test-libs)
   (source_tree test-data))
  (action
-  (run %{<})))
+  (run %{<} -shards 1)))


phated:
Running this now, and it nearly doubles the time of the tests (which would be expected) from ~100 seconds to ~200 seconds.

phated:
Superseded by #449

ospencer (approved):

phated:
Reopening this because OCaml's file lock mechanism doesn't seem to solve our issues. I opened #468 and #469 as enhancements that should allow us to run in parallel.

PR #441, opened 2020-12-01
refactor: use really_input instead of input + raising ourselves

phated:
While debugging our random test failure, I was reading docs about `input` and the way we were using it in these cases was the exact same as the way `really_input` works. With these changes, we no longer need to raise `End_of_file` ourselves!

### DIFF
--- a/compiler/src/utils/wasm_utils.re
+++ b/compiler/src/utils/wasm_utils.re
@@ -157,18 +157,14 @@ let read_leb128_u64_stream = stream =>
 
 let read_int32 = inchan => {
   let bytes = Bytes.create(4);
-  if (input(inchan, bytes, 0, 4) != 4) {
-    raise(End_of_file);
-  };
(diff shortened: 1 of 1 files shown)


ospencer (approved):
Sweet, great catch!

compiler/src/utils/wasm_utils.re
```diff
@@ -242,6 +238,7 @@ let get_wasm_sections = (~reset=false, inchan) => {
     let magic = [0x00, 0x61, 0x73, 0x6D];
     let version = [0x01, 0x00, 0x00, 0x00];
     let expect = (strbuilder, b) => {
+      // TODO: This can raise End_of_file. Should we really be letting that bubble all the way up?
```
phated, line 241:
```suggestion
```
With #449, do you think we can remove this TODO @ospencer ?

phated: I'm going to leave this in until I address #469

PR #442, opened 2020-12-01
chore: cleanup more generated files in stdlib

phated:
During debugging, I noticed some other files hanging around that weren't being cleaned, so I adjusted the script to clean those up.

### DIFF
--- a/stdlib/package.json
+++ b/stdlib/package.json
@@ -40,6 +40,6 @@
   "dependencies": {},
   "scripts": {
     "build": "node --harmony-top-level-await ./build.mjs",
-    "clean": "del-cli '**/*.wasm' '**/*.wasm.modsig' '**/*.wasm.wast' '!stdlib-external/**'"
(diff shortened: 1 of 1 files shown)


phated:
Heh, Cmi_error because this doesn't have my fix as the base 😛

stdlib/package.json
```diff
@@ -40,6 +40,6 @@
   "dependencies": {},
   "scripts": {
     "build": "node --harmony-top-level-await ./build.mjs",
-    "clean": "del-cli '**/*.wasm' '**/*.wasm.modsig' '**/*.wasm.wast' '!stdlib-external/**'"
+    "clean": "del-cli '**/*.wasm' '**/*.wasm.*' '**/*.gr.mashtree' '!stdlib-external/**'"
```
ospencer, line 43: Oh wow, I hadn't realized we were even generating `.mashtree` files. We probably don't need to generate those when running the tests.

ospencer (approved):

stdlib/package.json
phated: We don't by default, but I did something to create them and then I couldn't automatically clean them up.

PR #443, opened 2020-12-01
refactor: little cleanup around try/switch

phated:
I combined the `try` wrapping a `switch` and the additional `raise` into a single `switch` statement that matches on Option or exceptions.

Also, changed an OCaml reference.


ospencer (approved):
Much cleaner!

PR #447, opened 2020-12-05
Allow newline when creating array

### LINKED ISSUE #445: First element of an array literal shouldn't need to be on the same line
When working with a large array, I want to define it like this:
```
let data = [>
  1664,
  1909,
  1904,
  1638,
  1844,
]
```

But Grain currently requires me to write the first element on the same line, like:

```
let data = [> 1664,
  1909,
  1904,
  1638,
  1844,
```

I think this might just be a bug in the parser.

ospencer:
Fixes #445.


phated (approved):
🎊 Heck yeah!!

PR #448, opened 2020-12-05
feat: allow setting the initial memory via CLI option

phated:
This allows setting the initial amount of memory page via a command line option. It actually required a handful of workarounds due to the static initialization (and dependents upon it), which we should consider when taking a deeper look at the runtime in the future.

Using this CLI option, I am able to compile my Advent of Code solutions.

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -26,7 +26,7 @@ function onRespawn(ready, proc, argv) {
 
   let program = require('commander');
   // Workaround to defer loading of the grain runtime until the memory settings have been parsed
-  let imports = {
+  let actions = {
     get compile() {
       return require('./compile.js');
     },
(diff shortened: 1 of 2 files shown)


cli/bin/grain.js
```diff
@@ -25,9 +25,18 @@ function onRespawn(ready, proc, argv) {
   const path = require('path');
 
   let program = require('commander');
-  let compile = require('./compile.js');
-  let run = require('./run.js');
-  let lsp = require('./lsp.js');
+  // Workaround to defer loading of the grain runtime until the memory settings have been parsed
+  let imports = {
```
ospencer, line 29: Could we name this something else, like `actions`? Seeing `imports` here makes me think it's a WebAssembly import object.

ospencer (changes requested):
One comment, but otherwise looks good!

cli/bin/grain.js
phated: Done! Naming is one of the hardest problems in computer science

ospencer (approved):

PR #449, opened 2020-12-07
fix: Lock module reading/writing to prevent race conditions

ospencer:
Still need to run it through the paces, but I believe this should fix the issues we've been experiencing.


phated:
Docs for anyone (like me) interested in OCaml's locking mechanisms: https://caml.inria.fr/pub/docs/manual-ocaml/libref/Unix.html#VALlockf

phated (dismissed):
Haven't put it through rigorous pacing yet, but it seems like a great solution after reading about it!

💯 great work @ospencer!

phated:
Superseded by #440 #468 #469

PR #455, opened 2020-12-09
feat: add reduce/flatMap/every/some to Array stdlib

phated:
Needed some Array methods for AoC. I'll open an issue to document and assign to myself right now.

### DIFF
--- a/compiler/test/stdlib/array.test.gr
+++ b/compiler/test/stdlib/array.test.gr
@@ -201,10 +201,10 @@ assert Array.flatMap(x => fail "Shouldn't be called", [>]) == [>]
 
 assert Array.every(x => x < 4, arr) == true
 assert Array.every(x => x > 4, arr) == false
-assert Array.every(x => fail "Shouldn't be called", [>]) == true
+assert Array.every(x => fail "Shouldn't be called", [>])
 
 // Array.some
 
 assert Array.some(x => x > 1, arr) == true
 assert Array.some(x => x > 4, arr) == false
-assert Array.some(x => fail "Shouldn't be called", [>]) == false
+assert !Array.some(x => fail "Shouldn't be called", [>])
--- a/compiler/test/stdlib/array.test.gr
+++ b/compiler/test/stdlib/array.test.gr
@@ -186,3 +186,25 @@ assert Array.count(x => x > 0, [>]) == 0
 assert Array.count(x => x == 3, arr) == 1
 
 // TODO: test with wrong indexes and assert the exception.
+
+// Array.reduce
+
+assert Array.reduce((acc, x) => acc + x, 0, arr) == 6
+assert Array.reduce((acc, x) => fail "Shouldn't be called", 0, [>]) == 0
+
+// Array.flatMap
+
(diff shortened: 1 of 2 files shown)


compiler/test/stdlib/array.test.gr
```diff
@@ -186,3 +186,25 @@ assert Array.count(x => x > 0, [>]) == 0
...
+// Array.flatMap
+
+assert Array.flatMap(x => [> x, x * 2], arr) == [>1, 2, 2, 4, 3, 6]
+assert Array.flatMap(x => fail "Shouldn't be called", [>]) == [>]
+
+// Array.every
+
+assert Array.every(x => x < 4, arr) == true
+assert Array.every(x => x > 4, arr) == false
+assert Array.every(x => fail "Shouldn't be called", [>]) == true
```
ospencer, line 204: I think our convention has been to just leave the assertion as-is.
```suggestion
assert Array.every(x => fail "Shouldn't be called", [>])
```

compiler/test/stdlib/array.test.gr
```diff
@@ -186,3 +186,25 @@ assert Array.count(x => x > 0, [>]) == 0
...
+
+// Array.some
+
+assert Array.some(x => x > 1, arr) == true
+assert Array.some(x => x > 4, arr) == false
+assert Array.some(x => fail "Shouldn't be called", [>]) == false
```
ospencer, line 210:
```suggestion
assert !Array.some(x => fail "Shouldn't be called", [>])
```

stdlib/array.gr
```diff
@@ -92,6 +92,24 @@ export let mapi = (fn, array) => {
   })
 }
 
+export let reduce = (fn, initial, array) => {
+  let acc = box(initial);
```
ospencer, line 96: Why not `let mut` here?

stdlib/array.gr
```diff
@@ -92,6 +92,24 @@ export let mapi = (fn, array) => {
...
+  forEach((el) => acc := fn(unbox(acc), el), array);
+  unbox(acc)
+}
+
+export let flatMap = (fn, array) => {
+  reduce((result, value) => append(result, fn(value)), [>], array)
```
ospencer, line 102: I know that this makes the code a bit more complicated, but we should probably use `Array.concat` instead of append since it's a lot more efficient.

ospencer (commented):
Excited to have these!!

stdlib/array.gr
phated: I think I was getting a type bug with the `acc` value being mutable when passed into the function. Maybe I was doing something wrong?

stdlib/array.gr
phated: Can you explain how Array.concat is faster?

stdlib/array.gr
phated: I was doing something wrong. Just pushed the `let mut` changes.

stdlib/array.gr
ospencer: Chatted over discord.

ospencer (approved):

PR #456, opened 2020-12-09
feat: add initial Range stdlib module

phated:
Some very rudimentary Range stuff that I needed/wanted for AoC.

### DIFF
--- a/compiler/test/stdlib/range.test.gr
+++ b/compiler/test/stdlib/range.test.gr
@@ -6,17 +6,17 @@ let range = Range.Inclusive(1, 5)
 assert Range.isBetween(1, range)
 assert Range.isBetween(3, range)
 assert Range.isBetween(5, range)
-assert Range.isBetween(10, range) == false
-assert Range.isBetween(-1, range) == false
+assert !Range.isBetween(10, range)
+assert !Range.isBetween(-1, range)
 
 // Range.isBetween (Exclusive)
 let range = Range.Exclusive(1, 5)
 
 assert Range.isBetween(1, range)
 assert Range.isBetween(3, range)
-assert Range.isBetween(5, range) == false
-assert Range.isBetween(10, range) == false
-assert Range.isBetween(-1, range) == false
+assert !Range.isBetween(5, range)
+assert !Range.isBetween(10, range)
+assert !Range.isBetween(-1, range)
 
 // Range.forEach (Inclusive)
 let range = Range.Inclusive(1, 5)
--- a/compiler/test/stdlib/range.test.gr
+++ b/compiler/test/stdlib/range.test.gr
@@ -1,10 +1,41 @@
 import Range from 'range'
 
-// Range.isBetween
-let range = Range.inclusive(1, 5)
+// Range.isBetween (Inclusive)
+let range = Range.Inclusive(1, 5)
 
 assert Range.isBetween(1, range)
 assert Range.isBetween(3, range)
 assert Range.isBetween(5, range)
 assert Range.isBetween(10, range) == false
 assert Range.isBetween(-1, range) == false
+
+// Range.isBetween (Exclusive)
+let range = Range.Exclusive(1, 5)
+
+assert Range.isBetween(1, range)
+assert Range.isBetween(3, range)
+assert Range.isBetween(5, range) == false
+assert Range.isBetween(10, range) == false
+assert Range.isBetween(-1, range) == false
+
+// Range.forEach (Inclusive)
+let range = Range.Inclusive(1, 5)
+
+let mut results = []
+
+Range.forEach((idx) => {
+  results = [idx, ...results];
+}, range)
+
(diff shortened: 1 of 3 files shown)


stdlib/range.gr
```diff
@@ -0,0 +1,15 @@
+// TODO: Exclusive
+enum Range {
+  Inclusive(Number, Number),
+}
+
+export let inclusive = (lower, upper) => {
+  Inclusive(lower, upper)
+}
+
+export let isBetween = (value, range) => {
```
ospencer, line 10: Should this be data last?

phated: It is! The "range" is the data because it's the range module

ospencer: Oh interesting. I hadn't really thought of ranges as data, since I'd usually want to test some other data against the range. I see myself wanting to do stuff like `List.filter(partial Range.isBetween(myRange), myList)`. I can't really argue with your logic, though. I'll happily trust your judgment on what to do here though 👍

ospencer (approved):

stdlib/range.gr
phated: I definitely think your example will be a common case. I was thinking it would be achieved with placeholder values like `List.filter(partial Range.isBetween(_, myRange), myList)` but maybe that is too verbose?

compiler/test/stdlib/range.test.gr
```diff
@@ -0,0 +1,41 @@
+import Range from 'range'
+
+// Range.isBetween (Inclusive)
+let range = Range.Inclusive(1, 5)
+
+assert Range.isBetween(1, range)
+assert Range.isBetween(3, range)
+assert Range.isBetween(5, range)
+assert Range.isBetween(10, range) == false
+assert Range.isBetween(-1, range) == false
```
ospencer, line 10: Some quick style stuff:
```suggestion
assert !Range.isBetween(10, range)
assert !Range.isBetween(-1, range)
```

compiler/test/stdlib/range.test.gr
```diff
@@ -0,0 +1,41 @@
...
+
+// Range.isBetween (Exclusive)
+let range = Range.Exclusive(1, 5)
+
+assert Range.isBetween(1, range)
+assert Range.isBetween(3, range)
+assert Range.isBetween(5, range) == false
+assert Range.isBetween(10, range) == false
+assert Range.isBetween(-1, range) == false
```
ospencer, line 19:
```suggestion
assert !Range.isBetween(5, range)
assert !Range.isBetween(10, range)
assert !Range.isBetween(-1, range)
```

stdlib/range.gr
ospencer: As I look at this again, I wonder if this should be called `inRange`?

stdlib/range.gr
```diff
@@ -0,0 +1,25 @@
...
+
+export let forEach = (fn, range) => {
+  let (first, last) = match (range) {
+    Inclusive(lower, upper) => (lower, upper),
+    Exclusive(lower, upper) => (lower, upper - 1),
+  };
+
+  let mut idx = first;
+  while (idx <= last) {
+    fn(idx) == void
```
ospencer, line 22: Is this to force the return type to be `Void`? We might want to use an annotation instead.

stdlib/range.gr
```diff
@@ -0,0 +1,25 @@
...
+  Exclusive(Number, Number),
+}
+
+export let isBetween = (value, range) => {
+  match (range) {
+    Inclusive(lower, upper) when value >= lower && value <= upper => true,
+    Exclusive(lower, upper) when value >= lower && value < upper => true,
+    _ => false
+  }
+}
```
ospencer, line 12: This looks like it doesn't handle the cases where `Inclusive(upper, lower)` and `Exclusive(upper, lower)`.

stdlib/range.gr
```diff
@@ -0,0 +1,25 @@
...
+export let forEach = (fn, range) => {
+  let (first, last) = match (range) {
+    Inclusive(lower, upper) => (lower, upper),
+    Exclusive(lower, upper) => (lower, upper - 1),
+  };
+
+  let mut idx = first;
+  while (idx <= last) {
+    fn(idx) == void
+    idx += 1;
```
ospencer, line 23: And then this might need to be `-= 1` sometimes, yeah?

phated: Yep, but I completely changed the implementation to make it more clear

stdlib/range.gr
phated: Done!

stdlib/range.gr
phated: Done!

stdlib/range.gr
phated: Renamed! I'm not particular and `inRange` makes the range-last make more sense.

ospencer (approved):
Absolutely lovely!

Issue #159, opened 2020-06-27
Should fallible stdlib methods return Option?


phated:
I was looking at some methods in `lists.gr` and noticed that `hd`/`tl`/etc have a `fail` call in them. Once #158 lands, do you think that these methods should be changed to return Options? I really appreciate that from stdlibs I've used in the past.

ospencer:
I'm a big fan of having two versions, like `hd` and `hdOpt`, personally.

I would love for Grain to adopt a pattern where a function name that ends in `?` would return an option, so we could have `hd` and `hd?`. Just a thought.

But otherwise, I think for these we should have a version that throws and a version that returns an option.

phated:
I'm not a fan of having an exception and option returning version of the same function. It definitely feels like "legacy" problems in the OCaml stdlib.

As for `?`, I'd be careful with that symbol as part of the language, I think a lot of people coming from something like Rust would expect that to be an unwrap-or-early-return.

ospencer:
This has been open for a while... I think I vote that they just return options. Thoughts @grain-lang/core?

ohana54:
I'm not grain-lang/core, but I vote for options as well :)
From my experience with reasonml, having to be careful with which functions can throw was not ideal...

ng-marcus:
My preference is for options too.   Forcing the programmer to consider all potential code paths is a good thing and has raised the quality of my team's code a huge amount.  It can add a level of unwrapping overhead, but the alternative is unhandled exceptions in production.

Having both forms I'm more ambivalent about - it frequently leads to code review arguments that the value being used has already been checked (e.g. array length and unsafe array access), but that can nearly always be recoded to remove the manual check and use the safe check version anyway.

chriskrycho:
Strongly agree with returning `Option`. If you know that there’s nothing to check—for example because you just checked it in some way—you can always do the equivalent of this from Rust:

```rust
if (some_vec.len() > 1) {
    let item = some_vec
        .pop()
        .expect("we just checked that there's at least one element");

    // ...
}
```

This ends up working very well in practice: it means the normal path is the one that just checks the behavior, while giving end users an alternative which *will* blow up if they’re wrong… but they have to explicitly opt into it. There *are* times when this is the right thing to do, but because just using pattern matching or functional composition (`map`, `andThen`, etc.) is *easy*, you end up *usually* just doing it that way:

```rust
some_vec
    .pop()
    .and_then(|item| {
        // ...
    });
```

[...]

PR #459, opened 2020-12-10
feat!: change return type of Array.find/findIndex to Option

phated:
In #159, it seemed that most people were on-board with fallible stdlib methods to return Option.  During AoC, I actually found a use case for Array.find to return an Option (when I wrote my naive parseInt function), so I thought I'd get that started with this PR.

Just wanted to get the 👍 before I go through and convert the rest.


ospencer (approved):
LGTM!

PR #460, opened 2020-12-10
feat!: convert fallible List methods to return Option

phated:
List method conversions, which let me clean up one Queue method too.

I snuck some trailing whitespace removal in too.

I plan to add a contributor doc for stdlib patterns and update the website docs too.


ospencer (approved):
Excellent!

PR #461, opened 2020-12-11
fix!: Rework memory management

ospencer:
This PR addresses some issues we've been having working with large data sets.

Major changes: Grain will now properly grow the WebAssembly memory when out of space. Additionally, a majority of the memory manager has been rewritten from JS into AssemblyScript (which aligns with our goal of removing our dependency on JS).

Minor change: Removed support for garbage collection of cyclic data structures as it caused a major slowdown when trying to reclaim memory from data structures with many children. Grain doesn't (yet) have first-class support for cyclic data structures (see #96), so we don't believe this is a huge issue. If it does become an issue before WebAssembly has its own GC, we'll implement some variation of weak refs.

This is technically a breaking change because `GrainRunner` is no longer exported from the runtime. None of our tooling interacts with it directly, though.


phated (approved):
Tested this against all my Advent of Code stuff and it all works great! Truly amazing work @ospencer - I owe you a 🥃 !!!!!!!

ospencer:
I will definitely take a 🥃  any day of the week.

I also feel like I really need a 🥃 after the debugging to get this working 😂

PR #465, opened 2020-12-12
feat: add List.init to stdlib

phated:
I hacked together a `List.init` method that I wanted for day 5 of AoC. I'm pretty sure there is a better way to implement the recursion so I put a TODO asking for help making it better.

### DIFF
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -3,18 +3,17 @@
 export *
 
 let init = (length, fn) => {
-  if (length <= 0) {
-    []
-  } else {
-    // TODO: How can I make this better?
-    let rec iter = (idx, list) => {
-      match (idx) {
-        0 => [fn(idx), ...list],
-        _ => iter(idx - 1, [fn(idx), ...list])
-      }
+  // This method can be further improved by checking the length against a specific size
+  // and determining if it can be made tail-recursive, then use List.reverse on it.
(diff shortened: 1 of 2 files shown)


stdlib/list.gr
```diff
@@ -2,6 +2,21 @@
 
 export *
 
+let init = (length, fn) => {
+  if (length <= 0) {
+    []
+  } else {
+    // TODO: How can I make this better?
```
ospencer, line 9: Init for lists is annoying because there's a tradeoff on speed vs. not blowing the stack, i.e. do you just implement it the way you did here, or do you build a list and then reverse it at the end. OCaml picks a threshold and then chooses an implementation based on how many items you want: https://github.com/ocaml/ocaml/blob/03839754f46319aa36d9dad56940a6f3c3bcb48a/stdlib/list.ml#L79

Wasm (or maybe WASI would do this?) doesn't yet let you ask questions like "how big is the call stack?", so we can probably just pick something arbitrary, but fairly conservative, like a couple hundred. I think I actually ran a test a little while back to see how many stack frames you get in different browsers, and I think the lowest was something like 600. We could probably do some research.

Alternatively, if we just document that this is not tail-recursive, that's perfectly fine too 👍

phated: Updated with a note!

ospencer (approved):
🚀 initialization!

PR #466, opened 2020-12-12
feat: add Set stdlib module

zweimach:
Add initial implementation of Set module using ~~List as the~~ the same underlying data structure as Map.
It reuses many modified helper functions from the Map module.
Maybe we could unify those functions?

@ospencer, what is the convention for `init` and `make` family of functions?

### DIFF
--- a/compiler/test/stdlib/set.test.gr
+++ b/compiler/test/stdlib/set.test.gr
@@ -1,0 +1,223 @@
+import Set from "set"
+import List from "list"
+import Array from "array"
+
+// Data types used in multiple tests
+enum Resource { Grain, Sheep, Brick, Wood }
+record ResourceData { name: String, emoji: String }
+
+// Set.isEmpty
+
+let e = Set.make()
+
+assert Set.isEmpty(e)
+Set.add("🌾", e)
+assert !Set.isEmpty(e)
+
+// Set.size
+
+let s = Set.fromList(["🌾", "🐑", "🧱"])
+
+assert Set.size(s) == 3
+
+// Set.clear
+
+let c = Set.fromList(["🌾", "🐑", "🧱"])
+
+assert !Set.isEmpty(c)
+assert Set.clear(c) == void
+assert Set.isEmpty(c)
+
+// Set.contains
+
+let h = Set.fromList(["🌾", "🐑", "🧱"])
+
+assert Set.contains("🌾", h)
+assert Set.contains("🐑", h)
+assert Set.contains("🧱", h)
+assert !Set.contains("🌳", h)
+
+// Set.add
+
+let vars = Set.make()
+
+assert Set.add(Grain, vars) == void
+assert Set.add(Sheep, vars) == void
+assert Set.add(Grain, vars) == void
+
+assert Set.size(vars) == 2
+
+let recs = Set.make()
+
+assert Set.add({ name: "Grain", emoji: "🌾" }, recs) == void
+assert Set.add({ name: "Sheep", emoji: "🐑" }, recs) == void
+assert Set.add({ name: "Brick", emoji: "🧱" }, recs) == void
+assert Set.add({ name: "Grain", emoji: "🌾" }, recs) == void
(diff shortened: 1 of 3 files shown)


ospencer (changes requested):
This is awesome! Thanks for putting this together 😄 

Re: `make` vs. `init`, `make` creates an empty structure, and `init` allows you to create a structure with some initial data. Hopefully that answers your question, but let me know if that's not what you were asking.

I think this largely looks good, but for the backing data structure, I believe that we _do_ want it to be an Array, like Maps. Sets are essentially just `Map<t, Void>`, so the implementation can be pretty much exactly the same as Map, but the buckets would not have a `value` property. This way, your `Set.contains` and `Set.add` are both O(1) in the average case, versus O(n) to search/insert into a list.

Let me know if that makes sense or if there's anything I can clarify!

stdlib/set.gr
```diff
@@ -0,0 +1,305 @@
...
+  set
+}
+
+export let union = (set1, set2) => {
+  let set = make();
+  forEach((key) => {
+    add(key, set);
+  }, set1)
+  forEach((key) => {
+    if (!contains(key, set1)) {
```
phated, line 270: Is it more efficient to do this `contains` or to just call `add`?

stdlib/set.gr
```diff
@@ -0,0 +1,305 @@
...
+
+export let intersect = (set1, set2) => {
+  let set = make();
+  forEach((key) => {
+    if (contains(key, set2)) {
+      add(key, set);
+    }
+  }, set1)
+  forEach((key) => {
+    if (contains(key, set1) && !contains(key, set)) {
```
phated, line 300: Do we need the `!contains` or is that already handled by `add`? Which is faster?

phated (changes requested):
This is SICK! A few questions/changes that might be able to improve it.

stdlib/set.gr
```diff
@@ -0,0 +1,305 @@
...
+
+export let toList = (set) => {
+  reduce((list, key) => [key, ...list], [], set)
+}
+
+export let fromList = (list) => {
+  let set = make();
+  List.forEach((key) => {
+    add(key, set);
+  }, List.unique(list));
```
ospencer, line 248: We should be able to drop the `unique` call here, since `add` should take care of it (and it'll be faster!)

stdlib/set.gr
```diff
@@ -0,0 +1,305 @@
...
+
+export let toArray = (set) => {
+  Array.fromList(toList(set))
+}
+
+export let fromArray = (array) => {
+  let set = make();
+  Array.forEach((key) => {
+    add(key, set);
+  }, Array.unique(array));
```
ospencer, line 260: And here too.

compiler/test/stdlib/set.test.gr
```diff
@@ -0,0 +1,213 @@
...
+assert !Set.contains(Sheep, rejectTestSet)
+assert Set.contains(Brick, rejectTestSet)
+
+// Set.reduce
+
+let reduceTestSet = makeTestSet()
+
+let mut called = 0
+
+let result = Set.reduce((acc, key) => {
```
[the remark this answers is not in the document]
zweimach: Yes, we should. Thanks.

stdlib/set.gr
zweimach: How careless of me. Thanks.
They should be equally efficient as both use the same helper function.

stdlib/set.gr
zweimach: Thank you. Once again, I forgot that `add` should be sufficient.

stdlib/set.gr
zweimach: Yes, `add` should be enough. Thank you.

stdlib/set.gr
zweimach: Thank you.

phated (approved):
This looks great! Awesome work @zweimach!!

ospencer (approved):
Fantastic work! So excited to have this done! 🎉

PR #470, opened 2020-12-12
chore: Clean up old DOM implementation

ospencer:
I wanted to use the old DOM heap tag for characters, and I figured I should split out the DOM cleanup into a separate PR.


phated (approved):
You know me, always happy to see cleanup!

PR #473, opened 2020-12-14
feat: add reducei/counti/filter/filteri/unique to Array stdlib

zweimach:
Add filter, unique, and index variant of reduce & count to Array stdlib module.

### DIFF
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -283,3 +283,38 @@ export let counti = (fn, array) => {
   }
   count
 }
+
+export let filter = (fn, array) => {
+  let filtered = copy(array)
+  let mut position = 0
+  forEach((el) => {
+    if (fn(el)) {
+      filtered[position] := el
+      position = incr(position);
+    }
+  }, array)
+  init(position, (index) => {
+    filtered[index]
+  })
+}
+
+export let filteri = (fn, array) => {
+  let filtered = copy(array)
+  let mut position = 0
+  forEachi((el, index) => {
+    if (fn(el, index)) {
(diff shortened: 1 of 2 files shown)


stdlib/array.gr
```diff
@@ -262,3 +268,57 @@ export let count = (fn, array) => {
...
+      count = incr(count)
+      position = incr(position)
+    } else {
+      position = incr(position)
+    }
+  }
+  count
+}
+
+export let filter = (fn, array) => {
```
ospencer, line 287: I know this seems sort of annoying, but we can't really call `fn` more than once because of the possible side effects. It's really hard to write `Array.filter` efficiently, and that's probably why other functional languages don't really have it. What you could do instead is first make a copy of the array, place the filtered items at the start of the copied array, and then `init` a new array from the beginning of the copied array. It still kinda sucks because you need space for a full copy of the array, plus additional space for the array, potentially 3x the space of your array.

There's maybe the question of if you want to give your users a convenience function for something they probably shouldn't do much, as well. I don't have much of an opinion on this right now, but I'll chat with the core team and get back to you.

ospencer (commented):
Count and reduce look great! Left a note about filter, and I'll get back to you on that.

stdlib/array.gr
ospencer: We chatted, and we'll move forward with it!

phated:
I really need `Array.unique` for AoC. Super excited for this!

ospencer (approved):
Absolutely beautiful! Thanks for your hard work on this! 🎉

PR #477, opened 2020-12-15
feat: Char literals

ospencer:
This PR adds character literals to Grain. The character literal syntax is `'c'`. This introduces a breaking change, as previously single-quote strings were allowed.

### DIFF
--- a/compiler/test/test_end_to_end.re
+++ b/compiler/test/test_end_to_end.re
@@ -251,7 +251,7 @@ let basic_functionality_tests = [
   te("comp_bool2", "if (2 > true) {3} else {4}", "type"),
   te("comp_bool3", "if (true >= 4) {3} else {4}", "type"),
   te("comp_bool4", "let x = true; if (x < 4) {3} else {5}", "type"),
-  t("void", "{'foo';}", "void"),
+  t("void", "{\"foo\";}", "void"),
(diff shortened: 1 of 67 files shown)


compiler/test/test_end_to_end.re
```diff
@@ -251,7 +251,7 @@ let basic_functionality_tests = [
   te("comp_bool2", "if (2 > true) {3} else {4}", "type"),
   te("comp_bool3", "if (true >= 4) {3} else {4}", "type"),
   te("comp_bool4", "let x = true; if (x < 4) {3} else {5}", "type"),
-  t("void", "{'foo';}", "void"),
+  t("void", "{\"foo\";}", "void"),
```
phated, line 254: Opinions on using `{| |}` instead of escaping all these double quotes?
```suggestion
  t("void", {|{"foo";}|}, "void"),
```

Might look a little weird wrapping the record/block syntax

phated (approved):
Just 1 question, but looks great!

phated:
Oh, It seems like this needs an `yarn compiler format` too.

compiler/test/test_end_to_end.re
ospencer: Made https://github.com/grain-lang/grain/issues/479 to address!

PR #478, opened 2020-12-16
feat: Pattern matching on arrays

### LINKED ISSUE #212: Lang: Pattern match on Arrays
Today, I realized that we didn't have an actual issue for implementing pattern matching on Arrays (just #189 that mentions it).

This would make a lot of code that works with arrays cleaner.

ospencer:
This PR adds the ability to pattern match on Arrays.

Closes #212.

### DIFF
--- a/compiler/src/middle_end/matchcomp.re
+++ b/compiler/src/middle_end/matchcomp.re
@@ -36,6 +36,7 @@ type decision_tree =
       (constructor_tag, nested_decision_tree) tuples,
       with an optional default branch. */
     Switch(
+      switch_type,
       list((int, decision_tree)),
       option(decision_tree),
     )
@@ -56,8 +57,13 @@ type decision_tree =
 
 and matrix_type =
   | TupleMatrix(int)
+  | ArrayMatrix(int)
   | RecordMatrix(array(int))
-  | ConstructorMatrix(option(int));
+  | ConstructorMatrix(option(int))
+
+and switch_type =
+  | ConstructorSwitch
+  | ArraySwitch;
 
 type conversion_result = {
   tree: decision_tree,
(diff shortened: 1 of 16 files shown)


compiler/src/middle_end/matchcomp.re
```diff
@@ -485,6 +555,30 @@ let matrix_head_constructors = mtx => {
...
+  | TPatArray(ps) when !List.mem(List.length(ps), acc) => [
+      List.length(ps),
+      ...acc,
+    ]
+  | TPatOr(p1, p2) =>
+    pattern_head_arities_aux(p1, pattern_head_arities_aux(p2, acc))
+  | _ => acc
+  };
+
+let pattern_head_arities = patt => pattern_head_arities_aux(patt, []);
```
phated, line 569: What's the difference between an `_aux` recursive method outside the func and the `let rec help` method like below?

compiler/src/middle_end/matchcomp.re
```diff
@@ -627,6 +746,34 @@ let rec compile_matrix = mtx =>
...
+        (final_tree, [(arity, final_tree), ...switch_branches]);
+      };
+
+      switch (arities) {
+      | [] =>
+        failwith(
+          "Internal error: compile_matrix: non-wildcard column returned no arities",
+        )
+      | _ =>
+        let (result, switch_branches) =
```
phated, line 764: Am I blind or is this `result` never used?

phated (approved):
👏 🎊 You know how excited I am to have this!! Amazing work!

compiler/src/middle_end/matchcomp.re
ospencer: None, really. I just followed the pattern of `pattern_head_constructors`.

compiler/src/middle_end/matchcomp.re
ospencer: It is not. I'll get that updated.

PR #480, opened 2020-12-17
Remove use of memory.data

ospencer:
Per the research I did for #472, we can't use `memory.data` in AS. This updates our code to not use it anymore.

Could maybe optimize the env one to not need to call malloc an extra time, but this is much cleaner.


phated (approved):
LGTM! 👍

PR #482, opened 2020-12-17
fix: Multiline string locs

### LINKED ISSUE #454: Compiler: Large multiline strings seem to mess up the code locations
I've been embedding my data into my AoC problems, and the latest one is a 1000+ multi-line string. This seems to completely throw off the syntax & type error locations.

I just split it into a new file and now my locations look correct again.

ospencer:
Fixes #454


phated (approved):
❤️ you!!!

PR #484, opened 2020-12-17
fix: Multiline char locs

ospencer:
I realized that because 
```
'
'
```
is a valid character (newline), we also need to account for it.


phated (approved):
Make sense!

PR #487, opened 2020-12-17
docs: initial stdlib contributor docs

### LINKED ISSUE #464: [Docs]: Add a contributor document for our stdlib patterns
We've kinda been nailing down the stdlib patterns, so I think it's time we write it down.

phated:
This is a very rough draft of contributor docs for our standard library.

Closes #464 

@ospencer @peblair I think this needs a lot of input from you both. Maybe we can use Suggestions to collaboratively edit this!

### DIFF
--- a/docs/contributor/standard_library.md
+++ b/docs/contributor/standard_library.md
@@ -1,0 +1,28 @@
+# Standard Library
+
+Grain aims to have a comprehensive and consistent standard library.
+
+To us, this means that someone should be able to reach for the standard library to perform most of their day-to-day work, and it will always work the way they expect in the context of Grain.
+
+## Guidelines
+
+Here are some guidelines for making additions to the standard library while also keeping it high-quality and consistent!
+
+1. New data types should exist in their own modules. For example, the `Range` enum is the data type exported from `range.gr`. An exception to this are data types that are ubiquitous (`Option`/`Result`) or require compiler support (`List`).
+1. Prefer data constructors when possible. Only use separate constructor methods, like `make` or `init`, if needed to set an initialization value or to hide internals of your data structure.
+1. All methods that operate on a data type should exist in the same module as that data type. For example, all `Map` methods exist in the `map.gr` file.
+1. Fallible methods should almost always return an `Option` or `Result`. Usage of `fail` should be reserved for exeptional cases, such as argument validation producing an index out-of-bounds failure.
+    * `Option` should be preferred if you might or might not be able to get some value.
+    * `Result` is useful if you have multiple failures, or if the consumer might need additional context around the failure.
+1. If possible, keep dependencies on other standard library modules to a minimum. For example, `Array` re-implements some simple `List` operations to avoid depending on the entire `List` module.
+
+## AssemblyScript
+
+You maybe notice `.ts` files inside the `stdlib-external` directory. These are actually [AssemblyScript](https://www.assemblyscript.org/) files that we compile to WebAssembly and consume as externals.
+
+The primary reason for this is that AssemblyScript provides a nicer interface for writing WebAssembly than handcrafting it. It also gives us access to things Grain itself disallows, such as direct access into memory representations.
+
+However, our usage comes with some restrictions:
+
+* We compile it with their runtime disabled, which means we don't have access to a lot of the niceties they provide.
+* Anything that pre-populates or allocates memory cannot be used. This includes strings, even static ones.
--- a/docs/contributor/standard_library.md
+++ b/docs/contributor/standard_library.md
@@ -12,8 +12,8 @@ Here are some guidelines for making additions to the standard library while also
 1. Prefer data constructors when possible. Only use separate constructor methods, like `make` or `init`, if needed to set an initialization value or to hide internals of your data structure.
 1. All methods that operate on a data type should exist in the same module as that data type. For example, all `Map` methods exist in the `map.gr` file.
 1. Fallible methods should almost always return an `Option` or `Result`. Usage of `fail` should be reserved for exeptional cases, such as argument validation producing an index out-of-bounds failure.
-    * `Option` should be preferred if you might or might not be able to get some value.
-    * `Result` is useful if you have multiple failures, or if the consumer might need additional context around the failure.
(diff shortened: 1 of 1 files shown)


docs/contributor/standard_library.md
```diff
@@ -0,0 +1,28 @@
+# Standard Library
+
+Grain aims to have a comprehensive and consistent standard library.
+
+To us, this means that someone should be able to reach for the standard library to perform most of their day-to-day work, and it will always work the way they expect in the context of Grain.
```
ospencer, line 5: Love this.

docs/contributor/standard_library.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Guidelines
+
+Here are some guidelines for making additions to the standard library while also keeping it high-quality and consistent!
+
+1. New data types should exist in their own modules. For example, the `Range` enum is the data type exported from `range.gr`. An exception to this are data types that are ubiquitous (`Option`/`Result`) or require compiler support (`List`).
```
ospencer, line 11: `List` doesn't actually get much for compiler support—I think it lives in Pervasives for the same reasons that Option and Result do. You just wanna be able to make them/annotate using them without having to import anything.
```suggestion
1. New data types should exist in their own modules. For example, the `Range` enum is the data type exported from `range.gr`. An exception to this are data types that are ubiquitous (`Option`/`Result`/`List`), which should live in `pervasives.gr`.
```

docs/contributor/standard_library.md
```diff
@@ -0,0 +1,28 @@
...
+1. Prefer data constructors when possible. Only use separate constructor methods, like `make` or `init`, if needed to set an initialization value or to hide internals of your data structure.
+1. All methods that operate on a data type should exist in the same module as that data type. For example, all `Map` methods exist in the `map.gr` file.
+1. Fallible methods should almost always return an `Option` or `Result`. Usage of `fail` should be reserved for exeptional cases, such as argument validation producing an index out-of-bounds failure.
+    * `Option` should be preferred if you might or might not be able to get some value.
```
ospencer, line 15: Maybe we describe this as something that might be nullable in other languages?

docs/contributor/standard_library.md
```diff
@@ -0,0 +1,28 @@
...
+    * `Result` is useful if you have multiple failures, or if the consumer might need additional context around the failure.
```
ospencer, line 16: Should we add examples to these, like `List.find` should return `Option` and `Number.parseInt` should return `Result`?

ospencer (commented):
I think this is great! Thank you for putting this together 😄

docs/contributor/standard_library.md
phated: Hows that update look?

docs/contributor/standard_library.md
phated: Hows that update look?

ospencer (approved):
Updates look great!! Thanks for putting this together 😊

docs/contributor/standard_library.md
```diff
@@ -0,0 +1,28 @@
...
+Grain aims to have a comprehensive and consistent standard library.
+
+To us, this means that someone should be able to reach for the standard library to perform most of their day-to-day work, and it will always work the way they expect in the context of Grain.
+
+## Guidelines
+
+Here are some guidelines for making additions to the standard library while also keeping it high-quality and consistent!
+
+1. New data types should exist in their own modules. For example, the `Range` enum is the data type exported from `range.gr`. An exception to this are data types that are ubiquitous (`Option`/`Result`/`List`), which should live in `pervasives.gr`.
+1. Prefer data constructors when possible. Only use separate constructor methods, like `make` or `init`, if needed to set an initialization value or to hide internals of your data structure.
```
peblair, line 12:
```suggestion
1. Prefer data constructors when possible. Only use separate constructor functions, like `make` or `init`, if needed to set an initialization value or to hide internals of your data structure.
```
Grain doesn't have methods; we only have functions.

docs/contributor/standard_library.md
```diff
@@ -0,0 +1,28 @@
...
+1. All methods that operate on a data type should exist in the same module as that data type. For example, all `Map` methods exist in the `map.gr` file.
```
peblair, line 13:
```suggestion
1. All functions that operate on a data type should exist in the same module as that data type. For example, all `Map` methods exist in the `map.gr` file.
```

docs/contributor/standard_library.md
```diff
@@ -0,0 +1,28 @@
...
+1. Fallible methods should almost always return an `Option` or `Result`. Usage of `fail` should be reserved for exeptional cases, such as argument validation producing an index out-of-bounds failure.
```
peblair, line 14:
```suggestion
1. Fallible functions should almost always return an `Option` or `Result`. Usage of `fail` should be reserved for exceptional cases, such as argument validation producing an index out-of-bounds failure.
```

docs/contributor/standard_library.md
```diff
@@ -0,0 +1,28 @@
...
+    * `Option` should be preferred if you might or might not be able to get some value. Typically, these would be methods that could return `null` in other languages when they didn't produce a value. For example, `List.find` returns `None` when no item in the list matches the condition.
```
peblair, line 15:
```suggestion
    * `Option` should be preferred if you might or might not be able to get some value. Typically, these would be functions that could return `null` in other languages when they didn't produce a value. For example, `List.find` returns `None` when no item in the list matches the condition.
```

docs/contributor/standard_library.md
```diff
@@ -0,0 +1,28 @@
...
+    * `Result` is useful if you have multiple failures, or if the consumer might need additional context around the failure. Typically, these methods would throw exceptions in other languages. For example, `Number.parseInt` returns `Err(reason)` when it fails to parse a string into an integer.
```
peblair, line 16:
```suggestion
    * `Result` is useful if you have multiple failures, or if the consumer might need additional context around the failure. Typically, these functions would throw exceptions in other languages. For example, `Number.parseInt` returns `Err(reason)` when it fails to parse a string into an integer.
```

peblair (changes requested):
This is great! Just a minor semantic nit and one typo.

phated:
@peblair good catch! All suggestions applied 😄

PR #491, opened 2020-12-19
feat(stdlib): add Stack module

zweimach:
I realized we don't have a Stack module in the standard library.
Even though it's just a wrapper around List, it mirrors the Queue module.


ospencer (approved):
This looks great! I'm excited to land type aliases so we can just write `type Stack<a> = List<a>`. More work for me to do!

PR #492, opened 2020-12-19
chore: add git hooks to lint commit messages

zweimach:
Lately, we've been using [Conventional Commits][1] style for git commit messages.
We should [enforce][4] this style to make commit messages more meaningful, and we can use it to generate CHANGELOGs.

[husky][2] can also be used with [lint-staged][3] to lint and format the staging area before commits.

@ospencer, please discuss this with the core team.

[1]: https://www.conventionalcommits.org/en/v1.0.0/
[2]: https://github.com/typicode/husky
[3]: https://github.com/okonet/lint-staged
[4]: https://commitlint.js.org/#/

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,5 +21,20 @@
     "compiler": "yarn workspace @grain/compiler run",
     "stdlib": "yarn workspace @grain/stdlib run",
     "postcompiler": "yarn stdlib clean"
+  },
+  "devDependencies": {
+    "@commitlint/cli": "^11.0.0",
+    "@commitlint/config-conventional": "^11.0.0",
+    "husky": "^4.3.6"
+  },
+  "husky": {
+    "hooks": {
+      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
+    }
+  },
+  "commitlint": {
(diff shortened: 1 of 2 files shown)


phated:
We've discussed this before. I'm extremely against git hooks as they cause lag when executing commands and reduce the likelihood someone will contribute because they get frustrated by the rules (I've given up on contributions due to hooks).

As for conventional commits, that's just a thing we do when merging into master and don't expect anyone to write them when contributing.

zweimach:
@phated, thanks for the explanation.
If that is the rationale from the core team, I will close this.

phated:
I think our general rule is to keep friction of contributions to a minimum.

I tried really hard to avoid our refmt error in CI, but GH Actions doesn't allow us to run refmt against external PRs 😞

ospencer:
It's not so bad for us! GitHub prompts us to edit the commit message when we squash and merge, so it's always a great opportunity to make sure the commit is semantic. 

It's maybe not a bad idea to have CI run on master that catches if we ever don't do it though, so we can fix it.

PR #495, opened 2020-12-19
chore(runtime)!: Print strings without quotes

ospencer:
This makes it so `print("Hello, world!")` creates the program output `Hello, world!` instead of `"Hello, world!"`. As an added bonus, now `str == toString(str)` now holds (as it did not before).

Compound structures, i.e. `("Hello", 5)` still print with quotes.

### DIFF
--- a/stdlib/stdlib-external/runtime/string.ts
+++ b/stdlib/stdlib-external/runtime/string.ts
@@ -113,7 +113,7 @@ function grainListToString(ptr: u32, extraIndents: u32): u32 {
         decRef(oldRet)
       }
       isFirst = false
-      let itemString = grainToStringHelp(load<u32>(cur, 4 * 5), extraIndents)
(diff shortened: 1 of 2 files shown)


stdlib/stdlib-external/runtime/string.ts
```diff
@@ -546,7 +555,7 @@ function grainHeapValueToString(ptr: u32, extraIndents: u32): u32 {
   }
 }
 
-function grainToStringHelp(grainValue: u32, extraIndents: u32): u32 {
+function grainToStringHelp(grainValue: u32, extraIndents: u32, top: bool): u32 {
```
peblair, line 558: Maybe a better name than `top` here? I know for a fact in 6 months I won't have any idea what this means

peblair (changes requested):

peblair (approved):

PR #496, opened 2020-12-19
refactor(runtime): Move CharCode into its own file as a namespace

ospencer:


phated (approved):
Nice! I like this cleanup!

PR #498, opened 2020-12-20
chore: add regression test for #350

### LINKED ISSUE #350: Weird behavior with maps
Thanks, 2 notes about the implementation:
1. Removing elements inside a `forEach` caused the program to hang, I don't know if it's considered a limitation of the API or a bug that should be fixed.
2. In the tests, when I tried creating the `filterTestMap` using the following function the program hanged:
```
let makeFilterTestMap = () => {
  let map = Map.make();
  Map.set(Grain, "g", map);
  Map.set(Sheep, "s", map);
  Map.set(Brick, "b", map);
  map
}
```
Any idea why?

_Originally posted by @ohana54 in https://github.com/grain-lang/grain/pull/344#issuecomment-703132144_

phated:
Just a regression test now that the decRef bug is fixed.

Closes #350

### DIFF
--- a/compiler/test/stdlib/map.test.gr
+++ b/compiler/test/stdlib/map.test.gr
@@ -316,6 +316,24 @@ Map.filter((key, value) => true, filterTestMap);
 
 assert Map.size(filterTestMap) == 3;
 
+// Regression tests for https://github.com/grain-lang/grain/issues/350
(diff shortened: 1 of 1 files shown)


ospencer (approved):
🎉

PR #499, opened 2020-12-20
Regression tests for 497

### LINKED ISSUE #497: [bug] Resize fails with Map/Set on larger data sets
This is mostly a tracking issue so I can create & link regression tests.

When I try to create a Map (or Set) with size that grows past a certain number of elements, the program just hangs.

It was determined that this happens at the `resize` method used to grow the buckets. It was also determined that it was related to a decRef bug that also caused #152 and #350 (which is now fixed in master).

I will be created a regression test to check resize of large lists works.

phated:
This adds regression tests for Map & Set related to #497 - I needed to add a `getInternalStats` to Set and I copied the Map tests.

Closes #497


ospencer (approved):
Looks good!

PR #500, opened 2020-12-20
fix(runtime): Properly decRef record and array values

ospencer:


phated (approved):
Excellent! What's the best way to have a test for this?

runtime/src/core/memory.js
```diff
@@ -417,15 +428,33 @@ export class ManagedMemory {
...
+          case GRAIN_ARRAY_HEAP_TAG: {
+            let x = genericHeapValUserPtr / 4;
+            let arity = view[x + 1];
+            trace(`traversing ${arity} array vals on array 0x${this._toHex(userPtr)}`);
+            for (let i = 0; i < arity; ++i) {
+              yield view[x + 2 + i];
+            }
+            break;
+          }
           default:
```
peblair, line 458: Maybe we should emit some sort of warning here for unknown tags to prevent bugs like this in the future?

peblair (approved):

ospencer:
@phated The best way to have a test for this is to just add a GC test that verifies that after arrays and records are allocated, their memory is reclaimed. I believe that I've found a bug in the way we currently write GC tests that doesn't allow me to properly write one right now. I'm going to open an issue to address it, though.

ospencer:
https://github.com/grain-lang/grain/issues/504

PR #501, opened 2020-12-22
Better lenses and hover for LSP

ng-marcus:
Updated LSP mode to return the full set of statements, patterns and expressions in a JSON format to the LSP server.

This is much better than the previous version as I use the built-in typedtreeiterator to get what I need (so all the hard work was done for me thanks!)

Values are returned in a new field in the results (values) and the old field (lenses) is returned as empty so older LSP clients just report empty values and don't potentially crash.

One question - I build my own type description for the record type and just return "Variant" for the hover for variants.  What would you like to see here?  Having the hover show the whole record or variant is a little strange as that's just a repeat of the code, but it's what Reason does.  Let me know and I'll update the PR

### DIFF
--- a/compiler/src/diagnostics/lenses.re
+++ b/compiler/src/diagnostics/lenses.re
@@ -1,6 +1,6 @@
 open Grain_typed;
 
-// Left as a string not a variant as this is opaque to the 
+// Left as a string not a variant as this is opaque to the
 // work done here and gets parsed into a string immediately anyway
 
 // t:
@@ -11,14 +11,14 @@ open Grain_typed;
 
 [@deriving yojson]
 type lens_t = {
-  sl: int,  // start line
-  sc: int,  // start character
-  sb: int,  // start BOL
-  el: int,  // end line
-  ec: int,  // end character
-  eb: int,  // end BOL
+  sl: int, // start line
+  sc: int, // start character
+  sb: int, // start BOL
+  el: int, // end line
+  ec: int, // end character
+  eb: int, // end BOL
   s: string, // type signature
-  t: string,  // lens stype
+  t: string // lens stype
 };
 
 let get_raw_pos_info = (pos: Lexing.position) => (
@@ -142,7 +142,6 @@ let print_tree = (stmts: list(Grain_typed.Typedtree.toplevel_stmt)) => {
 
       lenses := List.append(lenses^, [lens]);
       Iterator.iter_toplevel_stmt(cur);
-      
     },
     stmts,
   );
--- a/compiler/src/diagnostics/lenses.re
+++ b/compiler/src/diagnostics/lenses.re
@@ -69,16 +69,18 @@ let create_record_signature = (labels: list(Types.record_field)) => {
         let ppf = Format.formatter_of_buffer(buf);
         Printtyp.type_expr(ppf, field.rf_type);
         Format.pp_print_flush(ppf, ());
-        acc
-        ++ comma
-        ++ Ident.name(field.rf_name)
-        ++ " : "
-        ++ Buffer.contents(buf);
+        Printf.sprintf(
+          "%s%s%s: %s",
+          acc,
+          comma,
+          Ident.name(field.rf_name),
+          Buffer.contents(buf),
+        );
       },
       "",
       labels,
     );
-  "{" ++ recSig ++ " }";
+  Printf.sprintf("{%s }", recSig);
 };
 
 let data_hover_val =
--- a/compiler/src/diagnostics/lenses.re
+++ b/compiler/src/diagnostics/lenses.re
@@ -4,22 +4,33 @@ open Grain_typed;
 type lens_t = {
   sl: int,
   sc: int,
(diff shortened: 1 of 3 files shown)


compiler/src/diagnostics/lenses.re
```diff
@@ -2,76 +2,154 @@ open Grain_typed;
 
 [@deriving yojson]
 type lens_t = {
-  line: int,
-  signature: string,
+  sl: int,
```
ospencer, line 5: I think it'd be really good to add a comment here that explains what each of these fields are.

compiler/src/diagnostics/lenses.re
```diff
@@ -2,76 +2,154 @@ open Grain_typed;
...
+  let recSig =
+    List.fold_left(
+      (acc, field: Types.record_field) => {
+        let comma = acc == "" ? "" : ", ";
+        let buf = Buffer.create(64);
+        let ppf = Format.formatter_of_buffer(buf);
+        Printtyp.type_expr(ppf, field.rf_type);
+        Format.pp_print_flush(ppf, ());
+        acc
+        ++ comma
```
ospencer, line 64: Maybe it might be better to use `Printf.sprintf` here?

compiler/src/diagnostics/lenses.re
```diff
@@ -2,76 +2,154 @@ open Grain_typed;
...
+        ++ Ident.name(field.rf_name)
+        ++ " : "
+        ++ Buffer.contents(buf);
       },
-      program.statements,
+      "",
+      labels,
     );
+  "{" ++ recSig ++ " }";
```
ospencer, line 72: And then for this too?

compiler/src/diagnostics/lenses.re
```diff
@@ -2,76 +2,154 @@ open Grain_typed;
...
+
+      lenses := List.append(lenses^, [lens]);
+      switch (cur.ttop_desc) {
+      | TTopException(_)
+      | TTopForeign(_)
+      | TTopImport(_)
+      | TTopExport(_)
+      | TTopExpr(_)
+      | TTopLet(_) => Iterator.iter_toplevel_stmt(cur)
+      // this is separate as the original emitted an enter data
```
ospencer, line 142: What does this mean?

compiler/src/diagnostics/output.re
```diff
@@ -39,9 +40,11 @@ let exn_to_lsp_error = (exn: exn): option(lsp_error) => {
   };
 };
 
+// kept lenses as an empty array so older LSPs don't crash
```
ospencer, line 43: I'm not super concerned about this at this stage since we're still iterating rapidly, but it's good to think ahead 🙂

ospencer (changes requested):
This looks really good! Left a couple of minor notes but otherwise this should be good to go!

compiler/src/diagnostics/lenses.re
ng-marcus: Where I copied this code from, it emitted values for entering and leaving a TTopData .  I don't need it, so wanted to check with you, I'm going to simplify this switch away, I can always revert to it if the data is needed in the future.

ospencer: Gotcha!

compiler/src/diagnostics/lenses.re
ng-marcus: Would love to!  Can you give me a code sample?  This pretty printer stuff confuses the heck of out me!

compiler/src/diagnostics/lenses.re
```diff
@@ -1,77 +1,156 @@
...
+        let comma = acc == "" ? "" : ", ";
+        let buf = Buffer.create(64);
+        let ppf = Format.formatter_of_buffer(buf);
+        Printtyp.type_expr(ppf, field.rf_type);
+        Format.pp_print_flush(ppf, ());
+        acc
+        ++ comma
+        ++ Ident.name(field.rf_name)
+        ++ " : "
+        ++ Buffer.contents(buf);
```
ospencer, line 76: Nothing too serious using the `Format` module or anything! You'll probably need to `yarn compiler format` again, but this is all I meant:
```suggestion
        Printf.sprintf("%s%s%s: %s", acc, comma, Ident.name(field.rf_name), Buffer.contents(buf));
```

compiler/src/diagnostics/lenses.re
ospencer:
```suggestion
  Printf.sprintf("{%s }", recSig);
```

compiler/src/diagnostics/lenses.re
ng-marcus: Comments added.

compiler/src/diagnostics/lenses.re
ng-marcus: Oh!  Yes, much nicer!

ospencer (approved):
This is huge!

PR #502, opened 2020-12-26
Runtime, CLI: Format JS with Prettier

ospencer:
Was honestly just too lazy to set this up years ago as it should have been done.

Adds `yarn runtime format` and `yarn cli format` to format the JS code. Should work with your editor so you shouldn't ever have to actually run the commands, but they're there if you need them.


phated (approved):
Seems solid! Not sure what the conflict is.

PR #503, opened 2020-12-30
Hover displays the enum and record name

ng-marcus:
I updated the hover provider slightly

it now returns.  enum <TypeName> for an enum and record <RecordName> { fields }

I'm not happy with either - parsing over the enum to produce the full enum could work, but I'm not sure what that would provide over the source code below.  Then again, that's what I've done for records...  What do you think?


ospencer:
@ng-marcus Usually when I'm writing OCaml or Reason code and I hover over an enum type, I _do_ want see all of the variants. It helps me remember what the type is and what the possible variants are. So I'd vote to produce the full enum 🙂

ospencer:
It's sort of the same way if you hovered over a module, you'd want to see the full module signature.

ng-marcus:
Sounds good, I'll get on !

ospencer:
@ng-marcus We've swapped our default branch from `master` to `main`. I'm going to rebase your branch and recreate the PR. Sorry for the blip!

ng-marcus:
No worries, and I'll get this done too!

Issue #504, opened 2021-01-02
Runtime: Find better way to test GC performance


ospencer:
Currently, to test that GC performs correctly, we use the flag `--limitMemory` to artificially limit the WebAssembly memory size below the 64k WebAssembly page size. We then have some tests that do some allocating and deallocating and then we verify that the program does not run out of memory.

Our artificial limit doesn't do enough, and the memory manager understands that there is more memory, so it does not attempt to immediately reuse the blocks that have just been freed. As such, our tests don't currently test the functionality we want to test.

Instead of trying to make `--limitMemory` smarter, I think we should explore other options. The solution I have in mind right now is changing `--limitMemory` to limit the number of WebAssembly pages used. A test that wants to constrict the amount of usable memory to 1k could just `malloc(64512)` and set `--maxPages=1`.

PR #510, opened 2021-01-16
feat: Use real wasm tail call instruction

### LINKED ISSUE #506: Compiler: Invalid tail call optimisation
The `Optimize_tail_calls` pass changes the behaviour of the following program:
```
let rec f = x => {
 let g = x => f(x - 1)
  if (x == 0) {x} else {f(g(x))}
}

print(f(1))
```
With all optimisations enabled, this gives a memory access out of bounds error. Rebuilding the compiler with `Optimize_tail_calls` removed from `optimization_passes`, the program prints 0 as expected. I think the `f(x - 1)` call is being tail call optimised, despite `g` not being rewritten as 2 functions.

ospencer:
Recreation of #488.

Closes #506.

This is actually a feature instead of a chore because now tail calls work across modules.


compiler/src/codegen/transl_anf.re
```diff
@@ -589,7 +596,14 @@ let rec compile_comp = (env, c) => {
       )
     | CLambda(args, body) =>
       MAllocate(MClosure(compile_lambda(env, args, body, c.comp_loc)))
-    | CApp(f, args) =>
+    | CApp(f, args, true) =>
+      /* TODO: Utilize MCallKnown */
```
phated, line 600: What does this mean?

compiler/src/codegen/transl_anf.re
```diff
@@ -589,7 +596,14 @@ let rec compile_comp = (env, c) => {
...
+
+      MReturnCallIndirect(
+        compile_imm(env, f),
+        List.map(compile_imm(env), args),
+      )
+    | CApp(f, args, _) =>
```
phated, line 606: Should we match `false` instead of `_` here?

compiler/src/middle_end/optimize_tail_calls.re
```diff
@@ -1,492 +1,26 @@
...
-  * Each function is transformed into two functions-- an invocation function and an iteree function. The invocation
-    function is called when the function is called from an outside source, and controls the loop that calls the iteree.
-    The iteree function is a thunk that first binds the would-be arguments from the references to names, then proceeds
-    with the original function body with minor alterations, the most important being a replacement of tail calls with a
-    setup to continue the iteration.
-  * Further optimizations (such as using only a single function) could be done for directly-recursive functions, but
-    these are not implemented at this time.
-*/;
-
 open Anftree;
```
phated, line 1: So this file still needs to exist?

compiler/test/test_end_to_end.re
```diff
@@ -922,7 +922,7 @@ let gc = [
     "(1, 2)",
   ),
   tgcfile("fib_gc_err", 1024, "fib-gc", "Out of memory"),
-  tgcfile("fib_gc", 3072, "fib-gc", "832040"),
+  tgcfile("fib_gc", 4000, "fib-gc", "832040"),
```
phated, line 925: What does it mean that this goes up to 4000?

phated (approved):
@ospencer you are truly a wizard! This is awesome. Just a few questions/comments, but overall looks great.

compiler/src/codegen/transl_anf.re
ospencer: It should actually say use `MReturnCallKnown` 😛

But it means that we should use `return_call` when we know what the function we're calling is instead of `return_call_indirect`. We just don't have that information available in the compiler right now. There's a similar comment for non-tail `CApp` (which is where I copied this from).

compiler/src/middle_end/optimize_tail_calls.re
ospencer: Yeah. We still need to tell the compiler when we need to do a tail call.

compiler/test/test_end_to_end.re
ospencer: It's just that this doesn't generate the same code anymore (because of the tail call instruction) and that changed the amount of memory this uses. It definitely uses _less_ memory than before, but the way we test this is currently broken (see #504), and the way our testing works makes this number a little arbitrary.

ospencer: I did a binary search to find how much memory spanned—that's a better term for what this is—happed and updated it.

PR #516, opened 2021-01-25
chore: Upgrade Binaryen.ml and ocaml-lsp-server

ospencer:
@phated let me know if it's better to do this as two separate PRs or if a single dependencies PR is fine.


phated (approved):
Tests pass, seems good! 👍 I think it can be squashed into a commit with the same name as this PR.

PR #518, opened 2021-01-30
chore: Use mutable globals instead of wrapper functions for function exports

### LINKED ISSUE #131: Export lambda locations as mutable globals
We currently create a webassembly function that just returns the location of a lambda in memory when functions are exported from modules.

Since the [mutable globals proposal](https://github.com/WebAssembly/mutable-global) has [been finished](https://github.com/WebAssembly/proposals/blob/master/finished-proposals.md) we should just export the global.

This will probably require updating our wasm spec.

ospencer:
Closes #131.

Before exportable mutable globals were a thing, the only way to simulate the behavior was to have a "getter" function that could return the value from your module. This reduces overall code size and is more efficient for module startup.

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -1266,10 +1266,7 @@ let compile_bind =
         typ,
       )
     };
-  | MGlobalBind(i, wasm_ty) =>
-    /* Global bindings need to be offset to account for any imports */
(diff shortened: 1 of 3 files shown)


phated (approved):
👍

phated (approved):
Still lgtm!

PR #523, opened 2021-02-02
fix: Don't optimize away expressions which could possibly contain side effects

ospencer:
This PR greatly simplifies our purity analysis to only identify expressions that cause (or could cause) side effects. Ideally, since it's much less clever it'll be easier to maintain and contribute fewer bugs in the future. I'm not particularly worried about the loss of common subexpression elimination, since I don't think we got much out of it, and Binaryen has similar optimizations.


compiler/test/test_end_to_end.re
```diff
@@ -920,7 +918,7 @@ let oom = [
 let gc = [
   tgc(
     "gc1",
-    256,
+    512,
```
ospencer, line 921: This previously was optimized into a single call of `f`, so now it makes more allocations than before. This change isn't an indication of the GC being broken, just the testing issue described in #504.

phated (approved):
You know I love code deletion! I'm not familiar with this, though, so it'd be good to hear from @peblair

peblair (approved):

PR #524, opened 2021-02-02
chore: Refactor bindings in the AST

ospencer:
This PR refactors `let` bindings in the AST to not have bodies, i.e. work the same way that toplevel bindings do. This makes things a little cleaner (as can be seen by the chunk of code deleted from `parser_header`). As an added bonus, compilation errors are now reported in top-down order, which can be seen in the test changes.

### DIFF
--- a/compiler/src/middle_end/linearize.re
+++ b/compiler/src/middle_end/linearize.re
@@ -409,18 +409,15 @@ let rec transl_imm =
     let (rest_ans, rest_setup) =
       transl_imm({...e, exp_desc: TExpBlock(rest)});
     (rest_ans, fst_setup @ [BSeq(fst_ans)] @ rest_setup);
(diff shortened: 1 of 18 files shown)


phated:
@ospencer does this enable/take away anything else?

ospencer:
@phated Nope! This is purely cleanup + a minor semantic error-reporting fix. This is something I've wanted to fix for a while and decided to do it now before fixing another bug. Turned out though that I'd already fixed that other bug some time ago.

Other than the error reporting fix, I think the main thing we get out of this is having our AST make more sense. If you look at the parser change, we were more or less warping what we parsed into something else (for _very_ legacy reasons).

phated (approved):
SGTM! Since it mentioned "blocks", I just wanted to make sure it wasn't removing support for like single branch if statements.

ospencer:
Ah! Blocks still have the same behavior :)

PR #526, opened 2021-02-06
chore!: Refactor value tags

ospencer:
This PR reworks our value tags a bit:

|           | old   | new   |
|-----------|-------|-------|
| numbers   | 0bxx0 | 0bxx1 |
| tuples    | 0b001 | none  |
| closures  | 0b101 | none  |
| constants | 0b111 | 0b110 |
| generic   | 0b011 | 0b000 |

The purpose of this change is to correct a small oversight we made long ago. It increases overall performance slightly and reduces compiled module size slightly by eliminating the need to tag or untag pointers to heap values. This structure is similar to how OCaml tags values. Additionally, this change makes the tags easier to understand for FFI users, and for those of us that write low-level compiler code, makes it less likely for us to make mistakes.

This is a breaking change for FFI users, but will hopefully make writing FFI bindings much easier in the future.

### DIFF
--- a/compiler/src/codegen/comp_utils.re
+++ b/compiler/src/codegen/comp_utils.re
@@ -8,7 +8,7 @@ let wasm_type =
   | F32Type => Type.float32
   | F64Type => Type.float64;
 
-let encoded_int32 = n => n * 2;
+let encoded_int32 = n => n * 2 + 1;
 
 let const_int32 = n => Literal.int32(Int32.of_int(n));
 let const_int64 = n => Literal.int64(Int64.of_int(n));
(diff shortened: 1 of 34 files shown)


peblair (commented):
I'm generally not opposed to this (I agree that not having to tag heap pointers is a plus), but do we really want to introduce the "generic heap value" layer of indirection to tuples and (especially) lambdas? I always felt that they deserved to be "privileged" values to avoid the overhead of the indirection. What's the rationale?

Also, this is a breaking change for FFI users, but it also breaks binary compatibility with previously-compiled Grain code, no?

ospencer:
@peblair For tuples and lambdas there's actually _less_ indirection now, since you don't need to untag the pointer. Because of WebAssembly load/store offsets, it's no less efficient than before. In the case of GC/print/equals you could maybe argue that the load for the tag makes it slightly less efficient, but again you don't need to spend instructions doing untagging.

Were you looking for more rationale other than what was in the PR description?

As for binary compatibility, yes, this is a breaking change.

peblair:
I see; I may have misunderstood. So, to be clear, we are going to have the following representations for these values, correct?
```
before:
<tuple_val>
      |
      v
| LENGTH | elts... |

<lambda_val>
      |
      v
| arity | other stuff ... |

after:
<tuple_val>
      |
      v
| TUPLE_TAG | LENGTH | elts... |

<lambda_val>
      |
      v
| LAMBDA_TAG | arity | other stuff ... |
```
If so, I think I understand now, and this looks good. Apologies for the confusion.

ospencer:
Yup, that's exactly right. Probably not easy to see in the diff, but I added that to the data representations doc, and updated the info around tags. Let me know if you think I could have made it clearer.

peblair (approved):

Issue #196, opened 2020-07-07
Enable Binaryen optimizations


ospencer:
Compilation breaks if Binaryen optimization is enabled. This is probably a problem with the Binaryen bindings and will need to be fixed there.

https://github.com/grain-lang/grain/blob/binaryen/compiler/src/codegen/compcore.re#L3956

PR #528, opened 2021-02-07
chore: Clean up function tables

ospencer:
We were putting way more than we needed to in our function tables, so this PR cleans that up. It also makes processing of function names a bit cleaner.

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -16,11 +16,9 @@ let memory_tracing_enabled = false;
 
 type codegen_env = {
   num_args: int,
-  func_offset: int,
   global_offset: int,
   stack_size,
   import_global_offset: int,
-  import_func_offset: int,
   import_offset: int,
   /* Allocated closures which need backpatching */
   backpatches: ref(list((Expression.t, closure_data))),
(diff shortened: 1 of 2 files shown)


peblair:
@ospencer Can you elaborate a little more on what is being removed here? It's not immediately obvious to me based on a read of the diff.

ospencer:
@peblair Previously, we put _all_ functions that appeared in a module in the function table. This includes function imports, functions defined by the module, `_start`, and other auxiliary functions. Really, all that needs to go in the table are the functions defined by the module that a closure will have a pointer to. All Grain function imports are already in the function table because they were put there by the module they were defined in. External functions don't need to be placed in the table because the version of the function that Grain creates a closure for is put in the table.

This is why we were seeing a sort of ballooning of our function table size, since every module duplicated the tables 🤦. Now the table size is equivalent to the number of functions you defined in your module.

The other huge benefit here is for Binaryen optimizations—since we put all imports in the function table, Binaryen couldn't remove unused imports 😭.

ospencer:
(Pending #196 of course.)

peblair (approved):
Looks like this will help a lot with compilation efficiency!

PR #529, opened 2021-02-07
feat: Export compiled functions from modules

ospencer:
This PR makes  (exported) Grain compiled functions have wasm function exports, i.e. if you compile the program `export let foo = x => 6` you'll see this in the compiled output:

image
image

This does two important things: 1) other languages (or the Grain compiler) can call Grain functions directly, as opposed to via loading the function pointer from the closure and using `call_indirect`, and 2) debugging compiled wasm output is much easier because the functions have their original source names.

There's not really a great way to write a test for this, but this functionality will be exercised heavily in the coming PRs.

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -3395,12 +3395,15 @@ let compile_imports = (wasm_mod, env, {imports}) => {
 };
 
 let compile_exports = (wasm_mod, env, {functions, imports, exports, globals}) => {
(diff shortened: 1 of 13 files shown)


peblair (approved):
@ospencer So, to be clear, the use case here is primarily for people interoping with Grain code, correct? That's my understanding, and, if so, this looks good (so, conditionally approved, but kick back to me if the answer is not "yes").

ospencer:
@peblair Yes, us included.

PR #532, opened 2021-02-13
feat: Allow pattern matching on unsafe wasm i32s

ospencer:
It's not necessarily _hard_ to get i64/f32/f64s working, but it's not a feature we desperately need right now.


phated (approved):
Whoa!!! 🤯

PR #533, opened 2021-02-17
fix: loop body typechecking

ospencer:
Similar to #531.


ospencer:
I realized this after I put up the PR—we don't actually have to restrict the type of the loop body to `Void`, and we can have the same behavior as before. All that's necessary is is the extra check that the entire loop evaluates to `Void`.

phated (approved):
Oh, heck yeah! 🎊

peblair (approved):

PR #534, opened 2021-02-17
feat: Grain implementation of memory manager

### LINKED ISSUE #504: Runtime: Find better way to test GC performance

ospencer:
Also closes #504.

This PR reimplements the Grain memory manager from JavaScript -> Grain.

That's pretty much all there is to it. Because of the Grain implementation, there was a fairly easy way to solve #504, which is in this PR as well.


compiler/src/codegen/compcore.re
```diff
@@ -1889,15 +1358,41 @@ let heap_allocate = (wasm_mod, env, num_words: int) =>
...
+          const_int32(round_to_even(num_words + 2) * 4),
         ),
       );
     Expression.tuple_extract(
       wasm_mod,
       Expression.tuple_make(
         wasm_mod,
         [
-          compile_bind(wasm_mod, env, ~action=BindGet, runtime_heap_ptr),
+          Expression.block(
```
ospencer, line 1369: This has changed to add a fake GC header to everything allocated in runtime land so values allocated here can work in GC land. The reference count is set to 1, and shouldn't ever fall below 1.

compiler/test/input/long_lists.gr
```diff
@@ -20,4 +20,4 @@ loop = (n) => {
   }
 };
 
-loop(45) // <- eats up a lot of heap
+loop(25) // <- eats up a lot of heap
```
ospencer, line 23: This was reduced because the test previously asked for 70,000 bytes of memory, but WebAssembly pages are only ~65,000 bytes. The test was reduced to only asking for 20,000 bytes.

ospencer: I suppose we could rework the test runner further to test with multiple pages, but I feel that can be saved for another PR.

compiler/test/test_end_to_end.re
```diff
@@ -21,12 +21,41 @@ let t = (~todo=?, name, program, expected) =>
...
+    import Memory from "runtime/unsafe/memory"
+    @disableGC
+    let leak = () => {
+      // find current memory pointer, subtract space for two malloc headers + 1 GC header
+      let offset = WasmI32.sub(Memory.malloc(8n), 24n)
+      // Calculate how much memory is left
+      let availableMemory = WasmI32.sub(offset, Malloc._RESERVED_RUNTIME_SPACE)
+      // Calculate how much memory to leak
+      let toLeak = WasmI32.sub(availableMemory, |}
+  ++ string_of_int(heap_size)
```
phated, line 38: Is it too nitpicky to want this to be a Sprintf on against the multiline string?

stdlib/runtime/gc.gr
```diff
@@ -0,0 +1,137 @@
...
+  } else {
+    userPtr
+  }
+}, decRefChildren = (userPtr: WasmI32) => {
+  match (WasmI32.load(userPtr, 0n)) {
+    t when t == Tags._GRAIN_ADT_HEAP_TAG => {
+      let arity = WasmI32.load(userPtr, 16n)
+      let maxOffset = arity * 4n
+      for (let mut i = 0n; WasmI32.ltU(i, maxOffset); i = i + 4n) {
+        decRef(WasmI32.load(userPtr + i, 20n), false);
```
phated, line 98: if these (like `20n`) are always `4n` more than the number used in `arity` (like `16n`), why are they hardcoded instead of calculated?

stdlib/runtime/gc.gr
```diff
@@ -0,0 +1,137 @@
...
+        decRef(WasmI32.load(userPtr + i, 16n), false);
+      }
+    },
+    t when t == Tags._GRAIN_TUPLE_HEAP_TAG => {
+      let arity = WasmI32.load(userPtr, 4n)
+      let maxOffset = arity * 4n
+      for (let mut i = 0n; WasmI32.ltU(i, maxOffset); i = i + 4n) {
+        decRef(WasmI32.load(userPtr + i, 8n), false);
+      }
+    },
```
phated, line 128: Exactly the same as _GRAIN_ARRAY_HEAP_TAG block above?

stdlib/runtime/malloc.gr
```diff
@@ -135,10 +133,7 @@ let growHeap = (nbytes: WasmI32) => {
  * @param ap: WasmI32 - The pointer to free
  */
 export let free = (ap: WasmI32) => {
-  // Let memory manager know we're freeing this pointer
-  extFree(ap)
-
-  let mut blockPtr = ap - 16n // 8 bytes for malloc header + 8 bytes for GC header
+  let mut blockPtr = ap - 8n // 8 bytes for malloc header
```
phated, line 136: Why is the 8 bytes for GC header no longer needed?

phated (approved):
Holy 🐮 this is a ton of code deletion! Great work @ospencer, love it!

compiler/src/codegen/compcore.re
ospencer: Yep, exactly. The problem that cropped up was GC-managed files would try to incRef/decRef runtime values, but since they didn't have GC headers, this was casually trashing runtime values 🙃 

The "fakeness" comes from the fact that calling `free()` on a runtime value wouldn't be _great_. Nothing would happen, except that a useless block would be added to the free list. (Size would be 0, so it would always be skipped.)

compiler/test/runner.re
```diff
@@ -108,26 +108,32 @@ let read_stream = cstream => {
...
   let testlibs = Sys.getcwd() ++ "/test-libs";
   let result = ref("");
-  let heap_args =
-    switch (heap_size) {
-    | Some(x) => ["--limitMemory", string_of_int(x)]
-    | None => []
+  let env =
+    switch (num_pages) {
+    | Some(x) =>
+      Array.append(Unix.environment()) @@
```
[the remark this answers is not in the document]
ospencer: Yeah. It already does this by default. If you _don't_ do this when you're trying to add to it, the tests can't even find `grain` 😂

compiler/test/test_end_to_end.re
ospencer: Nope, not at all. For some reason, I thought this was going to be more readable than a `sprintf` but that doesn't make any sense.

stdlib/runtime/gc.gr
ospencer: tl;dr the offset _has_ to be a constant and I didn't write an optimization that precomputes the additions, subtractions, multiplications, etc., so it breaks/complains if you don't provide a constant. 

(That's not _100%_ true, because I made it so if you don't do a constant it just manually adds the offset to your pointer, but that's bad for performance + messes with the instruction inlining).

We should write that optimization though (we already have the framework for it) and potentially go back and fix these.

stdlib/runtime/malloc.gr
ospencer: Ah yeah, I should have said something about this. Back in the day, this is how it worked, but it changed when I ported `malloc` to Grain. This PR restores the original behavior where this malloc doesn't (and shouldn't) have any knowledge of the GC, and the GC version of malloc (that calls this malloc) adds in the space for the GC header.

compiler/test/runner.re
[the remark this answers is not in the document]
phated: Ah!

stdlib/runtime/gc.gr
phated: Worth opening an issue for that optimizations so we don't lose it?

ospencer: I remembered why I didn't do that—Binaryen will do it. However, Binaryen won't allow you to have a non-constant offset. I didn't think it was worth it to duplicate that optimization _just_ to support this. I suppose it's worth it if it makes the code more readable. What do you think?

compiler/src/codegen/compcore.re
```diff
@@ -500,393 +301,80 @@ let call_malloc = (wasm_mod, env, args) => {
...
-    arg;
-  } else {
-    Expression.call(
-      wasm_mod,
-      get_imported_name(runtime_mod, incref_ident),
-      [arg],
-      Type.int32,
-    );
-  };
-let call_incref_adt = (wasm_mod, env, arg) =>
```
peblair, line 526: I perhaps would have left these variant functions as aliases to `call_incref`, since we could inject debugging information in the future.

compiler/src/typed/typemod.re
```diff
@@ -826,13 +826,8 @@ let type_module = (~toplevel=false, funct_body, anchor, env, sstr /*scope*/) =>
...
 
-let implicit_modules: ref(list((string, string, bool))) = (
-  ref([
-    ("Pervasives", "pervasives", true),
-    ("Malloc", "runtime/malloc", false),
-  ]):
-    ref(list((string, string, bool)))
-);
+let implicit_modules: ref(list((string, string, bool))) =
+  ref([("Pervasives", "pervasives", true), ("GC", "runtime/gc", false)]);
```
peblair, line 830: This makes me think: doesn't this mean that all modules, even in `Normal` mode, can just casually call `GC` functions? I suppose it's fine, but it gives me an odd gut reaction.

compiler/test/runner.re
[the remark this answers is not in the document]
peblair: I feel like this could affect reproducibility in the future (even if this has been the behavior before this PR). Maybe we should just find the appropriate `grain` from the `PATH` variable and then build a minimal environment from that?

stdlib/runtime/gc.gr
peblair: What about defining the values as constants at the top of the file (which are computed)? Something like
```
let _ADT_ARITY_OFFSET = 16n
let _ADT_SLOT_OFFSET = _ADT_ARITY_OFFSET + 4n
```
or will that not be sufficiently optimized either? This way would at least let readers understand where the magic numbers are coming from.

peblair (approved):
This looks good! Please review comments before merging.

compiler/test/runner.re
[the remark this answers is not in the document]
ospencer: https://github.com/grain-lang/grain/issues/548

stdlib/runtime/gc.gr
ospencer: Yeah, that won't get optimized either.

compiler/src/typed/typemod.re
ospencer: Oh, no they can't. That `false` at the end instructs the loader to not put any of the values in the environment.

PR #536, opened 2021-02-20
fix: and/or short circuiting

ospencer:
This looks like it broke when we made the primitives proper functions. I added support to transl_prim that makes sure this works if the primitive is added directly to a Grain program (as opposed to only importing from `Pervasives`), which is really useful for the runtime rewrite.

After the numbers rewrite, we can extend this for all primitives—that'll ensure that the shadowing behavior and constant folding optimization always works as expected. As an added bonus, it means that all of the primitives will be inlined (if we want them to be)!


phated (approved):
Hooray!!! So glad this will work as I expect now. 🙇 Love the tests, too ❤️

PR #538, opened 2021-02-21
feat: Grain implementation of equals

ospencer:


compiler/src/codegen/compcore.re
```diff
@@ -251,9 +261,17 @@ let grain_function_imports = [
...
   {
-    mimp_mod: stdlib_external_runtime_mod,
+    mimp_mod: equal_mod,
     mimp_name: equal_ident,
-    mimp_type: MFuncImport([I32Type, I32Type], [I32Type]),
+    mimp_type: MFuncImport([I32Type, I32Type, I32Type], [I32Type]),
+    mimp_kind: MImportWasm,
+    mimp_setup: MSetupNone,
+  },
+  // HACK: Depend on stdlib-external/runtime for the tests
```
ospencer, line 270: I'm actually not really sure what we're going to do about this—the only thing that always depends on `toString` is `Pervasives`, and programs that don't depend on `Pervasives` won't have loaded this module. This hack also only works until #196 lands.

I think I'd probably want a lightweight version of this implemented solely in JS, and any test that wants to test printing enums/records has to call `print` explicitly.

There's also a solid chance that we'll need to change how `_start` works to not return a value at all, which makes this a bit of a moot point as most tests will need to be rewritten to use `print` or some other technique.

phated: We're getting optimizations now, right? So what are we going to do?

stdlib/pervasives.gr
```diff
@@ -75,7 +77,7 @@ primitive fail : String -> a = "@fail"
 import foreign wasm print : a -> Void from "stdlib-external/runtime"
 
 // Checks the given items for structural equality.
-import foreign wasm equal : (a, a) -> Bool as (==) from "stdlib-external/runtime"
+export (==)
```
phated, line 80: We already `export *` at top?

stdlib/runtime/equal.gr
```diff
@@ -0,0 +1,209 @@
+/* compilation-mode: runtime */
```
phated, line 1: Mostly taking your word for this conversion. I don't have the energy to try to line up the AssemblyScript side-by-side with the Grain code.

phated (approved):

compiler/src/codegen/compcore.re
ospencer: I actually managed to resolve this in #540.

stdlib/pervasives.gr
ospencer: Doesn't work on imports.

peblair (approved):

PR #539, opened 2021-02-23
feat: Optimize number to string conversions by writing UTF8 directly

### LINKED ISSUE #475: [Optimization] UTF-8 number-to-string Implementations
To ease implementation of our WASM-based runtime, we modified the AssemblyScript implementations of `dtoa` (double-to-ASCII) and `itoa` (int-to-ASCII) functions to be compatible with the Grain runtime. This works well for our purposes, however, we pay the price of having to use this function to produce our final result (from `numberUtils.ts`):
```ts
function fixEncoding(utf16StringRaw: u32): u32 {
  let utf16String = utf16StringRaw & ~GRAIN_GENERIC_HEAP_TAG_TYPE
  let utf16Size = stringSize(utf16StringRaw)
  let utf8Size = utf16Size >> 1
  let ret = allocateString(utf8Size)
  String.UTF8.encodeUnsafe(<usize>(utf16String + 8), utf8Size, ret + 8, false)
  decRef(utf16StringRaw)
  return ret | GRAIN_GENERIC_HEAP_TAG_TYPE
}
```
It would be much more efficient if we directly wrote out a UTF-8 string into memory, as opposed to allocating this intermediate UTF-16 representation.

ospencer:
Closes #475. Figured this made sense to do before the rewrite.

I have to say—mad respect to the AssemblyScript team. This change was fairly straightforward to do because the offsets were kept as the number of digits instead of the number of bytes. 

To that, I'm fairly confident this is change is correct. Our tests pass, and my smoke testing looks good. We don't really have any tests _specifically_ for number `toString` functionality, but I added a couple for `Infinity/NaN` since I knew we didn't have any for those. Let me know if you think we should have more.

Also, for simplicity's sake, I left the same number of store functions for `DIGITS`/`HEX_DIGITS`/`ANY_DIGITS`, just because it's easier to read. It would be more efficient to store 4 pairs at once with a single `store<u64>` (by 4x), though. If we want to do that, I think it'd be best done in a later PR.

### DIFF
--- a/compiler/test/test_end_to_end.re
+++ b/compiler/test/test_end_to_end.re
@@ -1910,6 +1910,28 @@ let string_tests = {
       "let x = \"\\u{110000}\"",
       "Illegal unicode code point",
     ),
+    t("string_float1", {|import * from "float32"; div(0.0f, 0.0f)|}, "NaN"),
+    t(
+      "string_float2",
+      {|import * from "float32"; div(1.0f, 0.0f)|},
+      "Infinity",
+    ),
+    t(
+      "string_float3",
+      {|import * from "float32"; div(-1.0f, 0.0f)|},
+      "-Infinity",
+    ),
+    t("string_float4", {|import * from "float64"; div(0.0d, 0.0d)|}, "NaN"),
+    t(
+      "string_float5",
+      {|import * from "float64"; div(1.0d, 0.0d)|},
+      "Infinity",
+    ),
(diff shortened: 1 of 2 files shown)


phated (approved):
Mostly just skimmed this. Taking the tests and your word for it! 👍

peblair (approved):
Our hero 🙏

PR #540, opened 2021-02-25
feat: Grain implementation of toString/print

ospencer:
This PR provides a Grain implementation of `toString` and `print`.

Since we have access to Grain data structures now, I changed the algorithm of `toString` to create a list of strings of the subcomponents and concatenate them all at the end, but I only implemented it for each of the individual types. This could be made much better by having the helper function return lists of strings instead—this way each portion of a structure is only written once + copied once. I think that's a good change for a future PR.

Of course, if we wanted to be ultra-efficient, we could just allocate a buffer and write directly into it, resizing if necessary. That'd probably be a good change down the line.


compiler/src/codegen/compcore.re
```diff
@@ -2983,6 +2975,23 @@ let compile_imports = (wasm_mod, env, {imports}) => {
     };
   };
 
+  let imports =
+    if (Env.is_normal_mode()) {
+      [
+        // HACK: Depend on runtime/string for the tests
```
ospencer, line 2981: This was moved to only be included in normal mode, since the string module is now a Grain module and shouldn't depend on itself.

compiler/src/typed/env.re
```diff
@@ -665,6 +665,7 @@ let get_components = c =>
 type compilation_mode =
   | Normal
   | Runtime
+  | ManagedRuntime
```
ospencer, line 668: I added this mode as a sort of final layer between when GC is available but you don't have `Pervasives` yet. I'm not 100% thrilled with it, since it's essentially just a "don't include Pervasives" mode. We've talked about potentially being able to enable/disable compiler flags via directives in a file, but we never really landed on anything concrete.

compiler/test/test_end_to_end.re
```diff
@@ -1036,14 +1036,14 @@ let loop_tests = [
 
 let oom = [
   tgcerr("oomgc1", 70, "(1, (3, 4))", "Maximum memory size exceeded"),
-  tgc("oomgc2", 356, "(1, (3, 4))", "(1, (3, 4))"),
-  tgc("oomgc3", 256, "(3, 4)", "(3, 4)"),
+  tgc("oomgc2", 880, "(1, (3, 4))", "(1, (3, 4))"),
+  tgc("oomgc3", 440, "(3, 4)", "(3, 4)"),
```
ospencer, line 1040: These went up because the `toString` algorithm changed a bit. Thoughts to improve in the description.

As a side note, I think the GC tests would be better served if we didn't have the result printed—we could focus on the memory usage of the exact program written.

ospencer: Needed to fix the GC tests to work with now the CLI requires `toString` independently.

compiler/src/codegen/compcore.re
ospencer: No longer needed!

compiler/src/typed/env.re
peblair: Can we add some doc comments to the different `compilation_mode` entries (either here or in `env.rei`)? I feel like it's becoming fuzzy (to me) what each means...like, I feel like it's going to be easy to forget what the difference between `Runtime` and `ManagedRuntime` is once we context switch away from these PRs you've recently opened.

compiler/test/test_end_to_end.re
```diff
@@ -38,24 +38,37 @@ let make_gc_program = (program, heap_size) => {
   ++ string_of_int(heap_size)
   ++ {|n)
       // Memory is not reclaimed due to no gc context
-      Memory.malloc(toLeak);
+      // This will actually leak 16 extra bytes because of the headers
```
peblair, line 41: 🙃

stdlib/runtime/numberUtils.gr
```diff
@@ -0,0 +1,1334 @@
...
+  let mut k = k
+  let sign = WasmI32.ltS(k, 0n)
+  if (sign) { k = 0n - k; }
+  let decimals = decimalCount32(k) + 1n
+  utoa32_dec_core(buffer, k, decimals)
+  WasmI32.store8(buffer, (if (sign) _CHAR_CODE_MINUS else _CHAR_CODE_PLUS), 0n)
+  decimals
+}
+
+let grisu2 = (value, buffer, sign) => {
```
peblair, line 1135: 😳

peblair (changes requested):
Looks good! One action item before merging, but this is great stuff.

compiler/src/typed/env.re
ospencer: Yeah, I can copy the descriptions over.

stdlib/runtime/numberUtils.gr
ospencer: Fastest algorithm on the market!

peblair (approved):

stdlib/runtime/numberUtils.gr
```diff
@@ -0,0 +1,1334 @@
...
+
+  let mut wp_w_frc = WasmI64.sub(mp_frc, w_frc)
+
+  let mut p1 = WasmI32.wrapI64(WasmI64.shrU(mp_frc, WasmI64.extendI32U(one_exp)))
+  let mut p2 = WasmI64.and(mp_frc, mask)
+
+  let mut kappa = decimalCount32(p1)
+  let mut len = sign
+
+  let mut done = false
```
phated, line 1069: Issue for early return? Then cleanup this stuff

stdlib/runtime/numberUtils.gr
```diff
@@ -0,0 +1,1334 @@
...
+    } else {
+      let sign = WasmF64.lt(value, 0.W)
+      if (sign) {
+        WasmI32.store8(buffer, _CHAR_CODE_MINUS, 0n)
+        buffer += 1n;
+      }
+      // Would be nice if there were a way to specify alignment here as we
+      // don't know if this store is aligned. I don't know what's fastest—a
+      // misaligned store or an explict alignment check + branch to store with
+      // align=0 or align=3 _or_ just always align=0. I imagine just align=0?
```
phated, line 1326: Issue please

stdlib/runtime/numberUtils.gr
```diff
@@ -0,0 +1,1334 @@
...
+    let size = dtoa_core(get_dtoa_buf(), value)
+    let result = allocateString(size)
+    Memory.copy(result + 8n, get_dtoa_buf(), size)
+    result
+  }
+  WasmI32.toGrain(str): String
+}
+
+
+export let dtoa_buffered = (buffer: usize, value: f64) => {
```
phated, line 1304: @ospencer said that we don't use this right now, so should we open an issue to consider removing it and only add it back when needed? Heck, even just toss the code into the issue.

phated (approved):
Talked through this, thanks @ospencer!

stdlib/runtime/string.gr
```diff
@@ -0,0 +1,380 @@
...
+import foreign wasm recordTypeExists: (WasmI32, WasmI32) -> Bool from "grainRuntime"
+import foreign wasm getRecordFieldName: (WasmI32, WasmI32, WasmI32) -> String from "grainRuntime"
+
+import foreign wasm fd_write: (WasmI32, WasmI32, WasmI32, WasmI32) -> WasmI32 from "wasi_snapshot_preview1"
+
+primitive (!) : Bool -> Bool = "@not"
+primitive (&&) : (Bool, Bool) -> Bool = "@and"
+primitive (||) : (Bool, Bool) -> Bool = "@or"
+
+enum StringList { [], [...](String, StringList) }
```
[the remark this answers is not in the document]
ospencer: https://github.com/grain-lang/grain/issues/552

stdlib/runtime/numberUtils.gr
ospencer: https://github.com/grain-lang/grain/issues/553

stdlib/runtime/numberUtils.gr
ospencer: https://github.com/grain-lang/grain/issues/554

stdlib/runtime/numberUtils.gr
ospencer: https://github.com/grain-lang/grain/issues/555

ospencer:
🚂

PR #541, opened 2021-03-05
feat: Upgrade to Binaryen 100

ospencer:
Once this lands, it will be possible to finally do #196 (though not with this change, as there's some work to be done wrt the runtime).


compiler/src/codegen/compcore.re
```diff
@@ -1514,7 +1516,7 @@ let call_error_handler = (wasm_mod, env, err, args) => {
     gensym_label("call_error_handler"),
     [
       call_runtime_throw_error(wasm_mod, env, compiled_args),
-      Expression.unreachable(wasm_mod),
+      Expression.const(wasm_mod, const_void()),
```
ospencer, line 1519: This was causing an issue where we could generate tuples with invalid types. It would be nice to have Binaryen throw out code if the user wrote something like `fail "foo"; someFunc()` via `unreachable`, but I suppose those cases are pretty rare.

ospencer:
We've now got #543!

PR #549, opened 2021-03-08
Add JS FFI import support

ospencer:
This is a remake of #409. Draft until #512 lands. cc @z0w0 

This adds a `import foreign js funcName : Args -> Result from 'module'` syntax for
importing and calling JS functions directly from Grain. Any modules must be declared
when instantiating the runtime using a new `.addJSImports` method. The runtime then wraps those
JS imports so that the values going through as arguments and out as return values are coerced
between JS and Grain values as needed.

JS imports are outputted just like WASM imports in the compiled WASM,
except JS imports implicitly live under the "js/" module prefix. This module prefix
is then populated using the aforementioned `.addJSImports` method.


Alch-Emi:
With #995 out of the way (presumably having supplanted #512), is this back on the table?

Also, this might just be me being unfamiliar with the domain, but as I understood it, WASM FFI imports can import generic functions from whatever environment the WASM is running in, not necessarily just JavaScript.  (with some examples where this is relevant including the [Lunatic VM][] and the [TIC-80][], where WASM code calls into a predefined set of Rust and C functions, respectively.

Given this, it might be worth making the name more generic, to avoid confusion?  Unless I'm totally misunderstanding this PR, and I've got it all wrong.

[TIC-80]: https://tic80.com/
[WASM-4]: https://wasm4.org/
[Lunatic VM]: https://lunatic.solutions/

phated:
This PR is being superceded by wit-bindgen for grain code. In the future, we'll be integrating wit-bingen directly into the compiler, but that's a much future change.

ospencer:
Going to go ahead and close this for now since we'll have a way better solution pretty soon! If anyone finds this PR and is looking for wit-bindgen for Grain, you can find my WIP here: https://github.com/ospencer/witx-bindgen/tree/oscar/grain-bindgen

PR #550, opened 2021-03-09
chore: Call memory_size and memory_grow from within Grain

ospencer:


phated (approved):
LGTM! 👍 Looks like it can be straight-up merged since it based on main

PR #551, opened 2021-03-09
chore: Allow setting compiler flags at the start of a file

### LINKED ISSUE #544: Refactor compilation mode syntax
Currently, we set special compilation modes via a comment directive at the start of a file, i.e.:

```
/* compilation-mode: runtime */
```

We should do two things:
1. Refactor compilation modes to work via a compiler flag, and
2. Add the ability to set compiler flags at the start of a file.

Not sure exactly what this should look like, but probably something like this:

```
/* grainc-flags --compilation-mode=runtime */
```

ospencer:
Closes #544.

@phated Why yes, I _did_ stay up until 4am getting this done.

### DIFF
--- a/compiler/src/utils/config.re
+++ b/compiler/src/utils/config.re
@@ -22,7 +22,7 @@ type config_info = {
 };
 
 type config_spec =
-  | Spec(Cmdliner.Arg.t('a), ref('a)): config_spec;
+  | Spec(Cmdliner.Arg.t('a), list(string), ref('a)): config_spec;
 
 let opts: ref(list(config_opt)) = (ref([]): ref(list(config_opt)));
 let specs: ref(list(config_spec)) = (ref([]): ref(list(config_spec)));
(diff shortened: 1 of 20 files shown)


compiler/src/utils/config.re
```diff
@@ -268,6 +244,56 @@ let with_cli_options = (term: 'a): Cmdliner.Term.t('a) => {
   folded;
 };
 
+let with_unapplied_cli_options = (term: 'a): Cmdliner.Term.t('a) => {
```
ospencer, line 247: I don't love having to need this, but it was the only way I could get `with_used_args` to be populated.

compiler/src/utils/config.re
```diff
@@ -311,6 +337,22 @@ let principal =
     false,
   );
 
+let compilation_mode =
+  opt(
+    ~names=["compilation-mode"],
+    ~conv=
+      option_conv(
+        Cmdliner.Arg.enum([
+          ("normal", "normal"),
```
phated, line 346: Can cmdliner do something like this?
```suggestion
          ("normal", Env.normal),
```

phated (approved):
Just 1 question that might be informative of future cleanup, but this looks **great**!

compiler/src/utils/config.re
ospencer: Yes, but `Env` depends on `Config`. We could potentially move it into some separate file that they can both depend on, though.

PR #559, opened 2021-03-11
feat: Grain implementation of Char and String libraries

ospencer:


compiler/src/codegen/compcore.re
```diff
@@ -33,7 +33,7 @@ let gensym_label = s => {
 };
 
 /* Number of swap variables to allocate */
-let swap_slots_i32 = [|Type.int32, Type.int32|];
+let swap_slots_i32 = [|Type.int32, Type.int32, Type.int32|];
```
ospencer, line 36: The beauty of optimizations means that we can have as many swap slots as we want, and Binaryen will throw out the unused ones. I doubt we'll ever need more than three, though.

compiler/src/codegen/compcore.re
```diff
@@ -2465,6 +2465,102 @@ let compile_primn = (wasm_mod, env: codegen_env, p, args): Expression.t => {
         Expression.const(wasm_mod, const_void()),
       ],
     )
+  | WasmMemoryCompare =>
```
ospencer, line 2468: This is an implementation of `memcmp`: https://github.com/gcc-mirror/gcc/blob/master/libiberty/memcmp.c (public domain).

ospencer: inb4 why is this in the compiler and not just Grain code: Ideally this should always be inlined, and we can't yet guarantee that it gets inlined across files, but it will always be inlined as a compiler primitive.

compiler/src/middle_end/optimize_constants.re
```diff
@@ -3,7 +3,7 @@ open Grain_typed;
 open Grain_utils;
 open Types;
 
-let known_constants = ref(Ident.empty: Ident.tbl(constant));
+let known_constants = ref(Ident.empty: Ident.tbl(imm_expression_desc));
```
ospencer, line 6: We weren't previously optimizing a program like
```
let y = x
y
```
into
```
x
```
and it's needed for the compiler to do other optimizations on simple rebinds, like `let (+) = WasmI32.add`. (This worked previously for import aliases, i.e. `import { add as (+) } ...` because they're handled differently than regular bindings.)

ospencer: _Note:_ This optimization is only performed on immutable bindings. We don't yet have an optimization that rewrites `let mut`s into `let`s when no mutation is performed.

stdlib/string.gr
```diff
@@ -1,72 +1,522 @@
...
+      } else if ((byte & 0xE0n) == 0xE0n) {
+        ptr += 3n;
+      } else {
+        ptr += 2n;
+      }
+    }
+
+    if (result == -1n) {
+      let none = None
+      Memory.incRef(WasmI32.fromGrain(none))
```
phated, line 96: Why do you incRef `none` but not the `Some`?

phated (approved):
Mostly just a "I trust you" approval! My eyes glazed over at the string algorithms.

memory_compare is interesting!

stdlib/string.gr
ospencer: This just maintains GC semantics. A function is expected to return a value with a reference count of +1. Because of this rule, if I call a function and want to just return the result, I don't need to incRef it because its count is +1. If I've received a reference to something, like a parameter, its count is +0, so when I return it it needs an incRef to have a count of +1. (Writing `None` in Grain doesn't do an allocation, it just references the value None.)

Much easier to let the compiler figure it out for you 😛

PR #560, opened 2021-03-11
feat!: Update assignment semantics

ospencer:
This PR changes the semantics for assignment in Grain.

Previously, mutation via assignment would return the assigned value. Now, all assignments return `void`. As such, this is a breaking change.

Rationale: The majority of the time that we do a mutation in Grain, it's done in an `if` statement or within a loop. Single-sided `if` statements and loops both evaluate to `void`, and the bodies of each are expected to typecheck to `Void` as well. In the current state of Grain, this does not typecheck, as the assignment returns a `Number`:
```
let mut x = 5
if (true) x = 6
```

To resolve the typechecking error, the `if` can be rewritten as
```
if (true) ignore(x = 6)
```
or
```
if (true) { x = 6; }
```

It is easily seen how this can be frustrating—not only is this a common mistake, the solution is far less aesthetic than what we want to write.

It's fairly rare that code is written to take advantage of the current semantics, as writing code that does an assignment within an expression is often frowned upon.

Additionally, I've migrated the old array assignment syntax to the new assignment syntax. Regular assignment, record assignment, and array assignment now all use the `=` syntax. Box assignments continue to use the `:=` syntax.


stdlib/runtime/numberUtils.gr
```diff
@@ -918,23 +918,23 @@ export let itoa32 = (value, radix) => {
   let mut out = 0n
   let sign = WasmI32.shrU(value, 31n)
 
-  if (WasmI32.ne(sign, 0n)) { value = 0n - value; }
+  if (WasmI32.ne(sign, 0n)) { value = 0n - value }
```
phated, line 921: Why not drop the block?
```suggestion
  if (WasmI32.ne(sign, 0n)) value = 0n - value
```

stdlib/runtime/numberUtils.gr
```diff
@@ -982,29 +982,29 @@ export let itoa64 = (value, radix) => {
   let mut out = 0n
 
   let sign = WasmI32.wrapI64(WasmI64.shrU(value, 63N))
-  if (sign != 0n) { value = WasmI64.sub(0N, value); }
+  if (sign != 0n) { value = WasmI64.sub(0N, value) }
```
phated, line 985:
```suggestion
  if (sign != 0n) value = WasmI64.sub(0N, value)
```

stdlib/runtime/numberUtils.gr
```diff
@@ -1125,7 +1125,7 @@ let genDigits = (buffer, w_frc, w_exp, mp_frc, mp_exp, delta, sign) => {
 let genExponent = (buffer, k) => {
   let mut k = k
   let sign = WasmI32.ltS(k, 0n)
-  if (sign) { k = 0n - k; }
+  if (sign) { k = 0n - k }
```
phated, line 1128:
```suggestion
  if (sign) k = 0n - k
```

compiler/src/typed/typecore.re
```diff
@@ -977,18 +977,19 @@ and type_expect_ =
         Builtin_types.type_box @@
         newvar(~name="a", ()),
       );
-    let val_ = type_expect(env, sval, ty_expected_explained);
+    let val_ = type_expect(env, sval, mk_expected(newvar(~name="a", ())));
```
phated, line 980: What is with `~name="a"`? I assume it is just constructing a dummy var, but does `a` ever show up in errors or what if it is shadowed or something?

phated (changes requested):
You really hate my semicolon usage in the stdlib 🤣

compiler/test/test_end_to_end.re
```diff
@@ -865,7 +865,7 @@ let box_tests = [
     "let b = box(4);\n            {\n              b := unbox(b) - 1;\n              unbox(b)\n            }",
     "3",
   ),
-  t("test_set_extra1", "box(1) := 2", "2"),
+  t("test_set_extra1", "box(1) := 2", "void"),
```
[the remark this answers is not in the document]
ospencer: The tests below are actually doubled! There's a test for the assignment itself (which should return void) and the following test where it unboxes the value and checks that the value is correct. It's important to check that the return value is correct since even though typechecking is correct, we could accidentally generate the wrong code.

For this particular test, it seemed more like we were testing the return value of the statement, so I changed the expectation. I can also double this test to check the unboxed value as well, though.

ospencer (commented):
I don't hate your semicolons! Most of the ones I removed should have been the ones that we added to get around the typechecking limitation.

compiler/src/typed/typecore.re
ospencer: Ah yeah, there's no reason to provide a name here. It wouldn't ever show up in an error by virtue of everything will typecheck with it. As for shadowing, that's really only a concern during parsing. After typechecking, all identifiers are unique, even if they have the same name (since every identifier has its own id).

ospencer:
```suggestion
    let val_ = type_expect(env, sval, mk_expected(newvar()));
```

compiler/test/test_end_to_end.re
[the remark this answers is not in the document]
phated: Makes sense. Our test names leave a lot to be desired.

compiler/src/typed/typecore.re
```diff
@@ -834,14 +834,14 @@ and type_expect_ =
     };
     let (_, ty_arg, ty_res) = instance_label(false, label);
     unify_exp(env, record, ty_res);
-    let val_ = type_expect(env, sval, ty_expected_explained);
+    let val_ = type_expect(env, sval, mk_expected(newvar(~name="a", ())));
```
phated, line 837:
```suggestion
    let val_ = type_expect(env, sval, mk_expected(newvar()));
```

compiler/src/typed/typecore.re
```diff
@@ -977,18 +977,19 @@ and type_expect_ =
...
       exp_extra: [],
       exp_attributes: attributes,
-      exp_type: val_.exp_type,
+      exp_type: Builtin_types.type_void,
       exp_env: env,
     });
   | PExpAssign(sidexpr, sval) =>
-    let idexpr = type_expect(env, sidexpr, ty_expected_explained);
+    let idexpr =
+      type_expect(env, sidexpr, mk_expected(newvar(~name="a", ())));
```
phated, line 992:
```suggestion
      type_expect(env, sidexpr, mk_expected(newvar()));
```

compiler/src/typed/typecore.re
```diff
@@ -998,14 +999,14 @@ and type_expect_ =
     if (!val_mutable) {
       raise(Error(loc, env, Assign_not_mutable(id.txt)));
     };
-    let val_ = type_expect(env, sval, ty_expected_explained);
+    let val_ = type_expect(env, sval, mk_expected(newvar(~name="a", ())));
```
phated, line 1002:
```suggestion
    let val_ = type_expect(env, sval, mk_expected(newvar()));
```

phated (changes requested):
Can we change these too? And are we going to include the tests for #572 in here?

ospencer:
@phated should be all set 👍

phated (approved):
:approved: 🎉

PR #562, opened 2021-03-13
feat: Grain-based wasi bindings

ospencer:


stdlib/sys/file.gr
```diff
@@ -192,218 +343,698 @@ let filetypeFromNumber = (filetype) => {
...
-  }, dirents)
+export let fdReaddir = (fd: FileDescriptor) => {
+  let fd = match (fd) {
+    FileDescriptor(n) => WasmI32.fromGrain(n) >> 1n
+  }
+
+  let structWidth = 24n
+
+  let bufUsed = Memory.malloc(4n)
+  let mut cookie = 0N
```
phated, line 828: 🍪

stdlib/sys/process.gr
```diff
@@ -61,19 +78,130 @@ enum Signal {
...
+    Memory.copy(grainStrPtr + 8n, strPtr, strLength)
+
+    WasmI32.store(arr + i, grainStrPtr, 8n)
+  }
+
+  Memory.free(argcPtr)
+  Memory.free(argvPtr)
+  Memory.free(argvBufPtr)
+
+  WasmI32.toGrain(arr): Array<String>
```
phated, line 125: Do we still want this to be an array instead of a List now that it is in Grain?

stdlib/sys/process.gr
```diff
@@ -61,19 +78,130 @@ enum Signal {
...
+    Memory.copy(grainStrPtr + 8n, strPtr, strLength)
+
+    WasmI32.store(arr + i, grainStrPtr, 8n)
+  }
+
+  Memory.free(envcPtr)
+  Memory.free(envvPtr)
+  Memory.free(envvBufPtr)
+
+  WasmI32.toGrain(arr): Array<String>
```
phated, line 174: Same as above.

phated (changes requested):
As with most of the pointer math stuff, my eyes glazed over. Hopefully I caught some stuff and asked good questions.

Additionally, can you go through and make sure the comments are right now that you've changed stuff? I saw a few that were missing params or had extra, especially in `stdlib/sys/file.gr`.

I believe this should be a breaking change because some of the method arguments have been changed, yeah?

stdlib/sys/file.gr
```diff
@@ -102,7 +158,83 @@ enum Rights {
   SockShutdown,
 }
 
-enum FdFlag {
+// Grain currently cannot close over unsafe wasm i64s, so these are here in
+// module so they get inlined
```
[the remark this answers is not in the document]
ospencer: I don't want to say never, but there currently aren't any plans to be able to do this. Without a major hack, this is actually impossible until Long Grain (64-bit Grain). If you still want an issue to track it, I can make one, though I don't know how useful it'll be.

stdlib/sys/file.gr
```diff
@@ -192,218 +343,698 @@ let filetypeFromNumber = (filetype) => {
...
+export let pathOpen = (
+  dirfd: FileDescriptor,
+  dirflags: List<LookupFlag>,
+  path: String,
+  openflags: List<OpenFlag>,
+  fsRightsBase: List<Rights>,
+  fsRightsInheriting: List<Rights>,
+  fsFlags: List<FdFlag>
+) => {
+  let dirfdval = match (dirfd) {
```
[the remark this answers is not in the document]
ospencer: Copypasta.

stdlib/sys/process.gr
ospencer: I'm thinking no, specifically because the arguments are positional? OCaml seems to also keep them as an array: https://caml.inria.fr/pub/docs/manual-ocaml/libref/Sys.html

stdlib/sys/process.gr
ospencer: This one could probably be a list, yeah. Though honestly it's probably far more useful as a hash. That can probably happen in the higher level process module if we ever make it, though.

ospencer:
This shouldn't at all be a breaking change. The only places where arguments could have appeared to change should be for the functions that had extra processing within Grain. If you saw any specific examples, we should take a look.

As for the comment docs, fine if we do that later? I would care now if I made changes, but theoretically everything should be taking the same arguments and returning the same values.

stdlib/sys/file.gr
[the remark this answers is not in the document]
phated: Even if the issue just informs 1 per that tries to do something with wasm i64s that they can't be closed over, I think it is good to have.

stdlib/sys/file.gr
```diff
@@ -192,218 +343,698 @@ let filetypeFromNumber = (filetype) => {
...
+  }
+
+  WasmI32.toGrain(newoffset): Int64
+}
 
 // Read a file descriptor's offset
 // @param fd: FileDescriptor The file descriptor to read the offset
 // @returns Int64 The offset of the file descriptor, relative to the start of the file
-import foreign wasm fdTell : (FileDescriptor) -> Int64 from "stdlib-external/fs"
+export let fdTell = (fd: FileDescriptor) => {
```
phated, line 970: @ospencer this signature seems to have changed. No longer returns an Int64

stdlib/sys/file.gr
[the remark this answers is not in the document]
ospencer: https://github.com/grain-lang/grain/issues/575

phated (approved):
If we don't want to fix or remove the `@param`/`@returns` comments, can we please create a high-priority issue to get them fixed. I really dislike having incorrect inline doc comments (the whole point of having them next to the code is that you are supposed to update them when you make code changes).

ospencer:
Sure. https://github.com/grain-lang/grain/issues/576

I don't believe this code has ever changed, and they have probably always been wrong. These are used for the docs on the website because we don't currently auto-generate the docs. Once that's done, they can be deleted. I looked on the site, and at least one is incorrect there as well.

PR #563, opened 2021-03-13
chore: Refactor runtime setup

ospencer:
This PR refactors the runtime setup to be simpler and better handle the upcoming changes to exceptions and static linking.

This PR goes from 4 compilation modes down to just two, `Normal` and `Runtime`. `--compilation-mode=malloc` is now the same as `--compilation-mode=runtime`, since a static pointer has been allocated for the runtime heap pointer. `--compilation-mode=managed-runtime` has been superseded by `--no-pervasives` 🎉 

I also added some documentation about how the runtime works!

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -297,6 +297,16 @@ let var_of_ext_global = (env, modname, itemname) =>
 let lookup_ext_func = (env, modname, itemname) =>
   Ident.find_same(itemname, Ident.find_same(modname, env.imported_funcs));
 
+/** Static runtime values */
+
+// Static pointer to the runtime heap
+// Leaves low 1000 memory unused for Binaryen optimizations
+let runtime_heap_ptr = 0x400;
+// Start pointer for the runtime heap
+let runtime_heap_start = 0x410;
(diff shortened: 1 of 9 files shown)


compiler/src/codegen/compcore.re
```diff
@@ -3540,6 +3569,7 @@ let compile_wasm_module = (~env=?, ~name=?, prog) => {
         Features.mutable_globals,
       ],
     );
+  let _ = Module.set_low_memory_unused(1);
```
phated, line 3572: Is `1` the amount of memory unused?

phated (approved):
This was way less daunting to review that I expected from the title! LGTM 👍

compiler/src/codegen/compcore.re
ospencer: It's actually a number that represents a boolean. 😶

Binaryen v101 converts all these awkward number-as-boolean values to actual booleans, so that's something to look forward to.

phated: Good good! I thought that might be the case and was going to ask if we should do that massaging ourselves in binaryen.ml, but it seems it'll be solved for us!

PR #565, opened 2021-03-16
feat!: Use Grain exceptions instead of JS exceptions

### LINKED ISSUE #173: Show actual WASI error messages
We should show an actual message instead of `[GrainError]: System error: 2`.

### LINKED ISSUE #301: Better Overflow Messages
When adding our new number system in, the values we pass to our overflow exception were a bit of a rush job. These should ideally be improved, whether that be with more specific variants of `GRAIN_ERR_OVERFLOW` (e.g. `GRAIN_ERR_OVERFLOW_PLUS`) or just more consistency about what the user sees. Currently, we essentially just pick one of the numbers and display it to the user (which isn't super informative).

@grain-lang/core We can discuss what we'd like the messages to look like here.

### LINKED ISSUE #302: Fix Memory Leak in safeI64Multiply Overflow Check
Related: #301 

When producing the overflow message in `safeI64Multiply` in `numbers.ts`, we have an `i64` value that needs to be passed to the exception handler. As such, we first box it. The problem is that, when we eventually have overflow checking, we will actually leak the memory used by this boxed value, since there's no one around to clean it up. Maybe it's okay, though, depending on how we reify that exception back to the user? TBD.

ospencer:
feat: Add ability to `throw` exceptions
feat: Add ability to register custom exception printers
feat: Add Exception stdlib with Exception.registerPrinter
fix: Use Is instead of Eq for match variant comparison
fix: `export *` with exceptions
chore: Only emit MatchFailure exception for partial matches

Closes #173, closes #301, closes #302.

## Overview

This PR adds the ability to throw exceptions in Grain. (Previously, you could create exceptions, pattern match on them, use them with `Result`, etc., but you could not _throw_ them.)

```grain
throw InvalidArgument("Your argument is no good!")
```

The `toString` value of the exception is printed when an error is thrown, but thanks to a new standard library included in this PR, the message can be overridden:

```grain
import Exception from "exception"

exception ErrorCode(Number)

Exception.registerPrinter(e => match (e) {
  ErrorCode(1) => Some("Custom message 1"),
  ErrorCode(2) => Some("Custom message 2"),
  ErrorCode(_) => Some("Unknown error code!"),
  _ => None // Allow other printers to handle exceptions we don't know about!
})
```

## Details

Exceptions are managed by `stdlib/runtime/exception` (not to be confused with `stdlib/exception`). This module maintains a pointer to the list of available error printers, defines some exceptions to be used by the runtime, and defines a printer for those exceptions. As `toString` is not available yet, it exports a method that allows Pervasives to register the "base" exception printer. If Pervasives is not used (such as via the `--no-pervasives` flag), exceptions are printed as the generic "GrainException".

If you want to know why the base set of exceptions are defined in that module and not built directly into the compiler, we can have a virtual drink and I'll tell you the tale of how I went down that rabbit hole.

There were a handful of bugs that I needed to fix to get this together, but I'll point them out with some comments.

If this is approved by the core team, there's a few issues that I'll want to open:
- There's currently no way to re-export an exception. We've talked in the past about reworking how values vs types are imported/exported, and I believe we'll need to revisit that conversation. It's not a huge deal right now, as runtime functions can just import the exception and throw it. It will, however, be problematic in the future when WebAssembly exception handling lands and users will want to catch those errors—they should not be importing anything from `stdlib/runtime`.
- `MatchFailure` wasn't actually being thrown on a match failure—we have a custom node in the match compiler that handles that case. That was a bigger refactor than what I thought should go in this PR, so I left it alone for now.
- We should refactor most stdlib code to throw real errors instead of just `fail`ing.


compiler/src/codegen/compcore.re
```diff
@@ -1969,7 +1973,7 @@ let allocate_array_n = (wasm_mod, env, num_elts, elt) => {
           compiled_num_elts(),
           Expression.const(wasm_mod, encoded_const_int32(0)),
         ),
-        InvalidArgument,
+        IndexOutOfBounds,
```
ospencer, line 1976: For these two, it didn't seem worth it to me to set up calling an exception with arguments in the compiler, when this is just going to be rewritten into Grain soon anyway. I feel that IndexOutOfBounds is also an appropriate error here until this is moved over. If anyone really cares, I can do it sooner.

compiler/src/codegen/transl_anf.re
```diff
@@ -625,27 +626,34 @@ let transl_attributes = attrs => {
...
         Option.fold(
           ~none=I32Type,
           ~some=((_, exp)) => asmtype_of_alloctype(exp.anf_allocation_type),
           List.nth_opt(branches, 0),
         );
+      let default =
+        switch (partial) {
+        | Partial => [
+            {
+              instr_desc: MError(Runtime_errors.MatchFailure, [compiled_arg]),
```
ospencer, line 641: Like I mentioned in the description, this isn't actually the error that's happening on a match failure. I did have to do this (sensible) change though to allow the exception handler to use `match` and not depend on itself.

compiler/src/middle_end/matchcomp.re
```diff
@@ -375,13 +375,7 @@ module MatchTreeCompiler = {
       );
     | Fail =>
       /* FIXME: We need a "throw error" node in ANF */
```
ospencer, line 377: Even though I made this trap instead of moving on silently, I left this note here because that's what needs to happen to throw `MatchFailure` instead of just `unreachable`.

compiler/src/middle_end/matchcomp.re
```diff
@@ -502,7 +496,7 @@ module MatchTreeCompiler = {
                 cmp_id_name,
                 Comp.prim2(
                   ~allocation_type=StackAllocated(WasmI32),
-                  Eq,
+                  Is,
```
ospencer, line 499: Calling `Pervasives.(==)` was unnecessary here (and prevented pattern matching on variants to happen in runtime mode). This actually will speed up pattern matching by a non-trivial amount.

compiler/test/test_end_to_end.re
```diff
@@ -2182,5 +2204,4 @@ let tests =
   @ export_tests
   @ comment_tests
   @ number_tests
-  @ exception_tests
-  @ export_tests;
```
ospencer, line 2186: `export_tests` was in the list twice! Sneaky sneaky.

runtime/src/core/closures.js
```diff
@@ -1,7 +1,5 @@
 import { managedMemory, grainModule } from "../runtime";
```
ospencer, line 1: This file isn't used, but I kept it in because it can be useful for FFI.

stdlib/runtime/exception.gr
```diff
@@ -0,0 +1,103 @@
...
+  add as (+)
+} from "runtime/unsafe/wasmi32"
+
+import foreign wasm fd_write: (WasmI32, WasmI32, WasmI32, WasmI32) -> WasmI32 from "wasi_snapshot_preview1"
+
+enum Option<a> { Some(a), None }
+
+export let mut printers = 0n
+
+export let registerBasePrinter = (f) => {
```
ospencer, line 14: Arguably this should be `dangerouslyRegisterBasePrinter`.

stdlib/runtime/wasi.gr
```diff
@@ -161,3 +164,97 @@ let _ETIMEDOUT = 73n
 let _ETXTBSY = 74n
 let _EXDEV = 75n
 let _ENOTCAPABLE = 76n
+
+exception SystemError(Number)
+
+let stringOfSystemError = (code) => {
```
ospencer, line 170: This is honestly one the biggest things that I'm excited for.

compiler/src/parsing/lexer.mll
```diff
@@ -177,7 +177,7 @@ rule token = parse
   | "fail" { FAIL }
   | "exception" { EXCEPTION }
   | "try" { TRY }
-  | "raise" { RAISE }
+  | "throw" { THROW }
```
phated, line 180: Does this lexer change make this a breaking? (I forget if this PR includes a breaking note)

compiler/src/typed/builtin_types.re
```diff
@@ -176,15 +176,6 @@ let common_initial_env = (add_type, empty_env) =>
   |> add_type(ident_array, decl_array)
   |> add_type(ident_fd, decl_abstr(path_fd));
 
-let build_initial_env = (add_type, empty_env) => {
```
phated, line 179: 🧹 💯

runtime/src/core/closures.js
```diff
@@ -17,19 +15,10 @@ export class GrainClosure {
...
-      );
-    } else {
-      let grainVals = args.map((x) => JSToGrainVal(x, this.runtime));
-      grainVals.unshift(this.loc * 4);
-      // [TODO]: We should feed in a runtime here.
-      return grainToJSVal(null, this.func(...grainVals));
-    }
+    let grainVals = args.map((x) => JSToGrainVal(x, this.runtime));
+    grainVals.unshift(this.loc * 4);
+    // [TODO]: We should feed in a runtime here.
```
phated, line 20: Issue for this?

stdlib/runtime/exception.gr
```diff
@@ -0,0 +1,103 @@
...
+    }
+    current = next
+  }
+}
+
+export let dangerouslyRegisterPrinter = (f) => {
+  // This is dangerous because it leaks runtime memory.
+  // This should only be called by this module and Pervasives.
+  printers = WasmI32.fromGrain((f, printers))
+}
```
phated, line 34: Is there a way we can make these completely unavailable to consumers? Calling these a 2nd time totally screws everything up, right?

stdlib/runtime/exception.gr
```diff
@@ -0,0 +1,103 @@
...
+      None => void
+    }
+    current = next
+  }
+  result
+}
+
+// HACK: Allocate static buffer for printing (40 bytes)
+// Would be nice to have a better way to allocate a static block from
+// the runtime heap, but this is the only module that needs to do it
```
phated, line 60: Do we want to create an issue for the nice-to-have?

stdlib/runtime/numbers.gr
```diff
@@ -195,7 +178,7 @@ let reducedFraction64 = (x, y) => {
 let safeI32Multiply = (x, y) => {
   let prod = WasmI64.mul(WasmI64.extendI32S(x), WasmI64.extendI32S(y))
   if (WasmI64.gtS(prod, _I32_MAX) || WasmI64.ltS(prod, _I32_MIN)) {
-    throwOverflowError(x)
+    throw Exception.Overflow
```
phated, line 181: For these Overflows, why doesn't it contain the number that overflowed?

stdlib/runtime/wasi.gr
```diff
@@ -161,3 +164,97 @@ let _ETIMEDOUT = 73n
...
+  let n = WasmI32.shrS(WasmI32.fromGrain(code), 1n)
+  let (==) = WasmI32.eq
+  match (n) {
+    n when n == _ESUCCESS => "SystemError: No error occurred. System call completed successfully.",
```
phated, line 174: This one feels weird. Maybe it's just the message?

stdlib/runtime/wasi.gr
phated: 🔥

stdlib/string.gr
```diff
@@ -329,13 +328,13 @@ export let slice = (start: Number, to: Number, s: String) => {
 
   let mut start = WasmI32.fromGrain(start)
   if ((start & 1n) != 1n) {
-    throwError(Errors._GRAIN_ERR_INVALID_ARGUMENT, start, 0n)
+    throw InvalidArgument("invalid start index")
```
phated, line 331: Do we want to capitalize the error messages in this file to match most of the other messages that have been adjusted?
```suggestion
    throw InvalidArgument("Invalid start index")
```

stdlib/sys/process.gr
```diff
@@ -180,7 +177,7 @@ export let exit = (code: Number) => {
   let mut code = WasmI32.fromGrain(code)
 
   if ((code & 1n) == 0n) {
-    throwError(Errors._GRAIN_ERR_INVALID_ARGUMENT, code, 0n)
+    throw InvalidArgument("invalid exit code")
```
phated, line 180: capitalize?
```suggestion
    throw InvalidArgument("Invalid exit code")
```

stdlib/string.gr
```diff
@@ -329,13 +328,13 @@ export let slice = (start: Number, to: Number, s: String) => {
...
   }
   start = start >> 1n
 
   let mut to = WasmI32.fromGrain(to)
   if ((to & 1n) != 1n) {
-    throwError(Errors._GRAIN_ERR_INVALID_ARGUMENT, to, 0n)
+    throw InvalidArgument("invalid end index")
```
phated, line 337:
```suggestion
    throw InvalidArgument("Invalid end index")
```

stdlib/string.gr
```diff
@@ -347,11 +346,11 @@ export let slice = (start: Number, to: Number, s: String) => {
   }
 
   if (start > len || to > len) {
-    throwError(Errors._GRAIN_ERR_ARRAY_INDEX_OUT_OF_BOUNDS, 0n, 0n)
+    throw Exception.IndexOutOfBounds
   }
 
   if (to < start) {
-    throwError(Errors._GRAIN_ERR_INVALID_ARGUMENT, WasmI32.fromGrain(tagSimpleNumber(to)), 0n)
+    throw InvalidArgument("start index exceeds end index")
```
phated, line 353:
```suggestion
    throw InvalidArgument("Start index exceeds end index")
```

stdlib/runtime/gc.gr
```diff
@@ -74,7 +76,7 @@ let rec decRef = (userPtr: WasmI32, ignoreZeros: Bool) => {
       if (ignoreZeros) {
         userPtr
       } else {
-        fail "decRef called when reference count was zero."
+        throw DecRefError
```
phated, line 79: Does `DecRefError` not need a printer?

stdlib/char.gr
```diff
@@ -2,10 +2,11 @@ import WasmI32 from "runtime/unsafe/wasmi32"
 import Memory from "runtime/unsafe/memory"
 import Errors from "runtime/unsafe/errors"
 import { tagSimpleNumber, allocateChar, allocateString } from "runtime/dataStructures"
-import foreign wasm throwError: (WasmI32, WasmI32, WasmI32) -> a from "grainRuntime"
 
 export *
 
+exception MalformedUtf8
```
phated, line 8: Should this have a printer?

stdlib/runtime/exception.gr
```diff
@@ -0,0 +1,103 @@
...
+  match (e) {
+    IndexOutOfBounds => Some("IndexOutOfBounds: index out of bounds"),
+    DivisionByZero => Some("DivisionByZero: division by zero"),
+    ModuloByZero => Some("ModuloByZero: modulo by zero"),
+    Overflow => Some("Overflow: number overflow"),
+    NumberNotIntlike => Some("NumberNotIntlike: can't coerce number to integer"),
+    NumberNotRational => Some("NumberNotRational: can't coerce number to rational"),
+    MatchFailure => Some("MatchFailure: no matching pattern"),
+    AssertionError => Some("AssertionError: assertion failed"),
+    OutOfMemory => Some("OutOfMemory: maximum memory size exceeded"),
```
phated, line 97: Should we capitalize these?
```suggestion
    IndexOutOfBounds => Some("IndexOutOfBounds: Index out of bounds"),
    DivisionByZero => Some("DivisionByZero: Division by zero"),
    ModuloByZero => Some("ModuloByZero: Modulo by zero"),
    Overflow => Some("Overflow: Number overflow"),
    NumberNotIntlike => Some("NumberNotIntlike: Can't coerce number to integer"),
    NumberNotRational => Some("NumberNotRational: Can't coerce number to rational"),
    MatchFailure => Some("MatchFailure: No matching pattern"),
    AssertionError => Some("AssertionError: Assertion failed"),
    OutOfMemory => Some("OutOfMemory: Maximum memory size exceeded"),
```

phated (changes requested):
This PR is so spicy! 🌶️ Love it!

compiler/src/parsing/lexer.mll
ospencer: Mmm, yeah it does (if you used `throw` as an identifier).

runtime/src/core/closures.js
ospencer: Only annoyed because it's not my TODO 😛 
https://github.com/grain-lang/grain/issues/578

stdlib/char.gr
ospencer: I thought the default printer was fine, but I suppose I could see a printer for it.

stdlib/runtime/exception.gr
ospencer: Not as of right now, but generally people shouldn't ever be importing anything from `runtime`. We could achieve that though with a compiler flag or something similar that by default disables the ability to import from the runtime.

The only harm in this being called more than once is a memory leak. Not the worst thing in the world (but still not great).

stdlib/runtime/exception.gr
ospencer: I'd want to make the issue if this ever comes up again.

stdlib/runtime/exception.gr
ospencer: I dunno. I've seen languages do both. Would you prefer them capitalized?

stdlib/runtime/gc.gr
ospencer: I don't think so. This isn't actually an error that ever happens outside of GC debug mode. Since this module gets loaded for every program, I figure we can save on cycles by not registering a printer for it, and we can register one locally if we really want while debugging the GC.

stdlib/runtime/numbers.gr
ospencer: Largely it's because I don't think that information is at all useful. 99% of the time if you overflow the value is meaningless (and for what we were doing up until this point, it's definitely meaningless—see https://github.com/grain-lang/grain/issues/301). The other reasons are that it takes a lot to support it—it introduces a dependency on `itoa` for all programs to print that value, and also complicates the base printer quite a bit. I think this is perfectly fine for now, and if we really want we can keep #301 open (but I personally wouldn't bother).

stdlib/runtime/wasi.gr
ospencer: I think it's because these are error messages and this one isn't an error 😛 

The messages all come from the official spec: https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#-errno-variant

Grain currently doesn't ever throw a `SystemError` if it's `ESUCCESS`, but I left it in here for completeness.

stdlib/string.gr
ospencer: If we decide we're going that route!

runtime/src/core/closures.js
```diff
@@ -1,7 +1,5 @@
 import { managedMemory, grainModule } from "../runtime";
-import { throwGrainError } from "../errors/errors";
 import { grainToJSVal, JSToGrainVal } from "../utils/utils";
-import { GRAIN_ERR_ARITY_MISMATCH } from "../errors/error-codes";
```
phated, line 4: Do we still use `../errors/errors` or `../errors/error-codes` or can those be deleted now?

phated (approved):
Looks like the tests are passing now, so take this ⚔️ (it's dangerous to go alone)!

PR #566, opened 2021-03-17
Upgrade binaryen

phated:
I also locked ocamlfind since changing the esylock caused ocamlfind to try to upgrade and fail until esy has a patch. Some of the dune stuff got updated to a patch version, too.


ospencer (approved):
🙌

PR #567, opened 2021-03-17
fix: word-size independent LEB input byte checking

phated:
This removes the `MalformedLEB` exception and bit check and replaces it with an assert that checks to see a read byte is under 256.

This fix is needed in order for the compiler to run successfully when compiled to JS.

@ospencer is there a better way to phrase this commit message?

### DIFF
--- a/compiler/src/utils/wasm_utils.re
+++ b/compiler/src/utils/wasm_utils.re
@@ -40,7 +40,6 @@ let latest_abi = {major: 1, minor: 0, patch: 0};
 let identity: 'a. 'a => 'a = x => x;
 let i32_of_u64 = Int64.to_int32;
 
-exception MalformedLEB(bool, int);
 exception MalformedSectionType(int, option(int));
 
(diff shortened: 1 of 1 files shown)


compiler/src/utils/wasm_utils.re
```diff
@@ -58,12 +57,11 @@ let read_leb128:
...
         zero;
       } else {
-        switch (signed, next_byte()) {
+        let byte = next_byte();
         /* Check for invalid input */
-        | (false, n) when n >= 1 lsl maxbits =>
-          raise(MalformedLEB(false, maxbits))
-        | (true, n) when n >= 1 lsl (maxbits - 1) =>
-          raise(MalformedLEB(true, maxbits))
+        assert(byte < 256);
```
ospencer, line 62: This should probably be this:
```suggestion
        assert(byte >= 0 && byte < 256);
```

ospencer:
For commit message, I'd probably say `fix: word-size independent LEB input byte checking` or something like that.

ospencer (approved):

PR #570, opened 2021-03-18
feat: Compile grainc to JS & create binaries with pkg

phated:
All my current work to get GrainJS working.

I believe I have some other ideas to actually just compile `grainc` to JS, so I've marked this as draft.

### DIFF
--- a/cli/bin/compile.js
+++ b/cli/bin/compile.js
@@ -2,13 +2,9 @@ const exec = require("./exec");
 
 module.exports = (file, options) => {
   try {
-    // TODO: Should we console.log the output (stdout) of the exec?
-    exec(file, options);
+    exec(file, options, { stdio: "inherit" });
     return file.replace(/\.gr$/, ".gr.wasm");
   } catch (e) {
-    console.log(e.stdout.toString());
-    // TODO: Do we also want to print the stderr?
-    // console.log(e.stderr.toString());
     if (options.graceful) {
       process.exit();
     } else {
--- a/cli/bin/compile.js
+++ b/cli/bin/compile.js
@@ -1,18 +1,14 @@
-const path = require("path");
-const { execSync } = require("child_process");
-
-const grainc = path.join(__dirname, "grainc.exe");
+const exec = require("./exec");
 
 module.exports = (file, options) => {
   try {
-    execSync(
-      `${grainc} --stdlib=${options.stdlib} ${
-        options.cflags ? options.cflags : ""
-      } ${file}`
-    );
+    // TODO: Should we console.log the output (stdout) of the exec?
+    exec(file, options);
     return file.replace(/\.gr$/, ".gr.wasm");
   } catch (e) {
     console.log(e.stdout.toString());
+    // TODO: Do we also want to print the stderr?
+    // console.log(e.stderr.toString());
     if (options.graceful) {
       process.exit();
     } else {
--- a/cli/bin/exec.js
+++ b/cli/bin/exec.js
@@ -1,0 +1,26 @@
+const path = require("path");
+const { execSync } = require("child_process");
+const fs = require("fs");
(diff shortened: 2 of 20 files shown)


phated:
@ospencer @peblair I believe this is pretty much ready from the packaging/virtual file system perspective. We just need to figure out the String allocation bug that happens in the JS version of the compiler. I would be forever grateful if either of you could help me debug it.

I'll probably be splitting this PR into small stuff that is more related to cleanup of the monorepo, but needed for this stuff to wrap up!

ospencer:
Awesome work! 🎉  

I'll take a look at the allocation stuff.

hhugo:
Is there anything to contribute back to js_of_ocaml ? Better support for unix stubs with node ?

phated:
@hhugo yeah! If you think the unix stubs would be helpful to JSOO, I'm happy to contribute them back!

cli/bin/compile.js
```diff
@@ -1,18 +1,14 @@
...
-        options.cflags ? options.cflags : ""
-      } ${file}`
-    );
+    // TODO: Should we console.log the output (stdout) of the exec?
+    exec(file, options);
     return file.replace(/\.gr$/, ".gr.wasm");
   } catch (e) {
     console.log(e.stdout.toString());
+    // TODO: Do we also want to print the stderr?
+    // console.log(e.stderr.toString());
```
ospencer, line 11: Yup, let's go ahead and do these.

cli/bin/grain.js
```diff
@@ -42,9 +40,8 @@ function num(val) {
 }
 
 function graincVersion() {
-  const grainc = path.join(__dirname, "grainc.exe");
-
-  return execSync(`${grainc} --version`).toString().trim();
+  // TODO: This prints "unknown" for the JSOO version of grainc
```
ospencer, line 43: I think you said you wanted to just get rid of this? Probably fine if we're bundling the compiler in anyway.

cli/bin/compile.js
```diff
@@ -1,18 +1,10 @@
...
   try {
-    execSync(
-      `${grainc} --stdlib=${options.stdlib} ${
-        options.cflags ? options.cflags : ""
-      } ${file}`
-    );
+    exec(file, options, { stdio: "inherit" });
     return file.replace(/\.gr$/, ".gr.wasm");
   } catch (e) {
-    console.log(e.stdout.toString());
```
phated, line 15: @ospencer using `stdio: inherit` here instead of explicitly console.logging

cli/bin/exec.js
```diff
@@ -0,0 +1,26 @@
+const path = require("path");
+const { execSync } = require("child_process");
+const fs = require("fs");
+
+function getGrainc() {
+  const grainc = path.join(__dirname, "grainc.exe");
+
+  // TODO: Maybe make an installable path & check it?
```
phated, line 8: Needs an issue

phated: #589

cli/bin/lsp.js
```diff
@@ -1,23 +1,14 @@
...
-    execSync(
-      `${grainc} --stdlib=${options.stdlib} ${
-        options.cflags ? options.cflags : ""
-      } -g --lsp ${file}`,
-      { stdio: "inherit" }
-    );
+    exec(`-g --lsp ${file}`, options, { stdio: "inherit" });
     process.exit();
   } catch (e) {
-    console.log(e);
```
phated, line 20: @ospencer this uses `stdio: inherit` already, so we don't need to log the err here.

phated:
@ospencer hmm, there seems to be an issue with the lsp mode that I don't quite understand

ospencer (approved):
Approved with passing tests!

PR #572, opened 2021-03-23
fix: Parsing of conditional assignment with binary operators

### LINKED ISSUE #558: Compiler: Inline `if` assignment with binop yields ambiguous parse tree
The program in question is
```
if (foo) x = 3 + 5
```

The compiler doesn't know if this should be `(if (foo) x = 3) + 5` or `if (foo) (x = 3 + 5)`.

This should parse as the latter.

ospencer:
Fixes #558.

There's not really a great way to write tests for this without #560, since any inline assignment is a type error. Happy to add some tests after that's merged.

### DIFF
--- a/compiler/src/parsing/parser.dyp
+++ b/compiler/src/parsing/parser.dyp
@@ -154,6 +154,7 @@ binop_expr :
   | binop_expr colon typ { Exp.constraint_ ~loc:(symbol_rloc dyp) $1 $3 } p140
   | lam_expr { $1 } p0
   | non_assign_expr { $1 } p140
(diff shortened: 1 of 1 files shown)


ospencer:
I don't love this solution, and we can open an issue to figure out something better, but this is also probably fine.

phated (approved):
Yeah, let's get this merged and then some additional tests added in #560 and an issue opened to dig into a better solution :good-first-issue:

ospencer:
https://github.com/grain-lang/grain/issues/573

PR #574, opened 2021-03-23
feat: Add GrainDoc for markdown generation using doc comments

phated:
Rebased #410 on top of `main` instead of `master`

I believe this needs to be changed a bunch from my original implementation so marking it as Draft.

### DIFF
--- a/compiler/src/diagnostics/comments.re
+++ b/compiler/src/diagnostics/comments.re
@@ -224,14 +224,17 @@ module Doc = {
     };
   };
 
-  let rec ending_on_lnum = lnum => {
-    let data = Hashtbl.find_opt(comments.by_end_lnum, lnum);
-    switch (data) {
-    | Some((Doc({cmt_content}), _, _)) => data
-    // Hack to handle code that has an attribute on the line before, such as `@disableGC`
-    | None => ending_on_lnum(lnum - 1)
-    | _ => None
+  let ending_on_lnum = lnum => {
+    let rec ending_on_lnum_help = (lnum, check_prev) => {
+      let data = Hashtbl.find_opt(comments.by_end_lnum, lnum);
+      switch (data) {
+      | Some((Doc({cmt_content}), _, _)) => data
+      // Hack to handle code that has an attribute on the line before, such as `@disableGC`
+      | None when check_prev => ending_on_lnum_help(lnum - 1, false)
+      | _ => None
+      };
     };
+    ending_on_lnum_help(lnum, true);
   };
 
   let find_module = () => {
--- a/compiler/src/diagnostics/comments.re
+++ b/compiler/src/diagnostics/comments.re
@@ -31,11 +31,9 @@ module Attribute = {
 
   let extract = comment => {
     let attrs = ref([]);
-    // https://regexper.com/#%5E%40%28param%7Creturns%29%28%5B%20%5Ct%5D%28%5B%5E%3A%5D%2B%29%3A%29%3F%5B%20%5Ct%5D%28.*%29%24
-    let re =
-      Str.regexp(
-        "^@\\(param\\|returns\\|module\\|example\\|section\\)\\([ \t]\\([^:]+\\):\\)?[ \t]\\(.*\\)$",
-      );
+    // TODO: We should probably be using a less-janky RegExp library
+    // https://regexper.com/#%5E%40%28%5Ba-zA-Z_%5D%2B%29%28%5B%20%5D%2B%28%5B%5E%3A%5D%2B%29%3A%29%3F%5B%20%5D%2B%28.*%29%24
+    let re = Str.regexp({|^@\([a-zA-Z_]+\)\([ ]+\([^:]+\):\)?[ ]+\(.*\)$|});
     let out =
       Str.global_substitute(
         re,
--- a/compiler/src/diagnostics/comments.re
+++ b/compiler/src/diagnostics/comments.re
@@ -3,6 +3,9 @@ open Grain_typed;
 
 // Attributes in a comment are prefixed with `@` symbol, such as `@param`
 module Attribute = {
+  exception InvalidAttribute(string);
+  exception MalformedAttribute(string, string);
+
   type attr_name = string;
   type attr_desc = string;
   // The `attr_type` always starts as `None` and is applied later by something like Graindoc
@@ -29,6 +32,11 @@ module Attribute = {
         attr_desc,
       });
 
+  let try_or_malformed = (~attr, ~hint, fn) =>
+    try(fn()) {
+    | _ => raise(MalformedAttribute(attr, hint))
+    };
+
   let extract = comment => {
     let attrs = ref([]);
     // TODO: We should probably be using a less-janky RegExp library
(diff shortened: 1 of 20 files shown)


ospencer (commented):
Overall I think your approach looks good. The one major thing would be that the docs produced should match the signature of the file. I think the easiest way to achieve that is to maybe iterate the signature and build a hash with the idents that you need docs for... `val_fullpath` on a signature_item should give you the exact identifier (with the right stamp) to look for in the module, so you could then iterate the toplevel statements and check your hash if you need docs for that identifier. That should also handle the case of like 
```
/** my doc */
let foo = 5
export foo
```
where you need docs for `foo` but it doesn't have an export statement directly on it. Even better it also solves this case:
```
/** doc 1 */
export let foo = 5
/** doc 2 */
export let foo = 6
```
since it'll tell you to look for `foo` with the stamp that matches the second exported value.

(that might actually be a well-formedness error, but it's the same deal with two `let foo =` and an `export *`.)

But otherwise, looking stellar!

compiler/src/diagnostics/comments.re
```diff
@@ -0,0 +1,87 @@
...
+
+let get_docblock_from_location =
+    (loc: Grain_parsing.Location.t, comments: list(Typedtree.comment)) => {
+  let lineNumber = loc.loc_start.pos_lnum;
+  let prevLineNumber = lineNumber - 1;
+  List.find_map(
+    (comment: Grain_typed.Typedtree.comment) => {
+      switch (comment) {
+      | Doc({cmt_content, cmt_loc})
+          when cmt_loc.loc_end.pos_lnum == prevLineNumber =>
```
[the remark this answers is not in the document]
phated: So, I've been working with JSDoc/TSDoc a bunch and I believe that docblock comment shouldn't actually be documenting the `let foo` binding and instead would be considered a dangling docblock (such as some header information or an example of using the module) - this is annotated in TSDoc with `@packageDocumentation` but I think that's a bit verbose.

ospencer (commented):
Approach here looks great!

compiler/src/diagnostics/comments.re
```diff
@@ -0,0 +1,112 @@
...
+        ~args: option(list(Types.type_expr))=?,
+        ~returns: option(Types.type_expr)=?,
+        comment,
+      ) => {
+    let attrs = ref([]);
+    // https://regexper.com/#%5E%40%28param%7Creturns%29%28%5B%20%5Ct%5D%28%5B%5E%3A%5D%2B%29%3A%29%3F%5B%20%5Ct%5D%28.*%29%24
+    let re =
+      Str.regexp(
+        "^@\\(param\\|returns\\)\\([ \t]\\([^:]+\\):\\)?[ \t]\\(.*\\)$",
+      );
```
phated, line 73: @ospencer I tried to bust out Ocamllex but failed spectacularly. Maybe we can pair on that if you think it is better than this wild regexp.

phated:
@grain-lang/core GrainDoc is done!

compiler/src/diagnostics/comments.re
```diff
@@ -0,0 +1,206 @@
...
+    // Currently only accepts single-line examples
+    | Example({attr_desc})
+    | Section({
+        attr_name,
+        attr_desc,
+      });
+
+  let extract = comment => {
+    let attrs = ref([]);
+    // https://regexper.com/#%5E%40%28param%7Creturns%29%28%5B%20%5Ct%5D%28%5B%5E%3A%5D%2B%29%3A%29%3F%5B%20%5Ct%5D%28.*%29%24
```
ospencer, line 34: Updated link for the current regex:
```suggestion
    // https://regexper.com/#%5E%40%28param%7Creturns%7Cmodule%7Cexample%7Csection%29%28%5B%20%5Ct%5D%28%5B%5E%3A%5D%2B%29%3A%29%3F%5B%20%5Ct%5D%28.*%29%24
```

compiler/src/diagnostics/comments.re
```diff
@@ -0,0 +1,206 @@
...
+    let re =
+      Str.regexp(
+        "^@\\(param\\|returns\\|module\\|example\\|section\\)\\([ \t]\\([^:]+\\):\\)?[ \t]\\(.*\\)$",
```
ospencer, line 37: I'd make a tiny tweak here to consume more whitespace:
```suggestion
        "^@\\(param\\|returns\\|module\\|example\\|section\\)\\([ \t]+\\([^:]+\\):\\)?[ \t]+\\(.*\\)$",
```

Would we want to match word characters instead of the explicit param/returns/module/example/section? That'd have the benefit of being able to error/warn on unknown attributes, which would be handy and would also prevent an unknown one from showing up in a description.

compiler/src/diagnostics/comments.re
```diff
@@ -0,0 +1,206 @@
...
+    let module_comments = ref([]);
+    Hashtbl.iter(
+      (_, (_comment, _desc, attrs) as comment) =>
+        if (List.exists(Attribute.is_module, attrs)) {
+          module_comments := [comment, ...module_comments^];
+        },
+      comments.by_start_lnum,
+    );
+    if (List.length(module_comments^) > 1) {
+      failwith("More than one @module comment is not supported");
```
ospencer, line 189:
```suggestion
      failwith("More than one @module block is not supported");
```

ospencer (commented):
I think this is downright spectacular. I don't have much to say about it really. I really like the Markdown and Comments modules.

Only big thing would be the feature request I mentioned on Discord (which doesn't need to come in this PR) and throwing an error if a module doesn't have the necessary blocks. That, and some documentation 🙂 but happy to help with that.

marcusroberts (commented):
This is really great stuff!

I'm probably not going to spot issues yet like the others, so consider it a "soft" approval, but some great code in there I'm going to borrow :)

peblair (approved):
This is :fire:

compiler/src/diagnostics/comments.re
phated: Yeah, I like that idea! I'll make that change in this PR

compiler/src/diagnostics/comments.re
phated: Updated to a newer one.

compiler/src/diagnostics/comments.re
phated: Updated. Also added better error messages!

phated:
> Only big thing would be the feature request I mentioned on Discord (which doesn't need to come in this PR) and throwing an error if a module doesn't have the necessary blocks. 

I added support for printing names and type signatures, even if you don't have sections, in this PR for you @ospencer

compiler/src/diagnostics/comments.re
```diff
@@ -0,0 +1,264 @@
...
+
+  let rec ending_on_lnum = lnum => {
+    let data = Hashtbl.find_opt(comments.by_end_lnum, lnum);
+    switch (data) {
+    | Some((Doc({cmt_content}), _, _)) => data
+    // Hack to handle code that has an attribute on the line before, such as `@disableGC`
+    | None => ending_on_lnum(lnum - 1)
+    | _ => None
+    };
+  };
```
ospencer, line 235:
```suggestion
  let ending_on_lnum = lnum => {
    let rec ending_on_lnum_help = (lnum, check_prev) => {
      let data = Hashtbl.find_opt(comments.by_end_lnum, lnum);
      switch (data) {
      | Some((Doc({cmt_content}), _, _)) => data
      // Hack to handle code that has an attribute on the line before, such as `@disableGC`
      | None when check_prev => ending_on_lnum_help(lnum - 1, false)
      | _ => None
      };
    };
    ending_on_lnum_help(lnum, true);
  };
```

ospencer (commented):
Amazing, it does exactly what I want 😄 

I did run into one issue—when I tried getting docs for a file without an `@module` block, `grain doc` hung. I was able to track down the issue, and I suggested a fix. I don't know if you'd want to do something better, but this one works!

Not going to request changes since I won't be around later today, but consider that fixed as an approval!

ospencer (approved):
Let the docs flow!

PR #579, opened 2021-03-24
chore: Don't check for alignment of boxed numbers

### LINKED ISSUE #309: Compiler: Verify that boxed numbers are 8-byte aligned
We should also check that `malloc` always returns a block that is 8-byte aligned.

ospencer:
Closes #309


phated (approved):
Woohoo! ❤️ you!

PR #580, opened 2021-03-24
chore: fix mistakes in my sys/file docs

phated:
@ospencer I think I messed these up. Can you confirm?


ospencer (approved):
Yes, that's correct. 👍

PR #581, opened 2021-04-02
fix: Properly incRef exception printers

ospencer:
chore!: Rename `registerBasePrinter` to `dangerouslyRegisterBasePrinter` in runtime/exception
fix: Throwing an exception now traps immediately in all cases

I added a test for the last fix, but there isn't a great way to test that the incRefs weren't happening. However, it gets exercised pretty well by changes in the linker PR.


phated (approved):
:approved!!!:

PR #582, opened 2021-04-03
chore: upgrade binaryen.ml to 0.9.1

phated:
This version includes a patch for our Int64 const generations in JS.

Looks like we got some dune upgrades and stuff as well since esy did a new resolve on packages.


ospencer (approved):
Excellent!

PR #583, opened 2021-04-03
chore(cli): Remove flagged-respawn for built-in v8 method

phated:
I ripped out v8flags and flagged-respawn in favor of `v8.setFlagsFromString` because flagged-respawn doesn't work with pkg.


ospencer (approved):
Very nice. I imagine it's a tad faster, too, since a new process isn't spawned anymore?

phated:
Should be! GH Actions took 2m50s to run the tests instead of 3m30s from my other PR

PR #584, opened 2021-04-03
feat!: Static linking of Grain modules

### LINKED ISSUE #28: Multi-Module Packaging
Currently, we require modules to be dynamically linked by the Grain runtime. It would make deployment (and certain dependency management) significantly easier if there was an option to statically link a module's dependencies. Bonus points if we can do this for specific subsets of dependencies (as opposed to requiring all dependencies be statically linked).

ospencer:
Closes #28.

feat: Support for more WebAssembly runtimes, including Wasmtime and Wasmer
fix!: Correct type signature for `_start`
chore!: Introduce `_gmain` for old behavior of `_start`
chore!: Tail calls must be enabled explicitly via `--experimental-wasm-tail-call`

_Still a draft as this depends on a number of PRs in https://github.com/grain-lang/binaryen.ml/pulls, but everything is finished._

# Static Linking

This PR implements a static linker for Grain modules, which allows us to keep our separate compilation but produce a single wasm file that works with the existing runtime and more runtimes than ever before, including Wasmtime, Wasmer, and Wasm3. This PR may not be enormous, but it's riding on the back of more than two months of hard work! 🎉 

## Important info

`grainc` will perform linking by default. This means that `grain compile hello.gr && wasmtime hello.gr.wasm` will work without having to worry about flags / wondering why that _doesn't_ work. Linking can be disabled via the `--no-link` flag.

Since most wasm runtimes don't support tail calls yet, they're now off by default, and can be enabled via `--experimental-wasm-tail-call`.

`-p` in the `grain` CLI is now deprecated, as it doesn't (and can't) work for linked modules. The tests still rely on it, but once we implement snapshot testing, we can remove it.

## Other info

Given this is implemented via Binaryen, I learned a ton and can probably contribute some info back to https://github.com/WebAssembly/binaryen/issues/2767. The approach used isn't too far off from being able to work generically (and I even considered implementing the linker directly in Binaryen.ml), though modules would need to adhere to certain constraints.

## Other thoughts

We may want to consider renaming the `Compiled` step in `compile.re`. It's sort of an odd name now, but I can't quite think of anything better.

### DIFF
--- a/compiler/esy.json
+++ b/compiler/esy.json
@@ -29,6 +29,7 @@
     "@opam/fp": "0.0.1",
     "@opam/fs": "0.0.2",
     "@opam/grain_dypgen": "0.2",
+    "@opam/ocamlgraph": ">= 2.0.0",
     "@opam/ppx_deriving_yojson": ">= 3.5.2",
     "@opam/ppx_expect": ">= 0.14.0",
     "@opam/ppx_sexp_conv": ">= 0.14.0",
@@ -44,6 +45,7 @@
     "@opam/ounit": ">= 2.0.0"
   },
   "resolutions": {
+    "@grain/binaryen.ml": "grain-lang/binaryen.ml#8a21a8a6ce6abad71f36e5f13492a7c7a625cb00",
     "@opam/ocamlfind": "1.8.1",
     "@opam/fp": "facebookexperimental/reason-native:fp.opam#a33f1528a6dd86c67f365e226c81312733181c87",
     "@opam/fs": "facebookexperimental/reason-native:fs.opam#a33f1528a6dd86c67f365e226c81312733181c87",
--- a/compiler/src/compile.re
+++ b/compiler/src/compile.re
@@ -2,6 +2,7 @@ open Grain_parsing;
 open Grain_typed;
 open Grain_middle_end;
 open Grain_codegen;
+open Grain_linking;
 open Optimize;
 
 type input_source =
@@ -17,6 +18,8 @@ type compilation_state_desc =
   | Optimized(Anftree.anf_program)
   | Mashed(Mashtree.mash_program)
   | Compiled(Compmod.compiled_program)
+  | ObjectFileEmitted(Compmod.compiled_program)
+  | Linked(Compmod.compiled_program)
   | Assembled;
 
 type compilation_state = {
@@ -67,6 +70,8 @@ let log_state = state =>
       prerr_string("\nMashed program:\n");
       prerr_sexp(Mashtree.sexp_of_mash_program, mashed);
     | Compiled(compiled) => prerr_string("\nCompiled successfully")
+    | ObjectFileEmitted(compiled) => prerr_string("\nEmitted successfully")
+    | Linked(linked) => prerr_string("\nLinked successfully")
     | Assembled => prerr_string("\nAssembled successfully")
     };
     prerr_string("\n\n");
@@ -144,7 +149,14 @@ let next_state = ({cstate_desc, cstate_filename} as cs) => {
       | Some(outfile) => Emitmod.emit_module(compiled, outfile)
       | None => ()
       };
-      Binaryen.Module.dispose(compiled.asm);
+      ObjectFileEmitted(compiled);
+    | ObjectFileEmitted(compiled) =>
+      Linked(Linkmod.statically_link_wasm_module(compiled))
+    | Linked(linked) =>
+      switch (cs.cstate_outfile) {
+      | Some(outfile) => Emitmod.emit_module(linked, outfile)
+      | None => ()
+      };
       Assembled;
     | Assembled => Assembled
     };
@@ -229,6 +241,18 @@ let stop_after_compiled =
   fun
   | {cstate_desc: Compiled(_)} => Stop
   | s => Continue(s);
+let stop_after_object_file_emitted =
+  fun
+  | {cstate_desc: ObjectFileEmitted(_)} => Stop
+  | s => Continue(s);
+let stop_after_linked =
+  fun
+  | {cstate_desc: Linked(_)} => Stop
+  | s => Continue(s);
+let stop_after_assembled =
+  fun
+  | {cstate_desc: Assembled} => Stop
+  | s => Continue(s);
 
 let anf = Linearize.transl_anf_module;
 
@@ -267,5 +291,13 @@ let () =
 let () =
   Env.compile_module_dependency :=
     (
-      (input, outfile) => ignore(compile_file(~outfile, ~reset=false, input))
+      (input, outfile) =>
+        ignore(
+          compile_file(
+            ~outfile,
+            ~reset=false,
+            ~hook=stop_after_object_file_emitted,
+            input,
+          ),
+        )
     );
--- a/compiler/src/linking/link.re
+++ b/compiler/src/linking/link.re
@@ -1,0 +1,617 @@
+open Grain_codegen;
+open Compmod;
+open Grain_typed;
+open Cmi_format;
+open Binaryen;
+open Graph;
+
+module String = {
+  type t = string;
+  let compare = compare;
+  let hash = Hashtbl.hash;
+  let equal = (==);
+  let default = "";
+};
+
+module G = Imperative.Digraph.Concrete(String);
+module Topo = Topological.Make(G);
+
+let dependency_graph = G.create(~size=10, ());
+let modules: Hashtbl.t(string, Module.t) = Hashtbl.create(10);
+
+let main_module = "main";
+
+let grain_main = "_gmain";
+let grain_start = "_start";
+let function_table = "tbl";
+
+let resolve = mod_name => {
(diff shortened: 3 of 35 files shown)


compiler/esy.json
```diff
@@ -42,6 +43,7 @@
     "@opam/ounit": ">= 2.0.0"
   },
   "resolutions": {
+    "@grain/binaryen.ml": "link:../../binaryen.ml",
```
phated, line 46: Don't forget to use a git hash pointing at binaryen.ml so we can get the tests to pass here.

compiler/src/linking/link.re
```diff
@@ -0,0 +1,529 @@
...
+    globalize_names(local_names, Expression.memory_grow_get_delta(expr))
+  | Unreachable => ()
+  | Pop
+  | RefNull
+  | RefIs
+  | RefFunc
+  | RefEq
+  | Try
+  | Throw
+  | Rethrow => failwith("NYI")
```
phated, line 259: `failwith` or `raise`?

compiler/src/linking/link.re
```diff
@@ -0,0 +1,529 @@
...
+  | RttCanon
+  | RttSub
+  | StructNew
+  | StructGet
+  | StructSet
+  | ArrayNew
+  | ArrayGet
+  | ArraySet
+  | ArrayLen
+  | RefAs => failwith("NYI")
```
phated, line 285: `failwith` or `raise`?

compiler/src/linking/link.re
```diff
@@ -0,0 +1,529 @@
...
+let table_offset = ref(0);
+let module_id = ref(0);
+
+let link_all = (linked_mod, dependencies) => {
+  table_offset := 0;
+
+  let link_one = dep => {
+    let local_names: Hashtbl.t(string, string) = Hashtbl.create(128);
+
+    let wasm_mod = Hashtbl.find(modules, dep);
```
phated, line 298: Question: Does `.find()` raise if the key is not found? If so, do we want to defensively code around it?

compiler/src/linking/link.re
```diff
@@ -0,0 +1,529 @@
...
+                  wasm_mod,
+                  Literal.int32(Int32.of_int(table_offset^)),
+                )
+              | "moduleRuntimeId" =>
+                incr(module_id);
+                Expression.const(
+                  wasm_mod,
+                  Literal.int32(Int32.of_int(module_id^)),
+                );
+              | _ => assert(false)
```
phated, line 334: Why `assert(false)` instead of a `failwith` or `raise`?

compiler/src/linking/link.re
```diff
@@ -0,0 +1,529 @@
...
+        let imported_module = Import.function_import_get_module(func);
+        if (is_grain_module(imported_module)) {
+          let imported_name = Import.function_import_get_base(func);
+          let internal_name = Function.get_name(func);
+          let new_name =
+            Hashtbl.find(
+              Hashtbl.find(exported_names, imported_module),
+              imported_name,
+            );
+          Hashtbl.add(local_names, internal_name, new_name);
```
phated, line 382: This code is repeated a couple times.

compiler/src/linking/link.re
```diff
@@ -0,0 +1,529 @@
...
+        };
+      },
+      imported_funcs,
+    );
+
+    List.iter(
+      func => {
+        let internal_name = Function.get_name(func);
+        let new_name = gensym(internal_name);
+        Hashtbl.add(local_names, internal_name, new_name);
```
phated, line 408: Whenever you gensym a new name, you need to add it to the local_names hashtbl, right? Maybe those should be bundled so it isn't missed during a code cleanup or something.

compiler/src/linking/link.re
```diff
@@ -0,0 +1,529 @@
...
+        List.init(size, i =>
+          Hashtbl.find(
+            local_names,
+            Table.element_segment_get_data(segment, i),
+          )
+        );
+      ignore @@
+      Table.add_active_element_segment(
+        linked_mod,
+        "tbl",
```
phated, line 460: I've seen this string a couple times in this file. Should it be a constant at the top?

compiler/src/linking/link.re
```diff
@@ -0,0 +1,529 @@
...
+    Topo.fold((dep, acc) => [dep, ...acc], dependency_graph, []);
+  let linked_mod = Module.create();
+  link_all(linked_mod, dependencies);
+  let _ =
+    Module.set_features(
+      linked_mod,
+      [
+        Features.mvp,
+        Features.multivalue,
+        Features.tail_call,
```
phated, line 519: We can always enable tail_call, even it not supported in a runtime?

compiler/src/linking/link.re
```diff
@@ -0,0 +1,529 @@
...
+        Features.sign_ext,
+        Features.bulk_memory,
+        Features.mutable_globals,
+      ],
+    );
+  let _ = Module.set_low_memory_unused(1);
+  assert(Module.validate(linked_mod) == 1);
```
phated, line 526: `assert`, `failwith` or `raise`?

phated (commented):
Some thoughts, mostly about the binaryen.ml API design and about how to handle errors in the linker.

esy.lock/.gitattributes
```diff
@@ -0,0 +1,3 @@
+
+# Set eol to LF so files aren't converted to CRLF-eol on Windows.
+* text eol=lf linguist-generated
```
phated, line 3: @ospencer Oh, I just noticed that you accidentally ran `esy` in the top of the monorepo and committed an extra esy.lock directory here.

compiler/src/compile.re
```diff
@@ -267,5 +291,13 @@ let () =
 let () =
   Env.compile_module_dependency :=
     (
-      (input, outfile) => ignore(compile_file(~outfile, ~reset=false, input))
+      (input, outfile) =>
+        ignore(
+          compile_file(
+            ~outfile,
+            ~reset=false,
+            ~hook=stop_after_object_file_emitted,
```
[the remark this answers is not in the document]
ospencer: Previously, it would just compile completely, but now since linking is a step that also happens, this skips linking when compiling dependencies (since they'll all be linked together at the end).

compiler/src/linking/link.re
```diff
@@ -0,0 +1,529 @@
...
+
+let is_global_imported = global =>
+  Import.global_import_get_base(global) != "";
+let is_function_imported = func =>
+  Import.function_import_get_base(func) != "";
+
+let rec globalize_names = (local_names, expr) => {
+  let kind = Expression.get_kind(expr);
+  switch (kind) {
+  | Invalid => failwith("Invalid expression")
```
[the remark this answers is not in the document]
ospencer: We typically use `failwith` with things that are "impossible" and `raise` for errors related to the user's program. The only way we'd ever hit this case is if we generated a wasm module with instructions that the linker didn't know about.

We can probably do a round up on errors, but that's probably for another PR :)

compiler/src/linking/link.re
```diff
@@ -0,0 +1,529 @@
...
+  | Nop => ()
+  | Block =>
+    let internal_name = Expression.block_get_name(expr);
```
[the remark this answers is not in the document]
ospencer: We chatted about these and decided that we'll fix up the Binaryen API when there's a stable v101 release 🚀

compiler/src/linking/link.re
```diff
@@ -0,0 +1,529 @@
...
+  | SIMDReplace
+  | SIMDShuffle
+  | SIMDTernary
+  | SIMDShift
+  | SIMDLoad
+  | SIMDLoadStoreLane
+  | SIMDWiden
+  | Prefetch
+  | MemoryInit
+  | DataDrop => failwith("NYI")
```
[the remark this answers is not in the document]
ospencer: I would go with `failwith` here, but with a better message.

compiler/src/linking/link.re
ospencer: It does. This couldn't ever fail though because the hash is built from the list that is being iterated over. We could add a check here to raise an error, though it'd be a pretty similar one (some form of `Not_found`).

compiler/src/linking/link.re
ospencer: Draft PR 😛

compiler/src/linking/link.re
ospencer: It's similar, but not quite the same. They essentially get to the same end result, but it takes different functions to get there.

compiler/src/linking/link.re
ospencer: Not always, no. I suppose we could have some utility that does this for where we _do_ need to add it to the table.

phated (approved):
@ospencer you are an absolute genius. Amazing work on this!

It seems we both modified the esy.lock so you'll need to resolve that, but otherwise good to go 🎉

PR #585, opened 2021-04-03
Runtime cleanup in preparation of pkg

phated:
This PR moves our runtime deps to devDeps because the `pkg` tool tries to pull in everything in `dependencies`, which causes a bunch of issues. However, we just bundle and ship the bundles anyway, so I changed the `main` and `browser` fields to point at the produced `dist` files.

I also removed treeify because it didn't seem to be used. And I fixed an issue with the re-used identifier `GrainRunner`... no idea why babel casually allows that but real JS runtimes throw.


ospencer (approved):
Makes sense!

PR #586, opened 2021-04-05
chore(stdlib)!: Replace JS entrypoint with locator file

phated:
This replaces the `main` file in the stdlib's `package.json` to point to a valid JS file, which just exports the `__dirname` of the location of the stdlib.

This is necessary because the `pkg` tool parses the `main` entrypoint and it needs to be valid JS, otherwise the packaging fails. Having the `index.js` export the `__dirname` is just a neat trick so you can do `const stdlibLocation = require('@grain/stdlib')`

### DIFF
--- a/stdlib/index.js
+++ b/stdlib/index.js
@@ -1,0 +1,1 @@
+module.exports = __dirname;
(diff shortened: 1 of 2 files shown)


ospencer (approved):
Works for me!

PR #590, opened 2021-04-17
fix(compiler): Implement proper dependency resolver

peblair:
This has the added benefit of being far less harsh on the filesystem—the unit tests run about 40% faster.

### DIFF
--- a/compiler/src/parsing/driver.re
+++ b/compiler/src/parsing/driver.re
@@ -69,3 +69,46 @@ let parse = (~name=?, lexbuf): Parsetree.parsed_program => {
   | parses => raise(Error(startpos, AmbiguousParse(List.map(fst, parses))))
   };
 };
+
+let scan_for_imports = (~defer_errors=true, filename: string): list(string) => {
+  let ic = open_in(filename);
+  let lexbuf = Lexing.from_channel(ic);
+  try({
+    let {Parsetree.comments, Parsetree.statements} = parse(lexbuf);
+    let implicit_opens =
+      List.map(
+        o => {
+          switch (o) {
+          | Grain_utils.Config.Pervasives_mod => "pervasives"
+          | Grain_utils.Config.Gc_mod => "runtime/gc"
+          }
+        },
+        switch (comments) {
+        | [Block({cmt_content}), ..._] =>
+          Grain_utils.Config.with_inline_flags(
+            ~on_error=_ => (),
+            cmt_content,
+            Grain_utils.Config.get_implicit_opens,
+          )
+        | _ => Grain_utils.Config.get_implicit_opens()
+        },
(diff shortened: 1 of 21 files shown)


compiler/src/typed/dependency_graph.re
```diff
@@ -0,0 +1,297 @@
...
+  | Processed;
+
+module type G = {
+  type t;
+  module V: Sig.COMPARABLE;
+  let iter_vertex: (V.t => unit, t) => unit;
+  let iter_succ: (V.t => unit, t, V.t) => unit;
+};
+
+module Topo_make_stable = (G: {
```
peblair, line 31: Ignore this module (it's a copy of `Topological.Make_stable` that I was using for debugging); I'll delete it before marking ready for merge

compiler/src/parsing/driver.re
```diff
@@ -69,3 +69,74 @@ let parse = (~name=?, lexbuf): Parsetree.parsed_program => {
...
+          when
+            Str.string_match(
+              Str.regexp_string("grainc-flags"),
+              cmt_content,
+              0,
+            ) =>
+        // [FIXME] Kind of a gross hack. We should have less hard-coding.
+        let maybe_pervasives =
+          if (Str.string_match(
+                Str.regexp_string("grainc-flags --compilation-mode=runtime"),
```
ospencer, line 90: Since you already check that the string starts with `grainc-flags`, here we probably should just check that the string contains `--compilation-mode=runtime` or `--no-pervasives` since there can be multiple flags supplied.

ospencer: We actually should use Cmdliner to parse this, since it also allows various forms of the flags that would be too annoying to all handle here.

ospencer (approved):

phated (approved):
I tested this on a couple things, and the problems I ran into are happening on `main`, so I think this should be good to go. It shaves about ~10 seconds off the cold boot of Packaged Grain.

compiler/src/utils/config.re
```diff
@@ -446,11 +446,74 @@ let lsp_mode =
...
+
+let module_search_path = () => {
+  module_search_path_from_base_path(base_path^);
+};
+
+let apply_inline_flags = (~on_error, cmt_content) =>
+  if (Str.string_match(Str.regexp_string("grainc-flags"), cmt_content, 0)) {
+    let err_buf = Buffer.create(80);
+    let err = Format.formatter_of_buffer(err_buf);
+    let result = apply_inline_flags(~err, cmt_content);
```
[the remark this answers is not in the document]
ospencer: It's earlier in this file!

phated:
Merging this! Awesome work @peblair!

PR #595, opened 2021-04-19
chore: release testing-release

phated:
:robot: I have created a release \*beep\* \*boop\*
---
<details><summary>grain: 0.3.0</summary>

### ⚠ BREAKING CHANGES

* Correct type signature for `_start`
* Introduce `_gmain` for old behavior of `_start`
* Tail calls must be enabled explicitly via `--experimental-wasm-tail-call`
* Static linking of Grain modules (#584)
* **stdlib:** Replace JS entrypoint with locator file (#586)
* Streamline runtime dependencies & entrypoints (#585)
* Rename `registerBasePrinter` to `dangerouslyRegisterBasePrinter` in runtime/exception
* Use Grain exceptions instead of JS exceptions (#565)
* Update assignment semantics (#560)
* Grain implementation of memory allocator (#530)
* Refactor value tags (#526)
* **stdlib:** Ensure string methods are data-last (#494)
* **runtime:** Print strings without quotes (#495)
* Char literals (#477)
* Char (#474)
* Rework memory management (#461)
* convert fallible List methods to return Option (#460)
* change return type of Array.find/findIndex to Option (#459)
* upgrade binaryen to 0.4.0 & update test invocation to work with it (#435)
* Add support for bitwise operators (#425)
* remove `^` as unbox operator (fixes #183) (#426)

### Features

* `for` loops ([#525](https://www.github.com/grain-lang/grain/issues/525)) ([1032a3d](https://www.github.com/grain-lang/grain/commit/1032a3d4e92e8947b6406147e8992b9a0dd5b9e2))
* Add ability to `throw` exceptions ([1f1cd4a](https://www.github.com/grain-lang/grain/commit/1f1cd4a90f853a2c6290e736043d008643f768f7))
* Add ability to register custom exception printers ([1f1cd4a](https://www.github.com/grain-lang/grain/commit/1f1cd4a90f853a2c6290e736043d008643f768f7))
* add command line option to disable garbage collector ([#239](https://www.github.com/grain-lang/grain/issues/239)) ([7f27a0a](https://www.github.com/grain-lang/grain/commit/7f27a0a5b700eff0a0a80401262286a4762f8c44))
* Add Exception stdlib with Exception.registerPrinter ([1f1cd4a](https://www.github.com/grain-lang/grain/commit/1f1cd4a90f853a2c6290e736043d008643f768f7))
[...]

PR #601, opened 2021-04-20
chore: Prepare monorepo for release-please

phated:
This adds 2 files to the root of the monorepo to support creating releases with https://github.com/googleapis/release-please

I tested the setup with #595 and landed a few fixes upstream to ensure the OCaml releaser could be used in a monorepo.


ospencer (approved):
Do we not want pretty changelogs? (I imagine it probably messes with release-please, yeah?)

phated:
@ospencer we don't prettify in CI, we only run the "check" and then fail (because you can't prettify PRs)

PR #602, opened 2021-04-20
fix(runtime): Return correct pointer to morecore

ospencer:


phated (approved):
@ospencer Thank you so much for fixing this! Now both day 1 AoC problems run successfully! 🎊

PR #603, opened 2021-04-20
ci: Add most simple release workflow

phated:
This adds a Release workflow that currently just runs release-please. I'll expand upon this to generate the Packaged Grain binaries and attach them to the release.


ospencer (approved):
🤖

PR #604, opened 2021-04-20
chore: Restore ocamlfind 1.8.1

phated:
@ospencer's computer was acting hella weird and wouldn't build without this resolution still, so I guess we put it back for now.


ospencer (approved):
🙇

PR #605, opened 2021-04-20
chore: Use main branch name in workflows

phated:
Doesn't seem like that `$default-branch` thing works outside of templates. I also fixed the docker workflow branch name.


ospencer (approved):
Good find!

PR #606, opened 2021-04-20
ci: Add WIP binary uploader workflow

phated:
I need something like this in the default branch to test.


phated:
Realized I can work on this in my personal repo, so closing.

PR #608, opened 2021-04-20
chore(cli): Remove debug flag from lsp exec call

phated:
I'm not sure why the debug flag was added to the lsp command, but it causes .modsig and .wast files to be generated all over the place. I think we should remove it unless @ng-marcus or @ospencer has a reason it needs to be there.


ng-marcus (approved):
I checked, and it's not used by the LSP.   It may have been a hangover from when I was creating it.

ospencer (approved):

Issue #436, opened 2020-11-25
Re-enable inline tests once bug is fixed upstream


phated:
Disabled in #434 

Ref https://github.com/janestreet/ppx_inline_test/issues/23

PR #609, opened 2021-04-20
fix: Remove ppx_expect entirely to build on Windows

phated:
In #434, we disabled inline tests to help with Windows builds, however, the ppx still depended on time_now which is broken on Windows.

This comments out the tests and remove the dep, in order to build on Windows. We'll turn these back on with #436


ospencer (approved):
🏠

PR #610, opened 2021-04-20
ci: Add Binary creation & upload to release workflow

phated:
I tested this workflow in my personal fork: https://github.com/phated/grain/releases/tag/grain-v0.4.0

I think we should just manually do the npm publishes on stdlib and runtime for the 0.3.0 release (since they don't exist yet). I'll open an issue to automate that.


ospencer (approved):
That plan is fine with me!

PR #613, opened 2021-04-21
feat(cli)!: Pass compiler flags from the CLI to grainc directly

ospencer:
chore(cli)!: Remove `-f` CLI option
chore(compiler)!: Remove --principal-types and --recursive-types compiler flags
chore(compiler)!: Rename --cdebug to --verbose

### DIFF
--- a/cli/bin/exec.js
+++ b/cli/bin/exec.js
@@ -17,10 +17,15 @@ function getGrainc() {
 
 const grainc = getGrainc();
 
-function exec(commandOrFile = "", options = {}, execOpts = { stdio: "pipe" }) {
-  let cflags = options.cflags ? options.cflags : "";
-  let stdlib = options.stdlib ? `--stdlib=${options.stdlib}` : "";
-  return execSync(`${grainc} ${stdlib} ${cflags} ${commandOrFile}`, execOpts);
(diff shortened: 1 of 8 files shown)


cli/bin/exec.js
```diff
@@ -17,10 +17,28 @@ function getGrainc() {
...
+    if (value instanceof Array && value.length > 0) {
+      flags.push(`${option.long || option.short} ${value.join(",")}`);
+    } else if (typeof value === "string" || typeof value === "number") {
+      flags.push(`${option.long || option.short} ${value}`);
+    } else if (
+      (option.negate && value === false) ||
+      (!option.negate && value === true)
+    ) {
+      flags.push(option.long || option.short);
+    }
```
ospencer, line 38: Should we emit some sort of warning if there's a flag whose type we don't understand?

cli/bin/grain.js
```diff
@@ -51,33 +62,70 @@ program
     process.exit(0);
   })
   .description("Compile and run Grain programs. 🌾")
-  .option("-p, --print-output", "print the output of the program")
+  .addOption(new program.Option("-p, --print-output").hideHelp())
```
ospencer, line 65: This is to allow `-p` for the tests, but not expose it on the man page.

phated (approved):
Awesome work! 🎉

PR #614, opened 2021-04-21
fix(cli): Rework JSOO hacks to ensure stdin always gets data

phated:
Thanks to some amazing detective work by @ng-marcus - we discovered that doing `readSync` on the `0` file descriptor doesn't always work in node (depending on how values are pushed into stdin); however, `readFileSync` on the `0` file descriptor always works as expected!

With this discovery, I reworked the way we workaround JSOO's mock stdin by using their `Sys_js.set_channel_filler` API, which then just calls `fs.readFileSync(0)`. Unfortunately, those APIs can't be using in a native build, so I needed to do some dune trickery to concat the workaround to the top of `grainc.re` when we are building for JS.

I wonder if @hhugo has a better way to handle this. I'm happy to open an issue for cleanup.


ospencer (approved):
Huzzah!

hhugo:
It would be nice to have this working out of the box when using node. Tuning the Jsoo runtime should do the trick

hhugo:
Do you want to give it a go and open a PR ? (or at least open an issue)

hhugo:
it seems the patch to `caml_make_path` should also be upstreamed

PR #615, opened 2021-04-21
fix(compiler): Resolve names when doing CRC checks

ospencer:
This seems to be the underlying issue of why large portions of the dependency graph would be recompiled unnecessarily. With my checks, it looks like only changed files are recompiled.


phated (approved):
Whoa! This dropped a warm build from Packaged Grain down to 19s. Good catch 🎉

PR #620, opened 2021-04-24
feat(cli)!: Split -g into --debug and --wat

ospencer:
This PR splits the `-g` `grainc` flag into `--debug` and `--wat`. The WebAssembly Text files produced are now named with a `.gr.wat` extension instead of `.gr.wasm.wast` to no longer use the deprecated `wast` name, and to match `.gr.wasm` <-> `.gr.wat`.

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -108,7 +108,8 @@ program
     "--experimental-wasm-tail-call",
     "enables tail-call optimization"
   )
-  .graincOption("-g", "compile with debugging information")
+  .graincOption("--debug", "compile with debugging information")
+  .graincOption("--wat", "additionally produce a WebAssembly Text (.wat) file")
   .graincOption(
     "--hide-locs",
     "hide locations from intermediate trees. Only has an effect with `--verbose`"
--- a/.gitignore
+++ b/.gitignore
@@ -19,8 +19,8 @@ notes.org
 node_modules/
 *.wasm
 *.wasm.map
-*.wasm.wast
-*.wasm.modsig
+*.wat
+*.modsig
 .tern-port
 setup.data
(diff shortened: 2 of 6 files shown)


phated (changes requested):
I think this is technically a breaking change for the CLI.

ng-marcus:
We do have the conflict with exit gracefully with -g though, so it would be good to get some version of this in so we can both debug and exit gracefully.

phated:
@ng-marcus good call. Should we mark this as a fix then? Or should we get a fix version out that removes `-g` from debug only?

ng-marcus:
@phated I'm not sure of the impact of the --wat change.  Is that breaking?

cli/bin/grain.js
```diff
@@ -108,7 +108,8 @@ program
     "--experimental-wasm-tail-call",
     "enables tail-call optimization"
   )
-  .graincOption("-g", "compile with debugging information")
+  .graincOption("--debug", "compile with debugging information")
+  .graincOption("--wat", "additonally produce a WebAssembly Text (.wat) file")
```
follower, line 112: Typo: "additonally" -> "additionally"

ospencer: Good catch, thanks!

ospencer:
@marcusroberts It's breaking in the sense that the flags changed and the produced files have different names.

ng-marcus (approved):

ng-marcus:
Will be good to have this in the next version then, and good to get -g back for graceful exit

marcusroberts (approved):

peblair (approved):

phated (approved):
As you ask 🔥

PR #629, opened 2021-04-27
feat: Export number runtime functions as operators & deprecate old identifiers

### LINKED ISSUE #546: Runtime: Export number function names as the operator
We currently export the functions as names like `numberPlus` (which should have been `numberAdd`, but I digress) but it should instead just be exported as `(+)`.

jozanza:
Closes #546

### DIFF
--- a/stdlib/pervasives.gr
+++ b/stdlib/pervasives.gr
@@ -5,24 +5,24 @@ import Exception from "runtime/exception"
 import { equal as (==) } from "runtime/equal"
 
 import {
-  numberIncr as incr,
-  numberDecr as decr,
-  numberPlus as (+),
-  numberMinus as (-),
-  numberTimes as (*),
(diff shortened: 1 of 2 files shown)


ospencer:
@jozanza Looks great! Since you changed some of the functions to say `Add` instead of `Plus`, could you also change `Minus` to `Sub`? 🙏

jozanza:
You got it

phated (changes requested):
This is breaking so I'm putting a pause on it until 0.4.0 work begins to land

jozanza:
I just pushed another commit that continues to export the deprecated methods, similar to #635

phated (approved):
I think this looks solid to merge with the deprecation strategy and messaging!

ospencer (approved):

PR #630, opened 2021-04-27
chore: Use Option.value with default instead of switch statements

### LINKED ISSUE #520: Compiler: Refactor ast_helper.re to use Option.default
We've got a bunch of places where we're using simple `switch` statements instead of just using `Option.default`.

jozanza:
Closes #520


ospencer (approved):
Sweet!

ng-marcus (approved):

phated (approved):
Such pretty, clean code ❤️

PR #633, opened 2021-04-27
Bytes

jozanza:
# Summary
Add a new standard lib `Bytes` module

## API
```reason
// Sets a signed 8-bit integer starting at the given byte index.
setInt8: (Number, Int32, Bytes) -> Void

// Gets a signed 8-bit integer starting at the given byte index.
getInt8S: (Number, Bytes) -> Int32

// Gets an unsigned 8-bit integer starting at the given byte index.
getInt8U: (Number, Bytes) -> Int32

// Sets a signed 16-bit integer starting at the given byte index.
setInt16: (Number, Int32, Bytes) -> Void

// Gets a signed 16-bit integer starting at the given byte index.
getInt16S: (Number, Bytes) -> Int32

// Gets an unsigned 16-bit integer starting at the given byte index.
getInt16U: (Number, Bytes) -> Int32

// Sets a signed 32-bit integer starting at the given byte index.
setInt32: (Number, Int32, Bytes) -> Void

// Gets a signed 32-bit integer starting at the given byte index.
getInt32: (Number, Bytes) -> Int32

// Sets a 32-bit float starting at the given byte index.
setFloat32: (Number, Float32, Bytes) -> Void

// Gets a signed 32-bit float starting at the given byte index.
getFloat32: (Number, Bytes) -> Float32

// Sets a signed 64-bit integer starting at the given byte index.
setInt64: (Number, Int64, Bytes) -> Void

// Gets a signed 64-bit integer starting at the given byte index.
getInt64: (Number, Bytes) -> Int64

// Sets a 64-bit float starting at the given byte index.
setFloat64: (Number, Float64, Bytes) -> Void

// Gets a signed 64-bit float starting at the given byte index.
getFloat64: (Number, Bytes) -> Float64

// Return a new Bytes buffer that contains the same bytes as the argument.
copy: (Bytes) -> Bytes

// Returns a new Bytes buffer containing a subset of the original Bytes buffer.
slice: (Number, Number, Bytes) -> Bytes

// Add or remove bytes from the start and/or end of a byte sequence.
// A positive number represents bytes to add, while a negative number represents bytes to remove.appended.
resize: (Number, Number, Bytes) -> Bytes

// Copies a range of bytes from a source buffer to a given location in a destination buffer.
move: (Number, Number, Number, Bytes, Bytes) -> Void

// Get the byte length of a Bytes buffer.
length: (Bytes) -> Number

// Creates a new Bytes buffer that contains the bytes of both buffers a and b.
append: (Bytes, Bytes) -> Bytes

// Creates a new String from a Bytes buffer.
toString: (Bytes) -> String

// Creates a new byte sequence from a String.
fromString: (String) -> Bytes

// Fills a byte sequence with a given value.
fill: (Int32, Bytes) -> Void

// Make a new Bytes buffer of n-bytes size.
make: (Number) -> Bytes

// An empty Bytes buffer
empty: Bytes
```

## Things to look for

- [ ] Memory leaks
- [ ] Test coverage
- [ ] Documentation
- [ ] Exception throwing
- [ ] Performance
- [ ] Anything else I might have missed! 😅

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -10,6 +10,7 @@ const v8 = require("v8");
  *
  * This seems to work for our needs with Node 14, but we should be cautious when updating.
  */
+v8.setFlagsFromString("--experimental-wasm-bigint");
 v8.setFlagsFromString("--experimental-wasm-return-call");
 
 const program = require("commander");
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -10,7 +10,6 @@ const v8 = require("v8");
  *
  * This seems to work for our needs with Node 14, but we should be cautious when updating.
  */
-v8.setFlagsFromString("--experimental-wasm-bigint");
 v8.setFlagsFromString("--experimental-wasm-return-call");
 
 const program = require("commander");
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -1587,7 +1587,7 @@ let call_lambda = (~tail=false, wasm_mod, env, func, (argsty, retty), args) => {
   );
 };
 
-let allocate_bytes_from_buffer = (wasm_mod, env, buf, tag, label) => {
+let allocate_byte_like_from_buffer = (wasm_mod, env, buf, tag, label) => {
   let ints_to_push: list(int64) = buf_to_ints(buf);
   let get_swap = () => get_swap(wasm_mod, env, 0);
   let tee_swap = tee_swap(~skip_incref=true, wasm_mod, env, 0);
@@ -1632,7 +1632,7 @@ let allocate_bytes_from_buffer = (wasm_mod, env, buf, tag, label) => {
 let allocate_string = (wasm_mod, env, str) => {
   let buf = Buffer.create(80);
   Buffer.add_string(buf, str);
-  allocate_bytes_from_buffer(
+  allocate_byte_like_from_buffer(
     wasm_mod,
     env,
     buf,
@@ -1644,7 +1644,7 @@ let allocate_string = (wasm_mod, env, str) => {
 let allocate_bytes = (wasm_mod, env, bytes) => {
   let buf = Buffer.create(80);
   Buffer.add_bytes(buf, bytes);
-  allocate_bytes_from_buffer(wasm_mod, env, buf, BytesType, "allocate_bytes");
+  allocate_byte_like_from_buffer(wasm_mod, env, buf, BytesType, "allocate_bytes");
 };
 
 let allocate_char = (wasm_mod, env, char) => {
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -1587,7 +1587,7 @@ let call_lambda = (~tail=false, wasm_mod, env, func, (argsty, retty), args) => {
   );
 };
 
-let allocate_buffer = (wasm_mod, env, buf, len, tag, label) => {
+let allocate_bytes_from_buffer = (wasm_mod, env, buf, len, tag, label) => {
   let ints_to_push: list(int64) = buf_to_ints(buf);
   let get_swap = () => get_swap(wasm_mod, env, 0);
   let tee_swap = tee_swap(~skip_incref=true, wasm_mod, env, 0);
@@ -1632,7 +1632,7 @@ let allocate_buffer = (wasm_mod, env, buf, len, tag, label) => {
 let allocate_string = (wasm_mod, env, str) => {
   let buf = Buffer.create(80);
   Buffer.add_string(buf, str);
-  allocate_buffer(
+  allocate_bytes_from_buffer(
     wasm_mod,
     env,
     buf,
@@ -1645,7 +1645,7 @@ let allocate_string = (wasm_mod, env, str) => {
 let allocate_bytes = (wasm_mod, env, bytes) => {
   let buf = Buffer.create(80);
   Buffer.add_bytes(buf, bytes);
-  allocate_buffer(
+  allocate_bytes_from_buffer(
     wasm_mod,
     env,
     buf,
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -1587,10 +1587,7 @@ let call_lambda = (~tail=false, wasm_mod, env, func, (argsty, retty), args) => {
   );
 };
 
-let allocate_string = (wasm_mod, env, str) => {
-  let buf = Buffer.create(80);
-  Buffer.add_string(buf, str);
-
+let allocate_buffer = (wasm_mod, env, buf, len, tag, label) => {
   let ints_to_push: list(int64) = buf_to_ints(buf);
   let get_swap = () => get_swap(wasm_mod, env, 0);
   let tee_swap = tee_swap(~skip_incref=true, wasm_mod, env, 0);
@@ -1603,14 +1600,14 @@ let allocate_string = (wasm_mod, env, str) => {
       ),
       Expression.Const.make(
         wasm_mod,
-        const_int32(tag_val_of_heap_tag_type(StringType)),
+        const_int32(tag_val_of_heap_tag_type(tag)),
       ),
     ),
     store(
       ~offset=4,
       wasm_mod,
       get_swap(),
-      Expression.Const.make(wasm_mod, const_int32 @@ String.length(str)),
+      Expression.Const.make(wasm_mod, const_int32 @@ len),
     ),
   ];
   let elts =
@@ -1627,53 +1624,34 @@ let allocate_string = (wasm_mod, env, str) => {
     );
   Expression.Block.make(
     wasm_mod,
-    gensym_label("allocate_string"),
+    gensym_label(label),
     List.concat([preamble, elts, [get_swap()]]),
+  );
+};
+
+let allocate_string = (wasm_mod, env, str) => {
+  let buf = Buffer.create(80);
+  Buffer.add_string(buf, str);
+  allocate_buffer(
+    wasm_mod,
+    env,
+    buf,
+    String.length(str),
+    StringType,
+    "allocate_string",
   );
 };
 
 let allocate_bytes = (wasm_mod, env, bytes) => {
   let buf = Buffer.create(80);
   Buffer.add_bytes(buf, bytes);
-
-  let ints_to_push: list(int64) = buf_to_ints(buf);
-  let get_swap = () => get_swap(wasm_mod, env, 0);
-  let tee_swap = tee_swap(~skip_incref=true, wasm_mod, env, 0);
-  let preamble = [
-    store(
-      ~offset=0,
-      wasm_mod,
-      tee_swap(
-        heap_allocate(wasm_mod, env, 2 + 2 * List.length(ints_to_push)),
-      ),
-      Expression.Const.make(
-        wasm_mod,
-        const_int32(tag_val_of_heap_tag_type(BytesType)),
-      ),
-    ),
-    store(
-      ~offset=4,
-      wasm_mod,
-      get_swap(),
-      Expression.Const.make(wasm_mod, const_int32 @@ Bytes.length(bytes)),
-    ),
-  ];
-  let elts =
-    List.mapi(
-      (idx, i: int64) =>
-        store(
-          ~ty=Type.int64,
-          ~offset=8 * (idx + 1),
-          wasm_mod,
-          get_swap(),
-          Expression.Const.make(wasm_mod, wrap_int64(i)),
-        ),
-      ints_to_push,
-    );
-  Expression.Block.make(
+  allocate_buffer(
     wasm_mod,
-    gensym_label("allocate_bytes"),
-    List.concat([preamble, elts, [get_swap()]]),
+    env,
+    buf,
+    Bytes.length(bytes),
+    BytesType,
+    "allocate_bytes",
   );
 };
 
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -1644,13 +1644,7 @@ let allocate_string = (wasm_mod, env, str) => {
 let allocate_bytes = (wasm_mod, env, bytes) => {
   let buf = Buffer.create(80);
   Buffer.add_bytes(buf, bytes);
-  allocate_bytes_from_buffer(
-    wasm_mod,
-    env,
-    buf,
-    BytesType,
-    "allocate_bytes",
-  );
+  allocate_bytes_from_buffer(wasm_mod, env, buf, BytesType, "allocate_bytes");
 };
 
 let allocate_char = (wasm_mod, env, char) => {
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -1632,6 +1632,51 @@ let allocate_string = (wasm_mod, env, str) => {
   );
 };
 
+let allocate_bytes = (wasm_mod, env, bytes) => {
+  let buf = Buffer.create(80);
+  Buffer.add_bytes(buf, bytes);
+
+  let ints_to_push: list(int64) = buf_to_ints(buf);
+  let get_swap = () => get_swap(wasm_mod, env, 0);
+  let tee_swap = tee_swap(~skip_incref=true, wasm_mod, env, 0);
+  let preamble = [
+    store(
+      ~offset=0,
+      wasm_mod,
+      tee_swap(
+        heap_allocate(wasm_mod, env, 2 + 2 * List.length(ints_to_push)),
+      ),
+      Expression.const(
+        wasm_mod,
+        const_int32(tag_val_of_heap_tag_type(BytesType)),
+      ),
+    ),
+    store(
+      ~offset=4,
+      wasm_mod,
+      get_swap(),
+      Expression.const(wasm_mod, const_int32 @@ Bytes.length(bytes)),
+    ),
+  ];
+  let elts =
+    List.mapi(
+      (idx, i: int64) =>
+        store(
+          ~ty=Type.int64,
+          ~offset=8 * (idx + 1),
+          wasm_mod,
+          get_swap(),
+          Expression.const(wasm_mod, wrap_int64(i)),
+        ),
+      ints_to_push,
+    );
+  Expression.block(
+    wasm_mod,
+    gensym_label("allocate_bytes"),
+    List.concat([preamble, elts, [get_swap()]]),
+  );
+};
+
 let allocate_char = (wasm_mod, env, char) => {
   // Copy bytes into a fresh buffer so we can guarantee a copy of a full word
   let bytes = Bytes.make(4, Char.chr(0));
@@ -2362,6 +2407,7 @@ let compile_allocation = (wasm_mod, env, alloc_type) =>
   | MBox(elt) => allocate_box(wasm_mod, env, elt)
   | MArray(elts) => allocate_array(wasm_mod, env, elts)
   | MRecord(ttag, elts) => allocate_record(wasm_mod, env, ttag, elts)
+  | MBytes(bytes) => allocate_bytes(wasm_mod, env, bytes)
   | MString(str) => allocate_string(wasm_mod, env, str)
   | MChar(char) => allocate_char(wasm_mod, env, char)
   | MADT(ttag, vtag, elts) => allocate_adt(wasm_mod, env, ttag, vtag, elts)
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -1587,7 +1587,7 @@ let call_lambda = (~tail=false, wasm_mod, env, func, (argsty, retty), args) => {
   );
 };
 
-let allocate_bytes_from_buffer = (wasm_mod, env, buf, len, tag, label) => {
+let allocate_bytes_from_buffer = (wasm_mod, env, buf, tag, label) => {
   let ints_to_push: list(int64) = buf_to_ints(buf);
   let get_swap = () => get_swap(wasm_mod, env, 0);
   let tee_swap = tee_swap(~skip_incref=true, wasm_mod, env, 0);
@@ -1607,7 +1607,7 @@ let allocate_bytes_from_buffer = (wasm_mod, env, buf, len, tag, label) => {
       ~offset=4,
       wasm_mod,
       get_swap(),
-      Expression.Const.make(wasm_mod, const_int32 @@ len),
+      Expression.Const.make(wasm_mod, const_int32 @@ Buffer.length(buf)),
     ),
   ];
   let elts =
@@ -1636,7 +1636,6 @@ let allocate_string = (wasm_mod, env, str) => {
     wasm_mod,
     env,
     buf,
-    String.length(str),
     StringType,
     "allocate_string",
   );
@@ -1649,7 +1648,6 @@ let allocate_bytes = (wasm_mod, env, bytes) => {
     wasm_mod,
     env,
     buf,
-    Bytes.length(bytes),
     BytesType,
     "allocate_bytes",
   );
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -1644,7 +1644,13 @@ let allocate_string = (wasm_mod, env, str) => {
 let allocate_bytes = (wasm_mod, env, bytes) => {
   let buf = Buffer.create(80);
   Buffer.add_bytes(buf, bytes);
-  allocate_byte_like_from_buffer(wasm_mod, env, buf, BytesType, "allocate_bytes");
+  allocate_byte_like_from_buffer(
+    wasm_mod,
+    env,
+    buf,
+    BytesType,
+    "allocate_bytes",
+  );
 };
 
 let allocate_char = (wasm_mod, env, char) => {
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -1646,7 +1646,7 @@ let allocate_bytes = (wasm_mod, env, bytes) => {
       tee_swap(
         heap_allocate(wasm_mod, env, 2 + 2 * List.length(ints_to_push)),
       ),
-      Expression.const(
+      Expression.Const.make(
         wasm_mod,
         const_int32(tag_val_of_heap_tag_type(BytesType)),
       ),
@@ -1655,7 +1655,7 @@ let allocate_bytes = (wasm_mod, env, bytes) => {
       ~offset=4,
       wasm_mod,
       get_swap(),
-      Expression.const(wasm_mod, const_int32 @@ Bytes.length(bytes)),
+      Expression.Const.make(wasm_mod, const_int32 @@ Bytes.length(bytes)),
     ),
   ];
   let elts =
@@ -1666,11 +1666,11 @@ let allocate_bytes = (wasm_mod, env, bytes) => {
           ~offset=8 * (idx + 1),
           wasm_mod,
           get_swap(),
-          Expression.const(wasm_mod, wrap_int64(i)),
+          Expression.Const.make(wasm_mod, wrap_int64(i)),
         ),
       ints_to_push,
     );
-  Expression.block(
+  Expression.Block.make(
     wasm_mod,
     gensym_label("allocate_bytes"),
     List.concat([preamble, elts, [get_swap()]]),
--- a/compiler/test/stdlib/bytes.test.gr
+++ b/compiler/test/stdlib/bytes.test.gr
@@ -3,179 +3,179 @@ import Int32 from "int32";
 
 
 // Bytes.empty
-assert Bytes.empty == Bytes.empty;
+assert Bytes.empty == Bytes.empty
 
 
 // Bytes.length
-assert Bytes.length(Bytes.empty) == 0;
+assert Bytes.length(Bytes.empty) == 0
 
 
 // Bytes.make
 let bytes = Bytes.make(64)
-assert Bytes.length(bytes) == 64;
+assert Bytes.length(bytes) == 64
 
 
 // Bytes.setInt8, Bytes.getInt8S, Bytes.getInt8U
 let bytes = Bytes.make(1);
-Bytes.setInt8(bytes, 0, 0x000000ffl);
-assert Bytes.getInt8S(bytes, 0) == 0xffffffffl;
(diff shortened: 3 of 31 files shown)


compiler/test/stdlib/bytes.test.gr
```diff
@@ -0,0 +1,182 @@
+import Bytes from "bytes";
+import { coerceInt32ToNumber, coerceNumberToInt32 } from "runtime/numbers"
+import { tagSimpleNumber } from "runtime/dataStructures"
```
ospencer, line 3: Ideally, we don't import anything from the runtime in tests (unless they're tests for the runtime) since it's not how we want people to use the language. I'll drop some suggestions on how we can replace these.

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+/**
+ * Gets an unsigned 8-bit integer starting at the given byte index.
+ * @param b: Bytes - The Bytes buffer
+ * @param i: Number - The byte index
+ * @returns Int32
+ */
+@disableGC
+export let getInt8U = (b: Bytes, i: Number) => {
+  let byteSize = 1n;
+  let (ptr, offset) = getPointerAndOffset(b, byteSize, i);
```
ospencer, line 148: 💧

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+ * Sets a signed 16-bit integer starting at the given byte index.
+ * @param b: Bytes - The Bytes buffer
+ * @param i: Number - The byte index
+ * @param v: Int32 - The value to set
+ * @returns Void
+ */
+@disableGC
+export let setInt16 = (b: Bytes, i: Number, v: Int32) => {
+  let byteSize = 2n;
+  let (ptr, offset) = getPointerAndOffset(b, byteSize, i);
```
ospencer, line 163: 💧

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+/**
+ * Gets a signed 16-bit integer starting at the given byte index.
+ * @param b: Bytes - The Bytes buffer
+ * @param i: Number - The byte index
+ * @returns Int32
+ */
+@disableGC
+export let getInt16S = (b: Bytes, i: Number) => {
+  let byteSize = 2n;
+  let (ptr, offset) = getPointerAndOffset(b, byteSize, i);
```
ospencer, line 177: 💧

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+/**
+ * Gets an unsigned 16-bit integer starting at the given byte index.
+ * @param b: Bytes - The Bytes buffer
+ * @param i: Number - The byte index
+ * @returns Int32
+ */
+@disableGC
+export let getInt16U = (b: Bytes, i: Number) => {
+  let byteSize = 2n;
+  let (ptr, offset) = getPointerAndOffset(b, byteSize, i);
```
ospencer, line 191: 💧

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+ * Sets a signed 32-bit integer starting at the given byte index.
+ * @param b: Bytes - The Bytes buffer
+ * @param i: Number - The byte index
+ * @param v: Int32 - The value to set
+ * @returns Void
+ */
+@disableGC
+export let setInt32 = (b: Bytes, i: Number, v: Int32) => {
+  let byteSize = 4n;
+  let (bytesPtr, offset) = getPointerAndOffset(b, byteSize, i);
```
ospencer, line 207: 💧

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+/**
+ * Gets a signed 32-bit integer starting at the given byte index.
+ * @param b: Bytes - The Bytes buffer
+ * @param i: Number - The byte index
+ * @returns Int32
+ */
+@disableGC
+export let getInt32 = (b: Bytes, i: Number) => {
+  let byteSize = 4n;
+  let (ptr, offset) = getPointerAndOffset(b, byteSize, i);
```
ospencer, line 221: 💧

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+ * Sets a 32-bit float starting at the given byte index.
+ * @param b: Bytes - The Bytes buffer
+ * @param i: Number - The byte index
+ * @param v: Float32 - The value to set
+ * @returns Void
+ */
+@disableGC
+export let setFloat32 = (b: Bytes, i: Number, v: Float32) => {
+  let byteSize = 4n;
+  let (ptr, offset) = getPointerAndOffset(b, byteSize, i);
```
ospencer, line 236: 💧

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+/**
+ * Gets a 32-bit float starting at the given byte index.
+ * @param b: Bytes - The Bytes buffer
+ * @param i: Number - The byte index
+ * @returns Float32
+ */
+@disableGC
+export let getFloat32 = (b: Bytes, i: Number) => {
+  let byteSize = 4n;
+  let (ptr, offset) = getPointerAndOffset(b, byteSize, i);
```
ospencer, line 250: 💧

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+ * Sets a signed 64-bit integer starting at the given byte index.
+ * @param b: Bytes - The Bytes buffer
+ * @param i: Number - The byte index
+ * @param v: Int64 - The value to set
+ * @returns Void
+ */
+@disableGC
+export let setInt64 = (b: Bytes, i: Number, v: Int64) => {
+  let byteSize = 8n;
+  let (ptr, offset) = getPointerAndOffset(b, byteSize, i);
```
ospencer, line 265: 💧

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+/**
+ * Gets a signed 64-bit integer starting at the given byte index.
+ * @param b: Bytes - The Bytes buffer
+ * @param i: Number - The byte index
+ * @returns Int64
+ */
+@disableGC
+export let getInt64 = (b: Bytes, i: Number) => {
+  let byteSize = 8n;
+  let (ptr, offset) = getPointerAndOffset(b, byteSize, i);
```
ospencer, line 279: 💧

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+/**
+ * Gets a 64-bit float starting at the given byte index.
+ * @param b: Bytes - The Bytes buffer
+ * @param i: Number - The byte index
+ * @returns Float64
+ */
+@disableGC
+export let getFloat64 = (b: Bytes, i: Number) => {
+  let byteSize = 8n;
+  let (ptr, offset) = getPointerAndOffset(b, byteSize, i);
```
ospencer, line 308: 💧

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+ * @returns Bytes
+ */
+@disableGC
+export let sub = (b: Bytes, i: Number, len: Number) => {
+  let (>) = WasmI32.gtS;
+  let (%) = WasmI32.remS;
+  let (==) = WasmI32.eq;
+  let (<) = WasmI32.ltS;
+  let (+) = WasmI32.add;
+  let (src, size) = getPointerAndSize(b);
```
ospencer, line 345: 💧

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+  if ((i + len) > size) {
+    throw Exception.InvalidArgument("The given index and length do not specify a valid range of bytes");
+  }
+  let dst = allocateBytes(len);
+  let offset = i;
+  if (i % 4n == 0n) {
+    // If possible, copy 4 bytes at a time
+    for (let mut i = 0n; i < len; i += 4n) {
+      let n = WasmI32.load(src, 8n + i + offset);
+      WasmI32.store(dst, n, 8n + i);
```
ospencer, line 357: I got this far down and didn't catch it until now, but it's a tad more efficient to write loads/stores with a static offset:
```suggestion
      WasmI32.store(dst + i, n, 8n);
```

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+    (0n, left)
+  };
+  let end = if (right < 0n) {
+    size + right
+  } else {
+    size
+  };
+  // Copy relevant range of bytes
+  if (start < size && end > start) {
+    if (start % 4n == 0n && offset % 4n == 0n) {
```
ospencer, line 405: Looks like the wrong `==` here?

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+  let dstPos = numberToWasmI32(dstPos);
+  let len = numberToWasmI32(len);
+  if ((srcPos + len) > srcSize) {
+    throw Exception.InvalidArgument("Invalid source bytes range");
+  }
+  if ((dstPos + len) > dstSize) {
+    throw Exception.InvalidArgument("Invalid destination bytes range");
+  }
+  let end = srcPos + len;
+  if (srcPos % 4n == 0n && dstPos % 4n == 0n) {
```
ospencer, line 450: Wrong `==` here?

stdlib/bytes.gr
```diff
@@ -0,0 +1,505 @@
...
+ * @returns Void
+ */
+@disableGC
+export let blit = (src: Bytes, srcPos: Number, dst: Bytes, dstPos: Number, len: Number) => {
+  let (>) = WasmI32.gtS;
+  let (<) = WasmI32.ltS;
+  let (+) = WasmI32.add;
+  let (%) = WasmI32.remS;
+  let (src, srcSize) = getPointerAndSize(src);
+  let (dst, dstSize) = getPointerAndSize(dst);
```
ospencer, line 439: 💧

ospencer (changes requested):
Looking really good! Left a few notes and a couple of potential memory leaks to resolve, but this is looking great!

Also, it seems like about half of the compiler changes are to support bytes literals. I think it's fine to keep it in for now, but maybe important for us to know that those code paths aren't worked.

compiler/test/stdlib/bytes.test.gr
```diff
@@ -0,0 +1,182 @@
...
+Bytes.setFloat64(bytes, 0, 42.0d);
+assert Bytes.getFloat64(bytes, 0) == 42.0d;
+
+// Bytes.copy
+
+let bytes0 = Bytes.make(8);
+assert Bytes.length(bytes0) == 8;
+for (let mut i = 0; i < 8; i+= 1) {
+  Bytes.setInt8(bytes0, i, coerceNumberToInt32(i));
+}
```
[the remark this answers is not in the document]
jozanza: Looks like that one was already exported 😅 ! But I just made a PR for the other numeric wasm conversion utils in the bytes.gr module: <https://github.com/grain-lang/grain/pull/656>

cli/bin/grain.js
```diff
@@ -10,7 +10,6 @@ const v8 = require("v8");
  *
  * This seems to work for our needs with Node 14, but we should be cautious when updating.
  */
-v8.setFlagsFromString("--experimental-wasm-bigint");
```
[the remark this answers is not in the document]
jozanza: Cool I'll add that back

stdlib/bytes.gr
```diff
@@ -0,0 +1,549 @@
...
+@disableGC
+let wasmI64ToInt64 = (n: WasmI64) => {
+  WasmI32.toGrain(newInt64(n)) : Int64
+}
+
+@disableGC
+let wasmF64ToFloat64 = (n: WasmF64) => {
+  WasmI32.toGrain(newFloat64(n)) : Float64
+}
+
```
jozanza, line 69: @ospencer I can pretty much remove all these conversion utils if we can land https://github.com/grain-lang/grain/pull/656 or find some other solution 🎉

stdlib/bytes.gr
```diff
@@ -0,0 +1,441 @@
...
+  Memory.free(WasmI32.fromGrain(pointerAndSize))
+  let left = numberToWasmI32(left)
+  let right = numberToWasmI32(right)
+  let newLen = size + left + right
+  if (newLen < 0n) {
+    throw Exception.InvalidArgument("The resulting length is less than 0")
+  }
+  let dst = allocateBytes(newLen)
+  let (start, offset) = if (left < 0n) {
+    (left * -1n, 0n)
```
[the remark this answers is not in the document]
jozanza: Not my poor tuples...alright I just went with some `mut`s here https://github.com/grain-lang/grain/pull/633/commits/c96688e38141e0829ab9ca68cd15edaf92ea3279

compiler/src/codegen/compcore.re
```diff
@@ -1633,10 +1633,7 @@ let call_lambda = (~tail=false, wasm_mod, env, func, (argsty, retty), args) => {
   );
 };
 
-let allocate_string = (wasm_mod, env, str) => {
-  let buf = Buffer.create(80);
-  Buffer.add_string(buf, str);
-
+let allocate_buffer = (wasm_mod, env, buf, len, tag, label) => {
```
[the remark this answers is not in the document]
jozanza: Good call!

compiler/src/codegen/compcore.re
```diff
@@ -1649,14 +1646,14 @@ let allocate_string = (wasm_mod, env, str) => {
...
-        const_int32(tag_val_of_heap_tag_type(StringType)),
+        const_int32(tag_val_of_heap_tag_type(tag)),
       ),
     ),
     store(
       ~offset=4,
       wasm_mod,
       get_swap(),
-      Expression.Const.make(wasm_mod, const_int32 @@ String.length(str)),
+      Expression.Const.make(wasm_mod, const_int32 @@ len),
```
ospencer, line 1656: Any reason to do this over `Buffer.length`?

jozanza: I wasn't sure if `Bytes.length` and `String.length` would yield the same value. I will change that!

stdlib/runtime/unsafe/tags.gr
```diff
@@ -16,6 +16,8 @@ export let _GRAIN_BOXED_NUM_HEAP_TAG = 6n
 export let _GRAIN_LAMBDA_HEAP_TAG = 7n
 export let _GRAIN_TUPLE_HEAP_TAG = 8n
 
+export let _GRAIN_BYTES_HEAP_TAG = 9n
```
ospencer, line 19: I think you've got an extra newline above this here.

ospencer (approved):
At last, it is perfect 👌

marcusroberts:
This PR is a really interesting read for understanding how to implement a language feature!

compiler/src/codegen/compcore.re
```diff
@@ -1633,10 +1633,7 @@ let call_lambda = (~tail=false, wasm_mod, env, func, (argsty, retty), args) => {
   );
 };
 
-let allocate_string = (wasm_mod, env, str) => {
-  let buf = Buffer.create(80);
-  Buffer.add_string(buf, str);
-
+let allocate_bytes_from_buffer = (wasm_mod, env, buf, tag, label) => {
```
peblair, line 1636:
```suggestion
let allocate_byte_like_from_buffer = (wasm_mod, env, buf, tag, label) => {
```

peblair: I think this name might be a little more clear, since this is being used by both byte and string allocations

compiler/src/codegen/compcore.re
```diff
@@ -1673,11 +1670,29 @@ let allocate_string = (wasm_mod, env, str) => {
...
-    gensym_label("allocate_string"),
+    gensym_label(label),
     List.concat([preamble, elts, [get_swap()]]),
   );
 };
 
+let allocate_string = (wasm_mod, env, str) => {
+  let buf = Buffer.create(80);
+  Buffer.add_string(buf, str);
+  allocate_bytes_from_buffer(
```
peblair, line 1681:
```suggestion
  allocate_byte_like_from_buffer(
```

compiler/src/codegen/compcore.re
```diff
@@ -1673,11 +1670,29 @@ let allocate_string = (wasm_mod, env, str) => {
...
+    buf,
+    StringType,
+    "allocate_string",
+  );
+};
+
+let allocate_bytes = (wasm_mod, env, bytes) => {
+  let buf = Buffer.create(80);
+  Buffer.add_bytes(buf, bytes);
+  allocate_bytes_from_buffer(wasm_mod, env, buf, BytesType, "allocate_bytes");
```
peblair, line 1693:
```suggestion
  allocate_byte_like_from_buffer(wasm_mod, env, buf, BytesType, "allocate_bytes");
```

peblair (changes requested):
Thanks for adding this! This looks really great. I don't have much to critique, aside from one function name in `compcore.re`; with that change, I think this is good to merge!

jozanza:
> This PR is a really interesting read for understanding how to implement a language feature!
> 
> 
> 
> 

Yeah! I think I should write up some docs or a blog post on it so others can learn a bit. I know I learned a TON :)

compiler/src/codegen/compcore.re
jozanza: That's a good call!

peblair:
@ospencer Is the test failure here cause for alarm, or am I good to approve?

jozanza:
I forgot to mention, it'd be great if someone could look at the doc blocks in `bytes.gr`. There are a few different conventions for those comments in the codebase, and I'm not sure if chose the desired convention or not 👀

ospencer (changes requested):
@peblair You can approve (I know it's starting to get late in the day for you) and we'll get the test sorted out (I imagine it's something really small since the tests were just passing).

peblair (approved):

ospencer:
@jozanza I think the docs look great!

compiler/test/stdlib/bytes.test.gr
```diff
@@ -0,0 +1,191 @@
...
+Bytes.setFloat64(0, 42.0d, bytes)
+assert Bytes.getFloat64(0, bytes) == 42.0d
+
+
+// Bytes.copy
+let bytes0 = Bytes.make(8)
+assert Bytes.length(bytes0) == 8
+for (let mut i = 0; i < 8; i+= 1) {
+  Bytes.setInt8(i, Int32.fromNumber(i), bytes0)
+}
```
phated, line 76: I really want a `Bytes.init` for this pattern, maybe the syntax would be:
```
Bytes.init(8, (bytes0, i) => {
  Bytes.setInt8(i, Int32.fromNumber(i), bytes0)
})
```
@ospencer what do you think?

compiler/test/stdlib/bytes.test.gr
```diff
@@ -0,0 +1,191 @@
...
+let bytes1 = Bytes.copy(bytes0);
+assert Bytes.length(bytes1) == Bytes.length(bytes0)
+for (let mut i = 0; i < 8; i += 1) {
+  assert Bytes.getInt8S(i, bytes1) == Bytes.getInt8S(i, bytes0)
+}
```
phated, line 81: Should we have a `Bytes.forEach`?

compiler/test/stdlib/bytes.test.gr
```diff
@@ -0,0 +1,191 @@
...
+
+
+// Bytes.sub
+let bytes2 = Bytes.sub(0, 8, bytes1)
```
phated, line 85: What is `sub` short for?

compiler/test/stdlib/bytes.test.gr
```diff
@@ -0,0 +1,191 @@
...
+assert Bytes.length(bytes5) == 0
+
+
+// Bytes.move
+let srcBytes = Bytes.copy(bytes0)
+let dstBytes = Bytes.make(16)
+Bytes.move(0, 8, 8, srcBytes, dstBytes)
+for (let mut i = 8; i < 16; i += 1) {
+  if (i < 8) {
+    assert Bytes.getInt8S(i, dstBytes) == 0l
```
phated, line 143: What happens to the the bytes in `srcBytes` when you `move` them? Does it shorten that bytes?

compiler/test/stdlib/bytes.test.gr
```diff
@@ -0,0 +1,191 @@
...
+for (let mut i = 0; i < 16; i += 1) {
+  Bytes.setInt8(i, 1l, bytesA)
+}
+let bytesB = Bytes.make(16);
+for (let mut i = 0; i < 16; i += 1) {
+  Bytes.setInt8(i, 2l, bytesB)
+}
+let bytesC = Bytes.append(bytesA, bytesB)
+for (let mut i = 0; i < Bytes.length(bytesC); i += 1) {
+    if (i < 16) {
```
phated, line 161: weird indent
```suggestion
  if (i < 16) {
```

compiler/test/stdlib/bytes.test.gr
```diff
@@ -0,0 +1,191 @@
...
+    assert Bytes.getInt8S(i, bytesC) == 1l
+  } else {
+    assert Bytes.getInt8S(i, bytesC) == 2l
+  }
+}
+let bytesD = Bytes.append(bytesB, bytesA)
+for (let mut i = 0; i < Bytes.length(bytesD); i += 1) {
+    if (i < 16) {
```
phated, line 169: Weird indent
```suggestion
  if (i < 16) {
```

stdlib/bytes.gr
```diff
@@ -0,0 +1,466 @@
...
+    res
+  } else {
+    // fast path
+    WasmI32.shrS(m, 1n)
+  }
+}
+
+/** Throws an exception if the index specified is out-of-bounds */
+@disableGC
+let ensureIndexIsInBounds = (i, byteSize, max) => {
```
phated, line 37: Do we have a preference over `ensure` and `assert`?

stdlib/bytes.gr
```diff
@@ -0,0 +1,466 @@
...
+
+/**
+ * Returns a new Bytes buffer containing a subset of the original Bytes buffer.
+ * @param i: Number - The start position to copy from
+ * @param len: Number - The number of bytes to copy
+ * @param b: Bytes - The Bytes buffer get a subset of bytes from
+ * @returns Bytes
+ */
+@disableGC
+export let sub = (i: Number, len: Number, b: Bytes) => {
```
phated, line 319: Rename to `subset`?

stdlib/bytes.gr
```diff
@@ -0,0 +1,466 @@
...
+/**
+ * Returns a new Bytes buffer with uninitialized bytes prepended/appended to it.
+ * If the `left` or `right` parameters are negative, bytes will be removed from the corresponding side instead of appended.
+ * @param left: Number - The number of uninitialized bytes to prepend
+ * @param right: Number - The number of uninitialized bytes to append
+ * @param b: Bytes - The Bytes buffer get a subset of bytes from
+ * @returns Bytes
+ */
+@disableGC
+export let extend = (left: Number, right: Number, b: Bytes) => {
```
phated, line 345: Is this the replacement name for `blit`? I only think of "extend" as "to grow", so the removal seems really weird. I think I'd want to call this `splice`

stdlib/bytes.gr
phated: Maybe we should call this `slice`?

stdlib/bytes.gr
```diff
@@ -0,0 +1,466 @@
...
+  tagSimpleNumber(getSize(b))
+}
+
+/**
+ * Creates a new Bytes buffer that contains the bytes of both buffers a and b.
+ * @param a: Bytes - The buffer to be copied first
+ * @param b: Bytes - The buffer to be copied last
+ * @returns Bytes
+ */
+export let append = (a: Bytes, b: Bytes) => {
```
phated, line 428: No `prepend`?

phated (changes requested):
This PR is dreamy 😍 I had a few comments/questions

compiler/test/stdlib/bytes.test.gr
ospencer: We didn't go for a `Bytes.init` because we decided against adding a `Byte` type to Grain, at least for now. We could do it with `Int32`s that represent bytes, but that doesn't feel quite right.

compiler/test/stdlib/bytes.test.gr
ospencer: Sorta the same deal as `Bytes.init`. I could totally see a future API that had `forEach8S`, `forEach8U`, `forEach32`, etc.

compiler/test/stdlib/bytes.test.gr
ospencer: Basically substring, but subbytes sounds weird. I like your `slice` idea though.

compiler/test/stdlib/bytes.test.gr
ospencer: They stay there. We thought for a while about a name for this one and landed on `move`, which is a nod to C's [`memmove`](https://man7.org/linux/man-pages/man3/memmove.3.html). I do agree that `move` sorta implies that the bytes go away. We think it's far better than OCaml's `blit` 😂 

But obviously open to other suggestions.

stdlib/bytes.gr
ospencer: I don't quite like `assert`, but I could go for something like `checkIndexIsInBounds`.

compiler/test/stdlib/bytes.test.gr
phated: Hm, so it's more of a `copySlice` or `copySubset`? 🤔

stdlib/bytes.gr
ospencer: Nah, it's the same as OCaml's `extend`. I like the name `extend` because it's easy to pick out if reading API docs to go for a grow-y kind of action. I think this name is better for how it'll usually be used. That said, I think this could potentially be better as two separate functions, one that only extends and one that only shortens. (I think I'd want to avoid calling them `grow` and `shrink` since that could imply that it's in-place.)

stdlib/bytes.gr
ospencer: That's just the same but with the arguments reversed, no?

compiler/test/stdlib/bytes.test.gr
phated: 🤷

compiler/test/stdlib/bytes.test.gr
phated: 🤷

compiler/test/stdlib/bytes.test.gr
phated: Sweet, let's do `slice`

compiler/test/stdlib/bytes.test.gr
phated: Rust seems to call this `copy_to_slice` but they have a slice "type".  Nodejs has optional arguments for their copy method. Additionally, I think this "move" could be replicated with `Bytes.slice` and `Bytes.copy` so I really like `copySlice`

stdlib/bytes.gr
phated: I'd prefer `check` I think.

stdlib/bytes.gr
phated: Hooray to slice!

stdlib/bytes.gr
phated: I'd be happy with `extend` and `shorten`!

Btw, I noticed the docs say "uninitialized" and I remember that being a big security issue for nodejs because newly created buffers could have old data inside them. They started zero-ing out buffers when created - is that something we need to do?

phated: I also think we need to add `grow` and `shrink` because most of the usage of Bytes is mutative based on what I saw in the tests.

stdlib/bytes.gr
phated: Sure, but then the name just seems weird. Shall we use `concat` instead?

stdlib/bytes.gr
ospencer: I think zeroing is also nice from a user's perspective, just has an obvious performance hit. Since there's a bulk memory instruction for it though, I imagine it's not too bad.

ospencer: It's not really possible to implement grow/shrink with Bytes in its current form, since you'd need to be able to reallocate its memory block to a larger size, which isn't possible with our memory manager. I think if you wanted something like Bytes that could grow, you should instead reach for some Buffer type (which we'll also have to implement). They're definitely similar in a lot of ways, but having Bytes with a static size makes it quite a bit more performant.

I guess the behavior is pretty different as well, since a lot of Buffer implementations will automatically grow instead of having a method for it.

compiler/test/stdlib/bytes.test.gr
jozanza: One cool thing about `move` is that it can be used to shuffle bytes around within the same `Byte` instance by passing it as both the src and the dest.

stdlib/bytes.gr
jozanza: Yeah I would personally love to skip `grow/shrink` and defer `Buffer` to a separate PR 👀

phated: Sounds good! Sorry for my ignorance about the difference 🙈 

Let's not lose my `extend` and `shorten` comment ❤️

jozanza: @phated what should the signatures on those functions be?

compiler/test/stdlib/bytes.test.gr
ospencer: I think that I still have a preference for `move`, since we already use the "copy" terminology in our libraries to say that we're making a new copy of something and that could add some confusion. Since this utility is largely to move around bytes within a `Bytes`, I think `move` works here!

ospencer (approved):
I think this is fully ready to go!

compiler/test/stdlib/bytes.test.gr
phated: I do think we should assert that those bytes still remain then, but I'm not going to block on it.

phated (approved):
Absolute 🔥 PR! Great work @jozanza (and thanks for putting up with my pedantry)

jozanza:
Is it weird that I'm still super pumped about this getting merged? 🤣

ospencer:
I'm still hype about it!

PR #636, opened 2021-04-29
fix: Workaround for Windows drive letters with jsoo & pkg

### LINKED ISSUE #621: Grain binary on Windows 10 gives a JavaScript error at start of compiling
I downloaded the Windows binary: v0.3.0

c:\Grain>grain-win-x64.exe -v
Grain cli 0.3.0
Grain compiler 0.3.0

Grain is in PATH, grain-win-x64 -h works fine, I installed vscode-grain in VSCode.

When trying to compile a hello.gr  containing only print("Hello!"):

grain-win-x64 hello.gr
grain-win-x64 compile hello.gr

I immediately get the error message:

**C:\snapshot\grain\cli\bin\grainc.js: internal error, uncaught exception:
                                     TypeError: Cannot read property 'device' of undefined**

My node was version 14.15.3, updating to 16 didn't change the message.

Most probably the error will be gone when I build from source, but then I cannot see anymore what is needed for a clean Grain installation under Windows 10.

phated:
I believe this fixes #621 

I did a ton of logging and it seems that js_of_ocaml only makes a `jsoo_mount_point` for the drive letter of the `process.cwd()`; however, our `pkg` tool's virtual snapshot adds `C:` to all of the paths included in it. This checks to see if the caml_root is not `C:` and adds it as a `jsoo_mount_point` if not (only on Windows).

This is definitely not the best place for this to go, but it is really just a temporary solution.

### DIFF
--- a/compiler/grainc/jsoo_hacks.js
+++ b/compiler/grainc/jsoo_hacks.js
@@ -1,6 +1,12 @@
 //Provides: caml_make_path
 //Requires: caml_current_dir
 //Requires: caml_jsstring_of_string
+//Requires: caml_root, jsoo_mount_point, MlNodeDevice
+if (caml_root.charCodeAt(1) == 58 && caml_root[0] != 'C') {
+  // This ensure we can find our C:/snapshot/ stuff even when running in a diff drive
+  // Not the ideal place to do this but it is the easiest to avoid free variables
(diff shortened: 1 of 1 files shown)


ospencer (approved):
👍

ng-marcus (approved):

ng-marcus:
Nice work finding this!

PR #635, opened 2021-04-29
feat: Rename Queue/Stack functions & deprecate old ones

jozanza:
Reopening this one now that #632 is merged. Closes #493

PR #638, opened 2021-04-29
chore(releaser): Remove release-as & allow features in patch

### LINKED ISSUE #637: Remove release-as after release
Couple other cleanup things to do for our config file while I'm in there.

phated:
I removed the release-as for 0.3.1 - closes #637 

Since we want to be more conscious with breaking changes and only want to batch them into the next minor, I think we should allow new non-breaking features in patches. This will allow us to merge changes like #635 that add new things while deprecating old for the next breaking release.

Additionally, I tried adding the `node-workspace` plugin now that it works with monorepos to keep our `@grain/cli` in sync with version bumps happening in stdlib & runtime.

And finally, I removed the package names for the node releasers because they aren't actually used.

H/T @joeldodge79 for all the tips.

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -1,13 +1,14 @@
 {
-  "release-as": "0.3.1",
   "bump-minor-pre-major": true,
+  "bump-patch-for-minor-pre-major": true,
+  "plugins": [
+    "node-workspace"
+  ],
   "packages": {
     ".": {
-      "package-name": "grain",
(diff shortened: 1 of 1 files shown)


phated:
All of this worked slick in my fork: https://github.com/phated/grain/pull/5

ospencer (approved):

peblair (approved):

PR #639, opened 2021-04-30
Fix: Passing an absolute path to the file to compiler on Windows systems fails

ng-marcus:
Passing an absolute path fails to compile:

The file path that arrives in caml_make_path  still has Windows separators.

Make a sanity clean up of any filename arriving here to make sure subsequent parsing works


phated (changes requested):
Oh, interesting. I actually think this an issue inside the compiler itself. We run [`Grain_utils.Files.normalize_separators`](https://github.com/grain-lang/grain/blob/main/compiler/src/utils/files.re#L14) on any paths we resolve internally, but probably not on the first path a user sends to the compiler. Can you look into that?

ng-marcus:
I'll take a look.  My fix felt a dirty hack :)

marcusroberts:
Created as issue #640

PR #641, opened 2021-05-01
Fix: Normalise Windows paths in the JS components otherwise JSOO fails

### LINKED ISSUE #640: Passing an absolute path to compile to the pkg binary on Windows fails
Specifying an absolute path to the file to compile fails

@phated says " I actually think this an issue inside the compiler itself. We run Grain_utils.Files.normalize_separators on any paths we resolve internally, but probably not on the first path a user sends to the compiler. Can you look into that?"

marcusroberts:
PKGed Windows binaries fail because the specified file to compile/run is still in Windows separator style when it is parsed by the JSOO compiled component.

This fixes #640 by normalising the filename from the CLI before it is parsed

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -149,7 +149,6 @@ program
   .description("compile a grain program into wasm")
   .action(function (file) {
     var normalised_file = file.replace(/\\/g,'/');
-
     actions.compile(normalised_file, program);
   });
 
@@ -158,7 +157,6 @@ program
   .description("check a grain file for LSP")
   .action(function (file) {
     var normalised_file = file.replace(/\\/g,'/');
-
     actions.lsp(normalised_file, program);
   });
 
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -140,28 +140,34 @@ program
   // The root command that compiles & runs
   .arguments("<file>")
   .action(function (file) {
-    actions.run(actions.compile(file, program), program.opts());
+    var normalised_file = file.replace(/\\/g,'/');
+    actions.run(actions.compile(normalised_file, program), program.opts());
   });
 
 program
   .command("compile <file>")
   .description("compile a grain program into wasm")
   .action(function (file) {
-    actions.compile(file, program);
+    var normalised_file = file.replace(/\\/g,'/');
+
+    actions.compile(normalised_file, program);
   });
 
 program
   .command("lsp <file>")
   .description("check a grain file for LSP")
   .action(function (file) {
-    actions.lsp(file, program);
+    var normalised_file = file.replace(/\\/g,'/');
+
+    actions.lsp(normalised_file, program);
   });
 
 program
   .command("run <file>")
   .description("run a wasm file with the grain runtime")
   .action(function (wasmFile) {
-    actions.run(wasmFile, program.opts());
+    var normalised_wasm_file = wasmFile.replace(/\\/g,'/');
+    actions.run(normalised_wasm_file, program.opts());
   });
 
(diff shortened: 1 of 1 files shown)


peblair:
@marcusroberts From the issue:
> @phated says " I actually think this an issue inside the compiler itself. We run Grain_utils.Files.normalize_separators on any paths we resolve internally, but probably not on the first path a user sends to the compiler. Can you look into that?"

Was that ultimately not the issue? I can understand how this change fixes the problem, but we should also (instead?) probably fix any lingering normalization issues in the compiler itself.

marcusroberts:
The compiler was fine, and if you call the compiled compiler there was no crash.

My first attempt at fixing was to use normalize_separators but it made no difference as the crash happened before that, even with those fixes in place.

The crash was in the JS compiled version only, and fixing it here has solved this problem (which was JSOO not finding the files).

I'm happy to add normalize_separators into the compiler too if you think that's helpful, but I saw no problems without it.

cli/bin/grain.js
```diff
@@ -140,28 +140,32 @@ program
   // The root command that compiles & runs
   .arguments("<file>")
   .action(function (file) {
-    actions.run(actions.compile(file, program), program.opts());
+    var normalised_file = file.replace(/\\/g, "/");
+    actions.run(actions.compile(normalised_file, program), program.opts());
```
ospencer, line 144: We should probably have a linter, but we can change this slightly to match the file's code style:
```suggestion
    const normalisedFile = file.replace(/\\/g, "/");
    actions.run(actions.compile(normalisedFile, program), program.opts());
```

cli/bin/grain.js
```diff
@@ -140,28 +140,32 @@ program
...
   });
 
 program
   .command("compile <file>")
   .description("compile a grain program into wasm")
   .action(function (file) {
-    actions.compile(file, program);
+    var normalised_file = file.replace(/\\/g, "/");
+    actions.compile(normalised_file, program);
```
ospencer, line 152:
```suggestion
    const normalisedFile = file.replace(/\\/g, "/");
    actions.compile(normalisedFile, program);
```

cli/bin/grain.js
```diff
@@ -140,28 +140,32 @@ program
...
   });
 
 program
   .command("lsp <file>")
   .description("check a grain file for LSP")
   .action(function (file) {
-    actions.lsp(file, program);
+    var normalised_file = file.replace(/\\/g, "/");
+    actions.lsp(normalised_file, program);
```
ospencer, line 160:
```suggestion
    const normalisedFile = file.replace(/\\/g, "/");
    actions.lsp(normalisedFile, program);
```

cli/bin/grain.js
```diff
@@ -140,28 +140,32 @@ program
...
   });
 
 program
   .command("run <file>")
   .description("run a wasm file with the grain runtime")
   .action(function (wasmFile) {
-    actions.run(wasmFile, program.opts());
+    var normalised_wasm_file = wasmFile.replace(/\\/g, "/");
+    actions.run(normalised_wasm_file, program.opts());
```
ospencer, line 168:
```suggestion
    const normalisedWasmFile = wasmFile.replace(/\\/g, "/");
    actions.run(normalisedWasmFile, program.opts());
```

phated (changes requested):
I'm going to do some research of my own before we merge this. I don't feel as though I have a good understanding of *where* the issue is and I don't like to add hacks when I don't.

marcusroberts:
@phated  found the root cause, so this is no longer needed.

PR #654, opened 2021-05-10
fix(grainc): Apply separator normalization before cmdliner util

### LINKED ISSUE #640: Passing an absolute path to compile to the pkg binary on Windows fails

phated:
This is an alternative fix for #640 

As I mentioned in #641, I wanted to understand the fundamental issue, which came from our usage of the `Arg.non_dir_file` utility of cmdliner. This utility tries to use `Sys.exists` to see if a file exists, but we hadn't normalized the paths for our system yet.

This fix applies our path normalization inside `grainc` instead of in the JS CLI.

Closes #640 
Closes #641

### DIFF
--- a/compiler/grainc/grainc.re
+++ b/compiler/grainc/grainc.re
@@ -154,12 +154,22 @@ let output_file_conv = {
   (parse, Format.pp_print_string);
 };
 
+let input_file_conv = {
+  open Arg;
+  let (prsr, prntr) = non_dir_file;
+
+  (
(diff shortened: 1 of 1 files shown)


ospencer (approved):
Good find!

marcusroberts:
Nice work!

marcusroberts (approved):

Issue #479, opened 2020-12-16
Testing: Use alternate string syntax


ospencer:
All of our tests in compiler/test/test_end_to_end.re use normal OCaml strings, but we often write Grain programs containing strings, so they need to be escaped. It would be better if we converted these tests to use the `{| foo bar |}` syntax instead of the `" foo bar "` syntax.

Ultra-Instinct-05:
@ospencer Any chance I can take this up ?

ospencer:
Absolutely!

ashutoshvarma:
@ospencer Is anyone working on this? I can take this up if no one is?

ospencer:
@ashutoshvarma No one is currently working on this, though I'm actually in the middle of porting all of the tests to a different framework 😅 

This will probably be ready to be worked on again sometime after this weekend!

ashutoshvarma:
No problem, will take a look at it next week.

Meanwhile, will try contribute to other "good first issues" 😀

ospencer:
If you're also looking for other things to work on, feel free to ping me on the Grain Discord!

ashutoshvarma:
Sure, I just joined the Grain Discord.

PR #655, opened 2021-05-10
chore: Port tests from OUnit2 to Rely

ospencer:
The major motivation here was to gain access to snapshot testing to 1) make the tests run faster, and 2) support our transition away from the JS runtime (which uses the deprecated `-p` option in all of the tests).

Many of the tests now just check snapshots, and a majority of the tests that are actually run are done fully linked. (The only tests that don't are GC tests, as currently, only the JS runtime is able to set an upper limit on the number of pages a webassembly module is allowed to use.)

This approach also has the added benefit that the full test suite runs about 35-40% faster on my machine, finishing in about 90 seconds.

`yarn compiler test` now has some handy flags:

- `yarn compiler test -u`: if you're certain that your changes to the compiler are correct, you can run this command to have Rely update the snapshots.
- `yarn compiler test --filter <test-filter>`: only run tests matching the filter. Very handy. Here's the result when filtering by `string`:
```
λ grain git:(rely) yarn compiler test --filter string
yarn run v1.22.4
$ yarn workspace @grain/compiler run test --filter string
$ esy test --filter string
Done: 1200/1202 (jobs: 1)
Running 2 test suites matching \string\i
 PASS  stdlib
 PASS  strings

Test Suites: 0 failed, 2 passed, 2 total
Tests:       0 failed, 24 passed, 24 total
Time:        4.995s
```

Those two flags can be combined together.

Next step will be to get the tests running with Wasmtime bindings, which will probably greatly decrease test time since we won't have to spawn `node` processes for each test 😎 

I split the test porting and initial snapshot generation into two separate commits; hopefully, that makes this slightly easier to review.


compiler/src/codegen/compcore.re
```diff
@@ -31,6 +31,7 @@ let gensym_label = s => {
   gensym_counter := gensym_counter^ + 1;
   Printf.sprintf("%s.%d", s, gensym_counter^);
 };
+let reset_labels = () => gensym_counter := 0;
```
peblair, line 34: What's this about?

compiler/src/codegen/emitmod.re
```diff
@@ -13,6 +13,11 @@ let emit_module = ({asm, signature}, outfile) => {
     close_out(oc);
   };
   if (Config.wat^) {
+    if (Grain_utils.Config.color_enabled^) {
+      Binaryen.Settings.set_colors_enabled(true);
+    } else {
+      Binaryen.Settings.set_colors_enabled(false);
+    };
```
peblair, line 20:
```suggestion
    Binaryen.Settings.set_colors_enabled(Grain_utils.Config.color_enabled^);
```

compiler/test/input/recursive-equal-box.gr
```diff
@@ -44,3 +44,5 @@ b.r := Some(b)
 
 assert !(a is b)
 assert a == b
+
+print(void)
```
peblair, line 48: Why have this test print anything?

compiler/test/input/recursive-equal-mut.gr
```diff
@@ -32,3 +32,5 @@ b.r = Some(b)
 
 assert !(a is b)
 assert a == b
+
+print(void)
```
peblair, line 36: See other comment; why print?

compiler/test/runner.re
```diff
@@ -114,17 +21,38 @@ let read_stream = cstream => {
...
+let wasmfile = name => "test/output/" ++ name ++ ".gr.wasm";
 
-  let cli_flags = if (print_output) {"-pg"} else {"-g"};
+let run = (~num_pages=?, file) => {
+  let cli_flags = "-g";
 
   let stdlib = Option.get(Grain_utils.Config.stdlib_dir^);
-  let testlibs = Sys.getcwd() ++ "/test-libs";
-  let result = ref("");
+  let testlibs = Sys.getcwd() ++ "test-libs";
```
peblair, line 55: Maybe there's a safer API we should use which is slash-aware?

compiler/test/suites/boxes.re
```diff
@@ -0,0 +1,64 @@
...
+
+describe("boxes", ({test}) => {
+  let assertSnapshot = makeSnapshotRunner(test);
+  let assertCompileError = makeCompileErrorRunner(test);
+  let assertRun = makeRunner(test);
+  let assertFileRun = makeFileRunner(test);
+
+  assertRun(
+    "box1",
+    "let b = box(4);\n            {\n              print(unbox(b))\n            }",
```
peblair, line 12: Can we fix the weird spacing on these?

compiler/test/suites/gc.re
```diff
@@ -0,0 +1,93 @@
...
+    "Maximum memory size exceeded",
+  );
+  assertRunGC("oomgc2", 64, "(1, (3, 4))");
+  assertRunGC("oomgc3", 32, "(3, 4)");
+
+  // gc tests
+  assertRunGC(
+    "gc1",
+    160,
+    "let f = (() => (1, 2));\n       {\n         f();\n         f();\n         f();\n         f()\n       }",
```
peblair, line 72: Spacing is weird here too

compiler/test/suites/loops.re
```diff
@@ -0,0 +1,92 @@
...
+open Runner;
+
+describe("loops", ({test}) => {
+  let assertSnapshot = makeSnapshotRunner(test);
+  let assertCompileError = makeCompileErrorRunner(test);
+  let assertRun = makeRunner(test);
+
+  assertRun(
+    "loop1",
+    "let b = box(3);\n            {\n              while (unbox(b) > 0) {\n                b := unbox(b) - 1;\n              };\n              print(unbox(b))\n            }",
```
peblair, line 11: Spacing is weird here too

compiler/test/suites/optimizations.re
```diff
@@ -0,0 +1,349 @@
...
+            raise(e);
+          };
+        expect.string(result).toEqual(expected);
+      },
+    );
+  };
+
+  assertSnapshot(
+    "trs1",
+    "let f1 = ((x, y) => {x}),\n         f2 = ((x, y) => {y});\n       f1(1, 2)",
```
peblair, line 49: Spacing is weird here too (and other spots in this file)

compiler/test/test_end_to_end.re
```diff
@@ -1,2271 +0,0 @@
-open Grain.Compile;
```
peblair, line 1: 😍

peblair (changes requested):
This is beautiful stuff. I've got a couple of minor comments, but this is a major improvement over the existing test suite. Great work!

compiler/src/codegen/compcore.re
ospencer: This resets the label counter between runs (since before if you used the compiler as a library like we do in the tests, it just keeps incrementing) so it makes the snapshots more stable since the numbers show up in the snapshot. (If you ran test files in a different order or maybe filtered, you could maybe get different compiler output, but this takes care of that.)

compiler/src/codegen/emitmod.re
ospencer: 🥴

compiler/test/input/recursive-equal-box.gr
ospencer: All of our non-stdlib tests have to print _something_ to check that the test passed with the expected result (but the stdlib tests are just configured to have a zero exit code). I didn't really want to set up a brand new test runner just for these two tests, so this maintained the current test expectation (since before the module was expected to evaluate to `void`). I'll at least change these to `print("OK")` so it's more obvious what's going on.

compiler/test/suites/boxes.re
ospencer: Nah, I wanted to change the existing tests as little as possible in this PR. This was already planned to be fixed in #479.

peblair (approved):
Looks good!

phated (commented):
A few things. I didn't review any of the actual snapshot files and only skimmed the spec files.

compiler/esy.json
```diff
@@ -10,13 +10,14 @@
...
-      "CLI_BIN_DIR": "#{self.root / '..' / 'cli' / 'bin'}"
+      "CLI_BIN_DIR": "#{self.root / '..' / 'cli' / 'bin'}",
+      "TEST_EXEC_PATH": "#{self.target_dir / 'default' / 'test' / 'test.exe'}"
     }
   },
   "scripts": {
     "copy-compiler": "cp #{$GRAINC_BIN_PATH} #{$GRAINC_JS_PATH} #{$CLI_BIN_DIR}",
     "clean": "rm -rf #{self.root}/_esy",
-    "test": "dune runtest --force",
+    "test": "#{$TEST_EXEC_PATH}",
```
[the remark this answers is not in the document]
ospencer: Yeah I have no idea why I did it the other way 😛

compiler/test/TestFramework.re
```diff
@@ -0,0 +1,83 @@
...
+        err,
+      );
+      Format.pp_print_flush(formatter, ());
+      let s = Buffer.contents(buf);
+      Buffer.reset(buf);
+      Some(s);
+    }
+  );
+
+let starts_with = (string, prefix) => {
```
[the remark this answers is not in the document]
ospencer: Ooo, good idea.

compiler/test/dune
```diff
@@ -1,14 +1,17 @@
+(include_subdirs unqualified)
```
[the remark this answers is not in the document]
ospencer: Yeah, still needed.

compiler/esy.json
```diff
@@ -37,17 +38,21 @@
     "@opam/sexplib": ">= 0.14.0",
     "@opam/utf8": "0.1.0",
     "@opam/yojson": ">= 1.7.0",
+    "@opam/rely": "^3.2.1",
```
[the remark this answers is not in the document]
ospencer: The other reason native stuff we currently pull in via opam. Let's swap them all over in another PR?

phated (approved):
Ship it 🚢

PR #658, opened 2021-05-15
fix(compiler): Correct reporting of locations for CRLF files

ospencer:
Ref https://github.com/grain-lang/grain-language-server/issues/46

We don't really have a great way to test locations right now, so I just added some tests that the locations are correct when using comments.

I also dropped support for the single `CR` and `LFCR` line endings since a single `CR` hasn't been used since Mac OS 9 and I don't think `LFCR` was ever a thing (I did find one guy on Stack Overflow who memed about it, but that's it). I didn't count this as a breaking change because I'm _fairly_ sure Grain doesn't work on Mac OS 9. If anyone feels like we should mark it as breaking though, we can mark it as such.

### DIFF
--- a/compiler/src/parsing/lexer.mll
+++ b/compiler/src/parsing/lexer.mll
@@ -44,14 +44,17 @@
       raise (Error(loc, IllegalUnicodeCodePoint(str)));
   end
 
-  let newline_regex = Str.regexp "\\(\r\\|\n\\)"
+  let newline_regex = Str.regexp "\\(\r\n\\|\n\\)"
 
   let process_newlines lexbuf =
     let input = lexeme lexbuf in
-    String.iter (fun c ->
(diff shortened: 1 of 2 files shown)


peblair (approved):

phated (approved):
You, sir, are a gentleman **and** a scholar!  ❤️

marcusroberts (approved):

marcusroberts:
We can take the warning out of the language server too, excellent work!

phated:
@marcusroberts Do you think we should remove the default of `LF` in the extension config too?

marcusroberts:
@phated I think we should now the compiler is happy.  Windows folk will be used to CRLF and will know to switch to LF if working working with Mac/Linux devs on a project, but otherwise will stick with CRLF (from my experience at least)

PR #659, opened 2021-05-16
feat: Keep track of warnings and report them in LSP mode

marcusroberts:
To provide warnings to the LSP I just keep track of them in a mutable list as they are printed, and then pass them back to the LSP when the compile is finished.

It seems to work, and I'll make a PR for the language server next but would like feedback on the approach before I change this from draft.


ospencer (commented):
I think this general approach is fine. You'll want to have a way to reset the warnings for two reasons: the compiler compiles all files in one process so you'd end up with warnings from multiple files, and for a similar reason, when using the compiler as a library, you'd get warnings from unrelated programs. We don't generate warnings after typechecking, so the reset should probably live in `Typemod.type_implementation`.

marcusroberts:
Taking this back to draft. In LSP mode, I only want the warnings for the file I'm compiling, not the other files.    I'll cut the warnings down to just the compiled file.

compiler/src/diagnostics/output.re
```diff
@@ -43,9 +55,42 @@ let exn_to_lsp_error = (exn: exn): option(lsp_error) => {
...
+let convert_warnings = (warnings_this_run, topLevelFileName) => {
+  List.fold_left(
+    (acc, w) => {
+      let (loc: Grain_utils.Warnings.loc, warn: Grain_utils.Warnings.t) = w;
+      let (file, line, startchar) =
+        Grain_parsing.Location.get_pos_info(loc.loc_start);
+      let (_, endline, endchar) =
+        Grain_parsing.Location.get_pos_info(loc.loc_end);
+
+      if (file == topLevelFileName) {
```
peblair, line 78: Perhaps we should isolate the warnings arrays between files? Maybe something like this:
```reason
let with_preserve_warnings: (unit => 'a) => 'a;

// ...

// (in module_resolution.re:383)
with_preserve_unit^(~loc, chosen_unit_name, srcpath, () =>
  Grain_utils.Warnings.with_preserve_warnings(() => Grain_utils.Config.preserve_config(() =>
    compile_module_dependency^(srcpath, outpath)
  ))
);
```
This way, we shouldn't need to do this filtering step.

marcusroberts: Thanks for the discussion on this, looks a good way to handle it

marcusroberts: PR updated with the suggested method that works well.

compiler/src/utils/warnings.re
```diff
@@ -175,3 +175,25 @@ let check_fatal = () =>
     nerrors := 0;
     raise(Errors);
   };
+
+// track the warnings for the LSP
+let warnings_this_run: ref(list((loc, t))) = ref([]);
```
ospencer, line 180: We'll still want to reset this somewhere, since this will track all warnings across potentially unrelated programs that get compiled (you can think of the tests as a good example, since all of the tests get compiled individually without the compiler ever being "reset"). I think the place I suggested for this was `Typemod.type_implementation`, but with how you've changed it I think it might make more sense to do it in `compile.re` a similar way that we do `Env.clear_imports()`. It might also be useful to make a "reset" helper function that combines these.

marcusroberts: Oh, that's a good point.  I'd only tested individual programs and not see the warnings ever be carried across, but this gives me a good way to test now.  I'll generalise it into a reset helper too

marcusroberts: Just updated it with your suggestion @ospencer .  How does it look now?

compiler/src/compile.re
```diff
@@ -177,9 +177,15 @@ let rec compile_resume = (~hook=?, s: compilation_state) => {
   };
 };
 
+let reset_between_compiles = () => {
```
ospencer, line 180: I might call this `reset_compiler_state`, but otherwise looks great!

marcusroberts: Done

ospencer (approved):
Excellent!

peblair (approved):
Looks good!

Issue #490, opened 2020-12-18
Lang: Tail pattern on arrays discussion


phated:
In #478, we landed array pattern matching. However, this only matches individual elements and doesn't provide a "tail" pattern, like `...rest`` containing the tail of the array. It would be needed to provide something like #189, but it would/might have really poor performance because we would have to create a copy of the entire tail of the array on each match.

A couple questions:
* Do we want to support this for ergonomics?
* Are we okay if it is slow?
* Would there be a way to implement it in the compiler that wouldn't require that slowness?

Any other thoughts?

peblair:
(copied from #189)
Perhaps a slicing operation (since this is effectively an `array[1:]` operation) could be supported by adding a layer of indirection in front of arrays? Then, when doing a slice, we just return a new view of the same underlying array?

alex-snezhko:
To extend this idea, maybe a new `Slice` type could be added to allow slicing on other sequential collection types like Strings and Bytes as well? This type could simply wrap the underlying data and provide an immutable interface for accessing the underlying data, creating a new trimmed slice, etc. Special syntax could also be added for this type; for the pattern matching syntax it would probably have to be distinguished from list patterns so maybe something like `[# first, ...rest]`?

Possible standard library API:
```
// type definition
record Slice<collection, element> {
  data: collection,
  getElem: (Number, collection) -> element,
  start: Number,
  end: Number
}

// aliases for common use cases
type ArraySlice<a> = Slice<Array<a>, a>
type StringSlice = Slice<String, Char>

// converting to/from slice
fromArray(array: Array<a>) -> ArraySlice<a>
fromString/fromBytes/etc...
// can also create a slice from a custom type by specifying how it should behave
fromGeneric(data: a, getElem: (Number, a) -> b, start: Number, end: Number) -> Slice<a, b>
toArray(slice: ArraySlice<a>) -> Array<a>
toString/toBytes/etc...
// reduce/reduceRight could be used in place of something like "toGeneric"

// others
length(slice: Slice<a, b>) -> Number
get(index: Number, slice: Slice<a, b>) -> b  // maybe special syntax can be given here similar to Arrays
trim(trimLeft: Number, trimRight: Number, slice: Slice<a, b>) -> Slice<a, b>  // slice[trimLeft:trimRight] syntax perhaps
forEach/map/reduce/every/some/...
```
Example usage for Array.contains implementation:
```
let contains = (target, array) => {
  let rec containsInner = (slice) => {
    match (slice) {
[...]

PR #660, opened 2021-05-16
feat: Implement Grain array methods in Grain instead of the compiler

ospencer:
I left `ArrayLength` in the compiler since it's used by `match`.


ng-marcus (approved):
I'm not 100% qualified to approve this, but LGTM and great to see Grain implementing Grain!

marcusroberts (approved):
And now as the real me (sorry!)

I'm not 100% qualified to approve this, but LGTM and great to see Grain implementing Grain!

peblair (approved):
These code changes look good!

phated:
Something I'd like to know is how this will affect array pattern matching, specifically how we'll implement recursion with them in the future.

ospencer:
@phated I'm not quite sure what you mean. Can you give me an example?

phated:
@ospencer I'm afk, I just realized I needed to say something now that there were 2 approvals. We have an open issue about it, can you search for it?

ospencer:
@phated Yes, it's #490. No effect whatsoever! This PR is essentially just pulling some stdlib code that was baked into the compiler out since we didn't have the ability to do it this way when it was originally written.

PR #664, opened 2021-05-22
fix(compiler): properly handle special float equality

### LINKED ISSUE #663: Issue with NaN Equality
I was working on adding some constants to the floating point modules (we can set aside the question on whether they should be constants or literals until a later discussion for the sake of this issue), and I came across this bizarre behavior:
```grain
import WasmF32 from "runtime/unsafe/wasmf32";
import WasmI32 from "runtime/unsafe/wasmi32";
import {newFloat32} from "runtime/dataStructures";

let nan = {
  // 0/0 will produce NaN:
  let x = 0.0w
  let y = 0.0w
  let res = WasmF32.div(x, y)
  print(WasmF32.eq(res, res)) // --> false
  let ptr = newFloat32(res)
  WasmI32.toGrain(ptr) : Float32
}
print(nan) // -> NaN
print(nan == nan) // --> true
```
So, something about the boxing-unboxing cycle is causing the NaN-ness to be lost during the equality check (since the final `print` should be `false`). I took a quick look at the number equality function to see if I could find what was causing it, but nothing jumped out at me, so this will need more careful investigation.

ospencer:
Fixes #663

### DIFF
--- a/compiler/test/test_end_to_end.re
+++ b/compiler/test/test_end_to_end.re
@@ -2137,6 +2137,18 @@ let number_tests = [
   t("number_syntax8", "1.2e2", "120.0"),
   t("number_syntax9", "1l", "1"),
   t("number_syntax10", "1L", "1"),
+  // equality checks
+  t(
+    "nan_equality1",
+    {|import Float32 from "float32"; Float32.div(0.0f, 0.0f) == Float32.div(0.0f, 0.0f)|},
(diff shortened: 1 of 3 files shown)


peblair (approved):
🙏

ng-marcus (approved):

marcusroberts (approved):

PR #511, opened 2021-01-16
Hover displays the enum and record name

ospencer:
Recreation of #503

PR #667, opened 2021-05-23
feat(compiler): add better hover

marcusroberts:
Give more details on hover for enums and records.

(dupe of #511 recreated here against main)


compiler/src/diagnostics/output.re
```diff
@@ -43,9 +55,36 @@ let exn_to_lsp_error = (exn: exn): option(lsp_error) => {
...
+      ~values: list(Lenses.lens_t),
+    ) => {
+  let result: lsp_result = {errors, warnings, values, lenses: []};
   Yojson.Basic.pretty_to_string(
     Yojson.Safe.to_basic(lsp_result_to_yojson(result)),
   );
 };
+
+let convert_warnings = (warnings_this_run, topLevelFileName) => {
+  List.map(
```
peblair, line 70: @ospencer Do you think that warnings emitted at the dummy location should be included here?

marcusroberts: I just merged main on this PR, so this refers to a change that went through #659

ospencer (approved):

phated (approved):
🔥  Love it!

PR #668, opened 2021-05-23
feat(compiler)!: Universal WebAssembly initial and maximum pages flags

ospencer:
feat(compiler): Add --initial-memory-pages and --maximum-memory-pages to set module memory limits
feat(runtime)!: Removed --limitMemory, GRAIN_INIT_MEMORY_PAGES, and GRAIN_MAX_MEMORY_PAGES in favor of --initial-memory-pages and --maximum-memory-pages
feat(runtime)!: Removed `managedMemory`, `memory`, and `table` exports. These can be accessed via `runner.managedMemory`, `runner.managedMemory._memory`, and `runner.table` respectively.
feat(runtime)!: Removed `encoder` and `decoder` exports.

The important takeaway from this PR is that it adds two new flags, `--initial-memory-pages` and `--maximum-memory-pages`, which work in both linked and non-linked mode (versus the fragmented way it was done before).

### DIFF
--- a/runtime/src/core/closures.js
+++ b/runtime/src/core/closures.js
@@ -1,17 +1,16 @@
-import { managedMemory, grainModule } from "../runtime";
 import { grainToJSVal, JSToGrainVal } from "../utils/utils";
 
 export class GrainClosure {
   constructor(loc, runtime) {
-    const view = managedMemory.view;
+    const view = runtime.managedMemory.view;
 
     this.loc = loc;
     this.runtime = runtime;
-    this.arity = view[loc];
-    this.ptr = view[loc + 1];
-    this.closureSize = view[loc + 2];
-    this.closureElts = view.slice(loc + 3, loc + 3 + this.closureSize);
-    this.func = grainModule.instance.exports["GRAIN$LAM_" + this.ptr];
+    this.arity = view[loc + 1];
+    this.ptr = view[loc + 2];
+    this.closureSize = view[loc + 3];
+    this.closureElts = view.slice(loc + 4, loc + 4 + this.closureSize);
+    this.func = runtime.table.get(this.ptr);
   }
 
   jsFunc(...args) {
@@ -22,7 +21,7 @@ export class GrainClosure {
 }
 
 export function printClosure(c) {
-  const view = managedMemory.view;
+  const view = runtime.managedMemory.view;
 
   c /= 4;
   let arity = view[c];
--- a/runtime/src/runtime.js
+++ b/runtime/src/runtime.js
@@ -5,50 +5,34 @@ import { ManagedMemory } from "./core/memory";
 import { GrainRunner } from "./core/runner";
 import { defaultFileLocator } from "./utils/locators";
 
-export let grainModule;
-
-// Workaround for this memory being initialized statically on module load
-export const memory = new WebAssembly.Memory({
-  initial: process.env.GRAIN_INIT_MEMORY_PAGES || 64,
-  maximum: process.env.GRAIN_MAX_MEMORY_PAGES,
-});
-export const table = new WebAssembly.Table({
-  element: "anyfunc",
-  initial: 1024,
-});
-export const encoder = new TextEncoder("utf-8");
-export const decoder = new TextDecoder("utf-8");
-export const managedMemory = new ManagedMemory(memory);
-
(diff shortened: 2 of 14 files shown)


runtime/src/core/closures.js
```diff
@@ -1,17 +1,16 @@
...
     this.runtime = runtime;
     this.arity = view[loc];
-    this.ptr = view[loc + 1];
-    this.closureSize = view[loc + 2];
-    this.closureElts = view.slice(loc + 3, loc + 3 + this.closureSize);
-    this.func = grainModule.instance.exports["GRAIN$LAM_" + this.ptr];
+    this.ptr = view[loc + 2];
+    this.closureSize = view[loc + 3];
+    this.closureElts = view.slice(loc + 4, loc + 4 + this.closureSize);
+    this.func = runtime.table.get(this.ptr);
```
peblair, line 13: Just for my understanding, can you explain a little more about this change?

runtime/src/core/closures.js
```diff
@@ -1,17 +1,16 @@
...
+    const view = runtime.managedMemory.view;
 
     this.loc = loc;
     this.runtime = runtime;
     this.arity = view[loc];
-    this.ptr = view[loc + 1];
-    this.closureSize = view[loc + 2];
-    this.closureElts = view.slice(loc + 3, loc + 3 + this.closureSize);
-    this.func = grainModule.instance.exports["GRAIN$LAM_" + this.ptr];
+    this.ptr = view[loc + 2];
```
peblair, line 10: Was this just broken before, or did one of the changes in the compiler in this PR change the memory layout of closures?

peblair (commented):
I don't see any code issues with these changes. I believe I understand what's going on for the most part, but I've left a couple of questions. Any help would be appreciated!

runtime/src/core/closures.js
ospencer: This was just broken. This code isn't used or exercised at all and we should consider just deleting it 😶

runtime/src/core/closures.js
ospencer: The previous method here worked back before we had separate compilation, since there was only one Grain module being loaded and we exported all of the functions in the module. Now of course there's multiple modules and we only export the closures. This grabs the table index from the closure, and then just looks up the function in the global webassembly table. Pretty neat, and makes this more likely to work if we decide that we want to try to fix this.

peblair (approved):
This all looks really good! I believe I understand it enough to say it's good to go, but @phated should take a look too.

runtime/src/runtime.js
```diff
@@ -5,50 +5,34 @@ import { ManagedMemory } from "./core/memory";
 import { GrainRunner } from "./core/runner";
 import { defaultFileLocator } from "./utils/locators";
 
-export let grainModule;
-
-// Workaround for this memory being initialized statically on module load
-export const memory = new WebAssembly.Memory({
-  initial: process.env.GRAIN_INIT_MEMORY_PAGES || 64,
-  maximum: process.env.GRAIN_MAX_MEMORY_PAGES,
```
phated, line 13: Can we get rid of these env vars then (including removing from setting in the CLI)?

runtime/src/runtime.js
```diff
@@ -5,50 +5,34 @@ import { ManagedMemory } from "./core/memory";
...
-});
-export const table = new WebAssembly.Table({
-  element: "anyfunc",
-  initial: 1024,
-});
-export const encoder = new TextEncoder("utf-8");
-export const decoder = new TextDecoder("utf-8");
```
phated, line 20: Can we remove the `fast-text-encoding` import and dependency then?

runtime/src/utils/utils.js
```diff
@@ -1,12 +1,9 @@
-import { encoder, decoder, managedMemory } from "../runtime";
+const encoder = new TextEncoder("utf-8");
+const decoder = new TextDecoder("utf-8");
```
phated, line 2: What about `fast-text-encoding`? Even needed anymore?

phated (commented):
Just a few questions, but mostly looks good.

runtime/src/runtime.js
ospencer: Should be all set here.

phated (approved):
I am excite!

PR #670, opened 2021-05-25
fix: Properly escape quoted strings in toString/print

### LINKED ISSUE #669: Quoted strings should be escaped
This program
```grain
print(("foo\\bar", 2))
```
outputs the following:
```
("foo\bar", 2)
```
This is, in my opinion, a bit counterintuitive, since the quotes around the string make me believe that `\b` is an escape sequence in the string that is being rendered. In reality, we are seeing a single backslash simply because the printer does not escape the contents of quoted strings before printing them. This can have further unintended consequences:
```
print(("foo\nbar", 2))
-->
("foo
bar", 2)
```
While printing strings without escaping is preferable when the string is the top-level value being printed, we should properly escape these sequences when printing quoted strings.

ospencer:
feat: Support \b, \f, \t, and \v escapes
fix: Fix parsing of `'\''` char literal

Fixes #669


marcusroberts (approved):

phated (approved):
Excellent! And nice cleanup on that old code ❤️

PR #674, opened 2021-05-28
feat(stdlib): add Range.map

### LINKED ISSUE #616: Stdlib: Implement Range.map
```grain
map : ((Number -> a), Range) -> List<a>
```

Conaclos:
Fix #616

### DIFF
--- a/compiler/test/stdlib/range.test.gr
+++ b/compiler/test/stdlib/range.test.gr
@@ -75,3 +75,10 @@ assert results == [2, 3, 4, 5]
 Range.forEach((idx) => {
   fail "Shouldn't be called"
 }, exclusiveSameRange)
+
+assert Range.map(toString, inclusiveAscendingRange) == ["1", "2", "3", "4", "5"]
+assert Range.map(toString, inclusiveDescendingRange) == ["5", "4", "3", "2", "1"]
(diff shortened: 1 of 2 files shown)


compiler/test/stdlib/range.test.gr
```diff
@@ -75,3 +75,11 @@ assert results == [2, 3, 4, 5]
 Range.forEach((idx) => {
   fail "Shouldn't be called"
 }, exclusiveSameRange)
+
+let idfn = (a) => a
```
ospencer, line 79: Could we instead do something here that produces a new value? Mapping over `toString` could be a good candidate!

stdlib/range.gr
```diff
@@ -1,3 +1,4 @@
+
```
ospencer, line 1:
```suggestion
```

ospencer (changes requested):
Very nice, thanks @Conaclos! I've got one request but otherwise looks great 🙌

Conaclos:
Tanks for these suggestions. I updated the code :)

phated (approved):
Excellent!

ospencer (approved):
Perfect, thank you!

Issue #661, opened 2021-05-17
Char Unicode Data and Conversions


peblair:
It would be useful for `Char` to support a variety of Unicode-aware query functions and conversion functions (`toUpper`, `isPunctuation`, etc). For example, these are the ones supported by Racket [here](https://docs.racket-lang.org/reference/characters.html#%28part._.Classifications%29) and [here](https://docs.racket-lang.org/reference/characters.html#%28part._.Character_.Conversions%29).

We should try to add as many of these as possible, as one never knows what might be useful for libraries.

cician:
I'm currently working on emitting JSON in Grain and for escaping I need to generate an UTF-16 surrogate pair from a unicode codepoint.  And of course vice-versa, but parsing is still a long way off.

ospencer:
@cician My question is a tad unrelated to this issue, but I'm not sure I understand—for what you're trying to accomplish, why do you need to make surrogate pairs? Grain strings are UTF-8.

cician:
Actually I don't strictly need it because only ASCII codes 0-31 need to be escaped for conforming JSON output in UTF-8, but I've tentatively added an option to escape all non ASCII characters.

The ECMA-404 spec (https://www.ecma-international.org/publications-and-standards/standards/ecma-404/) says the escaping should be done in UTF-16 pairs, unless I misunderstand something. I'm learning in the process about both unicode and Grain. I think it's a consequence of the fact that JSON inherits some properties from JavaScript, which doesn't use UTF-8 internally. It spills to how escaping is done in JavaScript strings and thus JSON.

PS: I'm working on it [here](https://github.com/cician/grain-json).

ospencer:
Ah I see, it's the specification for unicode character escapes that appear within JSON object strings. Got it. That's interesting! So you'd want a utility like `Char.escapeSurrogatePair : Char -> String` that would take a char and return its unicode escape as a surrogate pair, e.g. `assert Char.escapeSurrogatePair('𝄞') == "\\uD834\\uDD1E"`? That'd differ from `Char.escape` which would just produce `"\\u{1D11E}"` for regular Grain strings, yeah?

ospencer:
Or I guess it could just be called `escapeUTF16`.

cician:
For now I've just copied a few lines from OpenJDK's source to do the job, but I should probably remove it to avoid copyright/licensing issues.

[...]

Issue #190, opened 2020-07-03
Compiler: Importing Records generates warnings


phated:
I exported a Record from my stdlib module, like so:
```
export data InternalStats = {
  nodeCount: Number,
  bucketCount: Number
}

export let getInternalStats = (map) => {
  {
    nodeCount: 0
    bucketCount: 1
  }
}
```

And when I try to use it in the tests:

```
import Map, { InternalStats } from 'map';

let resize = Map.makeSized(1);

assert Map.getInternalStats(resize) == { nodeCount: 0, bucketCount: 1 };
```

I get these warnings:

```
File "compiler/test/stdlib/map.test.gr", line 267, characters 39-71:
Warning 2: this type-based record disambiguation is not principal.
File "compiler/test/stdlib/map.test.gr", line 267, characters 39-71:
Warning 4: this record of type map.InternalStats contains fields that are 
not visible in the current scope: nodeCount bucketCount.
They will not be selected if the type becomes unknown.
```

cician:
I've stumbled upon this as well. In my case there's only the second warning about non visible fields and it's in plain modules, not in stdlib.

PR #680, opened 2021-05-30
feat(stdlib): Regular Expressions

### LINKED ISSUE #247: Stdlib: RegExp
I'll admit that I have no idea how we'd go about building this! RegExp support seems really difficult.

peblair:
This pull request adds a new module to the standard library for working with regular expressions. The flavor of regular expressions added here is adapted from Racket, providing a full Perl-like regex syntax for users to work with.

### Regular Expressions in Grain

```grain
// @param regexString: String - The regular expression to compile
// @returns Result<RegularExpression>
export let make = (regexString: String)
```
Compiles the given pattern string into a regular expression object.

For an overview of the theory regular expressions in general, readers are referred
to "Mastering Regular Expressions" by Friedl, or numerous alternative resources online.

Regular expressions are a combination of normal and special characters. A normal
character in a pattern will match a one-character string containing that character.
Moreover, if there are two regular expressions `A` and `B`, they can be concatenated
into a regular expression `AB`. If a string `p` matches `A` and `q` matches `B`,
then `pq` will match `AB`.

The special character sequences are as follows:

- `.` - Matches any character, except for a newline in multi-line mode
- `^` - Matches the beginning of the input, or after a newline (`\n`) in multi-line mode
- `$` - Matches the end of the input, or the before a newline (`\n`) in multi-line mode
- `«re»*` - Matches `«re»` zero or more times
- `«re»+` - Matches `«re»` one or more times
- `«re»?` - Matches `«re»` zero or one times
- `«re»{«n»}` - Matches `«re»` exactly `«n»` times
- `«re»{«n»,}` - Matches `«re»` `«n»` or more times
- `«re»{,«m»}` - Matches `«re»` zero to `«m»` times
- `«re»{«n»,«m»}` - Matches `«re»` between `«n»` and `«m»` times
- `«re»{}` - Matches `«re»` zero or more times
- `[«rng»]` - Matches any character in `«rng»` (see below)
- `[^«rng»]` - Matches any character not in `«rng»` (see below)
- `\«n»` - Matches the latest match for group `«n»` (one-indexed)
- `\b` - Matches the boundary of `\w*` (`\w` defined below, under "basic classes")
- `\B` - Matches where `\b` does not
- `\p{«property»}` - Matches any character with Unicode property `«property»` (see below)
- `\P{«property»}` - Matches any character without Unicode property `«property»` (see below)
- `(«re»)` - Matches `«re»`, storing the result in a group
- `(?:«re»)` - Matches `«re»` without storing the result in a group
- `(?«mode»:«re») - Matches `«re»` with the mode settings specified by `«mode»` using the following syntax:
  - `«mode»i` - The same as `«mode»`, but with case-insensitivity enabled (temporarily not supported until grain-lang/grain#661 is resolved)
  - `«mode»-i` - The same as `«mode»`, but with case-insensitivity disabled (the default)
  - `«mode»m` / `«mode»-s` - The same as `«mode»`, but with multi-line mode enabled
  - `«mode»-m` / `«mode»s` - The same as `«mode»`, but with multi-line mode disabled (the default)
  - An empty string, which will not change any mode settings
- `(?«tst»«re1»|«re2»)` - Will match `«re1»` if `«tst»`, otherwise will match `«re2»`. The following options are available for `«tst»`
  - `(«n»)` - Will be true if group `«n»` has a match
  - `(?=«re»)` - Will be true if `«re»` matches the next sequence
  - `(?!«re»)` - Will be true if `«re»` does not match the next sequence
  - `(?<=«re»)` - Will be true if `«re»` matches the preceding sequence
  - `(?<!«re»)` - Will be true if `«re»` does not match the preceding sequence
- `(?«tst»«re»)` - Equivalent to `(?«tst»«re»|)`
- Finally, basic classes (defined below) can also appear outside of character ranges.

Character ranges (referred to as `«rng»` above) have the following syntax:
- `«c»` - Matches the character `«c»` exactly
- `«c1»-«c2»` - Matches any character with a character code between the character code for `«c1»` and the code for `«c2»`

These forms can be repeated any number of times, which will construct a range of their union. That is, `[ba-c]` and `[a-c]` are equivalent ranges.
Additionally, there are the following special cases:
- A `]` as the first character of the range will match a `]`
- A `-` as the first or last character of the range will match a `-`
- A `^` in any position other than the first position will match a `^`
- `\«c»`, where `«c»` is a non-alphabetic character, will match `«c»`

Furthermore, ranges can include character classes, which are predefined commonly-used
sets of characters. There are two "flavors" of these: *basic* classes and *POSIX* classes.
Both are provided for ease of use and to maximize compatibility with other regular
expression engines, so feel free to use whichever is most convenient.

The *basic* classes are as follows:
- `\d` - Matches `0-9`
- `\D` - Matches characters not in `\d`
- `\w` - Matches `a-z`, `A-Z`, `0-9`, and `_`
- `\W` - Matches characters not in `\w`
- `\s` - Matches space, tab, formfeed, and return
- `\S` - Matches characters not in `\s`
The *POSIX* classes are as follows:
- `[:alpha:]` - Matches `a-z` and `A-Z`
- `[:upper:]` - Matches `A-Z`
- `[:lower:]` - Matches `a-z`
- `[:digit:]` - Matches `0-9`
- `[:xdigit:]` - Matches `0-9`, `a-f`, and `A-F`
- `[:alnum:]` - Matches `a-z`, `A-Z`, and `0-9`
- `[:word:]` - Matches `a-z`, `A-Z`, `0-9`, and `_`
- `[:blank:]` - Matches space and tab
- `[:space:]` - Matches space, tab, newline, formfeed, and return
- `[:graph:]` - Matches all ASCII characters which use ink when printed
- `[:print:]` - Matches space, tab, and all ASCII ink users
- `[:cntrl:]` - Contains all characters with code points < 32
- `[:ascii:]` - Contains all ASCII characters

Finally, the following is the list of supported Unicode properties. **Note** that
until grain-lang/grain#661 is resolved, matches using `\p` and `\P` are disabled,
as Grain is not currently able to determine membership in these character classes.
These class codes come from this portion of the Unicode standard:
https://www.unicode.org/reports/tr44/#General_Category_Values

- `Ll` - Letter, lowercase
- `Lu` - Letter, uppercase
- `Lt` - Letter, titlecase
- `Lm` - Letter, modifier
- `L&` - Union of `Ll`, `Lu`, `Lt`, and `Lm`
- `Lo` - Letter, other
- `L` - Union of `L&` and `Lo`
- `Nd` - Number, decimal digit
- `Nl` - Number, letter
- `No` - Number, other
- `N` - Union of `Nd`, `Nl`, and `No`
- `Ps` - Punctuation, open
- `Pe` - Punctuation, close
- `Pi` - Punctuation, initial quote
- `Pf` - Punctuation, final quote
- `Pc` - Punctuation, connector
- `Pd` - Punctuation, dash
- `Po` - Punctuation, other
- `P` - Union of `Ps`, `Pe`, `Pi`, `Pf`, `Pc`, `Pd`, and `Po`
- `Mn` - Mark, non-spacing
- `Mc` - Mark, spacing combining
- `Me` - Mark, enclosing
- `M` - Union of `Mn`, `Mc`, and `Me`
- `Sc` - Symbol, currency
- `Sk` - Symbol, modifier
- `Sm` - Symbol, math
- `So` - Symbol, other
- `S` - Union of `Sc`, `Sk`, `Sm`, and `So`
- `Zl` - Separator, line
- `Zp` - Separator, paragraph
- `Zs` - Separator, space
- `Z` - Union of `Zl`, `Zp`, and `Zs`
- `Cc` - Other, control
- `Cf` - Other, format
- `Cs` - Other, surrogate
- `Cn` - Other, not assigned
- `Co` - Other, private use
- `C` - Union of `Cc`, `Cf`, `Cs`, `Cn`, and `Co`
- `.` - Union of all Unicode categories

### Public API (for working with regular expressions)

```grain
/*
 The user-facing object which contains the results
 of a regular expression match.
 */
export record MatchResult {
  // Returns the contents of the given match group, raising
  // an exception if that group was not matched.
  group: Number -> String,
  // Returns the contents of the given group
  groupOptional: Number -> Option<String>,
  // Returns the position of the given match group, raising
  // an exception if that group was not matched.
  groupPosition: Number -> (Number, Number),
  // Returns the position of the given group
  groupPositionOptional: Number -> Option<(Number, Number)>,
  // Returns the number of defined groups in this match object (includes group 0)
  numGroups: Number,
  // Returns the contents of all groups matched in this match object
  allGroups: () -> Array<Option<String>>,
  // Returns the positions of all groups matched in this match object
  allGroupPositions: () -> Array<Option<(Number, Number)>>,
}

// Returns true if the given regular expression has a match in the given string
// @param rx: RegularExpression - The regular expression to search for
// @param string: String - The string to search within
// @returns Bool
export let isMatch = (rx: RegularExpression, string: String)

// Returns true if the given regular expression has a match in the given string within the given start/end offsets
// @param rx: RegularExpression - The regular expression to search for
// @param string: String - The string to search within
// @param start: Number - The start offset to search within
// @param end: Number - The end offset to search within
// @returns Bool
export let isMatchRange = (rx: RegularExpression, string: String, start: Number, end: Number)

// Returns the first match for the given regular expression contained within the given string
// @param rx: RegularExpression - The regular expression to search for
// @param string: String - The string to search within
// @returns Option<MatchResult>
export let find = (rx: RegularExpression, string: String)

// Returns the first match for the given regular expression contained within the given string,
// within the given start/end range.
// @param rx: RegularExpression - The regular expression to search for
// @param string: String - The string to search within
// @param start: Number - The start offset to search within
// @param end: Number - The end offset to search within
// @returns Option<MatchResult>
export let findRange = (rx: RegularExpression, string: String, start: Number, end: Number)

// Returns the all matches for the given regular expression contained within the given string
// @param rx: RegularExpression - The regular expression to search for
// @param string: String - The string to search within
// @returns List<MatchResult>
export let findAll = (rx: RegularExpression, string: String)

// Returns all matches for the given regular expression contained within the given string,
// within the given start/end range.
// @param rx: RegularExpression - The regular expression to search for
// @param string: String - The string to search within
// @param start: Number - The start offset to search within
// @param end: Number - The end offset to search within
// @returns List<MatchResult>
export let findAllRange = (rx: RegularExpression, string: String, start: Number, end: Number)

// Replaces the first match for the given regular expression contained within the given string with the given replacement.
// Replacement strings support the following syntax:
// - `&` - Replaced with the text of the matching portion of input (e.g. for `(foo)`, the search string `foo bar`, and the replacement `baz &`, the result will be `baz foo bar`)
// - `\n` / `\nn` (where `n` is a digit) - Replaced with the text of group `nn`
// - `\$` - Does nothing (this exists to support replacement strings such as `\4\$0`, which will place the contents of group 4 prior to a zero)
// - `\&` / `\\` - Literal `&` and `\`, respectively
// - Any other character will be placed as-is in the replaced output.
//
// @param rx: RegularExpression - The regular expression to search for
// @param toSearch: String - The string to search within
// @param replacement: String - The string to replace matches with
// @returns String
export let replace = (rx: RegularExpression, toSearch: String, replacement: String)

// Replaces all matches for the given regular expression contained within the given string with the given replacement.
// See `replace` for replacement string syntax.
//
// @param rx: RegularExpression - The regular expression to search for
// @param toSearch: String - The string to search within
// @param replacement: String - The string to replace matches with
// @returns String
export let replaceAll = (rx: RegularExpression, toSearch: String, replacement: String)
```

## Other Changes
(crossed-out changes have been delegated to linked separate PRs)
- ~Mutually recursive `record` and `enum` types are now supported, and can be defined by separating data definitions with a comma~ (#725)
- ~`String.join` has been added to the `String` module, which allows the concatenation of a list of strings~ (#722) 
- ~`String.charAt` has been added to the `String` module, which allows the retrieval of a character in a string by index~ (#721)
- ~A bug in record printing has been fixed which caused the closing brace of a nested record to have incorrect indentation~ (#724)
- ~`min` and `max` functions have been added to `Pervasives`, which compute the minimum and maximum of two numbers, respectively~ (obsoleted by concurrent work on `Number` library)
- ~`infinity` and `nan` constants have been added to `Float32` and `Float64`~ (#720)
- ~`Array.subArray` has been added to the `Array` module, which allows for slicing an array~ (#727)
- ~`Array.zip` has been added to the `Array` module, which allows for easily combining elements of two arrays into a single array~ (#719)

Closes #247.


ospencer:
Awesome work, and as you can imagine I love this API :D 

Haven't dove super deep into the code yet, but here's how I want to rename stuff (largely the names don't need `regex` in them because they're in the regex lib, but some other thoughts too):

```
export record MatchResult {
  group: Number -> String,
  groupOpt: Number -> Option<String>,
  groupPosition: Number -> (Number, Number),
  groupPositionOpt: Number -> Option<(Number, Number)>,
  numGroups: Number,
  allGroups: () -> Array<Option<String>>, // <-- also why is this Array<Option<String>> instead of Array<String>?
  allGroupPositions: () -> Array<Option<(Number, Number)>>, // Same question here
}

export let isMatch = (rx: RegularExpression, string: String) // `test` is also a good name here
export let isMatchRange = (rx: RegularExpression, string: String, start: Number, end: Number) // or `testRange`
export let match = (rx: RegularExpression, string: String)
export let matchRange = (rx: RegularExpression, string: String, start: Number, end: Number)
export let matchAll = (rx: RegularExpression, string: String)
export let matchAllRange = (rx: RegularExpression, string: String, start: Number, end: Number)
export let replace = (rx: RegularExpression, toSearch: String, replacement: String)
export let replaceAll = (rx: RegularExpression, toSearch: String, replacement: String)
```

And then the last thing would be renaming `makeRegex` -> `make` or something like `compile` 👍

peblair:
After discussing with @ospencer (we can't use `match` as a function name, unfortunately), I've pushed this API:
```
export record MatchResult {
  group: Number -> String,
  groupOpt: Number -> Option<String>,
  groupPosition: Number -> (Number, Number),
  groupPositionOpt: Number -> Option<(Number, Number)>,
  numGroups: Number,
  allGroups: () -> Array<Option<String>>, // <-- also why is this Array<Option<String>> instead of Array<String>?
  allGroupPositions: () -> Array<Option<(Number, Number)>>, // Same question here
}

export let isMatch = (rx: RegularExpression, string: String) // `test` is also a good name here
export let isMatchRange = (rx: RegularExpression, string: String, start: Number, end: Number) // or `testRange`
export let find = (rx: RegularExpression, string: String)
export let findRange = (rx: RegularExpression, string: String, start: Number, end: Number)
export let findAll = (rx: RegularExpression, string: String)
export let findAllRange = (rx: RegularExpression, string: String, start: Number, end: Number)
export let replace = (rx: RegularExpression, toSearch: String, replacement: String)
export let replaceAll = (rx: RegularExpression, toSearch: String, replacement: String)
```
I will update the main post to reflect these changes.

marcusroberts:
Wow!   Great work here!

stdlib/regex.gr
```diff
@@ -0,0 +1,2946 @@
...
+
+let configIncGroupNumber = (config: RegExParserConfig) => {
+  config.groupNumber := unbox(config.groupNumber) + 1;
+  config
+};
+
+record RegExBuf {
+  input: String,
+  inputExploded: Array<Char>,
+  cursor: Box<Number>,
```
ospencer, line 73:
```suggestion
  mut cursor: Number,
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2946 @@
...
+};
+
+let withConfig = (buf: RegExBuf, config: RegExParserConfig) => {
+  {input: buf.input, inputExploded: buf.inputExploded, cursor: buf.cursor, config: config}
+};
+
+// Parsing internals for recursive descent
+
+let parseErr = (buf: RegExBuf, msg: String, posShift) => {
+  "Invalid Regular Expression: " ++ msg ++ " (position " ++ toString(unbox(buf.cursor) + posShift) ++ ")"
```
ospencer, line 88: Should we have actual exceptions that people can pattern match on? Maybe something like this:
```suggestion
  RegexParseError("Invalid Regular Expression: " ++ msg ++ " (position " ++ toString(unbox(buf.cursor) + posShift) ++ ")")
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2946 @@
...
+
+=================================
+REGEX RANGE DEFINITIONS AND UTILS
+=================================
+
+Based on https://github.com/racket/racket/blob/0a9c70e95a69743dd5d219a395e995be4a4bfd41/racket/src/regexp/common/range.rkt
+
+*/
+
+// [TODO] alias type RERange as List<(Number, Number)>
```
ospencer, line 157: Issue?

stdlib/regex.gr
```diff
@@ -0,0 +1,2946 @@
...
+    }
+  };
+  let rec loop = (mode, accum, l) => {
+    match(l) {
+      // flatten nested sequences
+      [(RESequence(rxs1, _)), ...tl] => loop(mode, accum, List.append(rxs1, tl)),
+      // drop empty elements
+      [REEmpty, ...tl] => loop(mode, accum, tl),
+      [RELiteralString(""), ...tl] => loop(mode, accum, tl),
+      // [TODO] we _really_ need to add or patterns...
```
ospencer, line 421: Issue to clean up?
```suggestion
      // [TODO] Clean up with or-patterns
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2946 @@
...
+      Ok('+') => Err(parseErr(buf, "`+` follows nothing in pattern", 0)),
+      Ok('?') => Err(parseErr(buf, "`?` follows nothing in pattern", 0)),
+      Ok('{') when buf.config.isPerlRegExp => Err(parseErr(buf, "`{` follows nothing in pattern", 0)),
+      Ok('\\') => {
+        ignore(eat(buf, '\\'));
+        parseBackslashLiteral(buf)
+      },
+      Ok(')') => Err(parseErr(buf, "Unmatched `)` in pattern", 0)),
+      Ok(c) when (buf.config.isPerlRegExp) && (c == ']' || c == '}') => Err(parseErr(buf, "unmatched `" ++ Char.toString(c) ++ "` in pattern", 0)),
+      // [TODO] case-insensitive
```
ospencer, line 1203: Issue for this?

stdlib/regex.gr
```diff
@@ -0,0 +1,2946 @@
...
+let charMatcherIterated = (toMatch, max) => iterateMatcher((buf: MatchBuf, pos: Number, start: Number, limit: Number, end: Number, state, stack) => {
+  match(matchBufChar(buf, pos)) {
+    Err(_) => false,
+    Ok(c) => toMatch == c
+  }
+}, 1, max);
+
+// string matching
+
+// [TODO] would be better to avoid copies here (replace subarray calls to something which just does an equality check on the array)
```
ospencer, line 1909: Issue?

stdlib/runtime/numbers.gr
```diff
@@ -520,9 +520,31 @@ let numberEqualFloat64Help = (x, y) => {
   }
 }
 
+/*
 let numberEqualFloat32Help = (x, y) => {
   numberEqualFloat64Help(WasmF64.promoteF32(x), y)
 }
+*/
+
+let numberEqualFloat32Help = (x, y) => {
```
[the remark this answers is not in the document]
peblair: It's required for the new NaN constant to work correctly. I'd go as far as to say it's a bugfix, so I am not sure how worried we should be about it being breaking (since the current behavior is incorrect).

stdlib/regex.gr
```diff
@@ -0,0 +1,2946 @@
...
+===============================
+
+*/
+
+record RegExParserConfig {
+  isPerlRegExp: Bool,
+  caseSensitive: Bool,
+  multiLine: Bool,
+  groupNumber: Box<Number>,
+  references: Box<Bool>,
```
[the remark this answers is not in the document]
peblair: It's actually important here to use boxes, due to the way that `withConfig` and friends work (since we create new records which share the underlying references). I can add a comment.

stdlib/regex.gr
peblair: We can, but I'm not sure I see the benefit unless there are multiple variants, and I cannot think of multiple useful variants to define.

peblair: For the record, I made this change locally and did not see any benefit from it. Remember that `Result`'s `Err` value is typed as well, in case you were thinking something along the lines of the open type patterns (e.g. warnings) that we use in the compiler's Reason code.

stdlib/regex.gr
peblair: #691

stdlib/regex.gr
peblair: #692

stdlib/float32.gr
```diff
@@ -74,3 +74,21 @@ export let gte = (x: Float32, y: Float32) => {
   let y = WasmF32.load(WasmI32.fromGrain(y), 8n)
   WasmF32.ge(x, y)
 }
+
+// floating-point constants:
```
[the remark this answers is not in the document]
peblair: #693

stdlib/array.gr
```diff
@@ -301,6 +305,13 @@ export let product = (a, b) => {
   })
 }
 
+export let zip = (a, b) => {
```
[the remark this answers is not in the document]
peblair: #694

compiler/test/stdlib/regex.test.gr
```diff
@@ -0,0 +1,616 @@
...
+assert testRegex("[\\w]+", "--ab_cd0123--") == Ok(Some(("ab_cd0123", [> ])));
+assert testRegex("\\D+", "1234abc5678") == Ok(Some(("abc", [> ])));
+assert testRegex("[\\D]+", "1234abc5678") == Ok(Some(("abc", [> ])));
+assert testRegex("[\\da-fA-F]+", "123abc") == Ok(Some(("123abc", [> ])));
+assert testRegex("([\\s]*)([\\S]*)([\\s]*)", " testing!1972") == Ok(Some((" testing!1972", [> Some(" "), Some("testing!1972"), Some("")])));
+assert testRegex("(\\s*)(\\S*)(\\s*)", " testing!1972") == Ok(Some((" testing!1972", [> Some(" "), Some("testing!1972"), Some("")])));
+assert testRegex("(([a-z]+):)?([a-z]+)$", "smil") == Ok(Some(("smil", [> None, None, Some("smil")])));
+// We handle this the same as Racket (returning None instead of an error); is that a mistake?
+// This isn't a well-defined regexp, so maybe we should error in the parser...
+//assert Result.isErr(testRegex("((.)\\1+)", ""));
```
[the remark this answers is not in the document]
peblair: #695

stdlib/regex.gr
peblair: #696

stdlib/regex.gr
peblair: I just went ahead and fixed it.

stdlib/runtime/numbers.gr
```diff
@@ -1402,3 +1418,19 @@ export let decr = (x) => {
...
+  }
+}
+
+export let max = (x, y) => {
+  if (x < y) {
+    y
+  } else {
+    x
+  }
+}
```
phated, line 1436: These are now in the `Number` library, so I think we should rely on those instead of runtime/numbers.

phated:
So, I haven't gotten to look through the implementation yet, but the first thought I had while looking over the tests was "what is the actual public API of this library?" - I appreciate that you have added that overview in the body of the PR, but as someone who usually jumps to tests to see the usage, it feels weird that our tests need to be written with helpers. I believe that indicates that the API might be too complex for ergonomic usage.

As for the public API, I believe that we shouldn't have functions in a `MatchResult` that throw exceptions. We can just provide functions that return Option and users can use `Option.unwrap` if they know for a fact that those will always be `Some`.

What are `(Number, Number)` for the groups positions? Is it group number and string position?

Why does the `MatchResult` retusn Arrays but the RegExp exported methods return Lists?

I'm also wondering why we return an `Option<MatchResult>`? It seems like this record could always be provided, even if nothing matched, maybe it would need to have a field that indicated the success status of the match? I'm not sure.

I'm not sure I like the replace syntax. It feels like too much of a deviation from the languages people are going to be coming from.

I'll probably have more, but I need to sign off right now.

peblair:
@phated:
> As for the public API, I believe that we shouldn't have functions in a MatchResult that throw exceptions. We can just provide functions that return Option and users can use Option.unwrap if they know for a fact that those will always be Some.

Deferring to @ospencer on this one.

> What are (Number, Number) for the groups positions? Is it group number and string position?

They are the position offsets (start and end offset). Capture groups are not always fixed-size, so we need to return both.

> Why does the MatchResult retusn Arrays but the RegExp exported methods return Lists?

This behavior was taken from Racket, and I think it makes sense. When you say "give me all of the matches in this string", that is an operation which returns zero+ matches, and users will want to simply iterate over that list. If you say "give me all of the groups in this match", an array makes more sense because (a) you know precisely the number of groups which are defined in the regex and (b) you are more likely to want to do random access on that list than iterate over it. An array is more conducive to that scenario.

> I'm also wondering why we return an Option<MatchResult>? It seems like this record could always be provided, even if nothing matched, maybe it would need to have a field that indicated the success status of the match? I'm not sure.

I suppose it could, but I feel like it is nicer style-wise to use `Option`s here, since that allows users to use pattern-matching to determine if the match was successful. FWIW, this is what Haskell does, and OCaml sort of does the same thing (except they throw a `Not_found` exception instead of using an Option).

> I'm not sure I like the replace syntax. It feels like too much of a deviation from the languages people are going to be coming from.

Can you elaborate more on this? My impression is that `\n` is pretty standard across many languages, while the other pieces of syntax are relatively niche enough that I am not sure what the equivalents are in other languages.

---------

I personally feel that, while the API is a bit verbose, it leans into the `Option` and `Result` types to the appropriate extent (they exist for use cases like they appear in here).

ospencer:
> > As for the public API, I believe that we shouldn't have functions in a MatchResult that throw exceptions. We can just provide functions that return Option and users can use Option.unwrap if they know for a fact that those will always be Some.
>
> Deferring to @ospencer on this one.

In this case, I thought exceptions over options here were more ergonomic since you always knew what groups would match (unless you were compiling user-defined regexes) and it'd be annoying to have to unwrap this option. That gave us both the exception and non-exception versions. I _did_ forget that groups can be optional, i.e. `(.*(gg)?)`, and in this case sometimes asking for group 2 here would lead to an exception. With that in mind, I suppose it makes sense to simplify to only having the optional version.

ospencer:
>>I'm not sure I like the replace syntax. It feels like too much of a deviation from the languages people are going to be coming from.
>
>Can you elaborate more on this? My impression is that \n is pretty standard across many languages, while the other pieces of syntax are relatively niche enough that I am not sure what the equivalents are in other languages.

Given that Perl has the world's premier regex engine (lmfao) I think @phated has a point here. Perl (and JS) use `$` instead of `\`, so replacement strings look like `$2, $1` instead of `\2, \1`. Python and Ruby however use the `\` form. I guess the difference is that Ruby has single quote strings that let you write `'\1'` without issue, and Python has regex strings like`r'\1'`, versus in Grain your _only_ choice is to always escape them `"\\1"`. I have mixed feelings about it—on the one hand, I like `\` better since it's 1-1 with backreferences. On the other hand, a lot of folks will come from JS land and Perl regex is very popular, plus there wouldn't be any need to have to escape. 🤔

peblair:
@ospencer @phated I've updated the module to contain JS-like replace syntax and to use the proper Graindoc syntax. Here is the updated doc for `replace`:

```grain
replace : (RegularExpression, String, String) -> String
```

Replaces the first match for the given regular expression contained within the given string with the given replacement.
Replacement strings support the following syntax:
- `$&` - Replaced with the text of the matching portion of input (e.g. for `(foo)`, the search string `foo bar`, and the replacement `baz $&`, the result will be `baz foo bar`)
- `$n` / `$nn` (where `n` is a digit) - Replaced with the text of group `nn`
- `$$` - Replaced with a literal `$`
- `$.` - Does nothing (this exists to support replacement strings such as `$4$.0`, which will place the contents of group 4 prior to a zero)
- `$\`` - Replaced with the text of the string prior to the matched subspan of text
- `$'` - Replaced with the text of the string after the matched subspan of text
- Any other character will be placed as-is in the replaced output.

Parameters:

|param|type|description|
|-----|----|-----------|
|`rx`|RegularExpression|RegularExpression - The regular expression to search for|
|`toSearch`|String|String - The string to search within|
|`replacement`|String|String - The string to replace matches with|

Returns:

|type|description|
|----|-----------|
|String|The `replacement` string with the appropriate replacement, if any|

stdlib/regex.gr
```diff
@@ -0,0 +1,2995 @@
...
+=========================
+REGEX PARSING DEFINITIONS
+=========================
+
+*/
+
+// Range parsing ("[a-z]")
+
+let rangeLimit = 0x10FFFF
+//let rangeLimit = 255
```
ospencer, line 507: What's up with this?

stdlib/regex.gr
```diff
@@ -0,0 +1,2995 @@
...
+    }
+  }
+}
+
+
+//
+//
+//============
+//REGEX SEARCH
+//============
```
ospencer, line 2625:
```suggestion
// ============
// REGEX SEARCH
// ============
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2995 @@
...
+
+record MatchBuf {
+  matchInput: String,
+  matchInputExploded: Array<Char>,
+}
+
+let makeMatchBuffer = (s) => {
+  {
+    matchInput: s,
+    matchInputExploded: String.explode(s),
```
ospencer, line 1836: This is still my only major worry, but I don't think it needs to be refactored right now (especially since it'd be a pretty big refactor).

ospencer (commented):
I have a couple of tiny things, but other than those I think this is solid enough to go in.

stdlib/regex.gr
peblair: @ospencer I apologize, can you remind me what your worry is here?

stdlib/regex.gr
peblair: Forgot to convert this to a TODO. Fixed.

stdlib/regex.gr
ospencer: It's that this is inefficient and may be problematic for larger string sizes since the memory requirement is 28x character length of string (24 bytes for each Char and 4 bytes for each array slot) + computation time. https://github.com/grain-lang/grain/issues/769 is possibly a fix here? I'm not sure what Racket does, but maybe it treats multiple-byte characters as just a sequence to match against. Gives you the byte offsets instead of the character offsets though—not sure how expensive it would be to convert them. But then there'd be no memory overhead and no conversion cost! Otherwise, I think converting to UTF-32 instead of exploding might be better. That'd be 4x character length of string with constant time indexing. 

Regardless, we've got some options for what we can do here (that don't need to happen right now). If you don't think #769 is a fix, could you make an issue for it?

ospencer (approved):
`/^(?=approved\!)/`

ospencer:
Though it does irk me to see those warnings in the test output. Could you send a PR that just disables that warning for now until we resolve that issue? I think it's #190 but I'm not sure.

stdlib/regex.gr
```diff
@@ -0,0 +1,2995 @@
...
+export record MatchResult {
+  /**
+   * Returns the contents of the given match group, raising
+   * an exception if that group was not matched.
+   */
+  group: Number -> String,
+  /**
+   * Returns the contents of the given group
+   */
+  groupOptional: Number -> Option<String>,
```
phated, line 2692:
```suggestion
  /**
   * Returns the contents of the given group
   */
  group: Number -> Option<String>,
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2995 @@
...
+  /**
+   * Returns the position of the given match group, raising
+   * an exception if that group was not matched.
+   */
+  groupPosition: Number -> (Number, Number),
+  /**
+   * Returns the position of the given group
+   */
+  groupPositionOptional: Number -> Option<(Number, Number)>,
```
phated, line 2701:
```suggestion
  /**
   * Returns the position of the given group
   */
  groupPosition: Number -> Option<(Number, Number)>,
```

phated: @ospencer You agreed on these right?

ospencer: Yeah, I agreed.

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+      }, state)))
+    }
+  }
+}
+
+/**
+ * Returns true if the given regular expression has a match in the given string
+ * @param rx: RegularExpression - The regular expression to search for
+ * @param string: String - The string to search within
+ * @returns Whether `rx` matches `string`
```
phated, line 2795:
```suggestion
 * Determines if the given regular expression has a match in the given string.
 * @param rx: The regular expression to search for
 * @param string: The string to search within
 * @returns `true` if the RegExp matches the string, otherwise `false`
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+  fastDriveRegexIsMatch(rx, string, 0, String.length(string))
+}
+
+/**
+ * Returns true if the given regular expression has a match in the given string within the given start/end offsets
+ * @param rx: RegularExpression - The regular expression to search for
+ * @param string: String - The string to search within
+ * @param start: Number - The start offset to search within
+ * @param end: Number - The end offset to search within
+ * @returns Whether `rx` matches `string` in the given range
```
phated, line 2807:
```suggestion
 * Determines if the given regular expression has a match in the given string between the given start/end offsets.
 * @param rx: The regular expression to search for
 * @param string: The string to search
 * @param start: The start offset to search between
 * @param end: The end offset to search between
 * @returns `true` if the RegExp matches the string in the given range, otherwise `false`
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+ */
+export let isMatchRange = (rx: RegularExpression, string: String, start: Number, end: Number) => {
+  fastDriveRegexIsMatch(rx, string, start, end)
+}
+
+/**
+ * Returns the first match for the given regular expression contained within the given string
+ * @param rx: RegularExpression - The regular expression to search for
+ * @param string: String - The string to search within
+ * @returns The match result, if any
```
phated, line 2817:
```suggestion
 * Returns the first match for the given regular expression contained within the given string.
 * @param rx: The regular expression to search for
 * @param string: The string to search
 * @returns The match result, if any
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+}
+
+/**
+ * Returns the first match for the given regular expression contained within the given string,
+ * within the given start/end range.
+ * @param rx: RegularExpression - The regular expression to search for
+ * @param string: String - The string to search within
+ * @param start: Number - The start offset to search within
+ * @param end: Number - The end offset to search within
+ * @returns The match result, if any
```
phated, line 2830:
```suggestion
 * Returns the first match for the given regular expression contained within the given string
 * between the given start/end range.
 * @param rx: The regular expression to search for
 * @param string: The string to search
 * @param start: The start offset to search between
 * @param end: The end offset to search between
 * @returns The match result, if any
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+ */
+export let findRange = (rx: RegularExpression, string: String, start: Number, end: Number) => {
+  fastDriveRegexMatch(rx, string, start, end)
+}
+
+/**
+ * Returns the all matches for the given regular expression contained within the given string
+ * @param rx: RegularExpression - The regular expression to search for
+ * @param string: String - The string to search within
+ * @returns The list of results
```
phated, line 2840:
```suggestion
 * Returns all matches for the given regular expression contained within the given string.
 * @param rx: The regular expression to search for
 * @param string: The string to search
 * @returns The list of matches
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+}
+
+/**
+ * Returns all matches for the given regular expression contained within the given string,
+ * within the given start/end range.
+ * @param rx: RegularExpression - The regular expression to search for
+ * @param string: String - The string to search within
+ * @param start: Number - The start offset to search within
+ * @param end: Number - The end offset to search within
+ * @returns The list of results
```
phated, line 2853:
```suggestion
 * Returns all matches for the given regular expression contained within the given string
 * between the given start/end range.
 * @param rx: The regular expression to search for
 * @param string: The string to search
 * @param start: The start offset to search between
 * @param end: The end offset to search between
 * @returns The list of matches
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+        String.concat(String.slice(searchPos, start, toSearch),
+          String.concat(computeReplacement(buf, replacement, start, end, state),
+            if (all) recur(start, end) else String.slice(end, String.length(toSearch), toSearch))),
+    }
+  }
+  loop(0)
+}
+
+/**
+ * Replaces the first match for the given regular expression contained within the given string with the given replacement.
```
phated, line 2953:
```suggestion
 * Replaces the first match for the given regular expression contained within the given string with the specified replacement.
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+ * Replacement strings support the following syntax:
+ * - `$&` - Replaced with the text of the matching portion of input (e.g. for `(foo)`, the search string `foo bar`, and the replacement `baz $&`, the result will be `baz foo bar`)
```
phated, line 2955: I'm not sure I understand this...

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+ * - `$n` / `$nn` (where `n` is a digit) - Replaced with the text of group `nn`
```
phated, line 2956: why do we have `nn`? is that a double digit number? Do we not support triple digit?

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+ * - `$$` - Replaced with a literal `$`
```
phated, line 2957: Why do we escape `$` with `$` instead of `\`?

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+ * - `$.` - Does nothing (this exists to support replacement strings such as `$4$.0`, which will place the contents of group 4 prior to a zero)
+ * - `$\`` - Replaced with the text of the string prior to the matched subspan of text
```
phated, line 2959: I don't understand this either... maybe I need to read the tests

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+ * - `$'` - Replaced with the text of the string after the matched subspan of text
```
phated, line 2960: Same

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+ * - Any other character will be placed as-is in the replaced output.
+ *
+ * @param rx: RegularExpression - The regular expression to search for
+ * @param toSearch: String - The string to search within
+ * @param replacement: String - The string to replace matches with
+ * @returns The `replacement` string with the appropriate replacement, if any
```
phated, line 2966:
```suggestion
 * @param rx: The regular expression to search for
 * @param toSearch: The string to search
 * @param replacement: The string that replaces matches
 * @returns The given string with the appropriate replacements, if any
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+ */
+export let replace = (rx: RegularExpression, toSearch: String, replacement: String) => {
+  regexReplaceHelp(rx, toSearch, replacement, false)
+}
+
+/**
+ * Replaces all matches for the given regular expression contained within the given string with the given replacement.
```
phated, line 2973:
```suggestion
 * Replaces all matches for the given regular expression contained within the given string with the specified replacement.
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+ * See `replace` for replacement string syntax.
+ *
+ * @param rx: RegularExpression - The regular expression to search for
+ * @param toSearch: String - The string to search within
+ * @param replacement: String - The string to replace matches with
+ * @returns The `replacement` string with the appropriate replacements, if any
```
phated, line 2979:
```suggestion
 * @param rx: The regular expression to search for
 * @param toSearch: The string to search
 * @param replacement: The string that replaces matches
 * @returns The input string with the appropriate replacements, if any
```

stdlib/regex.gr
phated: Just to be clear, this returns the `toSearch` string with the first match replaced by the replacement?

stdlib/regex.gr
phated: Just to be clear, this returns the `toSearch` string with all matches replaced by the replacement?

phated (commented):
Holy cow, batman! Most of my comments are surface things because I barely followed the regexp parser/matcher logic.

I think there are still a few things we need to bikeshed (and I really need to review the tests still)

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
+/**
+ * @module Regex: Regular Expressions.
+ * @example import Regex from "regex"
```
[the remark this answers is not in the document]
peblair: No opinion here. Deferring to @ospencer

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+ */
+
+/*
+ * This library provides support for regular expressions in Grain.
+ * Its parser and analyzer are largely ported from Racket (https://racket-lang.org/),
+ * which is licensed under Apache 2.0. Racket's regular expression
+ * engine is itself inspired by the Spencer engine, as found in Tcl.
+ */
```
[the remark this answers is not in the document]
peblair: It doesn't particularly bother me with the `*`s, but we can remove them if you think it makes things more clear.

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+}
+
+let configWithCaseSensitive = (config: RegExParserConfig, caseSensitive: Bool) => {
+  {
+    isPerlRegExp: config.isPerlRegExp,
+    caseSensitive: caseSensitive,
+    multiline: config.multiline,
+    groupNumber: config.groupNumber,
+    references: config.references,
+  }
```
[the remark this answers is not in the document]
peblair: We do not have record spread, but we could pun here, yes.

stdlib/regex.gr
peblair: I hear you, but do we expect that we will be adding other types of errors to this module? I honestly expect this to not really evolve after it is merged (except for the byte-regexp matching, which would not introduce any new error types).

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+    }
+  }
+},
+
+parseClass = (buf: RegExBuf) => {
+  if (!more(buf)) {
+    Err("no chars") // caught in handler (we use a Result to cleanly mesh with the Result type below)
+  } else {
+    match(peek(buf)) {
+      Err(e) => Err(e),
```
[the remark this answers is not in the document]
peblair: You can see below (in `parseRangeRest`) that we actually ignore what's in the error altogether and just replace it with a generic "invalid class" error. More importantly, this `Err` from `peek` should never happen given that `more(buf)` must be true inside of this `if` branch. This could probably be streamlined into just the `match(peek(buf))` (without the enclosing `if`), but it looked just like this as a result of the port from Racket.

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+  let processState = ((cs, ml)) => {
+    let withCs = match(cs) {
+      None => buf.config,
+      Some(true) => configWithCaseSensitive(buf.config, true),
+      Some(_) => configWithCaseSensitive(buf.config, false),
+    }
+    match(ml) {
+      None => withCs,
+      Some(true) => configWithMultiLine(buf.config, true),
+      Some(_) => configWithMultiLine(buf.config, false),
```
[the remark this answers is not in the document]
peblair: Oops. Yes.

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+    Err(parseErr(buf, "Expected unicode category", 0))
+  } else {
+    match(peek(buf)) {
+      Err(e) => Err(e),
+      Ok('{') => {
+        ignore(eat(buf, '{'))
+        let catNegated = if (peek(buf) == Ok('^')) {
+          ignore(eat(buf, '^'))
+          true
+        } else false
```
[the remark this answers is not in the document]
peblair: 🙃

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+    REConditional(_, rx1, rx2, _, _, _) => isAnchored(rx1) && Option.mapWithDefault(isAnchored, false, rx2),
+    REGroup(rx, _) => isAnchored(rx),
+    RECut(rx, _, _, _) => isAnchored(rx),
+    _ => false,
+  }
+}
+
+// must-string:
+
+let rec somethingExpensive = (re: ParsedRegularExpression) => {
```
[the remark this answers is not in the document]
peblair: Correct, I ended up not needing this. I'll remove it.

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+  MightBeEmpty,
+  DoesNotMatchBounded,
+  BackreferenceTooBig,
+  InternalError(ParsedRegularExpression),
+}
+
+let rec validate = (re: ParsedRegularExpression, numGroups) => {
+  let groupSizes = Map.make()
+  let mut dependsSizes = Map.make()
+  let mut mustSizes = Map.make()
```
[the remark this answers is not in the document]
peblair: Yes. You can see below where they are swapped out for alternative maps during the validation process.

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+        }, (0, 0, 0), elts)
+      },
+      REGroup(re, n) => {
+        let (min1, max1, maxL1) = loop(re)
+        Map.set(n, min1, groupSizes)
+        (min1, max1, maxL1)
+      },
+      RERepeat(re, repeatMin, repeatMax, nonGreedy) => {
+        let oldDependsSizes = dependsSizes
+        dependsSizes = Map.make()
```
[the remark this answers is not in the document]
peblair: I'm not sure I follow? We are still using the old maps below, so clearing them would destroy the contents...

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+          thrownError = Some(MightBeEmpty)
+        }
+      }
+    }
+  }, mustSizes)
+  match(thrownError) {
+    Some(MightBeEmpty) => Err("`*`, `+`, or `{...}` operand could be empty"),
+    Some(DoesNotMatchBounded) => Err("lookbehind pattern does not match a bounded length"),
+    Some(BackreferenceTooBig) => Err("backreference number is larger than the highest-numbered cluster"),
+    Some(InternalError(re)) => Err("regex validate: Internal error: " ++ toString(re)),
```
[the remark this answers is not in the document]
peblair: I suppose they could be. I am just not sure it is warranted to make the API so verbose by including all of these variants...at the end of the day, it's a parse error, and I cannot think of a realistic use case in which someone is going to want to conditionally branch on whether `BackreferenceTooBig` was thrown...

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+
+let makeMatchBuffer = (s) => {
+  {
+    matchInput: s,
+    matchInputExploded: String.explode(s),
+  }
+}
+
+let matchBufMore = (buf: MatchBuf, pos: Number) => {
+  pos < Array.length(buf.matchInputExploded)
```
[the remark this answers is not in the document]
peblair: Sure :+1:

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+
+let rStack = [SEPositionProducer(pos => Some(pos))]
+
+let arrayCopy = (dest, destStart, src, srcStart, srcEnd) => {
+  let mut count = srcStart
+  while (count < srcEnd) {
+    dest[destStart + (count - srcStart)] = src[count]
+    count = count + 1
+  }
+}
```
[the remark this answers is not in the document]
peblair: I don't see it in there. I didn't add it because I wasn't sure if we wanted this precise API included in `Array`. Let's leave that to a separate PR.

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+      if ((pos + len <= limit) && subArraysEqual(buf.matchInputExploded, refStart, buf.matchInputExploded, pos, len)) {
+        next_m(buf, pos + len, start, limit, end, state, stack)
+      } else None
+    }
+  }
+}
+
+let referenceMatcher = makeReferenceMatcher(((a, b)) => (a == b))
+
+let asciiCharToLower = (c) => {
```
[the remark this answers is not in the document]
peblair: Ultimately, yes, but let's move it in a separate PR. Additionally, I'd want prefer if that version was unicode-aware, which is much more involved than this function (though I suppose we could merge this version in and make it unicode-aware later).

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+      REEnd => endMatcher(next_m),
+      RELineStart => lineStartMatcher(next_m),
+      RELineEnd => lineEndMatcher(next_m),
+      REWordBoundary => wordBoundaryMatcher(next_m),
+      RENotWordBoundary => notWordBoundaryMatcher(next_m),
+      RESequence(res, _) => {
+        List.reduceRight(compile, next_m, res)
+      },
+      REAlts(re1, re2) => altsMatcher(compile(re1, next_m), compile(re2, next_m)),
+      REMaybe(re, nonGreedy) when nonGreedy => altsMatcher(next_m, compile(re, next_m)),
```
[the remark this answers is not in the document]
peblair:
```suggestion
      REMaybe(re, true) => altsMatcher(next_m, compile(re, next_m)), // non-greedy
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+              } else {
+                repeatMatcher(r_m, min, max, next_m)
+              }
+            }
+          }
+        }
+      },
+      REGroup(re, n) => groupPushMatcher(n, compile(re, groupSetMatcher(n, next_m))),
+      REReference(n, _) when n == 0 => neverMatcher,
+      REReference(n, caseSensitive) when caseSensitive => referenceMatcher(n - 1, next_m),
```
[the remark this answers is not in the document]
peblair:
```suggestion
      REReference(n, true) when caseSensitive => referenceMatcher(n - 1, next_m), // case-sensitive
```

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+}
+
+/**
+ * @section Values: Functions for working with the regular expressions.
+ */
+
+/**
+ * Compiles the given pattern string into a regular expression object.
+ *
+ * For an overview of the theory regular expressions in general, readers are referred
```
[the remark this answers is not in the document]
peblair: There is a typo in my version. What I meant to say was this:
```suggestion
 * For an overview of the theory of regular expressions in general, readers are referred
```
Alternatively,
```suggestion
 * For an overview of the general theory of regular expressions, readers are referred
```
@phated Do either of those sound okay to you?

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+ * to "Mastering Regular Expressions" by Friedl, or numerous alternative resources online.
```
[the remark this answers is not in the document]
peblair:
```suggestion
 * to ["Mastering Regular Expressions"](http://regex.info/book.html) by Friedl, or numerous alternative resources online.
```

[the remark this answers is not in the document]
peblair: > Additionally, one of Grain's goals is to teach harder concepts to people, so linking out might not be akin to that?

I hear what you're saying. We do provide an overview below this, which does provide guidance for people who are not familiar with the concepts. I personally do not feel that anything more in-detail is appropriate for this doc-comment, but maybe instead in a separate tutorial. Maybe we could write something later and replace this with a link to that once we have it?

stdlib/regex.gr
peblair: I'm not sure how to clarify 😓...the example explains it as clearly as I can think to explain it...

stdlib/regex.gr
peblair: Correct.

stdlib/regex.gr
peblair: Correct.

stdlib/regex.gr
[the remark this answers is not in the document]
ospencer: I am a `Regex` fan!

stdlib/regex.gr
ospencer: I think some slight word changes might make it more clear (and tweaking the example):
```suggestion
 * - `$&` - Replaced with the matched substring (e.g. for the regex `\d+`, the search string `clouds: 99`, and the replacement `$& clouds`, the result will be `99 clouds`)
```

@phated does this make sense?

stdlib/regex.gr
ospencer: I don't know any regex engine that supports 3 numbers. That'd be quite the thicc regex!

stdlib/regex.gr
ospencer: The #1 reason would be to match other languages that use this, but beyond that it's for consistency/uncomplicated parsing. Using a different character would really make this a parsing nightmare.

stdlib/regex.gr
ospencer: I think the MDN doc here is solid, and we can do something similar. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_a_parameter

```suggestion
 * - `$\`` - Replaced with the text preceding the matched substring
```

stdlib/regex.gr
ospencer:
```suggestion
 * - `$'` - Replaced with the text following the matched substring
```

stdlib/regex.gr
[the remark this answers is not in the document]
phated: 😭 maybe one day. The issue has been opened so resolving.

stdlib/regex.gr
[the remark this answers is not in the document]
phated: I'd love to see the `match(peek(buf))` changes land in this PR, but I also understand if you want to put the cleanup off until later since it is a port.

stdlib/regex.gr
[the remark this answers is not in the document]
phated: Yeah, my branch was fried by line 1674 😛

stdlib/regex.gr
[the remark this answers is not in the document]
phated: `when caseSensitive` won't be defined here in the guard, right?

stdlib/regex.gr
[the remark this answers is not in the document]
phated: I like the first one 😄

stdlib/regex.gr
phated: Yep! Also, I just learned something about JS regexps too 🙃

stdlib/regex.gr
phated: Sounds good. I notice the MDN docs say "where n under 100" and I think our docs are better 👍

stdlib/regex.gr
phated: Makes sense! TIL about JS regexp replace too!

marcusroberts (approved):
LGTM!

stdlib/regex.gr
```diff
@@ -0,0 +1,2985 @@
...
+}
+
+/**
+ * The user-facing object which contains the results
+ * of a regular expression match.
+ */
+export record MatchResult {
+  /**
+   * Returns the contents of the given group
+   */
```
phated, line 2688: Remember that we lose these comments in the doc generator.

stdlib/regex.gr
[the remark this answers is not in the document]
phated: Please make an issue!

stdlib/regex.gr
peblair: Yep, manually writing them in now

stdlib/regex.gr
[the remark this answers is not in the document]
phated: Please make an issue 🙏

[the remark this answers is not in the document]
peblair: @phated What should the issue be? We need to discuss whether we want to have the API and what the design should be, so it wouldn't be appropriate to have an issue saying to implement it.

stdlib/regex.gr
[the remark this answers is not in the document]
peblair: Resolving per discussion in discord/standup

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+    }
+  }) next_m(buf, pos + 1, start, limit, end, state, stack) else None
+}
+
+let charTailMatcher = (toMatch) => (buf: MatchBuf, pos: Number, start: Number, limit: Number, end: Number, state, stack) => {
+  if ({
+    pos < limit && match(matchBufChar(buf, pos)) {
+      Err(_) => false,
+      Ok(c) => toMatch == c
+    }
```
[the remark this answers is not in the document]
peblair: To avoid a proliferation of helper functions (the only clean way to do it that I can see will require allocating a bunch of lambdas, and this code path is hot enough that I think the speedup of `match` warrants the sacrificed cleanliness), I am going to resolve this as-is.

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+}
+
+// match byte in set (range)
+
+let rangeMatcher = (rng, next_m) => (buf: MatchBuf, pos: Number, start: Number, limit: Number, end: Number, state, stack) => {
+  if ({
+    pos < limit && match(matchBufChar(buf, pos)) {
+      Err(_) => false,
+      Ok(c) => rangeContains(rng, Char.code(c))
+    }
```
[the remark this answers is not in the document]
peblair: See above

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+  }) next_m(buf, pos + 1, start, limit, end, state, stack) else None
+}
+
+let rangeTailMatcher = (rng) => (buf: MatchBuf, pos: Number, start: Number, limit: Number, end: Number, state, stack) => {
+  if ({
+    pos < limit && match(matchBufChar(buf, pos)) {
+      Err(_) => false,
+      Ok(c) => rangeContains(rng, Char.code(c))
+    }
```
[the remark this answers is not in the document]
peblair: See above

stdlib/regex.gr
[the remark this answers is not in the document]
phated: How about "Implement Array.fillSlice"? That seems to be aligned with our Buffer API naming

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+  }
+  bloop(pos2, n)
+}
+
+let lazyRepeatMatcher = (r_m, min, max, next_m) => (buf: MatchBuf, pos: Number, start: Number, limit: Number, end: Number, state, stack) => {
+  let rec rloop = (pos, n, min) => {
+    if (n < min) {
+      let newStack = [SEPositionProducer(pos => rloop(pos, n + 1, min)), ...stack]
+      r_m(buf, pos, start, limit, end, state, newStack)
+    } else if (match(max) { None => false, Some(max) => max == n }) {
```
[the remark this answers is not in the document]
peblair: To avoid the lambda allocation in this hot code path, let's leave it as-is.

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+}
+
+let lazyRepeatSimpleMatcher = (r_m, min, max, next_m) => (buf: MatchBuf, pos: Number, start: Number, limit: Number, end: Number, state, stack) => {
+  let rec rloop = (pos, n, min) => {
+    if (n < min) {
+      match(r_m(buf, pos, start, limit, end, state, stack)) {
+        Some(p) => rloop(p, n + 1, min),
+        None => None
+      }
+    } else if (match(max) { None => false, Some(max) => max == n }) {
```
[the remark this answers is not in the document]
peblair: See above

stdlib/regex.gr
ospencer: Mind adding the issue number to the comment?

stdlib/regex.gr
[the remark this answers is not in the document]
peblair: I'm going to renege on this. (1) It's a constant-time operation with minimal additional overhead, (2) storing it in a separate field introduces more possibilities to make a mistake (the length should be tightly coupled with the array), and (3) it wouldn't yield much gain in terms of readability/conciseness.

stdlib/regex.gr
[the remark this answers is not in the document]
peblair: Again, my point is that the question of "should we do this" is one which we should discuss, rather than directly opening an issue saying "Implement this thing". For example, I don't immediately see how `fillSlice` is in line with buffer's naming. I would say that `move` makes more sense, since `Bytes.move` does essentially the same thing as this function.

stdlib/regex.gr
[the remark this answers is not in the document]
peblair: #900

stdlib/regex.gr
```diff
@@ -0,0 +1,2983 @@
...
+                ignore(eat(buf, '('))
+                ignore(eat(buf, '?'))
+                ignore(eat(buf, '>'))
+                let preNumGroups = unbox(buf.config.groupNumber)
+                match(parseRegex(buf)) {
+                  Err(e) => Err(e),
+                  Ok(rx) => {
+                    let postNumGroups = unbox(buf.config.groupNumber)
+                    ignore(eat(buf, ')'))
+                    match(eat(buf, ')')) {
```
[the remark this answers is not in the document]
peblair: fixed

stdlib/regex.gr
peblair: Resolving per offline discussion

stdlib/regex.gr
peblair: #901

Issue #388, opened 2020-10-15
[Lang] Named (Optional) arguments


phated:
This would open up some amazing APIs! I really like the `~something` syntax, but I'm open to suggestions.

ospencer:
I'm cool with `~name` for named arguments, but I've never been a huge fan of the `(~x=?)` syntax in function definitions. I'd wanna do some thinking around that.

alex-snezhko:
Would it be possible to allow all positional arguments to be used as named arguments? For example:
```
let f = (x, y, z) => ...

f(1, 2, 3) // using all positional
<-> f(z=3, y=2, x=1) // all named
<-> f(y=2, 1, 3) // allow mixture; since y is named then 1, 3 correspond to x, z (the missing args)

// possible syntax sugar
let x = 1, y = 2, z = 3
f(~z, ~y, ~x) == f(z=z, y=y, x=x)
```
This may or may not cause problems depending on how currying is done, although I don't think it should be a big issue unless I'm missing something
```
// let cf = x => y => z => ...
let cf = curry f

cf(1)(2)(3) == cf(x=1)(y=2)(z=3)
// cf(y=2)(...) not allowed since x is the first curried argument
```
Then for optional arguments maybe the OCaml-like `?x` syntax could be used?
```
let f1 = (x, y, ?z) => ... // z=None by default
let f2 = (x, y, ?z=3) => ... // z=3 by default

f1(1, 2, z=Some(3))
f2(1, 2) == f2(1, 2, z=3)
```
Currying for this would be a bit trickier, although perhaps the curried function would somehow be able to maintain the list of optional arguments to be applied at any point before the function gets called
```
let cf = curry f2
cf(1)(2) // function gets called with z=3
cf(z=4)(1)(2) == cf(1)(z=4)(2) // optional argument can be given at any point before all non-optional arguments are given
cf(1)(2)(z=3) // illegal as cf(1)(2) would complete the argument list
```
Not sure how a curried function with all optional arguments would be handled though...

ospencer:
I'm actually working on labeled arguments right now 🙂 

I've thought about this! Where it gets a little tricky is making the whole system make sense in respect to required labeled arguments. Without required labeled arguments, it's all fine _and_ your type signatures are actually a bit nicer, since the inferred type of a function like `let join = (list, sep) => ...` would be `(list: List<String>, sep: String) -> String`, you can call the function with positional arguments or name them if you like, e.g. `join(sep="|", [])`.

[...]

PR #683, opened 2021-05-31
feat(stdlib): Add String.encode and String.decode functions to standard library

peblair:
This pull request adds two new functions to the `String` module in the standard library:

```
export enum Encoding {
  UTF8,
  UTF16_BE,
  UTF16_LE,
  UTF32_BE,
  UTF32_LE,
}

// Encodes the given string using the given encoding scheme
// @param s: String - The input string
// @param encoding: Encoding - The encoding to use
// @param dest: Bytes - The bytes object to write the encoded output into
// @param destPos: Number - The location in the byte array to write the output
// @returns Bytes - Returns `dest`
export let encodeInto = (s, encoding, dest, destPos)

// Encodes the given string using the given encoding scheme
// @param s: String - The input string
// @param encoding: Encoding - The encoding to use
// @param dest: Bytes - The bytes object to write the encoded output into
// @param destPos: Number - The location in the byte array to write the output
// @returns Bytes - Returns `dest`
export let encodeIntoWithBom = (s, encoding, dest, destPos)

// Encodes the given string using the given encoding scheme. A byte-order marker
// will not be included in the output.
// @param s: String - The input string
// @param encoding: Encoding - The encoding to use
// @returns Bytes
export let encode = (s: String, encoding: Encoding)

// Encodes the given string using the given encoding scheme, including a byte-order marker
// @param s: String - The input string
// @param encoding: Encoding - The encoding to use
// @returns Bytes
export let encodeWithBom = (s: String, encoding: Encoding)

// -----

// Decodes the given byte sequence into a string using the given encoding scheme, skipping
// the byte-order marker, if it's present.
// @param bytes: Bytes - The input bytes
// @param encoding: Encoding - The encoding to use
// @param start: Number - The byte offset to begin decoding from
// @param size: Number - The maximum number of bytes to decode
// @returns String
export let decodeRange = (bytes: Bytes, encoding: Encoding, start: Number, size: Number)

// Decodes the given byte sequence into a string using the given encoding scheme, including
// the byte-order marker, if it's present.
// @param bytes: Bytes - The input bytes
// @param encoding: Encoding - The encoding to use
// @param start: Number - The byte offset to begin decoding from
// @param size: Number - The maximum number of bytes to decode
// @returns String
export let decodeRangeKeepBom = (bytes: Bytes, encoding: Encoding, start: Number, size: Number)

// Decodes the given byte sequence into a string using the given encoding scheme,
// skipping the byte-order marker, if it's present.
// @param bytes: Bytes - The input bytes
// @param encoding: Encoding - The encoding to use
// @returns String
export let decode = (bytes: Bytes, encoding: Encoding)

// Decodes the given byte sequence into a string using the given encoding scheme,
// including the byte-order marker, if it's present
// @param bytes: Bytes - The input bytes
// @param encoding: Encoding - The encoding to use
// @returns String
export let decodeKeepBom = (bytes: Bytes, encoding: Encoding)
```
These help enable use cases such as those described by @cician in #661, in which users need to marshal Grain strings into nonstandard encodings.


stdlib/string.gr
```diff
@@ -519,3 +519,596 @@ export let endsWith = (p: String, s: String) => {
...
+      2n
+    }
+    if (n == 4n) {
+      size += 2n
+    } else {
+      size += 1n
+    }
+    ptr += n
+  }
+  // multiply by two to get number of bytes (equivalent to left-shift)
```
ospencer, line 569: I read this and expected to see multiplication instead of a shift, and then I saw a shift and it made me think that the code was originally a multiplication and then it was changed. Regardless I think I wouldn't have batted an eye if it didn't have this last part 😂 
```suggestion
  // multiply by two to get number of bytes
```

stdlib/string.gr
```diff
@@ -519,3 +519,596 @@ export let endsWith = (p: String, s: String) => {
...
+    UTF16_BE => utf16Length(s),
+    UTF16_LE => utf16Length(s),
+    UTF8 => byteLength(s),
+  }
+}
+
+exception MalformedUnicode
+
+@disableGC
+let getCodePoint = (ptr: WasmI32) => {
```
ospencer, line 586: This already exists in the Char module. I wonder if it's worth trying to share them.

stdlib/string.gr
```diff
@@ -519,3 +519,596 @@ export let endsWith = (p: String, s: String) => {
...
+// @param includeBom: Bool - Whether to include the byte-order marker in the encoded output
+// @returns Bytes
+@disableGC
+export let encode = (s: String, encoding: Encoding, includeBom: Bool) => {
+  let (>>>) = WasmI32.shrU
+  let (-) = WasmI32.sub
+  let (&) = WasmI32.and
+  let (<) = WasmI32.ltU
+  let (<=) = WasmI32.leU
+  let (==) = WasmI32.eq
```
ospencer, line 660: Could we move them down next to the add?

stdlib/string.gr
```diff
@@ -519,3 +519,596 @@ export let endsWith = (p: String, s: String) => {
...
+        WasmI32.store8(bytes + bytesIdx + 3n, 0n, _BYTES_OFFSET)
+        bytesIdx += 4n
+      }
+    }
+  }
+
+  while (ptr < end) {
+    let byte = WasmI32.load8U(ptr, 0n)
+    // number of bytes in Grain UTF-8 string
+    // ptr+ offset + 8n
```
ospencer, line 722: I got pretty confused trying to understand what this comment was trying to say. I think this is more clear:
```suggestion
    // number of bytes spanning this UTF-8-encoded scalar value
```

stdlib/string.gr
```diff
@@ -519,3 +519,596 @@ export let endsWith = (p: String, s: String) => {
...
+        bytesIdx += 4n
+      },
+    };
+    ptr += n
+  }
+  WasmI32.toGrain(bytes): Bytes
+}
+
+// Byte->String decoding and helper functions:
+import Char from "char"
```
ospencer, line 798: It doesn't look like this is used (and would be best if String did not depend on Char)
```suggestion
```

stdlib/string.gr
```diff
@@ -519,3 +519,596 @@ export let endsWith = (p: String, s: String) => {
...
+
+@disableGC
+let writeUtf8CodePoint = (ptr, codePoint) => {
```
ospencer, line 801: This one also more or less exists in Char and could maybe have a shared utility.

ospencer (changes requested):
Largely looks good. Curious to hear your thoughts on the shared utilities.

stdlib/string.gr
peblair: Sorry, this half-included some commented out copy-paste stuff.

stdlib/string.gr
peblair: Yes, this was for debugging at one point.

stdlib/string.gr
peblair: Per discussion offline, deferring to #684.

stdlib/string.gr
peblair: Per discussion offline, deferring to #684.

stdlib/string.gr
peblair: Fixed.

cician:
May a suggest that it would be useful to have a version of encode that writes to an existing instance of Bytes into a specific offset?

I'm currently using Bytes as a big buffer where I append multiple strings and have to create a temporary Bytes with Bytes.fromString for each write operation, only to then copy it over into my big buffer with Bytes.move.

Also I think it would be useful to add offset and length parameters to the decode function for things like reading multiple strings from structured binary formats.

peblair:
@cician I am not opposed (in principle, at least), but there is a question about what the API would look like:

Option 1 would be to add two new functions which do the read/write based on the existing buffer. The design question here is what these variants would be named.

Option 2 would be to wait until #388 is resolved, and then have a buffer/offset be optional arguments to the existing functions. I lean towards this option, but, of course, this means waiting until #388 is resolved.

@ospencer, what are your thoughts?

ospencer:
I'd say go ahead and add them in now, as we'll end up refactoring the whole stdlib once we've got optional args.

I'd call them `encodeBuffered` and `decodeBuffered`.

cician:
For encode i think it would make sense to just have a separate function like this:
``` grain
export let encodeInto = (s: String, encoding: Encoding, includeBom: Bool, dstPos: Number, dst: Bytes)
```

For decode optional parameters would be nice, though maybe a bit strange if the user could pass one, but not the other.
``` grain
export let decodeRange = (bytes: Bytes, encoding: Encoding, skipBom: Bool, srcPos: Number, srcLen: Number)
```
Or encodeBuffered/decodeBuffered I'm not picky about names.

peblair:
I've added `encodeInto` and `decodeRange`. cc @ospencer for re-review and feedback on naming.

stdlib/string.gr
```diff
@@ -519,3 +519,654 @@ export let endsWith = (p: String, s: String) => {
...
+  }
+}
+
+// Decodes the given byte sequence into a string using the given encoding scheme
+// @param bytes: Bytes - The input bytes
+// @param encoding: Encoding - The encoding to use
+// @param skipBom: Bool - Whether to include the byte-order marker (if present) in the decoded output
+// @returns String
+@disableGC
+export let decode = (bytes: Bytes, encoding: Encoding, skipBom: Bool) => {
```
phated, line 1168: Since we don't have optional arguments yet, I believe we should have separate methods for `withBOM`.

This would become `decode` and `decodeWithBOM` that passes that argument into the "help" function for `decodeRange`.

phated (changes requested):
I'll leave the implementation review to @ospencer since I'm very dumb when it comes to the ptr logic; however, I do think we need to split up the APIs a bit since we don't support optional arguments yet.

peblair:
@phated Everything should be addressed now. I've updated the original post with the present user-facing API.

peblair:
cc @ospencer on the updated API ☝️

phated (approved):
Awesome! Thanks for updating the API 🎊

ospencer (approved):
I'm still not a fan of the `encodeInto` name, because to me it's confusing what "into" means. To me it _really_ sounds like `encode` has some default behavior and `encodeInto` lets you pick what encoding you want to encode the string into, not having anything to do with offsets within a `Bytes` instance. I'm not going to block on it though—once #710 is merged, we should refactor `encode` to write into a Buffer. Then `encodeInto` isn't needed and we can just delete it.

peblair:
@ospencer Would `encodeAt` be better in your opinion? I am not sure I 100% agree with your proposal re: #710 (in terms of how it will change the APIs added here...I think that's a discussion for later, though), so I would feel better if we merged in a name you're happy with.

ospencer (approved):
We chatted and largely aligned on what the future API would look like. In the meanwhile, we're gonna go with `encodeAt`. This PR has my blessing once that change is done 👼 (approving now for those beautiful European summer mornings 🌞)

peblair:
:eu:

PR #698, opened 2021-06-04
feat(stdlib): Add Array.reverse() function

technosophos:
This adds an `Array.reverse()` function that takes a source array and returns a new array with the elements reversed. It also adds a test in the compiler tests.

This helps bring `Array` and `List` closer to parity.

I am not sure where to put the documentation for this function.

Signed-off-by: Matt Butcher <matt.butcher@microsoft.com>

### DIFF
--- a/compiler/test/stdlib/array.test.gr
+++ b/compiler/test/stdlib/array.test.gr
@@ -94,6 +94,11 @@ Array.forEachi(addToSum, arr1)
 
 assert sum == 8
 
+// Array.reverse
+
+assert Array.reverse([> ]) == [> ] // Empty
+assert Array.reverse([> 1, 2, 3, 4]) == [> 4, 3, 2, 1]
+
 // Array.toList
 
 assert Array.toList([>]) == []
--- a/compiler/test/stdlib/array.test.gr
+++ b/compiler/test/stdlib/array.test.gr
@@ -97,7 +97,9 @@ assert sum == 8
 // Array.reverse
 
 assert Array.reverse([> ]) == [> ] // Empty
-assert Array.reverse([> 1, 2, 3, 4]) == [> 4, 3, 2, 1]
+let testArr = [> 1, 2, 3, 4]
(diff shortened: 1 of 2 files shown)


ospencer (approved):
Looks great! There's maybe a question of if this should happen in place, but I think we'll probably want both anyway. I think @phated might have ideas about this (since no one seems to like my idea of following the Ruby tradition of bang names, i.e. `Array.reverse!()` for functions that mutate the argument 😂).

Also, no need to worry about the test failure—that's fixed in #700. You may just need to rebase to bring in that change.

compiler/test/stdlib/array.test.gr
```diff
@@ -94,6 +94,11 @@ Array.forEachi(addToSum, arr1)
 
 assert sum == 8
 
+// Array.reverse
+
+assert Array.reverse([> ]) == [> ] // Empty
+assert Array.reverse([> 1, 2, 3, 4]) == [> 4, 3, 2, 1]
```
peblair, line 100: Based on @ospencer's review, I might suggest an explicit test for this not being in-place; just so we can make sure that behavior doesn't change in the future
```suggestion
let testArr = [> 1, 2, 3, 4]
assert Array.reverse(testArr) == [> 4, 3, 2, 1] // array is reversed properly
assert testArr == [> 1, 2, 3, 4] // original array is not modified
```

peblair (changes requested):
Thanks for adding this! The function itself looks great, but I think we should go ahead and check that we don't modify the original array in the test suite. Outside of that, this looks good to go!

compiler/test/stdlib/array.test.gr
technosophos: Good idea.

technosophos:
I was never a huge Ruby fan, but I did like the concise nature of the `!` for modify-in-place. That said, I'm cool with whatever convention you adopt.

I can also verify that after rebasing with #700, I no longer get the failed test.

phated (approved):
Hooray! 🎉

PR #700, opened 2021-06-05
fix: Properly initialize malloc free list

ospencer:
This PR fixes a pesky bug that only appears under absolutely perfect™️ conditions.

Currently, if the free list only contains one block and it's exactly the same size as the amount of requested memory, it will not be released from the free list. As such, further requests for that same size will result in the same pointer being allocated over and over again.

We hadn't noticed this because the stars really have to align to hit this case! In the case of [this run](https://github.com/grain-lang/grain/pull/698/checks?check_run_id=2750176540), it was the addition of a closure that happened to be just the right size that later calls to malloc would be perfectly sized to trigger the bug.

I reviewed the reference implementation and realized this was because we weren't setting up the free list correctly, though I don't fault us for it because it's one of the few changes that needed to be made for the C -> Wasm port. I also added a regression test for this.

### DIFF
--- a/compiler/test/input/mallocTight.gr
+++ b/compiler/test/input/mallocTight.gr
@@ -1,0 +1,30 @@
+/* grainc-flags --compilation-mode=runtime */
+
+import Malloc from "runtime/malloc"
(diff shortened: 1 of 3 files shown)


phated (approved):
🔥

peblair (approved):
Nice find 👏

PR #701, opened 2021-06-05
fix(compiler): Record function-call associativity

ospencer:


compiler/src/parsing/parser.dyp
```diff
@@ -485,6 +483,14 @@ non_assign_expr :
   | array_get   { $1 }
   | array_expr  { $1 }
 
+left_accessor_expr :
+  | app_expr    { $1 }
+  | simple_expr { $1 }
+  /*| array_get   { $1 }*/
```
ospencer, line 489: This doesn't work and I don't understand why it doesn't work. Uncommenting this line makes the parser give really cryptic error messages. I think it's fine to move forward without this for now (since it's not a super common use case) and if this PR is approved, I'll make an issue for this and reference it here.

peblair (changes requested):
Can we add a test case to catch any regressions?

ospencer:
@peblair I added tests, were you looking for something else?

peblair:
Oh, I apologize. I somehow missed that when reading the changes.

peblair (approved):
Thanks for fixing this!

phated (approved):
🔥

ospencer:
Issue for the array function call support is #709.

PR #705, opened 2021-06-06
fix(stdlib): return proper values from getClockTime for Sys/Time

cician:
The getClockTime function was overwriting the number tag and thus returning invalid instances of Int64.


ospencer (approved):
Definitely a bug! I made an issue to figure out a way to test this library, as even a smoke test probably wouldn't have caught this.

peblair (approved):

PR #710, opened 2021-06-08
feat(stdlib): Add buffer module

### LINKED ISSUE #707: Add `Buffer` module to stdlib
From Discord conversation in `#development`. We want this library to be à la https://ocaml.org/api/Buffer.html.

jozanza:
Closes #707 

More detailed description to come soon!

### DIFF
--- a/stdlib/buffer.gr
+++ b/stdlib/buffer.gr
@@ -429,7 +429,7 @@ export let clear = (buffer) => {
  * 
  * This operation resizes the underlying byte sequence to the initial size of the buffer.
  * 
- * @param buffer: The buffer to mutate
+ * @param buffer: The buffer to reset
  * 
  * @since v0.4.0
  */
--- a/stdlib/buffer.gr
+++ b/stdlib/buffer.gr
@@ -387,7 +387,7 @@ export let addFloat64 = (value, buffer) => {
  * That byte sequence is automatically reallocated when more than `initialSize` bytes are stored in the buffer, but shrinks back to `initialSize` characters when reset is called.
  * 
  * @param initialSize: The initial size of the buffer
- * @returns A buffer
+ * @returns The new buffer
  * 
  * @since v0.4.0
  */
--- a/stdlib/buffer.gr
+++ b/stdlib/buffer.gr
@@ -1,7 +1,7 @@
 /**
  * @module Buffer: Utilities for working with buffers.
  *
- * Buffers are an automatically growing data structure useful for storing and operating on an unknown number of bytes. When working with buffers, all set or append operations will be mutating.
+ * Buffers are data structures that automatically expand as more data is appended. They are useful for storing and operating on an unknown number of bytes. All set or append operations mutate the buffer.
  * @example import Buffer from "buffer"
  *
  * @since v0.4.0
--- a/stdlib/buffer.gr
+++ b/stdlib/buffer.gr
@@ -225,7 +225,7 @@ export let getInt32 = (index, buffer) => {
 /**
  * Sets a signed 32-bit integer starting at the given byte index.
  * 
- * @param index: The byte index to access
+ * @param index: The byte index to update
  * @param value: The value to set
  * @param buffer: The buffer to mutate
  * 
--- a/stdlib/buffer.gr
+++ b/stdlib/buffer.gr
@@ -381,7 +381,7 @@ export let addFloat64 = (value, buffer) => {
 }
 
 /**
- * Returns a fresh buffer, initially empty.
+ * Creates a fresh buffer, initially empty.
  * 
  * The `initialSize` parameter is the initial size of the internal byte sequence that holds the buffer contents.
  * That byte sequence is automatically reallocated when more than `initialSize` bytes are stored in the buffer, but shrinks back to `initialSize` characters when reset is called.
--- a/stdlib/buffer.gr
+++ b/stdlib/buffer.gr
@@ -36,11 +36,11 @@ let _32BIT_LEN = 4
 
 let _64BIT_LEN = 8
 
-/** Gets the size of a Bytes via its ptr */
+/* Gets the size of a Bytes via its ptr */
 @disableGC
 let getSize = (ptr) => WasmI32.load(ptr, _SIZE_OFFSET)
 
-/** Doubles the size of buffer's underlying byte sequence, if the given size is larger than the size of a buffer's underlying byte sequence */
+/* Doubles the size of buffer's underlying byte sequence, if the given size is larger than the size of a buffer's underlying byte sequence */
 let autogrow = (len, buf) => {
   while (buf.len + len > Bytes.length(buf.data)) {
     let mut n = Bytes.length(buf.data)
@@ -49,14 +49,14 @@ let autogrow = (len, buf) => {
   }
 }
 
-/** Gets the pointer for a char's bytes following the char's tag */
+/* Gets the pointer for a char's bytes following the char's tag */
 @disableGC
 let rec getCharAsWasmI32 = (char) => {
   let c = WasmI32.fromGrain(char)
   WasmI32.load(c, 4n)
 }
 
-/** Gets the UTF-8 byte length of a char */
+/* Gets the UTF-8 byte length of a char */
 @disableGC
 let rec getCharByteLength = (byte) => {
   let (+) = WasmI32.add
@@ -73,14 +73,14 @@ let rec getCharByteLength = (byte) => {
   }
 }
 
-/** Memcopies bytes from a source byte sequence to a destination byte sequence via pointers */
+/* Memcopies bytes from a source byte sequence to a destination byte sequence via pointers */
 @disableGC
 let rec appendBytes = (srcOff, dstOff, len, src, dst) => {
   let (+) = WasmI32.add
   Memory.copy(dst + _VALUE_OFFSET + dstOff, src + _VALUE_OFFSET + srcOff, len)
 }
 
-/** Throws if a given index + len is not within the bounds of a buffer's current length */
+/* Throws if a given index + len is not within the bounds of a buffer's current length */
 let checkIndexIsInBounds = (i, len, buf) => {
   if (i < 0) throw Exception.IndexOutOfBounds
   if (i >= buf.len || i + len > buf.len) throw Exception.IndexOutOfBounds
--- a/stdlib/buffer.gr
+++ b/stdlib/buffer.gr
@@ -16,9 +16,6 @@ import Bytes from "bytes"
 import String from "string"
 import { coerceNumberToWasmI32 } from "runtime/numbers"
 
-/**
- * Buffers are an automatically growing data structure useful for storing and operating on a unknown number of bytes.
- */
 record Buffer {
   mut len: Number,
   initialSize: Number,
--- a/stdlib/buffer.gr
+++ b/stdlib/buffer.gr
@@ -454,7 +454,7 @@ export let truncate = (length, buffer) => {
 /**
  * Returns a copy of the current contents of the buffer as a byte sequence.
  * 
- * @param buffer: The buffer to stringify
+ * @param buffer: The buffer to copy into a byte sequence
  * @returns A string made from copied buffer data
  * 
  * @since v0.4.0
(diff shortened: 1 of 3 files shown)


stdlib/buffer.gr
```diff
@@ -0,0 +1,463 @@
...
+  checkIndexIsInBounds(buf.size, _8BIT_LEN, buf)
+  setInt8(buf.size, value, buf)
+}
+
+// 16-bit
+// -----------------------------------------------------------------------------
+
+let _16BIT_LEN = 2
+
+export let getInt16S = (i, buf) => {
```
peblair, line 116: I'd love to see support for both little-endian and big-endian ints in the `Buffer` module (perhaps requires support from the `Bytes` module).

stdlib/buffer.gr
```diff
@@ -0,0 +1,463 @@
...
+ * @param char: Char - the char to end
+ * @param buf: Buffer - The buffer
+ * @returns Void
+ */
+export let addChar = (char, buf) => {
+  // TODO: Should I just @disableGC memcopy stuff in this function to improve perf 👀
+  let n = getCharAsWasmI32(char)
+  let byteLen = getCharByteLength(n)
+  let len = match (byteLen) {
+    1n => {
```
peblair, line 330: @ospencer your thoughts on whether this should depend on `String.encode` (once it lands) or maybe vice versa?

ospencer: It's probably pretty close to the time that we figure out how we want to handle shared utilities in the stdlib. I'd like if they could both pull the same utility so neither has to depend on the other.

cician:
I'm excited for this hope to rebase my JSON stuff on top of it soon.

If think the autogrow function needs a correction. It should allocate new Bytes of size at least equal to `buf.size + len` and all the functions that write to the buffer should call autogrow instead of grow() directly. If I understand correctly, the OCaml's buffer allocates to nearest power of two that is greater than new required length.

Also, maybe we want a cap on exponential growth in wasm32 environment?

stdlib/buffer.gr
peblair: That's fine, but it will likely be a breaking change. Additionally, we should document in this module the endianness of the operations in this PR, even if we don't support both.

stdlib/buffer.gr
```diff
@@ -0,0 +1,575 @@
...
+  let (+) = WasmI32.add
+  Memory.copy(dst + _VALUE_OFFSET + dstOff, src + _VALUE_OFFSET + srcOff, len)
+}
+
+/**
+ * Gets a signed 8-bit integer starting at the given byte index.
+ * @param i: Number - The byte index
+ * @param b: Buffer - The buffer
+ * @returns Int32
+ */
```
phated, line 104: Need to update these. The syntax for docblocks is currently documented at https://grain-lang.org/blog/2021/08/27/tooling-preview-graindoc/

Here's the gist. I'd also prefer external facing arguments not be 1-letter. We may need to rework my verbiage.
```suggestion
/**
 * Gets a signed 8-bit integer starting at the given byte index.
 *
 * @param index: The byte index to access
 * @param buffer: The buffer to access
 * @returns The 8-bit integer that starts at the given index
 *
 * @since v0.4.0
 */
```

stdlib/buffer.gr
```diff
@@ -0,0 +1,680 @@
...
+import WasmI32 from "runtime/unsafe/wasmi32"
+import Conv from "runtime/unsafe/conv"
+import Exception from "runtime/exception"
+import Int32 from "int32"
+import Bytes from "bytes"
+import String from "string"
+import { coerceNumberToWasmI32 } from "runtime/numbers"
+
+/**
+ * @section Types: Type declarations included in the Buffer module.
```
ospencer, line 19: Just a quick note—we might not want this section since the record type isn't exported.

peblair (commented):
This PR looks great! One change I would request before we merge: for the private helper functions, there's not really a reason to prefix them with `_` (since they aren't exported), and it isn't in line with the rest of the stdlib's style. Could we maybe remove those?

stdlib/buffer.gr
```diff
@@ -0,0 +1,742 @@
...
+
+let _8BIT_LEN = 1
+
+let _16BIT_LEN = 2
+
+let _32BIT_LEN = 4
+
+let _64BIT_LEN = 8
+
+/** Gets the size of a Bytes via its ptr */
```
phated, line 39:
```suggestion
/* Gets the size of a Bytes via its ptr */
```
Shouldn't (I hope) be an issue for graindoc, but `/**` is actually the indicator to parse as a docblock instead of a multiline comment. Switch to `/*` to parse as the multiline comment.

stdlib/buffer.gr
```diff
@@ -0,0 +1,742 @@
...
+@disableGC
+let getSize = (ptr) => WasmI32.load(ptr, _SIZE_OFFSET)
+
+/** Doubles the size of buffer's underlying byte sequence, if the given size is larger than the size of a buffer's underlying byte sequence */
```
phated, line 43:
```suggestion
/* Doubles the size of buffer's underlying byte sequence, if the given size is larger than the size of a buffer's underlying byte sequence */
```

stdlib/buffer.gr
```diff
@@ -0,0 +1,742 @@
...
+let autogrow = (len, buf) => {
+  while (buf.len + len > Bytes.length(buf.data)) {
+    let mut n = Bytes.length(buf.data)
+    if (n == 0) n = 4 // Make sure bytes of 0 length grow too
+    buf.data = Bytes.resize(0, n, buf.data)
+  }
+}
+
+/** Gets the pointer for a char's bytes following the char's tag */
```
phated, line 52:
```suggestion
/* Gets the pointer for a char's bytes following the char's tag */
```

stdlib/buffer.gr
```diff
@@ -0,0 +1,742 @@
...
+@disableGC
+let rec getCharAsWasmI32 = (char) => {
+  let c = WasmI32.fromGrain(char)
+  WasmI32.load(c, 4n)
+}
+
+/** Gets the UTF-8 byte length of a char */
```
phated, line 59:
```suggestion
/* Gets the UTF-8 byte length of a char */
```

stdlib/buffer.gr
```diff
@@ -0,0 +1,742 @@
...
+  } else if ((byte & 0xF0n) == 0xF0n) {
+    4n
+  } else if ((byte & 0xE0n) == 0xE0n) {
+    3n
+  } else {
+    2n
+  }
+}
+
+/** Memcopies bytes from a source byte sequence to a destination byte sequence via pointers */
```
phated, line 76:
```suggestion
/* Memcopies bytes from a source byte sequence to a destination byte sequence via pointers */
```

stdlib/buffer.gr
```diff
@@ -0,0 +1,742 @@
...
+@disableGC
+let rec appendBytes = (srcOff, dstOff, len, src, dst) => {
+  let (+) = WasmI32.add
+  Memory.copy(dst + _VALUE_OFFSET + dstOff, src + _VALUE_OFFSET + srcOff, len)
+}
+
+/** Throws if a given index + len is not within the bounds of a buffer's current length */
```
phated, line 83:
```suggestion
/* Throws if a given index + len is not within the bounds of a buffer's current length */
```

stdlib/buffer.gr
```diff
@@ -0,0 +1,742 @@
...
+ * @param index: The byte index to access
+ * @param buffer: The buffer to access
+ * @returns A 32-bit integer representing a signed 8-bit integer that starts at the given index
+ * 
+ * @since v0.4.0
+ */
+export let getInt8S = (i, b) => {
+  checkIndexIsInBounds(i, _8BIT_LEN, b)
+  Bytes.getInt8S(i, b.data)
+}
```
phated, line 105: Graindoc needs to match up the names you document with the parameters of your functions, so you need to update the signatures and implementations throughout this file.
```suggestion
export let getInt8S = (index, buffer) => {
  checkIndexIsInBounds(index, _8BIT_LEN, buffer)
  Bytes.getInt8S(index, buffer.data)
}
```
You should try running the `grain doc` command on this to catch this sort of thing 🔥

phated (changes requested):
Going to submit this "request changes" right now but then keep reviewing. This is because the graindoc tool won't generate on the code as it currently is and the functions will need to be updated.

stdlib/buffer.gr
```diff
@@ -0,0 +1,744 @@
...
+  void
+}
+
+/**
+ * Appends the bytes of a source buffer to destination buffer.
+ * 
+ * The source buffer is not mutated by this operation. The destination buffer, however, is mutated.
+ * 
+ * @param srcBuffer: The buffer to append
+ * @param dstBuffer: The buffer to append to
```
phated, line 727:
```suggestion
 * @param dstBuffer: The buffer to mutate
```
Not sure about this one. Thoughts?

stdlib/buffer.gr
```diff
@@ -0,0 +1,744 @@
...
+
+/**
+ * Appends the bytes of a part of a buffer to the end of the buffer
+ * 
+ * The source buffer is not mutated by this operation. The destination buffer, however, is mutated.
+ * 
+ * @param start: The byte offset into the buffer
+ * @param length: The number of bytes to append
+ * @param srcBuffer: The buffer to append
+ * @param dstBuffer: The buffer to append to
```
phated, line 743:
```suggestion
 * @param dstBuffer: The buffer to mutate
```
Same as above, whatever the choice

phated (changes requested):
This is looking 🌶️ - I have a few more docs nits and 1 implementation question/issue with truncate

stdlib/buffer.gr
```diff
@@ -0,0 +1,744 @@
...
+ * 
+ * @param length: The number of bytes to truncate the buffer to
+ * @param buffer: The buffer to mutate
+ * 
+ * @since v0.4.0
+ */
+export let truncate = (length, buffer) => {
+  if (length < 0 || length > buffer.len) throw Exception.IndexOutOfBounds
+  buffer.len = length
+}
```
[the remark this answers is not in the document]
jozanza: I was using the ocaml implementation of this method as a reference. Looks like they just leave the underlying bytes as-is.

stdlib/buffer.gr
```diff
@@ -0,0 +1,744 @@
...
+
+/**
+ * Returns a copy of the current contents of the buffer as a byte sequence.
+ * 
+ * @param buffer: The buffer to stringify
```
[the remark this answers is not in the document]
jozanza: A good catch. Just bad copy-pasta

stdlib/buffer.gr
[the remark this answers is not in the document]
phated: Ping @peblair and @ospencer to weigh in. It feels like a security risk to leave a Bytes around with initialized data.

stdlib/buffer.gr
```diff
@@ -0,0 +1,747 @@
...
+
+/*
+ * If given index is negative, returns a positive index.
+ * Throws if a given index + len is not within the bounds of a buffer's current length
+ */
+let getSafeIndex = (index, len, buf) => {
+  let i = index % buf.len
+  if (i >= buf.len || i + len > buf.len) throw Exception.IndexOutOfBounds
+  i
+}
```
phated, line 89: Ping @ospencer and @peblair - are we good/happy to apply the logic of "a negative index is the index from the end" to Buffers?

phated: @jozanza depending on the answer here, we may need some negative index tests.

ospencer: Yes, that's correct.

stdlib/buffer.gr
```diff
@@ -0,0 +1,760 @@
+/**
+ * @module Buffer: Utilities for working with buffers.
+ *
+ * Buffers are an automatically growing data structure useful for storing and operating on an unknown number of bytes. When working with buffers, all set or append operations will be mutating.
```
ospencer, line 4: Wording suggestion:
```suggestion
 * Buffers are data structures that automatically expand as more data is appended. They are useful for storing and operating on an unknown number of bytes. All set or append operations mutate the buffer.
```

ospencer (approved):
Looks really straightforward! Just one suggestion on some wording.

stdlib/buffer.gr
phated: We decided that negative indexing doesn't make sense for Buffer because the data isn't constant size, like an array or list.

phated (approved):
:approved: ‼️

Amazing work @jozanza - thanks for bearing with the iteration! 🎉

jozanza:
Thanks for the reviews, guys. Big shout out to @peblair for putting up with me while explaining the new calling convention 😂

PR #713, opened 2021-06-08
feat (stdlib): coerceNumberToWasmI32

cician:
This PR adds a simple function for Number to WasmI32 conversion that just wraps an existing one. The new function adds a bit of type safety instead of just exposing coerceInt32. It also replaces the duplicate function in Bytes module, which is replicated in the Buffer module currently proposed in PR #710 as well.

@ospencer briefly mentioned on Discord the possibility of just exporting coerceInt32 directly instead of a wrapper function, so if that is preferable I don't mind changing it.

### DIFF
--- a/stdlib/bytes.gr
+++ b/stdlib/bytes.gr
@@ -7,6 +7,7 @@ import Conv from "runtime/unsafe/conv"
 import { tagSimpleNumber, allocateBytes, allocateString } from "runtime/dataStructures"
 import Exception from "runtime/exception"
 import Int32 from "int32"
+import { coerceNumberToWasmI32 as numberToWasmI32 } from "runtime/numbers"
 
 let _SIZE_OFFSET = 4n;
 let _VALUE_OFFSET = 8n;
@@ -16,21 +17,6 @@ let _INT32_BYTE_SIZE = 4n;
 let _FLOAT32_BYTE_SIZE = 4n;
 let _INT64_BYTE_SIZE = 8n;
 let _FLOAT64_BYTE_SIZE = 8n;
-
-/** Converts Number to WasmI32 */
-@disableGC
-let numberToWasmI32 = n => {
-  let m = WasmI32.fromGrain(n)
-  if (WasmI32.eqz(WasmI32.and(m, 1n))) {
-    let i32 = Int32.fromNumber(n)
-    let res = Conv.fromInt32(i32)
-    Memory.free(WasmI32.fromGrain(i32))
-    res
-  } else {
-    // fast path
-    WasmI32.shrS(m, 1n)
-  }
-}
 
 /** Throws an exception if the index specified is out-of-bounds */
 @disableGC
--- a/stdlib/bytes.gr
+++ b/stdlib/bytes.gr
@@ -7,7 +7,7 @@ import Conv from "runtime/unsafe/conv"
 import { tagSimpleNumber, allocateBytes, allocateString } from "runtime/dataStructures"
 import Exception from "runtime/exception"
 import Int32 from "int32"
-import { coerceNumberToWasmI32 as numberToWasmI32 } from "runtime/numbers"
+import { coerceNumberToWasmI32 } from "runtime/numbers"
 
 let _SIZE_OFFSET = 4n;
 let _VALUE_OFFSET = 8n;
@@ -47,7 +47,7 @@ export let setInt8 = (i: Number, v: Int32, b: Bytes) => {
   let (+) = WasmI32.add
   let ptr = WasmI32.fromGrain(b)
   let size = getSize(ptr)
-  let offset = numberToWasmI32(i)
+  let offset = coerceNumberToWasmI32(i)
   checkIndexIsInBounds(offset, _INT8_BYTE_SIZE, size)
   let value = Conv.fromInt32(v)
   WasmI32.store8(ptr + offset, value, _VALUE_OFFSET)
@@ -64,7 +64,7 @@ export let getInt8S = (i: Number, b: Bytes) => {
   let (+) = WasmI32.add
   let ptr = WasmI32.fromGrain(b)
   let size = getSize(ptr)
-  let offset = numberToWasmI32(i)
+  let offset = coerceNumberToWasmI32(i)
   checkIndexIsInBounds(offset, _INT8_BYTE_SIZE, size)
   let n = WasmI32.load8S(ptr + offset, _VALUE_OFFSET)
   Conv.toInt32(n)
@@ -81,7 +81,7 @@ export let getInt8U = (i: Number, b: Bytes) => {
   let (+) = WasmI32.add
   let ptr = WasmI32.fromGrain(b)
   let size = getSize(ptr)
-  let offset = numberToWasmI32(i)
+  let offset = coerceNumberToWasmI32(i)
   checkIndexIsInBounds(offset, _INT8_BYTE_SIZE, size)
   let n = WasmI32.load8U(ptr + offset, _VALUE_OFFSET)
   Conv.toInt32(n)
@@ -99,7 +99,7 @@ export let setInt16 = (i: Number, v: Int32, b: Bytes) => {
   let (+) = WasmI32.add
   let ptr = WasmI32.fromGrain(b)
   let size = getSize(ptr)
-  let offset = numberToWasmI32(i)
+  let offset = coerceNumberToWasmI32(i)
   checkIndexIsInBounds(offset, _INT16_BYTE_SIZE, size)
   let value = Conv.fromInt32(v)
   WasmI32.store16(ptr + offset, value, _VALUE_OFFSET)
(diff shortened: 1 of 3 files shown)


ospencer (commented):
@cician Thanks for the thought leadership on this! I thought about this for a bit, and I think the right solution here would be to rename `coerceInt32` and friends to `coerceNumberToWasmI32` and change the input type of these functions to be `Number` instead of `WasmI32`. The call sites in this file will need to be updated to be `coerceNumberToWasmI32(WasmI32.toGrain(x) : Number)` (which is admittedly verbose, but I think that's okay for this file).

That'll let us have these nicely-typed functions that can be exported elsewhere (like Bytes) and also make the use in this module more clear. What do you think?

cician:
My pleasure.

I've made the suggested adjustments.

stdlib/runtime/numbers.gr
```diff
@@ -1240,17 +1241,17 @@ export let coerceNumberToInt64 = (x: Number) => {
...
+    let tag = boxedNumberTag(y)
     if (WasmI32.eq(tag, Tags._GRAIN_RATIONAL_BOXED_NUM_TAG)) {
-      x
+      y
     } else if (WasmI32.eq(tag, Tags._GRAIN_INT32_BOXED_NUM_TAG)) {
-      newRational(boxedInt32Number(x), 1n)
+      newRational(boxedInt32Number(y), 1n)
     } else if (WasmI32.eq(tag, Tags._GRAIN_INT64_BOXED_NUM_TAG)) {
-      newRational(coerceInt32(x), 1n)
+      newRational(coerceNumberToWasmI32(x), 1n)
```
ospencer, line 1254: I think it would have been fine to leave the variable name as `x` and cast `x` back to a Number here:
```suggestion
      newRational(coerceNumberToWasmI32(WasmI32.toGrain(x) : Number, 1n)
```

ospencer (commented):
Looks good! Can we do the same for `coerceFloat32`, `coerceFloat64`, and `coerceInt64`?

stdlib/runtime/numbers.gr
cician: Changed back to x as suggested.

cician:
> Looks good! Can we do the same for `coerceFloat32`, `coerceFloat64`, and `coerceInt64`?

Sure. Would you prefer as part of this PR?

ps: Should I worry about each commit message in the PR or will they get quashed when merging? Sorry if I'm asking something obvious, but it's first time for me iterating on a pull request on github.

ospencer:
Yeah, that'd be great! And no need to worry about the messages, it'll get squashed :)

cician:
Note that unlike other of these functions, coerceFloat64 is already exported. Should I keep it for compatibility?

ospencer:
Nope! Let's get rid of it and we'll mark this is a breaking change.

cician:
> Nope! Let's get rid of it and we'll mark this as a breaking change.

Removed.

stdlib/number.gr
```diff
@@ -49,7 +49,7 @@ export let div = (/)
 @disableGC
 export let sqrt = (x: Number) => {
   let x = WasmI32.fromGrain(x)
-  let xval = coerceFloat64(x)
+  let xval = coerceNumberToWasmF64(WasmI32.toGrain(x) : Number)
```
phated, line 52: Is there a reason the `x` is shadowed here and then converted back into a Grain number?

stdlib/bytes.gr
```diff
@@ -7,6 +7,7 @@ import Conv from "runtime/unsafe/conv"
 import { tagSimpleNumber, allocateBytes, allocateString } from "runtime/dataStructures"
 import Exception from "runtime/exception"
 import Int32 from "int32"
+import { coerceNumberToWasmI32 as numberToWasmI32 } from "runtime/numbers"
```
phated, line 10: Can we just rename the `numberToWasmI32` to `coerceNumberToWasmI32` in this file?

stdlib/runtime/numbers.gr
```diff
@@ -701,14 +705,14 @@ let numberAddSubFloat32Help = (xval, y, isSub) => {
     let result = if (isSub) WasmF64.sub(xval, yval) else WasmF64.add(xval, yval)
     newFloat64(result)
   } else {
-    let yval = coerceFloat32(y)
+    let yval = coerceNumberToWasmF32(WasmI32.toGrain(y) : Number)
```
phated, line 708: This coercion back to a Grain number is a bit weird to me.

stdlib/runtime/numbers.gr
```diff
@@ -701,14 +705,14 @@ let numberAddSubFloat32Help = (xval, y, isSub) => {
...
     let result = if (isSub) WasmF32.sub(xval, yval) else WasmF32.add(xval, yval)
     newFloat32(result)
   }
 }
 
 let numberAddSubFloat64Help = (xval, y, isSub) => {
-  let yval = coerceFloat64(y)
+  let yval = coerceNumberToWasmF64(WasmI32.toGrain(y) : Number)
```
phated, line 715: This coercion back to a Grain number is a bit weird to me.

stdlib/runtime/numbers.gr
```diff
@@ -952,7 +956,7 @@ let numberTimesDivideRationalHelp = (x, y, isDivide) => {
 }
 
 let numberTimesDivideFloat64Help = (x, y, isDivide) => {
-  let yAsFloat = coerceFloat64(y)
+  let yAsFloat = coerceNumberToWasmF64(WasmI32.toGrain(y) : Number)
```
phated, line 959: This coercion back to a Grain number is a bit weird to me.

stdlib/runtime/numbers.gr
```diff
@@ -969,7 +973,7 @@ let numberTimesDivideFloat32Help = (x, y, isDivide) => {
       newFloat64(WasmF64.mul(WasmF64.promoteF32(x), boxedFloat64Number(y)))
     }
   } else {
-    let yAsFloat = coerceFloat32(y)
+    let yAsFloat = coerceNumberToWasmF32(WasmI32.toGrain(y) : Number)
```
phated, line 976: This coercion back to a Grain number is a bit weird to me.

stdlib/runtime/numbers.gr
```diff
@@ -1022,8 +1026,8 @@ let numberDivide = (x, y) => {
 let i64abs = (x) => if (WasmI64.geS(x, 0N)) x else WasmI64.sub(0N, x)
 
 let numberMod = (x, y) => {
-  let xval = coerceInt64(x)
-  let yval = coerceInt64(y)
+  let xval = coerceNumberToWasmI64(WasmI32.toGrain(x) : Number)
+  let yval = coerceNumberToWasmI64(WasmI32.toGrain(y) : Number)
```
phated, line 1030: This coercion back to a Grain number is a bit weird to me.

stdlib/runtime/numbers.gr
```diff
@@ -1070,14 +1070,14 @@ export let (>) = (x: Number, y: Number) => {
...
   // Equality is finicky, so delegate
-  let xval = coerceFloat64(x)
-  let yval = coerceFloat64(y)
+  let xval = coerceNumberToWasmF64(x)
+  let yval = coerceNumberToWasmF64(y)
   if (WasmF64.lt(xval, yval)) {
     true
   } else {
+    let x = WasmI32.fromGrain(x)
+    let y = WasmI32.fromGrain(y)
```
phated, line 1080: This is technically doing a `fromGrain` twice in the `else` case?

stdlib/runtime/numbers.gr
```diff
@@ -1088,14 +1088,14 @@ export let (<=) = (x: Number, y: Number) => {
...
   // Equality is finicky, so delegate
-  let xval = coerceFloat64(x)
-  let yval = coerceFloat64(y)
+  let xval = coerceNumberToWasmF64(x)
+  let yval = coerceNumberToWasmF64(y)
   if (WasmF64.gt(xval, yval)) {
     true
   } else {
+    let x = WasmI32.fromGrain(x)
+    let y = WasmI32.fromGrain(y)
```
phated, line 1098: This is technically doing a `fromGrain` twice in the `else` case?

phated (commented):
Some thoughts but they probably need @ospencer's responses.

stdlib/number.gr
ospencer: Yeah, in this case we could swap this line with the line above to avoid the need to cast back to `Number`.

stdlib/runtime/numbers.gr
ospencer: I asked them to do it like this because I thought it was better than having two separate values since `toGrain` and `fromGrain` are just no-ops.

stdlib/runtime/numbers.gr
ospencer: What do you mean?

phated: the coerce call is made at the top of this function, which does a `fromGrain` inside that function, then when we hit this else case, we are calling fromGrain again.

stdlib/bytes.gr
cician: Renamed.

stdlib/number.gr
cician: Swapped the lines and removed the cast.

phated (approved):
Since it seems @ospencer asked for the implementation this way, I'll just approve (resolved all my comments)

ospencer (approved):
Thanks @cician for bearing with us on this one!

cician:
> Thanks @cician for bearing with us on this one!

No problem. It has been an interesting learning experience.

PR #715, opened 2021-06-08
chore(tests): Add dune files to configure tests properly

phated:
My understanding is that `(include_subdirs unqualified)` is a bad stanza and we shouldn't be using it, so I dove into setting up the tests repository more correctly. This required adding dune files for the suites and utils and setting up the framework/runner library differently, but things seem to be working.


ospencer (approved):
Nice refactor!

peblair (approved):
👏 👏 👏

PR #717, opened 2021-06-09
chore(stdlib): Update clean command to delete new files

phated:
When the CLI changes landed, it looks like this didn't get updated. I'm unsure if I should have kept the mashtree glob.


ospencer (approved):
Mister. Clean.

peblair (approved):
😎

PR #718, opened 2021-06-09
chore(tests): Fix test name that created directory

phated:
The `//` in the name of this test caused a directory called `comment_lone_` to be created in the `test/output` directory and then it wrote files named **just** `.gr.wasm`, `.gr.wat`, and `.gr.modsig` into it.

This just renames the test and updates the snapshot so that doesn't happen.


ospencer (approved):
Solid, readable, less buggy, love it.

peblair (approved):

PR #711, opened 2021-06-08
chore(docs): Add contributor docs for commits, PRs & merging

phated:
First draft of core team docs for commits, pull requests, and merging. Please make suggestions!

Closes #685

PR #699, opened 2021-06-05
add Array.unzip and tests for zip and unzip

MihailITPlace:
Hello. I tried to solve the problem from issue #694. This PR adds Array.unzip function and some tests for Array.zip and Array.unzip.

PR #719, opened 2021-06-09
feat(stdlib): Add Array.zip function

phated:

I started with `Array.zip` in order to land #699

I attributed the commit to @peblair but I also made a change to explicitly guard on the array lengths since a longer `b` wouldn't error. I'm not sure how much I like this API being a "fail" API but I think it is fine for now.

I didn't add tests since they are added in #699

### DIFF
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -373,3 +373,14 @@ export let unique = (array) => {
     array
   )
 }
+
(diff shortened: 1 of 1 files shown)


ospencer (approved):

peblair (approved):
Created #730 as a follow-up on the `fail` behavior. This looks good to merge.

PR #722, opened 2021-06-09
feat(stdlib): Add List.join and Array.join functions

phated:

This originally was `String.join`, but I changed it to `List.join` since it is actually joining a List into a String. I added `Array.join` too. And then I regenerated the snapshots because they are crazy.


phated:
> I don't think we should block on it for now, but we should probably make an issue to rewrite these to do the join the "fast" way via low-level instructions.

#728

peblair (approved):

PR #723, opened 2021-06-09
chore(stdlib): Tweak Number library docs

ospencer:


stdlib/number.gr
```diff
@@ -114,7 +114,7 @@ export let trunc = (x: Number) => {
 }
 
 /**
- * Returns its operand rounded to its nearest integer. `.5` and above return `1`, `.4` and below return `0`
```
ospencer, line 117: I got rid of this end part because I think it adds more confusion than it adds clarity. It sort of implies that `round` is only meant to be used on numbers between 0 and 1, and gives an incorrect explanation for numbers greater than 1 or less than 0 (or between .4 and .5). I thought about trying to rephrase this to explain when it rounds up versus down, but even that language isn't helpful in the context of negative numbers. Ultimately, `round` returns the nearest whole number, and that's captured in the first part of the doc.

ospencer:
If this PR is accepted, I'll also update https://github.com/grain-lang/grain-lang.org/pull/223.

phated (approved):
👍

marcusroberts (approved):

PR #726, opened 2021-06-10
fix!: Adjust Number equality, thus that 5 == 5.0

phated:

(I'm pretty sure this commit name will need to be changed)

I split out the change for Number equality and labeled it as a breaking fix. However, when writing tests for this, I noticed that `5 == 5.0` is false but `5.0 == 5` is true. Need someone's help to fix that, but it highlights the importance of writing tests for all our changes.

### DIFF
--- a/stdlib/runtime/equal.gr
+++ b/stdlib/runtime/equal.gr
@@ -17,6 +17,7 @@ import Tags from "runtime/unsafe/tags"
 
 primitive (!) : Bool -> Bool = "@not"
 primitive (||) : (Bool, Bool) -> Bool = "@or"
+primitive (&&) : (Bool, Bool) -> Bool = "@and"
 
 import { isNumber, numberEqual } from "runtime/numbers"
 
@@ -186,12 +187,12 @@ let rec heapEqualHelp = (heapTag, xptr, yptr) => {
     }
   }
 }, equalHelp = (x, y) => {
-  if ((x & Tags._GRAIN_GENERIC_TAG_MASK) != 0n) {
+  if (((x & Tags._GRAIN_GENERIC_TAG_MASK) != 0n) && ((y & Tags._GRAIN_GENERIC_TAG_MASK) != 0n)) {
     // Short circuit for non-pointer values
     x == y
   } else if (isNumber(x)) {
     // Numbers have special equality rules, e.g. NaN != NaN
-    numberEqual(x, y)
+    isNumber(y) && numberEqual(x, y)
   } else {
     // Handle all other heap allocated things
     // Can short circuit if pointers are the same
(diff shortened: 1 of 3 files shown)


ospencer:
@phated Adjusted your title since this affects the Number type rather than the Float32 type.

peblair:
Note: IIRC, this change was originally meant to facilitate f32 `NaN` equality (which begs the question why it's not broken in the `NaN` PR 🙂 ). The int/f32 equality was a good catch, though.

peblair:
@phated I've rebased this branch and fixed the lingering `equal` bug.

peblair (approved):

marcusroberts (approved):

stdlib/runtime/equal.gr
```diff
@@ -186,12 +187,12 @@ let rec heapEqualHelp = (heapTag, xptr, yptr) => {
...
   }
 }, equalHelp = (x, y) => {
-  if ((x & Tags._GRAIN_GENERIC_TAG_MASK) != 0n) {
+  if (((x & Tags._GRAIN_GENERIC_TAG_MASK) != 0n) && ((y & Tags._GRAIN_GENERIC_TAG_MASK) != 0n)) {
     // Short circuit for non-pointer values
     x == y
   } else if (isNumber(x)) {
     // Numbers have special equality rules, e.g. NaN != NaN
-    numberEqual(x, y)
+    isNumber(y) && numberEqual(x, y)
```
ospencer, line 195: This check shouldn't be necessary. What's up with it?

peblair: Forgot that this function wasn't exported. Fixed.

peblair:
The now-failing tests make me think we are probably calling `equal` from somewhere with two `WasmI32` values, which essentially subverts the type-safety of `equal`. @ospencer , maybe we should just add a check in `equalHelp` that its arguments are of the same type, just to harden against this? I imagine that, even if we find it, it will be easy to accidentally trigger again in the future...

ospencer:
It looks like it's here: https://github.com/grain-lang/grain/blob/66319ca8f3bbeb80e18525d1c15e2b84f0abd0c1/stdlib/string.gr#L1175

I've actually thought about this some... I really don't want to have that sort of check in there. 😕 Feels bad to have the typechecker and then have to do that kind of check. As an alternative though, I've thought about adding a warning to the typechecker that yells at you if you pass a value of type `WasmI32` to `(==)`, since that shouldn't be too hard to add. It doesn't guard against having some sort of wrapper function (e.g. `let (==) = (a, b) => a == b`) but it _does_ prevent us from having that bug in the code. WDYT?

ospencer:
https://github.com/grain-lang/grain/pull/735

peblair:
@ospencer `isNumber` added back in, with #746 opened as a tracking issue

ospencer (approved):

ospencer:
@peblair Thank you!

PR #727, opened 2021-06-10
feat(stdlib): Add Array.slice function

phated:

This adds `slice` to the Array stdlib. I actually rewrote the function because I believe our guiding principals on the stdlib is to be graceful where possible. This change is likely to be a bit controversial for the team so I'd like feedback if anyone doesn't think we should handle indexes too large or negative indexing from the end.

### DIFF
--- a/compiler/test/stdlib/array.test.gr
+++ b/compiler/test/stdlib/array.test.gr
@@ -269,3 +269,20 @@ assert Array.unique([>]) == [>]
 // Array.join
 assert Array.join(", ", [> "a", "b", "c"]) == "a, b, c"
 assert Array.join(", ", [>]) == ""
+
+// Array.slice
+let  testChars = [> 'a', 'b', 'c']
+
+assert Array.slice(0, 1, testChars) == [> 'a']
+assert Array.slice(1, Array.length(testChars), testChars) == [> 'b', 'c']
+assert Array.slice(0, 0, testChars) == [>]
+// Purposefully huge number
+assert Array.slice(1, 10000, testChars) == [> 'b', 'c']
+// Negative indexing
+assert Array.slice(1, -1, testChars) == [> 'b']
+assert Array.slice(-2, -1, testChars) == [> 'b']
+// Bad order
+assert Array.slice(2, 1, testChars) == [>]
+assert Array.slice(-1, -2, testChars) == [>]
+// Empty
+assert Array.slice(1, 4, [>]) == [>]
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -420,7 +420,7 @@ export let join = (separator: String, items: Array<String>) => {
  * at the end index will not be included in the result.
  *
  * If either index is a negative number, it will be treated as a reverse index from
- * the end of the array. e.g. slice(1, -1, [> 'a', 'b', 'c']) === 'b'].
+ * the end of the array. e.g. slice(1, -1, [> 'a', 'b', 'c']) == [> 'b'].
  *
  * @param startIndex: Number The index of the array where the slice will begin (inclusive)
  * @param endIndex: Number The index of the array where the slice will end (exclusive)
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -416,7 +416,7 @@ export let join = (separator: String, items: Array<String>) => {
 }
 
 /**
- * Slices an array given zero-based start and end indexes. The character
+ * Slices an array given zero-based start and end indexes. The value
  * at the end index will not be included in the result.
  *
  * If either index is a negative number, it will be treated as a reverse index from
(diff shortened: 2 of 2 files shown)


stdlib/array.gr
```diff
@@ -373,3 +373,47 @@ export let unique = (array) => {
     array
   )
 }
+
+/**
+ * Slices an array given zero-based start and end indexes. The character
```
ospencer, line 378:
```suggestion
 * Slices an array given zero-based start and end indexes. The value
```

stdlib/array.gr
```diff
@@ -373,3 +373,47 @@ export let unique = (array) => {
...
+ * at the end index will not be included in the result.
+ *
+ * If either index is a negative number, it will be treated as a reverse index from
+ * the end of the array. e.g. slice(1, -1, [> 'a', 'b', 'c']) === 'b'].
```
ospencer, line 382:
```suggestion
 * the end of the array. e.g. slice(1, -1, [> 'a', 'b', 'c']) == [> 'b'].
```

stdlib/array.gr
```diff
@@ -373,3 +373,47 @@ export let unique = (array) => {
...
+
+/**
+ * Slices an array given zero-based start and end indexes. The value
+ * at the end index will not be included in the result.
+ *
+ * If either index is a negative number, it will be treated as a reverse index from
+ * the end of the array. e.g. slice(1, -1, [> 'a', 'b', 'c']) == [> 'b'].
+ *
+ * @param stardIndex: Number The index of the array where slicing starts
+ * @param endIndex: Number The index of the array where slicing ends, but the element will not be included
```
peblair, line 385: Two things:
1. Typo in `startIndex`
2. This might not be as spelled-out as the version you wrote, but I think it is more clear and precise.
```suggestion
 * @param startIndex: Number The index of the array where the slice will begin (inclusive)
 * @param endIndex: Number The index of the array where the slice will end (exclusive)
```

peblair (changes requested):
> This change is likely to be a bit controversial for the team
Nope, you implemented exactly the semantics I'd expect from `slice()`. This looks good to me! I just have some minor feedback on the documentation.

stdlib/array.gr
marcusroberts: I understood both, but the second is more precise.

compiler/test/stdlib/array.test.gr
```diff
@@ -250,3 +250,20 @@ assert Array.unique([> 1, 1, 2, 2, 3, 3]) == [> 1, 2, 3]
 assert Array.unique([> 0, 1, 2]) == [> 0, 1, 2]
 assert Array.unique([> 0, 0, 0, 0, 0, 0]) == [> 0]
 assert Array.unique([>]) == [>]
+
+// Array.slice
+
+let  testChars = [> 'a', 'b', 'c']
+
+assert Array.slice(0, 1, testChars) == [> 'a']
+assert Array.slice(1, Array.length(testChars), testChars) == [> 'b', 'c']
```
marcusroberts, line 259: What's the expected behaviour for Array.slice(0,0,testChars) ?

phated: I believe it's an empty array. Let's add the test 😄 
```suggestion
assert Array.slice(1, Array.length(testChars), testChars) == [> 'b', 'c']
assert Array.slice(0, 0, testChars) == [>]
```

phated:
Rebased, and I believe I got all of the comments I lost in the force push experiment.

stdlib/array.gr
```diff
@@ -414,3 +414,47 @@ export let join = (separator: String, items: Array<String>) => {
...
+ * the end of the array. e.g. slice(1, -1, [> 'a', 'b', 'c']) == [> 'b'].
+ *
+ * @param startIndex: Number The index of the array where the slice will begin (inclusive)
+ * @param endIndex: Number The index of the array where the slice will end (exclusive)
+ * @param array: Array<a> The array to be sliced
+ * @returns Array<a> The subset of the array that was sliced
+ */
+export let slice = (startIndex, endIndex, array) => {
+  // Note(blaine): Implementation taken directly from the String.slice method
+  // I wrote in Reason for Grain_utils.String
```
ospencer, line 432: I personally don't feel like we need this note here, but I don't feel crazy strong about it.
```suggestion
```

Alternatively, if you want to keep it, it might be better to put `phated` here instead of `blaine` since it might be easier for people to find you:
```suggestion
  // NOTE(phated): Implementation taken directly from the String.slice method
  // I wrote in Reason for Grain_utils.String
```

ospencer (approved):

stdlib/array.gr
phated: I think it's fine to remove.

peblair (approved):

PR #729, opened 2021-06-11
feat(stdlib): isNaN and isFinite utility functions for Number

cician:
This pull request adds two functions to the number module:
``` grain
isFinite: (Number) -> Bool
isNaN: (Number) -> Bool
```
So we want explicit isInfinity and isNegativeInfinity as well?

Internally they depend on isBoxedNumber from runtime/numbers so if you don't mind I'd export it.

Technically there's duplicate code in numberUtils, but it's only for WasmF64 and just two lines of actual code. So I don't know if we want to bother reusing it.

Tests could benefit from waiting for PR #720 I guess.


stdlib/number.gr
```diff
@@ -135,3 +137,61 @@ export let abs = (x: Number) => if (0 > x) x * -1 else x
...
+ */
+@disableGC
+export let isNaN = (n: Number) => {
+	let x = WasmI32.fromGrain(n)
+	if (isBoxedNumber(x)) {
+		// Boxed numbers can have multiple subtypes. Of which float32 and float64 can be NaN,
+		let tag = WasmI32.load(x, 4n)
+		if (WasmI32.eq(tag, Tags._GRAIN_FLOAT64_BOXED_NUM_TAG)) {
+			let wf64 = WasmF64.load(x, 8n)
+			WasmF64.ne(wf64, wf64)
```
peblair, line 184: This comment might help with readability:
```suggestion
			// uses the fact that NaN is the only number not equal to itself
			WasmF64.ne(wf64, wf64)
```

stdlib/number.gr
```diff
@@ -135,3 +137,61 @@ export let abs = (x: Number) => if (0 > x) x * -1 else x
...
+ */
+@disableGC
+export let isFinite = (x: Number) => {
+	let asPtr = WasmI32.fromGrain(x)
+	if (isBoxedNumber(asPtr)) {
+		// Boxed numbers can have multiple subtypes, Of which float32 and float64 can be infinite,
+		let tag = WasmI32.load(asPtr, 4n)
+		if (WasmI32.eq(tag, Tags._GRAIN_FLOAT64_BOXED_NUM_TAG)) {
+			let wf64 = WasmF64.load(asPtr, 8n)
+			WasmF64.eq(WasmF64.sub(wf64, wf64), 0.W)
```
peblair, line 155: This comment might help with readability:
```suggestion
			// uses the fact that all finite floats minus themselves are zero
			// (NaN - NaN == NaN, inf - inf == NaN, 
			//  -inf - -inf == NaN, inf - -inf == inf, -inf - inf == -inf)
			WasmF64.eq(WasmF64.sub(wf64, wf64), 0.W)
```

peblair (changes requested):
Thanks, this looks great! A couple of notes (aside from the suggested comments):
- We indent the code in the standard library with two spaces, so if you could update these functions to match that style, it would be very helpful.
- With respect to other useful functions, I think an `isInfinite` function (returning true for positive and negative infinity) would round all of this out nicely. This is what other languages (e.g. Python, Racket) appear to do. I'm not strictly opposed to `isPositiveInfinity`/`isNegativeInfinity` as well, but I am not sure it's necessary when users can do `isInfinite(x) && x > 0`. Wondering what @ospencer's thoughts are.
- Finally, when making the above change(s), can you rebase this onto the latest `main`?

stdlib/number.gr
cician: Added the comment.

stdlib/number.gr
cician: Added the comment.

phated (approved):
I love how readable this is, even to me!

stdlib/number.gr
```diff
@@ -135,3 +137,94 @@ export let abs = (x: Number) => if (0 > x) x * -1 else x
...
+ * Checks if a number is finite.
+ * All values are finite exept for floating point NaN, infinity or negative infinity.
+ * @param x: Number - The number to check
+ * @returns Bool
+ */
+@disableGC
+export let isFinite = (x: Number) => {
+  let asPtr = WasmI32.fromGrain(x)
+  if (isBoxedNumber(asPtr)) {
+    // Boxed numbers can have multiple subtypes, Of which float32 and float64 can be infinite,
```
ospencer, line 151:
```suggestion
    // Boxed numbers can have multiple subtypes, of which float32 and float64 can be infinite.
```

stdlib/number.gr
```diff
@@ -135,3 +137,94 @@ export let abs = (x: Number) => if (0 > x) x * -1 else x
...
+ * Checks if a number contains the NaN value (Not A Number).
+ * Only boxed floating point numbers can contain NaN.
+ * @param x: Number - The number to check
+ * @returns Bool
+ */
+@disableGC
+export let isNaN = (x: Number) => {
+  let asPtr = WasmI32.fromGrain(x)
+  if (isBoxedNumber(asPtr)) {
+    // Boxed numbers can have multiple subtypes. Of which float32 and float64 can be NaN,
```
ospencer, line 183:
```suggestion
    // Boxed numbers can have multiple subtypes, of which float32 and float64 can be NaN.
```

ospencer (changes requested):
Looks awesome! I agree with @peblair that `isInfinite` is a fine choice here 👍 

Left you some tiny tweaks to your code comments, but otherwise this is perfect!

peblair (approved):

ospencer (approved):
Thanks so much @cician!

PR #731, opened 2021-06-14
chore(compiler): Attach comments to the typed_program

phated:
In preparation of "graindoc", I need to submit a few cleanup PRs.

This attaches the comments from the `parsed_program` to the `typed_program` because the documentation generator runs the compiler until typechecking is complete and then combines the type signatures with the comments.

I'm not sure this needs to be tested because it is typechecked and doesn't actually modify the comments at all. We already test the comments are parsed correctly in the Parsetree.

### DIFF
--- a/compiler/src/typed/typedtree.re
+++ b/compiler/src/typed/typedtree.re
@@ -504,11 +504,28 @@ type toplevel_stmt = {
   ttop_env: [@sexp.opaque] Env.t,
 };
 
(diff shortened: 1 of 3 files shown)


marcusroberts (approved):

marcusroberts:
The code looks good to me.   As you say, not sure if needed here, but the more info the better for diagnostics etc I say!

PR #734, opened 2021-06-14
chore(compiler): Trim each line of a block or doc comment

phated:
In preparation of "graindoc", I need to submit a few cleanup PRs.

This improves our comment string trimming to trim each line of a Doc or Block comment which allows for a more precise regexp in the documentation attribute parser (so each attribute line will start with an `@`-symbol).

### DIFF
--- a/compiler/src/utils/string_utils.re
+++ b/compiler/src/utils/string_utils.re
@@ -13,6 +13,13 @@ let starts_with = (string, prefix) => {
   };
 };
 
+let trim_each_line = str => {
+  str
+  |> Str.split(Str.regexp("[\n\r]"))
+  |> List.map(String.trim)
+  |> String.concat("\n");
+};
+
 /**
 Slices a string given optional zero-based [~first] and [~last] indexes. The character
 at the [~last] index will not be included in the result.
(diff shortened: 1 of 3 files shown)


marcusroberts (approved):

compiler/src/utils/string_utils.re
```diff
@@ -13,6 +13,13 @@ let starts_with = (string, prefix) => {
   };
 };
 
+let trim_each_line = str => {
+  str
+  |> Str.split(Str.regexp("[\n\r]"))
```
ospencer, line 18: Could this give us issues on Windows, since this would end up adding an extra newline for each line, since there's a split for both `\r` and `\n`? If so, I think this would fix:

```suggestion
  |> Str.split(Str.regexp("\\(\r\n\\|\n\\)"))
```

phated: Yep! I wrote this extremely blurry-eyed and even then knew something was wrong, but didn't know how to fix it. Thanks! Should I add a test?

ospencer: Probably a good idea!

peblair (approved):

ospencer (approved):
Up to you if you want to write a test for the line endings thing!

phated:
Test added! Merging.

PR #736, opened 2021-06-16
chore(stdlib): Add doc comments to Range

phated:
This adds doc comments to the `range.gr` file in our standard library. It was my test file for all my Graindoc work and I wanted to submit this PR before Graindoc to find out if anyone has qualms with the doc comment syntax/attributes/structure/whatever.

### DIFF
--- a/stdlib/range.gr
+++ b/stdlib/range.gr
@@ -1,8 +1,34 @@
+/**
+  @module Range: Utilities for working with ranges. A range represents an interval, or a set of values with a beginning and an end.
+  @example import Range from "range"
+*/
+
+/**
+  @section Types: Type declarations included in the Range module.
+*/
+
+/**
+  Ranges can be inclusive or exclusive. When `Inclusive`, the end value will be included in operations. When `Exclusive`, the end value will be excluded from operations.
+*/
 export enum Range {
   Inclusive(Number, Number),
   Exclusive(Number, Number),
 }
 
+/**
+  @section Values: Functions and constants included in the Range module.
+*/
+
+/**
+  Checks if the given number is within the range.
+
+  @param value: The number being checked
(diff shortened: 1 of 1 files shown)


marcusroberts (commented):
I really love this new graindoc!

stdlib/range.gr
```diff
@@ -1,8 +1,34 @@
+/**
+  @module Range: Utilities for working with ranges. A range represents an interval, or a set of values with a beginning and an end.
+  @example import Range from "range"
```
ospencer, line 3: I think this will even be compatible with multiline examples in the future! https://jsdoc.app/tags-example.html

Side note that's not relevant here... How do you escape within jsdoc to show an example with a decorator (thinking about Grain attributes) 🤔

ospencer (approved):
Beautiful!

stdlib/range.gr
phated: @ospencer multi-line anything is pretty tricky given my current regexp implementation. We probably won't want to look at a full-on lexer if we want to handle multi-line gracefully.

As for decorators, I only parse attributes at the very start of a line. I think multi-line examples should have to be indented (though, I also trim those lines...)

stdlib/range.gr
```diff
@@ -1,8 +1,34 @@
+/**
+  @module Range: Utilities for working with ranges. A range represents an interval, or a set of values with a beginning and an end.
```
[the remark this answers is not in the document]
phated: Thanks, I took this directly from the website.

phated:
I think @marcusroberts was meaning to do approve and @peblair gave a "LGTM!" in discord. I'm going to merge this so I can rebase on it for my GrainDoc PR (so everyone can test stuff!)

PR #737, opened 2021-06-16
feat(compiler)!: Improve assert messages to contain location information

peblair:


compiler/src/codegen/compcore.re
```diff
@@ -2054,7 +2077,21 @@ let compile_prim1 = (wasm_mod, env, p1, arg): Expression.t => {
...
+                wasm_mod,
+                env,
+                0,
+                allocate_string(wasm_mod, env, loc.Grain_parsing.Location.loc_start.pos_fname)
+              );
+              ([
+                Expression.Drop.make(wasm_mod, set_fname),
+               ], MImmBinding(MSwapBind(Int32.zero, I32Type)))
+            }
+          ],
```
ospencer, line 2094: You shouldn't need the swap or the delayed functions here; this should be able to just be
```suggestion
          [
            MImmConst(MConstI32(Int32.of_int(loc.Grain_parsing.Location.loc_start.pos_lnum))),
            allocate_string(wasm_mod, env, loc.Grain_parsing.Location.loc_start.pos_fname)
          ],
```
(and updating `call_error_handler` to just take regular args)

peblair: The problem is that `MImmConst` is an immediate, when `allocate_string` is a Binaryen expression. I did try to clean it up (just pushed), but it is still giving me the same issue as I described on Discord.

stdlib/runtime/exception.gr
```diff
@@ -82,7 +82,7 @@ export exception Overflow
 export exception NumberNotIntlike
 export exception NumberNotRational
 export exception MatchFailure
-export exception AssertionError
+export exception AssertionError(Number, String)
```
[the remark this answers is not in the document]
peblair: Yes, for the final draft, I plan to switch these.

stdlib/runtime/exception.gr
```diff
@@ -95,7 +95,7 @@ let runtimeErrorPrinter = (e) => {
     NumberNotIntlike => Some("NumberNotIntlike: Can't coerce number to integer"),
     NumberNotRational => Some("NumberNotRational: Can't coerce number to rational"),
     MatchFailure => Some("MatchFailure: No matching pattern"),
-    AssertionError => Some("AssertionError: Assertion failed"),
+    AssertionError(_, _) => Some("AssertionError: Assertion failed at <unknown>"), // <- actual printer located in pervasives.gr, since malloc depends on this module
```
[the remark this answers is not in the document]
peblair: I'm not particularly opposed; I tried this at one point, but ran into issues (which I no longer recall). Now that the line number wiring is correctly set up, maybe it's worth revisiting this idea.

ospencer:
Seems to be working just fine for me locally! Maybe since the cleanup made it simpler it just works now?

compiler/src/codegen/compcore.re
```diff
@@ -7,6 +7,13 @@ open Grain_utils;
 open Comp_utils;
 open Comp_wasm_prim;
 
+/*
+ GENERAL NOTES (extend as desired)
+ =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+
+ Make sure that all nodes in the Binaryen AST are distinct instances! See note in link.re.
+ */
```
ospencer, line 15: I kind of have a lot of thoughts and feelings about this, so as not to block this change can we pull this and the note out of this PR? I can get a PR up with some info around this.

stdlib/runtime/exception.gr
```diff
@@ -82,7 +82,7 @@ export exception Overflow
 export exception NumberNotIntlike
 export exception NumberNotRational
 export exception MatchFailure
-export exception AssertionError
+export exception AssertionError(String) // <- since we can't malloc in this module, we put the full error message in the Exception object
```
ospencer, line 85: I don't think we need this comment here since I don't think there's any expectation that it should be any other way:
```suggestion
export exception AssertionError(String)
```

compiler/src/codegen/compcore.re
```diff
@@ -2054,7 +2079,17 @@ let compile_prim1 = (wasm_mod, env, p1, arg): Expression.t => {
             Expression.Const.make(wasm_mod, const_false()),
           ),
           AssertionError,
-          [],
+          [
+            allocate_string(
+              wasm_mod,
+              env,
+              Printf.sprintf(
+                "AssertionError: Assertion failed at %s:%d",
```
ospencer, line 2087: Small tweak here to be a little closer to how our other location messages look:
```suggestion
                "AssertionError: Assertion failed in %s, line %d",
```

ospencer (changes requested):
Hype for this one! Could you also update the two assert run tests in `basic_functionality` to look for the new messages?

stdlib/runtime/exception.gr
peblair: I disagree. I think that it merits some sort of comment, since (a) it explains what the `String` here is, (b) a much more natural choice of Exception object would be `AssertionError(String, Number)` (containing the location information of the failure), so this comment helps explain why the exception is not formatted like that, and (c) I don't think there's a compelling reason to not err on the side of having too much explanation.

ospencer: I disagree with (c) because of the "made you think" factor. If you already understand the constraints of this module and how assertions work this makes sense, but otherwise it's confusing because this doesn't explain how the AssertionError gets thrown or that the information is automatically populated. Without the comment, the thought is just "this assertion error has a message with it", but with the comment and no knowledge of how `assert` works, you might question "Why can't we malloc? Why does it matter that we can't malloc, and how would this be different? What do you mean by 'we' put the full error message in there? Does this mean users need to provide a full error message? What does a 'full' error message entail? Does that mean this is a hack of some kind?"

You can have a comment here, but you'd have to explain a bit more:
```suggestion
/*
 * `assert <condition>` is essentially syntactic sugar for:
 * `if (!<condition>) throw AssertionError(<message>)`
 * where <message> is a string injected by the compiler, containing file and line information.
 * This would be better as `AssertionError(String, Number)` to allow the exception printer in
 * this file to construct the error, but constructing a string in this module is impractical as we
 * don't have access to memory allocation since malloc.gr depends on this module.
 */
export exception AssertionError(String)
```

I think this information is useful, but none of it is really a "need to know" for this module IMO. I think it could live somewhere else, though.

compiler/src/codegen/compcore.re
peblair: #743

stdlib/runtime/exception.gr
```diff
@@ -82,7 +82,16 @@ export exception Overflow
...
+/*
+ * `assert <condition>` is essentially syntactic sugar for:
+ * `if (!<condition>) throw AssertionError(<message>)`
+ * where <message> is a string injected by the compiler, containing file and line information.
+ * This would be better as `AssertionError(String, Number)` to allow the exception printer in
+ * this file to construct the error, but constructing a string in this module is impractical as we
+ * don't have access to memory allocation since malloc.gr depends on this module.
+ */
+export exception AssertionError(String)
+export exception AssertionError(String)
```
ospencer, line 94: Looks like we got an extra `AssertionError` in here:
```suggestion
```

ospencer (approved):
Most impactful feature of 0.4??? 😤😤😤

phated (approved):
🔥

marcusroberts (approved):

PR #739, opened 2021-06-17
chore(compiler)!: Avoid building JS bundles by default

phated:
Abuses profiles and `enabled_if` to disable the jsoo targets during normal development flow. This also updates our release workflow to build only the JS artifacts.

I marked this as breaking because it probably affects end-user workflow if they are building from source (and want the JS version for some reason?)

Needed to bump the dune language version to get `(enabled_if)`

I know that @ospencer and @marcusroberts wanted this for awhile.

I'm not sure if not building the JSOO without `(profile release)` will cause any issues 🤷


ospencer (approved):
Woohoo!

peblair (approved):

marcusroberts (approved):
This is great and will speed up my dev cycle!

PR #740, opened 2021-06-18
fix(compiler): Strip leading line asterisks in block and doc comments

ospencer:
This was the intended behavior of how `cmt_content` would work (and why we have both `cmt_source` and `cmt_content` in the first place 🤦). Added bonus: this makes writing `* @param` work with Graindoc 😄

### DIFF
--- a/compiler/src/parsing/parser_header.re
+++ b/compiler/src/parsing/parser_header.re
@@ -39,6 +39,7 @@ let make_shebang_comment = (source, loc) => {
 let make_block_comment = (source, loc) => {
   let content =
     String_utils.slice(~first=2, ~last=-2, source)
+    |> String_utils.deasterisk_each_line
     |> String_utils.trim_each_line;
   Block({cmt_content: content, cmt_source: source, cmt_loc: loc});
 };
@@ -46,6 +47,7 @@ let make_block_comment = (source, loc) => {
 let make_doc_comment = (source, loc) => {
   let content =
     String_utils.slice(~first=3, ~last=-2, source)
+    |> String_utils.deasterisk_each_line
     |> String_utils.trim_each_line;
   Doc({cmt_content: content, cmt_source: source, cmt_loc: loc});
 };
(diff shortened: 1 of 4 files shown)


compiler/src/parsing/parser_header.re
```diff
@@ -39,13 +39,15 @@ let make_shebang_comment = (source, loc) => {
 let make_block_comment = (source, loc) => {
   let content =
     String_utils.slice(~first=2, ~last=-2, source)
+    |> String_utils.deasterisk_each_line
```
phated, line 42: Do you actually want to support this form?
```
/*
 *
 */
```

I don't think that block comments should actually use asterisks (only doc comments)

ospencer: I wouldn't say that this is an endorsement that we should style it that way in our Grain code, just that `cmt_content` should remove them since people will probably write them. I don't really have any stats on it, but it's in [some official style guides](https://google.github.io/styleguide/jsguide.html#formatting-block-comment-style), the [default for some linting rules](https://eslint.org/docs/rules/multiline-comment-style), and [often found in the wild](https://github.com/reasonml/reason-native/blob/master/src/fp/Fp.re#L334).

If you want though, I could make this a little more robust by stripping them only if _every_ line starts with an asterisk. That'd mean that you could write a comment like this if you wanted, and `cmt_content` would keep the asterisks:
```
/*
Lunch options for today:
* pizza
* salad
* burger
*/
```

phated: Makes sense, I just wanted to call out that it was a form that we were "supporting" (and that I think it looks weird).

phated (approved):
LGTM 👍 Let's get this in so I can update my LSP PR

ng-marcus (approved):

ng-marcus:
Ignore the ghost approval!

marcusroberts (approved):

PR #741, opened 2021-06-18
chore: Simplify the badges, square them & fix version

phated:
I noticed the fury.io stuff was broken again, so I went to update the badges. I switched to shields.io to read the latest version and I squared them up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,9 +5,15 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


peblair (approved):

ospencer (approved):
For anyone who doesn't want to find a preview:

image

Lookin' mighty sharp!

PR #735, opened 2021-06-15
fix(stdlib): decodeRangeHelp equality check in String module

ospencer:

PR #742, opened 2021-06-19
feat(compiler): Emit a warning when using Pervasives.(==) with unsafe Wasm types

### LINKED ISSUE #738: Compiler: Warn when unsafe wasm types are passed to `Pervasives.(==)`
It looks like it's here: https://github.com/grain-lang/grain/blob/66319ca8f3bbeb80e18525d1c15e2b84f0abd0c1/stdlib/string.gr#L1175

I've actually thought about this some... I really don't want to have that sort of check in there. 😕 Feels bad to have the typechecker and then have to do that kind of check. As an alternative though, I've thought about adding a warning to the typechecker that yells at you if you pass a value of type `WasmI32` to `(==)`, since that shouldn't be too hard to add. It doesn't guard against having some sort of wrapper function (e.g. `let (==) = (a, b) => a == b`) but it _does_ prevent us from having that bug in the code. WDYT?

_Originally posted by @ospencer in https://github.com/grain-lang/grain/issues/726#issuecomment-861588009_

peblair:
Closes #738.

This pull request adds a QOL improvement for low-level standard WASM development by emitting a warning when `Pervasives.(==)` is used to compare any of the following types: `WasmI32`, `WasmI64`, `WasmF32`, and `WasmF64`. This is accomplished by a new well-formedness pass on the Typed AST, which we can add things to in the future. To illustrate, consider the following `tst.gr` file:
```grain
import WasmI32 from "runtime/unsafe/wasmi32"
assert 9n == WasmI32.fromGrain(4)
assert WasmI32.toGrain(9n) == 4
let (==) = WasmI32.eq
// safe, since we have shadowed (==) with the WasmI32 variant
assert 9n == WasmI32.fromGrain(4)
```
The compiler now emits the following:
```
 λ grain git:(warn-wasm-unsafe-equal) > grain tst.gr                             
File "tst.gr", line 2, characters 7-33:
Warning 17: it looks like you are using (==) on two unsafe WASM values here.
Please ensure that this is intentional.
```

I've verified that backing out the change from #735 causes the warning to be emitted when running the test:
```
 λ grain git:(warn-wasm-unsafe-equal) ✗ > grain compiler/test/stdlib/string.test.gr
File "/Users/pblair/views/grain/stdlib/string.gr", line 1176, characters 6-22:
Warning 17: it looks like you are using (==) on two WasmI32 values here.
Please ensure that this is intentional.
```
I haven't found any warnings emitted by the standard library, which is good news in general, but bad news for #726 .

### DIFF
--- a/compiler/src/typed/typed_well_formedness.re
+++ b/compiler/src/typed/typed_well_formedness.re
@@ -1,0 +1,44 @@
+open Grain_parsing;
+open Types;
+open Typedtree;
+open TypedtreeIter;
+
+/*
+ * This module provides well-formedness checks which are type/value resolution-dependent.
+ * For the moment, this is just a warning on (==), but feel free to refactor once more checks
+ * are desired.
+ */
+
+let wasm_unsafe_types = ["WasmI32", "WasmI64", "WasmF32", "WasmF64"];
+
+let exp_is_wasm_unsafe = ({exp_type: {desc}}) => {
+  switch (desc) {
+  | TTyConstr(path, [], _) => List.mem(Path.name(path), wasm_unsafe_types)
+  | _ => false
+  };
+};
+
+module WellFormednessArg: TypedtreeIter.IteratorArgument = {
+  include TypedtreeIter.DefaultIteratorArgument;
+
+  let enter_expression: expression => unit =
+    ({exp_desc, exp_loc}) => {
+      switch (exp_desc) {
+      | TExpApp({exp_desc: TExpIdent(val_fullpath, _, _)}, args)
+          when Path.name(val_fullpath) == "Pervasives.==" =>
+        if (List.exists(exp_is_wasm_unsafe, args)) {
+          let warning = Grain_utils.Warnings.EqualWasmUnsafe;
+          if (Grain_utils.Warnings.is_active(warning)) {
+            Grain_parsing.Location.prerr_warning(exp_loc, warning);
+          };
+        }
+      | _ => ()
+      };
+    };
+};
+
+module WellFormednessIterator = TypedtreeIter.MakeIterator(WellFormednessArg);
+
+let check_well_formedness = ({statements}) => {
+  List.iter(WellFormednessIterator.iter_toplevel_stmt, statements);
+};
--- a/compiler/src/typed/typed_well_formedness.re
+++ b/compiler/src/typed/typed_well_formedness.re
@@ -5,8 +5,6 @@ open TypedtreeIter;
 
 /*
  * This module provides well-formedness checks which are type/value resolution-dependent.
- * For the moment, this is just a warning on (==), but feel free to refactor once more checks
- * are desired.
  */
 
 let wasm_unsafe_types = [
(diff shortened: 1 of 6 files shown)


ospencer (changes requested):

compiler/src/utils/warnings.re
```diff
@@ -102,7 +104,8 @@ let message =
       msg,
     )
   | NonClosedRecordPattern(s) =>
-    "the following fields are missing from the record pattern: " ++ s;
+    "the following fields are missing from the record pattern: " ++ s
+  | EqualWasmUnsafe => "it looks like you are using (==) on two unsafe WASM values here.\nPlease ensure that this is intentional.";
```
ospencer, line 108:
```suggestion
  | EqualWasmUnsafe => "it looks like you are using Pervasives.(==) on two unsafe Wasm values here.\nThis is generally unsafe and will cause errors. Use `WasmI32.eq`, `WasmI64.eq`, `WasmF32.eq`, or `WasmF64.eq` instead.";
```

compiler/src/typed/typed_well_formedness.re
```diff
@@ -0,0 +1,44 @@
...
+};
+
+module WellFormednessArg: TypedtreeIter.IteratorArgument = {
+  include TypedtreeIter.DefaultIteratorArgument;
+
+  let enter_expression: expression => unit =
+    ({exp_desc, exp_loc}) => {
+      switch (exp_desc) {
+      | TExpApp({exp_desc: TExpIdent(val_fullpath, _, _)}, args)
+          when Path.name(val_fullpath) == "Pervasives.==" =>
```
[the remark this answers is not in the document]
peblair: I made the change, but I'm unsure how it is less fragile. If you had something else in mind, please submit a suggestion.

compiler/src/typed/typed_well_formedness.re
```diff
@@ -0,0 +1,53 @@
...
+    ({exp_desc, exp_loc}) => {
+      switch (exp_desc) {
+      | TExpApp(
+          {
+            exp_desc:
+              TExpIdent(Path.PExternal(Path.PIdent(mod_), fnc, _), _, _),
+          },
+          args,
+        )
+          when Ident.name(mod_) == "Pervasives" && fnc == "==" =>
```
ospencer, line 37: Can simplify this a tad bit:
```suggestion
      | TExpApp(
          {
            exp_desc:
              TExpIdent(Path.PExternal(Path.PIdent({name: "Pervasives"}), "==", _), _, _),
          },
          args,
        ) =>
```

compiler/src/typed/typed_well_formedness.re
```diff
@@ -0,0 +1,53 @@
...
+let wasm_unsafe_types = [
+  Builtin_types.path_wasmi32,
+  Builtin_types.path_wasmi64,
+  Builtin_types.path_wasmf32,
+  Builtin_types.path_wasmf64,
+];
+
+let exp_is_wasm_unsafe = ({exp_type: {desc}}) => {
+  switch (desc) {
+  | TTyConstr(path, _, _) => List.mem(path, wasm_unsafe_types)
```
ospencer, line 19: 🙌

ospencer (commented):
Love it! I'm excited to add a warning for `print` too 😄

phated (approved):
Heck yeah! 🥇

ospencer (approved):

Issue #745, opened 2021-06-19
Memory leak returning objects from function


cician:
This simple program reaches memory limit and never prints "end". The used memory peaks at 3 GiB before crashing.

Same when executed with Wasmtime or Wasmer, except for the latter the runtime allows 4 GiB so more iterations or larger string is needed.

The leak occurs as long as f() returns a heap allocated object.

``` grain
let f = () => {
  " "
}

for (let mut i=0; i<100000000; i += 1) {
  let s = f()
}

print("end")
```

cician:
I've done a little bit of testing in hope of narrowing down on the issue.

The following program prints 2. Meaning the string returned by leaky() has a reference count of two despite being referenced only by "let s" (assuming passing it into getStrRefCnt doesn't increase the count because of `@disableGC`).

``` grain
import Memory from "runtime/unsafe/memory"
import WasmI32 from "runtime/unsafe/wasmi32"
import { tagSimpleNumber } from "runtime/dataStructures"

let leaky = () => {
  toString('c')
}

@disableGC
let getStrRefCnt = (s) => {
  // based on getRefCount from gc module, for debugging
  let _HEADER_SIZE = 8n
  let ptr = WasmI32.fromGrain(s)
  let cnt = WasmI32.load(WasmI32.sub(ptr, _HEADER_SIZE), 0n)
  tagSimpleNumber(cnt)
}

let s = leaky()
print(getStrRefCnt(s))
```


Also with the following otherwise meaningless change the result is 1 and leak doesn't occur. I don't know if it's my misunderstanding of `@disableGC` or another bug. Without `@disableGC` on `foo` the result is 2. Intuitively i would think the reference count would remain the same in both cases.
``` grain
@disableGC
let foo = (a) => a

let leaky = () => {
  foo(toString('c'))
}
```

ospencer:
Hey @cician, I've been investigating this and that's more or less the issue. It's tad tricky though, so I don't yet have a fix 😅

cician:
No pressure! I just want to help within my abilities.

PR #747, opened 2021-06-21
Reduce reduntant memory loads in Malloc.free.

cician:
This PR should speed up all grain programs at least a little bit by avoiding redundant memory accesses in Malloc.free.

I hope someone can carefully triple check this change so we can feel a little bit greener 💚

Synthetic test:
just allocate and free 1000000 strings  of 10 bytes each in a loop (with @disableGC).
207ms-216ms unoptimized
198ms-207ms optimized

Slightly less synthetic test:
Emit 15KiB json.
127ms-132ms unoptimized
74ms-81ms optimized

Emit 150KiB json.
5969ms-6364ms unoptimized
2955ms-3188ms optimized

Side note: Emit json time seems to grow exponentially with input size, but its not the case. It's because of a memory leak in GC (#745) so my benchmark results are heavily skewed with each iteration of the same being slower. 1.5MiB json takes 14 minutes (compare that to 5ms for the same in node.js, not exactly a fair comparison, but still 🐢). The synthetic test instead uses @disableGC to only test malloc. Sorry for side-tracking.

```
starting benchmark emit example json
iteration 0
iteration time: 3896ms
iteration 1
iteration time: 4438ms
iteration 2
iteration time: 5207ms
iteration 3
iteration time: 5742ms
iteration 4
iteration time: 5648ms
iteration 5
iteration time: 6961ms
iteration 6
iteration time: 7296ms
iteration 7
iteration time: 7985ms
iteration 8
iteration time: 7983ms
iteration 9
iteration time: 8478ms
emit example json: 6364ms
```

### DIFF
--- a/stdlib/runtime/malloc.gr
+++ b/stdlib/runtime/malloc.gr
@@ -172,7 +172,7 @@ export let free = (ap: WasmI32) => {
   // is actually already pointing to this node, so we don't do anything.
   if (blockPtr != freePtr) {
     // Find the location to insert this block into the free list
-    for ( ; ; ) {
+    while (true) {
       let nextp = getNext(p)
       if ( ((blockPtr > p) && (blockPtr < nextp)) || ((p >= nextp) && ((blockPtr > p) || (blockPtr < nextp))) ) {
         break
--- a/stdlib/runtime/malloc.gr
+++ b/stdlib/runtime/malloc.gr
@@ -172,23 +172,27 @@ export let free = (ap: WasmI32) => {
   // is actually already pointing to this node, so we don't do anything.
   if (blockPtr != freePtr) {
     // Find the location to insert this block into the free list
-    for ( ; !((blockPtr > p) && (blockPtr < getNext(p))); p = getNext(p) ) {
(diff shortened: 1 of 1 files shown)


stdlib/runtime/malloc.gr
```diff
@@ -172,23 +172,27 @@ export let free = (ap: WasmI32) => {
   // is actually already pointing to this node, so we don't do anything.
   if (blockPtr != freePtr) {
     // Find the location to insert this block into the free list
-    for ( ; !((blockPtr > p) && (blockPtr < getNext(p))); p = getNext(p) ) {
-      if ((p >= getNext(p)) && ((blockPtr > p) || (blockPtr < getNext(p)))) {
+    for ( ; ; ) {
```
phated, line 175: Would we want this to just be `while(true)`?

cician: Changed into while(true). Looks a bit less cryptic, but the worse offender for readability remains the condition for the break. I'd compensate with a comment, but I lack a proper understanding of the algorithm.

peblair (approved):
Thanks for making this better!

ospencer (approved):
Sorry for the delay on the review; I wanted to be able to sit and really look it over. Looks good!

PR #749, opened 2021-06-22
fix(compiler): Print parens for argument-less functions

phated:
While writing docs, I noticed that the LSP was looking weird for functions that had no arguments. This was due to my oprint PR that skipped parens if arguments were <= 1 - this was incorrect and we should **only** skip the parens for single-argument functions.

This restores parens to argument-less functions.


ospencer (approved):
Ah yeah, I saw this and meant to make a note of it. But now it's fixed!

marcusroberts (approved):

PR #751, opened 2021-06-22
feat(graindoc): Add support for deprecated attribute

phated:
This adds support in GrainDoc for the `@deprecated` attribute. I decided to generate this markdown when the attribute is encountered:
```md
> **Deprecated:** Message for deprecation
```
(This currently doesn't style well on the website, but we can fix that.)

The deprecation markdown is inserted after the function name but before the type signature or any other documentation.

I also added my Stack documentation to this PR so y'all can test the generation to see if you like it. If you'd like me to split that documentation out before merge, let me know.

### DIFF
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -123,6 +123,28 @@ let signature_item_in_range =
 let to_markdown = docblock => {
   let buf = Buffer.create(0);
   Buffer.add_string(buf, Markdown.heading(~level=3, docblock.name));
+  let deprecations =
+    docblock.attributes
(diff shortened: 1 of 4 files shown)


ospencer:
Re: website styling, I actually kind of like the system that ReadMe uses, and maybe in the future we'll do something like that: https://rdmd.readme.io/docs/callouts 

It'll be super easy to update the generated markdown as well 🤘

ospencer (approved):

marcusroberts (approved):

peblair (approved):

phated:
Nobody requested me to remove the Stack docs, so I'm going to merge this as

PR #752, opened 2021-06-22
chore(cli): Remove workaround now that env vars are gone

phated:
When @ospencer removed the env var hack to apply memory settings, we missed removing this workaround. Let's get rid of it now.

cc @marcusroberts because you'll need to update the import in your formatter code.


ospencer (approved):

marcusroberts (approved):
Thanks for the heads up

peblair (approved):

PR #754, opened 2021-06-22
chore(compiler): Refactor test paths & clean up more output

phated:
This refactors some of our test pathing and makes sure we cleanup the test-libs directory, since having old code there caused some test failures when changing compiler internals.


ospencer (approved):
Looks great!

peblair (approved):

PR #755, opened 2021-06-22
chore!: Refactor out references to "runtime" that aren't related to `stdlib/runtime`

phated:
This PR has 2 parts.

* The first part is renaming `@grain/runtime` to `@grain/js-runner` since it only exists to run in a JavaScript environment and isn't needed to run compiler Grain programs in wasmtime or wasmer. The "runtime" name was misleading because references to "runtime" in contributor docs and stuff all refer to `stdlib/runtime`
* Updates `grainRuntime` references injected by the compiler and provided by the `js-runner` to be named `_grainEnv` (as workshopped with @ospencer).

The result of this PR is that "runtime" will only refer to Grain's `stdlib/runtime` within the project. This also sets us up to decouple that runtime from the stdlib!

This doesn't require #754 - but I encountered a false test failure because not all the generated files were being cleaned up. So ideally that will be merged before anyone tries to run this locally (thus merged...)

The one thing I'm not sure about is if renaming this project will be handled gracefully by `release-please`. cc @joeldodge79 do you know?


ospencer (approved):
Excited about this change! I also think it's funny how this is like the _one_ change you could make that would affect every single snapshot 😂 

Nothing that the file filter couldn't handle though!

.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: Checkout project
         uses: actions/checkout@v2
 
-      - name: Set up runtime and CLI
+      - name: Set up JS runner and CLI
```
peblair, line 31: Nothing gets past @phated 🕵️‍♂️

compiler/src/codegen/comp_utils.re
```diff
@@ -84,3 +84,7 @@ let load =
   let align = Option.value(~default=sz, align);
   Expression.Load.make(~signed, wasm_mod, sz, offset, align, ty, ptr);
 };
+
+let grain_env_name = "_grainEnv";
+
+let is_grain_env = str => grain_env_name == str;
```
peblair, line 90: Feel free to reject this, but this really spoke to me:
```suggestion
let is_grain_env = (==)(grain_env_name);
```

peblair (approved):
Awesome stuff!

compiler/src/codegen/comp_utils.re
phated: I generally would accept, but in thinking towards Whole Grain, I don't believe that our partial application will look like this, so I'd prefer to keep it just the generic function.

PR #756, opened 2021-06-23
feat(compiler)!: Remove semantic meaning of semicolons

ospencer:
Up until this PR, ending a statement with a semicolon had the same effect as calling `ignore` on that statement. We found this feature to be too subtle, often causing more confusion than the feature was worth. This PR removes that behavior.

### DIFF
--- a/compiler/src/parsing/parser.dyp
+++ b/compiler/src/parsing/parser.dyp
@@ -503,7 +503,7 @@ block_body_expr :
   | expr  { $1 }
 
 block_body_stmt :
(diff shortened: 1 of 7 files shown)


phated (approved):
👍 I like this

marcusroberts (approved):

marcusroberts:
I always stumbled over semicolons!   Good change!

peblair (approved):
@ospencer Any warnings generated by the stdlib with this change?

ospencer:
@peblair No warnings! We're all set here.

PR #757, opened 2021-06-23
chore(stdlib): Implement Map.toArray without intermediate list

### LINKED ISSUE #187: Compiler(?): Initialize an array with a size but no data
While trying to implement `Map.toArray`, I found that it wasn't possible to initialize a sized array with no data. I ended up using `Array.fromList(Map.toList(map))` to implement this currently but that seems very under-performant.

I *think* this needs to be implemented in the compiler, but I'm not sure.

phated:
This implements Map.toArray without converting it to an intermediate list. This avoids blowing the stack on huge Maps when you don't have experimental tail calls, and it increases performance by 25%.

Closes #187 (which this implementation inspired, and proves that we can allocate an array without items in it).

### DIFF
--- a/stdlib/map.gr
+++ b/stdlib/map.gr
@@ -291,7 +291,7 @@ let setInArray = (array) => {
 @disableGC
 export let toArray = (map) => {
   let length = WasmI32.shrS(WasmI32.fromGrain(map.size), 1n)
-  // TODO: Removing parens around Array<a> causes a parse error
+  // TODO(#783): Removing parens around Array<a> causes a parse error
   let array = WasmI32.toGrain(allocateArray(length)): (Array<a>)
   let setInArray = setInArray(array)
   reduce(setInArray, 0, map)
--- a/stdlib/map.gr
+++ b/stdlib/map.gr
@@ -277,9 +277,15 @@ export let fromList = (list) => {
   map
 }
 
-let setInArray = (array) => (i, key, value) => {
-  array[i] = (key, value)
-  i + 1
+let setInArray = (array) => {
+  @disableGC
+  let iter = (i, key, value) => {
+    Memory.incRef(WasmI32.fromGrain(key))
+    Memory.incRef(WasmI32.fromGrain(value))
+    array[i] = (key, value)
+    i + 1
+  }
+  iter
 }
 
 @disableGC
--- a/stdlib/map.gr
+++ b/stdlib/map.gr
@@ -3,6 +3,9 @@
 import List from "list"
 import Array from "array"
 import { hash } from "hash"
+import Memory from "runtime/unsafe/memory"
+import WasmI32 from "runtime/unsafe/wasmi32"
+import { allocateArray } from "runtime/dataStructures"
 
 // TODO: Consider implementing this as List<(Box<k>, Box<v>)>
 record Bucket<k, v> {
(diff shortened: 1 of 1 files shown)


cician:
Correct me if I'm wrong, but I think that tuples are always allocated on the heap. Since a new tuple is allocated for each call from reduce, this makes it so there are at least as many allocations as in the case of creating an intermediate linked list.

stdlib/map.gr
```diff
@@ -274,8 +277,48 @@ export let fromList = (list) => {
...
+print(("toArray took", Int64.div(Int64.sub(afterToArray, beforeToArray), 1000L)))
+
+*/
+
+@disableGC
+let setInArray = ((i, array), key, value) => {
+  Memory.incRef(WasmI32.fromGrain(array))
+  array[i] = (key, value);
+  (i + 1, array)
+}
```
peblair, line 309:
```suggestion
let setInArray = (array) => (i, key, value) => {
  Memory.incRef(WasmI32.fromGrain(array))
  array[i] = (key, value);
  i + 1
}
```

stdlib/map.gr
```diff
@@ -274,8 +277,48 @@ export let fromList = (list) => {
...
+
+@disableGC
 export let toArray = (map) => {
-  Array.fromList(toList(map))
+  let length = WasmI32.shrS(WasmI32.fromGrain(map.size), 1n)
+  // TODO: Removing parens around Array<a> causes a parse error
+  let array = WasmI32.toGrain(allocateArray(length)): (Array<a>)
+  Memory.incRef(WasmI32.fromGrain(array))
+  let acc = (0, array)
+  let res = reduce(setInArray, acc, map)
```
peblair, line 318:
```suggestion
  let func = setInArray(array)
  let res = reduce(func, acc, map)
  Memory.free(WasmI32.fromGrain(func))
```

peblair (commented):
I think @cician is on the right track. I made a couple of suggestions which may alleviate the issue (untested, so they may need tuning).

stdlib/map.gr
```diff
@@ -274,8 +277,20 @@ export let fromList = (list) => {
   map
 }
 
+let setInArray = (array) => (i, key, value) => {
+  array[i] = (key, value)
+  i + 1
+}
```
phated, line 283: I'm not exactly sure why this doesn't need `@disableGC` anymore??

stdlib/map.gr
```diff
@@ -274,8 +277,20 @@ export let fromList = (list) => {
...
+
+@disableGC
 export let toArray = (map) => {
-  Array.fromList(toList(map))
+  let length = WasmI32.shrS(WasmI32.fromGrain(map.size), 1n)
+  // TODO: Removing parens around Array<a> causes a parse error
+  let array = WasmI32.toGrain(allocateArray(length)): (Array<a>)
+  let setInArray = setInArray(array)
+  reduce(setInArray, 0, map)
```
phated, line 291: Why don't we need to free the result of reduce?

stdlib/map.gr
```diff
@@ -274,8 +277,20 @@ export let fromList = (list) => {
...
+let setInArray = (array) => (i, key, value) => {
+  array[i] = (key, value)
+  i + 1
+}
+
+@disableGC
 export let toArray = (map) => {
-  Array.fromList(toList(map))
+  let length = WasmI32.shrS(WasmI32.fromGrain(map.size), 1n)
+  // TODO: Removing parens around Array<a> causes a parse error
```
phated, line 288: Need to create an issue for this.

phated:
Alright, this was updated to include @peblair and @ospencer's changes - I squashed them into a single commit.

This new implementation gains about 25% efficiency!

stdlib/map.gr
ospencer: We do, my mistake. It needs to go on the inner function though, and we'll need to incRef both `key` and `value` here.

stdlib/map.gr
ospencer: It's a simple number, so there's no need to do anything here.

peblair (approved):
@phated This looks good, but please create the issue you mentioned before merging.

stdlib/map.gr
```diff
@@ -274,8 +277,20 @@ export let fromList = (list) => {
...
+  let array = WasmI32.toGrain(allocateArray(length)): (Array<a>)
```
phated, line 289: Parse error when using a generic type parameter to annotate a function call.

stdlib/map.gr
phated: #783

stdlib/map.gr
phated: Should be good-to-go!

ospencer (approved):
Excellent!

PR #758, opened 2021-06-23
feat(compiler): Emit a warning when using Pervasives.(!=) with unsafe Wasm types

peblair:
Fixes issue found by @phated .

### DIFF
--- a/compiler/src/utils/warnings.re
+++ b/compiler/src/utils/warnings.re
@@ -105,7 +105,10 @@ let message =
     )
   | NonClosedRecordPattern(s) =>
     "the following fields are missing from the record pattern: " ++ s
-  | FuncWasmUnsafe(func) => "it looks like you are using " ++ func ++ " on two unsafe Wasm values here.\nThis is generally unsafe and will cause errors. Use one of the equivalent functions in `WasmI32`, `WasmI64`, `WasmF32`, or `WasmF64` instead.";
+  | FuncWasmUnsafe(func) =>
+    "it looks like you are using "
+    ++ func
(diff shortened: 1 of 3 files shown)


compiler/src/utils/warnings.re
```diff
@@ -105,7 +105,7 @@ let message =
     )
   | NonClosedRecordPattern(s) =>
     "the following fields are missing from the record pattern: " ++ s
-  | EqualWasmUnsafe => "it looks like you are using Pervasives.(==) on two unsafe Wasm values here.\nThis is generally unsafe and will cause errors. Use `WasmI32.eq`, `WasmI64.eq`, `WasmF32.eq`, or `WasmF64.eq` instead.";
+  | FuncWasmUnsafe(func) => "it looks like you are using " ++ func ++ " on two unsafe Wasm values here.\nThis is generally unsafe and will cause errors. Use `WasmI32.eq`, `WasmI64.eq`, `WasmF32.eq`, or `WasmF64.eq` instead.";
```
phated, line 108:
```suggestion
  | FuncWasmUnsafe(func) => "it looks like you are using " ++ func ++ " on two unsafe Wasm values here.\nThis is generally unsafe and will cause errors. Use the function from `WasmI32`, `WasmI64`, `WasmF32`, or `WasmF64` instead.";
```

peblair: @phated I fixed this independently. Feel free to re-suggest another wording if you don't like what I changed it to.

phated (approved):
This is similar to the hacks I was doing locally. Thanks!!

ospencer (approved):

PR #759, opened 2021-06-23
fix(stdlib): Correct usages of `==` and `!=` on WasmI32 values

### LINKED ISSUE #746: Figure out why `isNumber(y)` is needed in `equalHelp`
Commit 1705d1da (part of #726) reintroduced an extra `isNumber` check into `equalHelp` as a patch around failing test cases in #726. Since `(==)` is typed, this should not be required. We thought it might be caused by an accidental usage of `Pervasives.(==)` using unsafe Wasm types (which is undefined behavior), but the changes in #742 indicate that there are no such usages in the standard library. We should figure out why this call to `isNumber` is needed, and remove it once it is safe to do so.

phated:
Closes #746

The new warnings helped catch some more references to `Pervasives.(==)` and `Pervasives.(!=)` in the stdlib and tests. This fixes those issues and then removes the `isNumber(y)` workaround.

Probably should land after #758

### DIFF
--- a/stdlib/runtime/equal.gr
+++ b/stdlib/runtime/equal.gr
@@ -192,7 +192,7 @@ let rec heapEqualHelp = (heapTag, xptr, yptr) => {
     x == y
   } else if (isNumber(x)) {
     // Numbers have special equality rules, e.g. NaN != NaN
-    isNumber(y) && numberEqual(x, y)
+    numberEqual(x, y)
   } else {
     // Handle all other heap allocated things
     // Can short circuit if pointers are the same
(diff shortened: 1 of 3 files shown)


stdlib/runtime/equal.gr
```diff
@@ -192,7 +192,7 @@ let rec heapEqualHelp = (heapTag, xptr, yptr) => {
     x == y
   } else if (isNumber(x)) {
     // Numbers have special equality rules, e.g. NaN != NaN
-    isNumber(y) && numberEqual(x, y)
+    numberEqual(x, y)
```
peblair, line 195: 🔥

peblair (approved):

ospencer (approved):
Excellent!

PR #761, opened 2021-06-24
chore(graindoc): Print param/return types as inline-code

phated:
While previewing some GrainDoc output, I noticed that words coming after `Array<a>` got turned into a link (because that's an html element!), so I think we should always wrap types in an inline code block.

### DIFF
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -156,7 +156,7 @@ let to_markdown = docblock => {
          switch (attr) {
          | Param({attr_name, attr_type, attr_desc}) => [
              Markdown.code(attr_name),
-             Option.value(~default="", attr_type),
+             Option.fold(~none="", ~some=Markdown.code, attr_type),
              attr_desc,
            ]
          | _ =>
@@ -176,7 +176,7 @@ let to_markdown = docblock => {
     |> List.map((attr: Comments.Attribute.t) => {
          switch (attr) {
          | Returns({attr_type, attr_desc}) => [
-             Option.value(~default="", attr_type),
+             Option.fold(~none="", ~some=Markdown.code, attr_type),
              attr_desc,
            ]
          | _ =>


ospencer (approved):
Good find!

peblair:
@phated This is a good fix, but it is a little alarming to me that this was possible by mistake. Should we have the `Markdown` module HTML-sanitize inputs before generating the markdown? If not, why not?

phated:
@peblair I reviewed all of the markdown libraries on opam and none of them print markdown, they all parse it and output HTML. The markdown printing is extremely naive and I don't really want to write a full-featured markdown printer in OCaml/Reason if our goal is just to be self-hosted eventually.

peblair:
@phated I'm not sure if we're talking past each other. I understand that our Markdown printer is hand-rolled and naive, but I'm just saying that maybe we should throw something like this into it: https://discuss.ocaml.org/t/html-encoding-of-string/4289/4

Then we could make tweak the functions like so:
```diff
let paragraph = str => {
-  Format.sprintf("%s\n\n", str);
+  Format.sprintf("%s\n\n", html_escape(str));
};
```
Or maybe put a kwarg which controls it, or something else. This would preclude any accidental HTML output. It's just a thought. I'm not pushing that we really need to do it; it just feels like a code smell to me that this bug was possible.

phated:
@peblair I see your point, but I also am not sure where to apply this logic, as HTML entities are valid in markdown. Where do we draw the line at allowing users to insert HTML in their code comments? This one instance was me pulling a type signature from the AST and not thinking about the rendering.

phated:
While looking into the `@since` attribute, I realized that we will need to insert HTML directly into the markdown file `<details><summary> ...` to have a collapsible summary of the history. Also, the GFM spec identifies that all HTML elements are allowed (though they have a markdown extension that escapes some, like `<script>`). Since we are just generating markdown to be consumed by other tools, I think we should choose to not escape HTML.

peblair (approved):

ospencer:
Sorry for not chiming in sooner; I agree with @phated—I don't think we want to escape HTML since it'll make things a lot easier for us.

PR #762, opened 2021-06-24
chore(compiler): Remove extra space when printing tuples

phated:
While working on some docs, I noticed an extra space in tuples when being printed. This updates oprint to remove where I believe it was coming from.

### DIFF
--- a/compiler/src/typed/oprint.re
+++ b/compiler/src/typed/oprint.re
@@ -380,7 +380,7 @@ and print_out_type_2 = ppf =>
       fprintf(
         ppf,
         "@[<0>%a@]",
-        print_typlist(print_simple_out_type, ", "),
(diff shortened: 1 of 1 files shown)


ospencer (approved):
Thank you!!

peblair:
@phated I'm having a hard time grokking what the before/after behavior of this change is (I'm unclear whether there is an extra space compared to what I would expect or if what I would expect contains an extra space). Would you maybe be able to post an example?

phated:
@peblair I wish I could, but github markdown compresses spaces, even in code blocks.

peblair:
Okay, based on that answer, I think that I'm okay with the change. Thanks!

peblair (approved):

PR #764, opened 2021-06-26
feat(stdlib): String.encodeAtHelp bounds checks and optimization

cician:
This PR optimizes String.encodeAtHelp with bulk memory copy for specific case of UTF8 encoding. This mirrors what String.decodeRangeHelp already does, but in the opposite direction. More importantly, it adds bounds checks against size of the destination Bytes instance.


peblair (approved):
Great addition! Looks good to me, and nice catch with the bounds checking.
Other people's GitHub-fu is probably better than mine, but here's a tip to make reviewing this PR a breeze: check this box:
image

ospencer (approved):
Thanks @cician!

PR #765, opened 2021-06-29
fix(compiler): Refactor function return value incRef logic

### LINKED ISSUE #745: Memory leak returning objects from function

ospencer:
Fixes #745.

The tl;dr of the issue here was that we always would incRef the return value of a function, which could make some simple functions return a value with a ref count of +2. This PR adds some logic to insert appropriate incRefs.

Since this changes the wasm emitted for nearly all functions, nearly all of the snapshots were updated 😬 
I put all of the snapshots into a separate commit, so it should be easy to use the file filter or filter out that commit when reviewing the changes. I added a test so we can catch it if it happens again.

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -611,6 +611,33 @@ let cleanup_local_slot_instructions = (wasm_mod, env: codegen_env) => {
     );
   flatten(instrs);
 };
+let appropriate_incref = (wasm_mod, env, arg, b) =>
+  switch (b) {
+  | MArgBind(_, I32Type)
+  | MLocalBind(_, I32Type)
+  | MSwapBind(_, I32Type)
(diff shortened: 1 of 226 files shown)


peblair (approved):
Nice solution :+1:

marcusroberts (approved):
A cautious approval from me in that the code and approach looks good but I may easily miss a bug in the details.

PR #768, opened 2021-07-04
chore: Document memory management in Grain

peblair:
This will probably need some revisions based on reviewer feedback, but I wanted to have a go at documenting the Grain memory system, since a lot of details are not recorded anywhere and contributors such as @cician and @jozanza are interacting with low-level runtime code on a regular basis.

### DIFF
--- a/docs/contributor/memory_management.md
+++ b/docs/contributor/memory_management.md
@@ -1,0 +1,145 @@
+# Memory Management in Grain
+
+This guide documents Grain's memory management system. This document is aimed at those who are developing the code-generation pass of the compiler or writing low-level code inside of the Grain runtime; most users should not need to interact with the memory manager, and doing so risks corrupting the memory of your program (if the guidelines laid out here are not adhered to).
+
+We ultimately aim to replace Grain's in-house memory management with the WebAssembly GC instructions once [the proposal](https://github.com/WebAssembly/gc/blob/master/proposals/gc/Overview.md) is implemented, but, in the meantime, Grain implements its own memory allocator and garbage collector.
+
+## Memory Allocator
+Grain uses a [memory allocator](https://github.com/grain-lang/grain/blob/main/stdlib/runtime/malloc.gr) derived from the `malloc`/`free` example given in Kernighan and Ritchie's "The C Programming Language" (K&R C). This module exports the following values:
+
+```grain
+/**
+ * The amount of space reserved for runtime code which needs to be executed
+ * at boot before the malloc module is loaded
+ */
+export let _RESERVED_RUNTIME_SPACE: WasmI32
+
+/**
+ * Allocates the requested number of bytes, returning a pointer.
+ *
+ * @param nbytes: WasmI32 - The number of bytes to allocate
+ * @return WasmI32 - The pointer to the allocated region (8-byte aligned), or -1 if the allocation failed.
+ */
+export let malloc: (nbytes: WasmI32) => WasmI32
+
+/**
+ * Frees the given allocated pointer.
+ *
+ * @param ap: WasmI32 - The pointer to free
+ */
+export let free = (ap: WasmI32) => Void
+
+/**
+ * Returns the current free list pointer
+ * (used for debugging)
+ *
+ * @return WasmI32 - The free list pointer
+ */
+export let getFreePtr = () => WasmI32
+```
+
+These functions should be familiar to programmers who have used `malloc` and `free` in C (and C-like languages). For further reading, refer to this Wikipedia page: [C dynamic memory allocation](https://en.wikipedia.org/wiki/C_dynamic_memory_allocation). The semantics of these functions align near-identically with those of C's corresponding functions.
+
+Users should not call these functions directly in 99% of circumstances, and should instead use the `GC` module's wrappers (described below).
+
+## Garbage Collector
+
+Functional programming patterns often result in a large number of allocations, so it is imperative for developer experience that the language manages the creation and deletion of objects for users. Grain uses a reference-counting [garbage collector](https://github.com/grain-lang/grain/blob/main/stdlib/runtime/gc.gr) in order to determine when objects are no longer in active use by the program (and therefore able to have their memory freed).
+
+At a technical level, this is represented in-memory in the following way, for an n-byte heap object:
+```
+ [ 32-bit counter ][ 32-bit padding ][ n-byte payload ]
+ ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
+ {start address}                    {pointer used by Grain runtime}
+```
+Here, `{start address}` is the value returned by `Malloc.malloc` (and passed to `Malloc.free` when the object is cleaned up), and the `{pointer used by Grain runtime}` is what everything outside of the `GC` wrapper uses as the pointer to the object. In other words, we "hide" the reference count in the 8 bytes prior to the pointer.
+
+The interface provided by the `GC` module is similar (but not identical) to that which is provided by the `Malloc` module:
+```grain
+/**
+ * Allocates the requested number of bytes, returning a pointer with a reference count of 1.
+ *
+ * @param size: WasmI32 - The number of bytes to allocate
+ * @return WasmI32 - The pointer to the allocated region
+ */
+export let malloc = (size: WasmI32) => WasmI32
+
+/**
+ * Frees the given pointer. Using this pointer after it has been freed will result in undefined behavior.
+ *
+ * @param userPtr: WasmI32 - The pointer to free
+ * @return Void
+ */
+export let free = (userPtr: WasmI32) => Void
+
+/**
+ * Increments the reference count of the given pointer.
+ *
+ * @param userPtr: WasmI32 - The pointer whose reference count should be incremented
+ * @return WasmI32 - The given pointer
+ */
+export let incRef = (userPtr: WasmI32) => WasmI32
+
+/**
+ * Decrements the reference count of the given pointer if it is greater than zero
+ *
+ * @param userPtr: WasmI32 - The pointer whose reference count should be decremented
+ * @return WasmI32 - The given pointer
+ */
+export let decRefIgnoreZeros = (userPtr: WasmI32) => WasmI32
+
+/**
+ * Decrements the reference count of the given pointer. An error is thrown if the
+ * reference count is not greater than zero.
+ *
+ * @param userPtr: WasmI32 - The pointer whose reference count should be decremented
+ * @return WasmI32 - The given pointer
+ */
+export let decRef = (userPtr: WasmI32) => WasmI32
+```
+
+When writing code in the runtime, `decRefIgnoreZeros` should never be used. It exists solely to support the cleanup step of compiled functions.
+
+### Garbage Collection in Compiled Programs
+
+In most Grain programs, the above functions do not need to be invoked by users. Instead, the compiler automatically inserts calls to them where required. When developing in `compcore.re`, it is important to understand the conventions used by the garbage collector.
+
+**References from the local frame still count as references.** In other words, when a variable is stored inside of a WASM local, its reference count should be incremented. There is one exception to this rule: so-called "swap slots". These are special locals that Grain uses to store temporary results within a single compiled `Mashtree` instruction (similar to how CPU registers are used in native compilers). For performance reasons, we do not increment the reference counter when storing values in these slots. This is safe due to the fact that these values do not escape beyond a single `Mashtree` instruction; while they may still be physically present in the local, they are not read by later pieces of the program.
+
+For non-swap locals, this gives rise to the following pseudocode when storing a value in a local variable (this order is important to account for situations in which `val` is already in `local0`):
+```
+// We want to put `val` in `local0`
+local0 := {
+  tmp := incRef(val);
+  decRef(local0);
+  tmp
+}
+```
+In practice, this is facilitated by `Tuple` instructions Binaryen, similar to this:
+```
+// We want to put `val` in `local0`
+local0 := {
+  (incRef(val), decRef(local0))[0]
+}
+```
+
+**When a function returns a pointer, that pointer already has a reference count of at least one, which includes the reference from the local frame.** This means that, if the returned pointer does not escape the function, then a (net) single `decRef` should be invoked on that pointer in order to not leak memory.
+
+When a function is exiting, it is imperative to clean up all references to values in the local frame while preserving the return value. We do this by doing the following:
+
+- Calling `incRef` on the return value (which is in a local)
+- Calling `decRefIgnoreZeros` on all locals
+
+This should result in a net change of zero references on the return value, and a net change of -1 references on all other locals.
+
+## Disabling the Garbage Collector
+
+There are times in which the garbage collector is not desired in Grain code. The typical example of this is the implementation of the core runtime (for example the garbage collector itself), which involves working with native WebAssembly values. Consequently, Grain provides two mechanisms for disabling the garbage collector, either on a per-function or a per-file basis.
+
+First, it is possible to annotate functions with `@disableGC`. This will compile the annotated function with no garbage collection instructions. Second, it is possible to pass the `--compilation-mode=runtime` flag to the compiler (**NOTE** that this is not the only effect of this flag; see `runtime.md` for more details), which is typically done by placing this at the beginning of the file:
+```grain
+/* grainc-flags --compilation-mode=runtime */
+```
+When this flag is enabled, no functions in the file will be compiled with garbage collection instructions.
+
+Naturally, this means that it is incumbent upon the programmer to make the appropriate calls to `GC.incRef` and `GC.decRef` in order to not leak memory.
--- a/docs/contributor/memory_management.md
+++ b/docs/contributor/memory_management.md
@@ -5,7 +5,7 @@ This guide documents Grain's memory management system. This document is aimed at
 We ultimately aim to replace Grain's in-house memory management with the WebAssembly GC instructions once [the proposal](https://github.com/WebAssembly/gc/blob/master/proposals/gc/Overview.md) is implemented, but, in the meantime, Grain implements its own memory allocator and garbage collector.
 
 ## Memory Allocator
-Grain uses a [memory allocator](https://github.com/grain-lang/grain/blob/main/stdlib/runtime/malloc.gr) derived from the `malloc`/`free` example given in Kernighan and Ritchie's "The C Programming Language" (K&R C). This module exports the following values:
+Grain uses a [memory allocator](https://github.com/grain-lang/grain/blob/main/stdlib/runtime/malloc.gr) derived from the `malloc`/`free` example given in Kernighan and Ritchie's ["The C Programming Language"](https://kremlin.cc/k&r.pdf) (K&R C), pages 185-188 (PDF page 199). This module exports the following values:
 
 ```grain
 /**
@@ -104,7 +104,7 @@ When writing code in the runtime, `decRefIgnoreZeros` should never be used. It e
 
 In most Grain programs, the above functions do not need to be invoked by users. Instead, the compiler automatically inserts calls to them where required. When developing in `compcore.re`, it is important to understand the conventions used by the garbage collector.
 
-**References from the local frame still count as references.** In other words, when a variable is stored inside of a WASM local, its reference count should be incremented. There is one exception to this rule: so-called "swap slots". These are special locals that Grain uses to store temporary results within a single compiled `Mashtree` instruction (similar to how CPU registers are used in native compilers). For performance reasons, we do not increment the reference counter when storing values in these slots. This is safe due to the fact that these values do not escape beyond a single `Mashtree` instruction; while they may still be physically present in the local, they are not read by later pieces of the program.
+**References from the local frame still count as references.** In other words, when a variable is stored inside of a Wasm local, its reference count should be incremented. There is one exception to this rule: so-called "swap slots". These are special locals that Grain uses to store temporary results within a single compiled `Mashtree` instruction (similar to how CPU registers are used in native compilers). For performance reasons, we do not increment the reference counter when storing values in these slots. This is safe due to the fact that these values do not escape beyond a single `Mashtree` instruction; while they may still be physically present in the local, they are not read by later pieces of the program.
 
 For non-swap locals, this gives rise to the following pseudocode when storing a value in a local variable (this order is important to account for situations in which `val` is already in `local0`):
 ```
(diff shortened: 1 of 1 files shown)


docs/contributor/memory_management.md
```diff
@@ -0,0 +1,145 @@
+# Memory Management in Grain
+
+This guide documents Grain's memory management system. This document is aimed at those who are developing the code-generation pass of the compiler or writing low-level code inside of the Grain runtime; most users should not need to interact with the memory manager, and doing so risks corrupting the memory of your program (if the guidelines laid out here are not adhered to).
+
+We ultimately aim to replace Grain's in-house memory management with the WebAssembly GC instructions once [the proposal](https://github.com/WebAssembly/gc/blob/master/proposals/gc/Overview.md) is implemented, but, in the meantime, Grain implements its own memory allocator and garbage collector.
+
+## Memory Allocator
+Grain uses a [memory allocator](https://github.com/grain-lang/grain/blob/main/stdlib/runtime/malloc.gr) derived from the `malloc`/`free` example given in Kernighan and Ritchie's "The C Programming Language" (K&R C). This module exports the following values:
```
ospencer, line 8: Might as well link the book 😂 
```suggestion
Grain uses a [memory allocator](https://github.com/grain-lang/grain/blob/main/stdlib/runtime/malloc.gr) derived from the `malloc`/`free` example given in Kernighan and Ritchie's ["The C Programming Language"](https://kremlin.cc/k&r.pdf) (K&R C), pages 185-188 (PDF page 199). This module exports the following values:
```

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+When writing code in the runtime, `decRefIgnoreZeros` should never be used. It exists solely to support the cleanup step of compiled functions.
+
+### Garbage Collection in Compiled Programs
+
+In most Grain programs, the above functions do not need to be invoked by users. Instead, the compiler automatically inserts calls to them where required. When developing in `compcore.re`, it is important to understand the conventions used by the garbage collector.
+
+**References from the local frame still count as references.** In other words, when a variable is stored inside of a WASM local, its reference count should be incremented. There is one exception to this rule: so-called "swap slots". These are special locals that Grain uses to store temporary results within a single compiled `Mashtree` instruction (similar to how CPU registers are used in native compilers). For performance reasons, we do not increment the reference counter when storing values in these slots. This is safe due to the fact that these values do not escape beyond a single `Mashtree` instruction; while they may still be physically present in the local, they are not read by later pieces of the program.
```
ospencer, line 107:
```suggestion
**References from the local frame still count as references.** In other words, when a variable is stored inside of a Wasm local, its reference count should be incremented. There is one exception to this rule: so-called "swap slots". These are special locals that Grain uses to store temporary results within a single compiled `Mashtree` instruction (similar to how CPU registers are used in native compilers). For performance reasons, we do not increment the reference counter when storing values in these slots. This is safe due to the fact that these values do not escape beyond a single `Mashtree` instruction; while they may still be physically present in the local, they are not read by later pieces of the program.
```

ospencer (commented):
This is excellent! Thanks for putting this together 😄 

The only thing that I feel is missing here is an explanation that references are caller-owned and how the compiler injects incRef/decRef calls and what optimizations we do, e.g.

```
// instead of this
let f = x => {
  incRef(x)
  foo(x)
  decRef(x)
  incRef(x)
  bar(x)
  decRef(x)
}

// we just do this, since the calls all cancel each other out
let f = x => {
  foo(x)
  bar(x)
}

// and whoever called `f` is responsible for freeing `x` (or knowing that it will be freed)
```

If you feel like adding that in then 👏 but if you want me to do it then we can just merge this and you can make an issue for me 🤘

peblair:
@ospencer Section added! Let me know what you think.

ospencer (approved):
Looks great!

cician:
Can we have a mention of reference cycles? At least whether Grain's GC deals with them automatically or not.

peblair:
@cician I'll add a mention to it. For the moment, we do not support it (support was actually removed in the old garbage collector in https://github.com/grain-lang/grain/pull/461). We made the conscious decision to remove the cycle detection because Grain does not cleanly support cyclic data at the moment (it takes some extra ingenuity to assemble a cyclic structure currently) and the cycle detection made things very slow. We believe that it's a niche enough case that we can afford to wait until the Wasm GC lands. If it turns out to be a problem though, we'll explore ways to mitigate it.

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,168 @@
...
+ */
+export let _RESERVED_RUNTIME_SPACE: WasmI32
+
+/**
+ * Allocates the requested number of bytes, returning a pointer.
+ *
+ * @param nbytes: WasmI32 - The number of bytes to allocate
+ * @return WasmI32 - The pointer to the allocated region (8-byte aligned), or -1 if the allocation failed.
+ */
+export let malloc: (nbytes: WasmI32) => WasmI32
```
phated, line 23: Do we use `->` for signatures?
```suggestion
export let malloc: (nbytes: WasmI32) -> WasmI32
```

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,168 @@
...
+}
+```
+
+## Disabling the Garbage Collector
+
+There are times in which the garbage collector is not desired in Grain code. The typical example of this is the implementation of the core runtime (for example the garbage collector itself), which involves working with native WebAssembly values. Consequently, Grain provides two mechanisms for disabling the garbage collector, either on a per-function or a per-file basis.
+
+First, it is possible to annotate functions with `@disableGC`. This will compile the annotated function with no garbage collection instructions. Second, it is possible to pass the `--compilation-mode=runtime` flag to the compiler (**NOTE** that this is not the only effect of this flag; see `runtime.md` for more details), which is typically done by placing this at the beginning of the file:
+```grain
+/* grainc-flags --compilation-mode=runtime */
```
phated, line 164: I thought runtime mode has GC but no pervasives? Though there is a `--disable-gc` flag 🤔

ospencer: Runtime mode doesn't have GC (since it implements GC). We managed to get rid of `--compilation-mode=managed-runtime` which had that behavior, since you can just use `--no-pervasives` now.

phated: My mistake 🤦

cician:
I did stumble in a memory leak caused by cycles with recursive functions (#773), but maybe this could be worked around in the compiler without full cycles support in GC?

> @cician I'll add a mention to it. For the moment, we do not support it (support was actually removed in the old garbage collector in #461). We made the conscious decision to remove the cycle detection because Grain does not cleanly support cyclic data at the moment (it takes some extra ingenuity to assemble a cyclic structure currently) and the cycle detection made things very slow. We believe that it's a niche enough case that we can afford to wait until the Wasm GC lands. If it turns out to be a problem though, we'll explore ways to mitigate it.

phated:
@peblair were you planning to update this PR for the new memory management style? We should get this in sooner rather than later.

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Disabling the Garbage Collector
+
+There are times in which the garbage collector is not desired in Grain code. The typical example of this is the implementation of the core runtime (for example the garbage collector itself), which involves working with native WebAssembly values. Consequently, Grain provides two mechanisms for disabling the garbage collector, either on a per-function or a per-file basis.
+
+First, it is possible to annotate functions with `@disableGC`. This will compile the annotated function with no garbage collection instructions. Second, it is possible to pass the `--compilation-mode=runtime` flag to the compiler (**NOTE** that this is not the only effect of this flag; see `runtime.md` for more details), which is typically done by placing this at the beginning of the file:
```
ospencer, line 140: This isn't quite right—the flag to turn off GC at the file level is `--no-gc`. `--compilation-mode=runtime` doesn't emit GC instructions more because the GC doesn't exist rather than it being disabled. We can probably just mention that code written in runtime mode that interacts with user code needs manual management.

If we wanted we could also include some information like this:
`--no-gc`: disables the garbage collector, allocations are still done via `malloc`
`--compilation-mode=runtime`: GC doesn't exist, implicit allocations can't be reclaimed

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,177 @@
...
+
+**References are callee-owned.** This has a couple of implications:
+
+- If a function is called with an argument, it is the responsibility of that function to `decRef` that argument
+- If a value is passed to a function with a reference count of `n`, when it returns, it will have the a reference count of `n-1` (unless it was stored in an external location, in which case the reference count may be higher).
+
+**Functions contain references to themselves.** In order to support closures, Grain's compiler emits code which passes the current closure to each function as an extra argument. With respect to the calling convention, this extra argument is the same as normal arguments, so it follows the same `incRef`-before-function-call pattern as standard arguments (and must be `decRef`ed before functions return). This is particularly important to keep in mind when writing `@disableGC` code, as discussed below.
```
ospencer, line 134: I'd love to see this info before the `Reference counts are incremented upon access` section! I like having the concepts and then getting into our implementation of it. It might even be better to have two subsections of implementations—`Reference counts are incremented upon access` for how it's implemented in the compiler, and then this other bit below for how it should be implemented in Grain code 😎

peblair: @ospencer I did the rearrangement, but I am not sure what would be the best way to break this apart into subsections as you describe. Do you want to take another look and see if it now looks like what you'd expect?

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
+# Memory Management in Grain
+
+This guide documents Grain's memory management system. This document is aimed at those who are developing the code-generation pass of the compiler or writing low-level code inside of the Grain runtime; most users should not need to interact with the memory manager, and doing so risks corrupting the memory of your program (if the guidelines laid out here are not adhered to).
```
phated, line 3:
```suggestion
This guide documents Grain's memory management system. This document is aimed at those who are developing the code-generation pass of the compiler or writing low-level code inside of the Grain runtime; most users should not need to interact with the memory manager, and doing so risks corrupting the memory of your program if the guidelines laid out here are not adhered to.
```

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+
+We ultimately aim to replace Grain's in-house memory management with the WebAssembly GC instructions once [the proposal](https://github.com/WebAssembly/gc/blob/master/proposals/gc/Overview.md) is implemented, but, in the meantime, Grain implements its own memory allocator and garbage collector.
```
phated, line 5:
```suggestion
We ultimately aim to replace Grain's bespoke memory management with the WebAssembly GC instructions once [the proposal](https://github.com/WebAssembly/gc/blob/master/proposals/gc/Overview.md) is implemented—in the meantime, Grain implements its own memory allocator and garbage collector.
```
😛

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+ * The amount of space reserved for runtime code which needs to be executed
+ * at boot before the malloc module is loaded
+ */
+export let _RESERVED_RUNTIME_SPACE: WasmI32
+
+/**
+ * Allocates the requested number of bytes, returning a pointer.
+ *
+ * @param nbytes: WasmI32 - The number of bytes to allocate
+ * @return WasmI32 - The pointer to the allocated region (8-byte aligned), or -1 if the allocation failed.
```
phated, line 22:
```suggestion
 * @param nbytes: The number of bytes to allocate
 * @returns The pointer to the allocated region (8-byte aligned), or -1 if the allocation failed.
```
Might as well follow graindoc?

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+ */
+export let malloc: (nbytes: WasmI32) -> WasmI32
+
+/**
+ * Frees the given allocated pointer.
+ *
+ * @param ap: WasmI32 - The pointer to free
```
phated, line 29:
```suggestion
 * @param ap: The pointer to free
```
Might as well follow graindoc?

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+ */
+export let free = (ap: WasmI32) => Void
+
+/**
+ * Returns the current free list pointer
+ * (used for debugging)
```
phated, line 35:
```suggestion
 * Returns the current free list pointer (used for debugging)
```

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * @return WasmI32 - The free list pointer
```
phated, line 37:
```suggestion
 * @returns The free list pointer
```
Might as well follow graindoc?

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+ */
+export let getFreePtr = () => WasmI32
+```
+
+These functions should be familiar to programmers who have used `malloc` and `free` in C (and C-like languages). For further reading, refer to this Wikipedia page: [C dynamic memory allocation](https://en.wikipedia.org/wiki/C_dynamic_memory_allocation). The semantics of these functions align near-identically with those of C's corresponding functions.
+
+These functions generally should not be called directly in 99% of circumstances, and the `GC` module's wrappers should be used instead (described below).
```
phated, line 44:
```suggestion
These functions generally should not be called directly in 99% of circumstances, and the Garbage Collector wrappers should be used instead (described below).
```

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Garbage Collector
+
+Functional programming patterns often result in a large number of allocations, so it is imperative for developer experience that the language manages the creation and deletion of objects for users. Grain uses a reference-counting [garbage collector](https://github.com/grain-lang/grain/blob/main/stdlib/runtime/gc.gr) in order to determine when objects are no longer in active use by the program (and therefore able to have their memory freed).
```
phated, line 48:
```suggestion
Functional programming patterns often result in a large number of allocations, so it is imperative for developer experience that the language manages the creation and deletion of objects for users. Grain uses a reference-counting [garbage collector](https://github.com/grain-lang/grain/blob/main/stdlib/runtime/gc.gr) in order to determine when objects are no longer in active use by the program and therefore able to have their memory freed.
```
Just trying to remove parens

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+Here, `{start address}` is the value returned by `Malloc.malloc` (and passed to `Malloc.free` when the object is cleaned up), and the `{pointer used by Grain runtime}` is what everything outside of the `GC` wrapper uses as the pointer to the object. In other words, we "hide" the reference count in the 8 bytes prior to the pointer.
+
+The interface provided by the `GC` module is similar (but not identical) to that which is provided by the `Malloc` module:
+
+```grain
+/**
+ * Allocates the requested number of bytes, returning a pointer with a reference count of 1.
+ *
+ * @param size: WasmI32 - The number of bytes to allocate
+ * @return WasmI32 - The pointer to the allocated region
```
phated, line 67:
```suggestion
 * @param size: The number of bytes to allocate
 * @returns The pointer to the allocated region
```
Might as well follow graindoc?

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+ */
+export let malloc = (size: WasmI32) -> WasmI32
+
+/**
+ * Frees the given pointer. Using this pointer after it has been freed will result in undefined behavior.
+ *
+ * @param userPtr: WasmI32 - The pointer to free
+ * @return Void
```
phated, line 75:
```suggestion
 * @param userPtr: The pointer to free
```
Might as well follow graindoc?

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+ */
+export let free = (userPtr: WasmI32) -> Void
+
+/**
+ * Increments the reference count of the given pointer.
+ *
+ * @param userPtr: WasmI32 - The pointer whose reference count should be incremented
+ * @return WasmI32 - The given pointer
```
phated, line 83:
```suggestion
 * @param userPtr: The pointer whose reference count should be incremented
 * @returns The given pointer
```
Might as well follow graindoc?

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+ */
+export let incRef = (userPtr: WasmI32) -> WasmI32
+
+/**
+ * Decrements the reference count of the given pointer if it is greater than zero
+ *
+ * @param userPtr: WasmI32 - The pointer whose reference count should be decremented
+ * @return WasmI32 - The given pointer
```
phated, line 91:
```suggestion
 * @param userPtr: The pointer whose reference count should be decremented
 * @returns The given pointer
```
Might as well follow graindoc?

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+ */
+export let decRefIgnoreZeros = (userPtr: WasmI32) -> WasmI32
+
+/**
+ * Decrements the reference count of the given pointer. An error is thrown if the
+ * reference count is not greater than zero.
+ *
+ * @param userPtr: WasmI32 - The pointer whose reference count should be decremented
+ * @return WasmI32 - The given pointer
```
phated, line 100:
```suggestion
 * @param userPtr: The pointer whose reference count should be decremented
 * @returns The given pointer
```
Might as well follow graindoc?

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+ */
+export let decRef = (userPtr: WasmI32) -> WasmI32
+```
+
+When writing code in the runtime, `decRefIgnoreZeros` should never be used. It exists solely to support the cleanup step of compiled functions. The reference count-managing functions are safe to use with non-pointers; if a non-pointer is passed
+to them, it will be returned without any side effects. **However**, it should be noted that these functions are not safe to use with arbitrary (untagged) `WasmXX` (`WasmI32`/`WasmI64`/etc) values, which is why the use of such types is only permitted in `@disableGC` blocks (so that the compiler does not insert `incRef`/`decRef` calls).
```
phated, line 106:
```suggestion
to them, it will be returned without any side effects. **However**, it should be noted that these functions are not safe to use with arbitrary (untagged) `WasmXX` (`WasmI32`/`WasmI64`/etc) values, which is why the use of such types is only permitted in `@disableGC` blocks—so that the compiler does not insert `incRef`/`decRef` calls.
```

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Garbage Collection in Compiled Programs
+
+In most Grain programs, the above functions do not need to be invoked by users. Instead, the compiler automatically inserts calls to them where required. When developing in `compcore.re`, it is important to understand the conventions used by the garbage collector.
+
+**Functions contain references to themselves.** In order to support closures, Grain's compiler emits code which passes the current closure to each function as an extra argument. With respect to the calling convention, this extra argument is the same as normal arguments, so it follows the same `incRef`-before-function-call pattern as standard arguments (and must be `decRef`ed before functions return), as described in the following section. This is particularly important to keep in mind when writing `@disableGC` code, as discussed in "Disabling the Garbage Collector" below.
```
phated, line 112:
```suggestion
**Functions contain references to themselves.** In order to support closures, Grain's compiler emits code which passes the current closure to each function as an extra argument. With respect to the calling convention, this extra argument is the same as normal arguments, so it follows the same `incRef`-before-function-call pattern as standard arguments and must be `decRef`ed before functions return, as described in the following section. This is particularly important to keep in mind when writing `@disableGC` code, as discussed in "Disabling the Garbage Collector" below.
```

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+
+**Reference counts are incremented upon access.** With some exceptions in generated code, reference counts are incremented whenever a value is accessed (loaded from storage slots such as arguments, locals, and globals). This streamlines the conventions in generated code, as function calls require no specific `incRef` invocation for arguments (see "References are callee-owned" subsection below). For clarity, a couple of notes:
+
+- If a value is stored somewhere, no extra `incRef`/`decRef` calls are needed (again, there are rare exceptions to this in the compiler, specifically regarding "swap slots", which are local variables used as registers).
```
phated, line 116:
```suggestion
- If a value is stored somewhere, no extra `incRef`/`decRef` calls are needed. There are rare exceptions to this in the compiler, specifically regarding "swap slots", which are local variables used as registers.
```

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+for local_variable in local_variables:
+    decref(local_variable)
+return load_swap(return_value) # see note below
+```
+
+Note that `load_swap` is one of the aforementioned exceptions which does not `incRef` the value upon access. Swap slots are used by the compiler to temporarily store intermediate values during computation, and they do not need to have their contents `incRef`ed; i.e. since they are only used for intermediate values, any references they hold are treated as weak and do not impact the reference count.
+
+**References are callee-owned.** This has a couple of implications:
+
+- If a function is called with an argument, it is the responsibility of that function to `decRef` that argument
```
phated, line 133:
```suggestion
- If a function is called with an argument, it is the responsibility of that function to `decRef` that argument.
```
nit

phated (changes requested):
Just a bunch of nits from me. I generally understand this stuff at the high-level and this is helping to solidify it.

docs/contributor/memory_management.md
```diff
@@ -0,0 +1,182 @@
...
+ */
+export let decRef = (userPtr: WasmI32) -> WasmI32
+```
+
+When writing code in the runtime, `decRefIgnoreZeros` should never be used. It exists solely to support the cleanup step of compiled functions. The reference count-managing functions are safe to use with non-pointers; if a non-pointer is passed
+to them, it will be returned without any side effects. **However**, it should be noted that these functions are not safe to use with arbitrary (untagged) `WasmXX` (`WasmI32`/`WasmI64`/etc) values, which is why the use of such types is only permitted in `@disableGC` blocks (so that the compiler does not insert `incRef`/`decRef` calls).
+
+### Garbage Collection in Compiled Programs
+
+In most Grain programs, the above functions do not need to be invoked by users. Instead, the compiler automatically inserts calls to them where required. When developing in `compcore.re`, it is important to understand the conventions used by the garbage collector.
```
ospencer, line 110: Want to remove any thought that users sometimes need to do it:
```suggestion
In normal Grain programs, the above functions do not need to be invoked by users. Instead, the compiler automatically inserts calls to them where required. When developing in `compcore.re`, it is important to understand the conventions used by the garbage collector.
```

docs/contributor/memory_management.md
ospencer: Looks great!

marcusroberts (approved):
This helped my understanding a lot!

docs/contributor/memory_management.md
peblair: You sure about this one? I thought "`GC` module" would be better at making it explicit _where_ these wrappers are.

phated: I'm fine if we keep the current language.

phated (approved):
Looking great to me! Thanks @peblair 🎉

ospencer (approved):
I feel like Thanos sitting in the sun, taking this all in. 😤

PR #770, opened 2021-07-06
fix(stdlib): Fixed memory leak in the print function.

cician:
This PR fixes a memory leak in the print function.

The intermediate string created was not freed. I'm not entirely sure I'm using the "right" free. Or maybe I should call decRef instead?

ps: I was using print to debug memory leaks, which lead me to the memory leak in print. So then I used print to debug the memory leak in print. 🤟

### DIFF
--- a/stdlib/runtime/string.gr
+++ b/stdlib/runtime/string.gr
@@ -572,7 +572,8 @@ export let toString = (value) => {
 
 @disableGC
 export let print = (value) => {
-  let ptr = WasmI32.fromGrain(toString(value))
+  let s = toString(value)
+  let ptr = WasmI32.fromGrain(s)
   let buf = Memory.malloc(37n)
   let iov = buf
   let written = buf + 32n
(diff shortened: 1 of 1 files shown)


ospencer (changes requested):
Nice find! We shouldn't need the conditional check here though. It looks like we're missing a `Memory.incRef` on line 283, and once that's in you should be able to remove the `if`. And yeah, we should use `Memory.decRef` on the ptr here instead of free 😃 

Technically `free` is fine, but `decRef` would be more correct if we were to look at `print` in isolation.

stdlib/runtime/string.gr
```diff
@@ -280,6 +280,10 @@ let rec heapValueToString = (ptr, extraIndents, toplevel) => {
   match (tag) {
     t when t == Tags._GRAIN_STRING_HEAP_TAG => {
       if (toplevel) {
+        // We can just return the pointer since its already a string and
+        // doesn't need escaping in this case, but by convention its reference
+        // count needs to be incremented.
+        Memory.incRef(ptr)
```
peblair, line 286: I had to think about this one a bit, but I _think_ it doesn't leak memory 🙂

peblair (approved):
Great fix! Thanks!

ospencer (approved):
Thanks again @cician!

PR #771, opened 2021-07-06
fix(stdlib): Add @disableGC annotation to WasmXX-using getSize function in Bytes

peblair:

PR #772, opened 2021-07-06
feat(compiler)!: Add typed well-formedness pass which forbids usage of WasmXX values outside of `@disableGC` context

peblair:
Implementation based on strategy discussed on Discord. Marking this as a draft until we decide that this is the approach we want to take.

(Blocked by #771)


phated (approved):
Looks great! And I'd like to get this in 0.4 since it is breaking. Would hate for it to linger and need lots of rebase before 0.5

Thanks for getting it out of draft @peblair!!!

marcusroberts (approved):
I think the stdlib changes are in this PR because without them these files would now fail the new well formed pass?

If so LGTM!

ospencer (approved):

jozanza (approved):

PR #776, opened 2021-07-10
feat(compiler): Emit unoptimized wat file when `--debug` and `--wat` are enabled

peblair:
This is a QOL improvement PR which facilitates easier debugging of `compcore.re`. When `--wat` and `--debug` are passed to the compiler, it will emit an additional `<file>.unoptimized.wat` file containing the WebAssembly module _prior_ to the call to `Module.optimize` (which directly reflects the instructions emitted in `compcore.re`).

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -3501,7 +3501,6 @@ let compile_wasm_module = (~env=?, ~name=?, prog) => {
     Bytes.to_string(serialized_cmi),
   );
   validate_module(~name?, wasm_mod);
-  Module.optimize(wasm_mod);
   wasm_mod;
 };
 
--- a/compiler/src/codegen/emitmod.re
+++ b/compiler/src/codegen/emitmod.re
@@ -12,6 +12,16 @@ let emit_module = ({asm, signature}, outfile) => {
     output_string(oc, sig_string);
     close_out(oc);
   };
+  // Emit unoptimized version of module if debug and wat set:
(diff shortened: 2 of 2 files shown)


phated (approved):
I think this is a solid change. Are we happy with the Module.optimize to be moved out of compcore and into emitmod?

ospencer (commented):
I can't say I'm a fan of moving `Module.optimize` into emitmod, since I wouldn't expect that to be where optimizations would get run. I think instead of this, I'd want a flag to disable Binaryen optimizations that would affect the `wat` like other flags do. Maybe `-O3` (default) is Grain + Binaryen optimizations, `-O2` is just Grain optimizations, and we reserve `-O1` for some limited set of Grain optimizations or something like that.

Then you could pass something like `grain -O2 --wat test.gr`, and you could get whatever level of optimizations you want for your `.wat` file.

peblair:
I understand that the new location of `Module.optimize` is not what would necessarily be expected. We can make the change described by @ospencer, but I just want to clarify that the reason it was done this was was so that the emitted Grain program is still optimized (and runnable). By moving to the `-O2`-style, you would need to run a separate command to generate the appropriate wat file (since Binaryen optimizations are currently required).

ospencer:
Right. That's a separate (fixable) bug. I'll send a PR to get that working again.

marcusroberts (approved):

peblair:
@ospencer We should close this, yes?

ospencer:
@peblair Yes, let's close.

PR #777, opened 2021-07-10
feat(compiler): Call known functions directly

ospencer:
This PR changes wasm output to emit a `call` instruction (rather than `call_indirect`) for all functions we know about. (You may ask, what's a function we don't know about? In `let foo = f => f()`, we don't statically know what function a user could pass to `foo`.)

While this technically could be considered an optimization, I didn't implement it as a standard optimization pass since I thought it'd complicate the ANF IR and make other optimizations harder (and be quite expensive to compute compared to this simpler way, _and_ I don't think this is something you should be able to turn off). It does, however, open up the doors for more Binaryen optimizations.

The easiest way to see the effect of this change is to look at the snapshots. You'll see there's an added bonus that the function names used in the code are now _always_ in the wasm output, which makes debugging much easier.


compiler/test/__snapshots__/basic_functionality.61c58118.0.snapshot
```diff
@@ -17,55 +17,49 @@ basic functionality › block_no_expression
...
-        (local.tee $0
-         (tuple.extract 0
-          (tuple.make
-           (call $import_GRAIN$MODULE$runtime/gc_0_malloc_0
-            (global.get $import_GRAIN$MODULE$runtime/gc_0_GRAIN$EXPORT$malloc_0)
-            (i32.const 16)
-           )
-           (i32.const 0)
-          )
+  (drop
```
ospencer, line 25: In this case, Binaryen just inlined the call, so there's no `call` at all anymore. You'll notice that `f` here is still in the output even though it was inlined—that's what I'll be tackling next.

phated (approved):
This is **super** cool! 🎉  I reviewed the impl, but mostly just skimmed the snapshots.

peblair (commented):
This change looks good! I think it may be prudent to add some test cases which cover shadowed function names (maybe some funky stuff with combinations of shadowing/indirect calls too), just to cover our bases, but with that, I think this is good to go.

ospencer:
@peblair Added some tests. Lots of `foo`.

marcusroberts (approved):

PR #779, opened 2021-07-16
feat: Add format command to the Grain CLI

marcusroberts:
Here's the first version of the Grain formatter.

We also need to implement a reformat in place mode for the CLI, for now this works with format on save with the LSP

You can grain format filename or grain format  without a filename to reformat stdin.   Both modes currently send the reformatted code to stdout.

We use the Rescript print engine with attribution (converted from OCaml to Reason).  Should this be a separate module?

There's some useful debugging code left in (but commented out) as I suspect I'll be needing it once the comments come in.


marcusroberts:
One issue to solve is adding line breaks in, and handling LR/CR for Windows.

compiler/test/formatter_outputs/list_sugar.gr
```diff
@@ -0,0 +1,36 @@
+let list = [1]
+
+let list2 = [1,2]
+
+let list3 = [1,2,3]
+
+let list4 = [[1,2],[3,4]]
+
+let list5 = [[1, 2, 3 ], [3, 4, 5 ], [5, 6, 7 ] ]
```
peblair, line 9: @marcusroberts What's going on with the spaces here? (from the filename, this is what the formatter produces, right?)

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2773 @@
...
+module Doc = Res_doc;
+
+let get_raw_pos_info = (pos: Lexing.position) => (
+  pos.pos_fname,
+  pos.pos_lnum,
+  pos.pos_cnum - pos.pos_bol,
+  pos.pos_bol,
+);
+
+let getOriginalCode =
```
peblair, line 15: I think we should try to be consistent with the rest of our Reason codebase and stick to snake_case (for all functions and variables)

compiler/grainformat/walktree.re
```diff
@@ -0,0 +1,311 @@
...
+
+type node_t =
+  | Code(Grain_parsing.Location.t)
+  | Comment((Grain_parsing.Location.t, Grain_parsing.Parsetree.comment));
+
+type node_tree =
+  | Empty
+  | Node(node_tree, node_t, node_tree);
+
+let allLocations: ref(list(node_t)) = ref([]);
```
peblair, line 11: snake_case also

peblair (changes requested):
This looks awesome! I have gone ahead and done a preliminary review, but I think this will be ready to land soon.

compiler/test/formatter_outputs/list_sugar.gr
marcusroberts: That is a good question, taking a look now

marcusroberts: Well spotted, it was an out of date test not in the suite!

compiler/grainformat/reformat.re
marcusroberts: Fixed. Let me know if you spot any more

phated:
I'm going to hijack the title a bit, this isn't a new feature of the compiler, but instead a new tool. 🍻

peblair (commented):
This is beautiful stuff! I've left a few comments (mostly performance-related), but this is really awesome work!

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+and is_empty_array = (expr: Parsetree.expression) => {
+  switch (expr.pexp_desc) {
+  | PExpId(loc) =>
+    let loc_txt =
+      switch (loc.txt) {
+      | IdentName(name) => name
+      | _ => ""
+      };
+
+    if (loc_txt == "[]") {
```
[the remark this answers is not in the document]
ospencer: It's impossible to make a `PExpId` for an empty array, so this is definitely for a list. We should probably rename this function 🙂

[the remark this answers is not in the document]
ospencer: Also, the `if` statement here isn't necessary.

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+        Doc.softLine,
+        Doc.rbracket,
+      ]),
+    ),
+  );
+}
+
+and resugar_list_inner =
+    (expressions: list(Parsetree.expression), parent_loc) => {
+  let arg1 = List.nth(expressions, 0);
```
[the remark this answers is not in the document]
ospencer: Grain syntax makes it impossible to construct a list cons without the right number of arguments, so it's probably fine. A comment would be good here!

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+            } else {
+              Doc.nil;
+            },
+          ]),
+          false,
+        );
+      };
+
+    | PPatOr(pattern1, pattern2) => (
+        Doc.text("/* Formatter PPatOr error */"),
```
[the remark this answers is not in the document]
ospencer: Should we instead be throwing an error/refuse to format the file? This looks like it'd break the code if we added support for this without formatter support.

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+    Doc.concat([Doc.lparen, clean_pattern, Doc.rparen]);
+  } else {
+    clean_pattern;
+  };
+}
+and print_constant = (c: Parsetree.constant) => {
+  let print_c =
+    switch (c) {
+    | PConstNumber(PConstNumberInt(i)) => Printf.sprintf("%s", i)
+    | PConstNumber(PConstNumberFloat(f)) => Printf.sprintf("%s", f)
```
ospencer, line 684: Do we want to do any reformatting of numbers? Remove/add trailing zeros for a float, etc?

ospencer: FWIW I'd like to preserve the form a number is written in, i.e. `0x77` should stay in hex. (Looks like this is the case, but I figure I'd call it out if we were to make other changes.)

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+  | IdentExternal(externalIdent, second) =>
+    Doc.concat([
+      print_ident(externalIdent),
+      Doc.text("."),
+      Doc.text(second),
+    ])
+  };
+}
+
+and print_imported_ident = (ident: Identifier.t) => {
```
ospencer, line 721: Any reason why this should be separate from the regular ident printer?

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+            ),
+            Doc.rparen,
+          ]);
+        },
+      ),
+      Doc.text(" -> "),
+      print_type(parsed_type),
+    ])
+
+  | PTyTuple(parsed_types) =>
```
ospencer, line 834: We'll need to add a trailing comma for single-argument tuples.

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+        Doc.text("<"),
+        Doc.join(
+          Doc.concat([Doc.comma, Doc.space]),
+          List.map(typ => {print_type(typ)}, parsedtypes),
+        ),
+        Doc.text(">"),
+      ]);
+    };
+  | PTyPoly(locationstrings, parsed_type) =>
+    Doc.text("/* Formatter error PTyPoly*/")
```
ospencer, line 855: Same deal, we should probably refuse to format the file?

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+      Doc.text(function_name),
+      Doc.space,
+      second_brackets,
+    ]);
+  } else {
+    let func_name = get_function_name(func);
+
+    if (func_name == "[...]") {
+      resugar_list(~expressions, ~parent_loc, ~original_source);
+    } else if (func_name == "throw"
```
ospencer, line 994: Let's add a constant for these that we can check against.

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+            ~parentIsArrow=false,
+            ~endChar=None,
+            ~parent_loc,
+            ~original_source,
+          );
+
+        if (left == first) {
+          // +=, -=, *=, /=, and %=
+          switch (trimmed_operator) {
+          | "+"
```
ospencer, line 1937: An infix list would also be really good.

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+        let expression =
+          print_expression(
+            ~expr=vb.pvb_expr,
+            ~parentIsArrow=false,
+            ~endChar=None,
+            ~original_source,
+            ~parent_loc,
+          );
+
+        // fix needed here?, we may  need to work out if all arguments fit on the line or will break
```
ospencer, line 2086:
```suggestion
        // TODO: fix needed here? we may need to work out if all arguments fit on the line or will break
```

Probably also good to open an issue.

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+let toplevel_print =
+    (
+      ~data: Parsetree.toplevel_stmt,
+      ~original_source: list(string),
+      ~previous_line: int,
+    ) => {
+  let attributes = data.ptop_attributes;
+
+  let (leading_comments, _trailing_comments) =
+    Walktree.partition_comments(data.ptop_loc, None); //
```
ospencer, line 2514:
```suggestion
    Walktree.partition_comments(data.ptop_loc, None);
```

compiler/test/formatter_outputs/ignores.gr
```diff
@@ -0,0 +1,20 @@
+let x = 5
+
+// formatter-ignore
+
+let x2 = [ 1,
```
ospencer, line 5: This really caught me off guard before I realized there was a formatter-ignore haha

compiler/test/formatter_outputs/list_sugar.gr
```diff
@@ -0,0 +1,35 @@
...
+let list4 = [[1, 2], [3, 4], [5, 6]]
+
+let list5 = [[1, 2, 3], [3, 4, 5], [5, 6, 7]]
+
+let arr = [
+  "dsdsdshdfjkhdsjhfdsjkhfjdshjfkhsdjkfh",
+  "dsdsdshdfjkhdsjhfdsjkhfjdshjfkhsdjkfh",
+  "dsdsdshdfjkhdsjhfdsjkhfjdshjfkhsdjkfh",
+  "dsdsdshdfjkhdsjhfdsjkhfjdshjfkhsdjkfh"
+  ]
```
ospencer, line 16: I'd prefer if this weren't indented, personally.

compiler/test/formatter_outputs/variants.gr
```diff
@@ -0,0 +1,12 @@
+enum Animals {
+  VeryVeryLong,
+  VeryVeryLongVeryVeryLong,
+  VeryVeryLongVeryVeryLongVeryVeryLong,
+}
+
+enum SmallAnimals { Abc, Def, }
+
+enum CabbageColor { Red, Green, }
+enum Veggie { Squash, Cabbage(CabbageColor), Broccoli, }
```
ospencer, line 10: I'm still in the no trailing comma if they all fit on the same line camp 😛

ospencer (changes requested):
Really looking forward to this! Definitely the tool I've been waiting for 😎

compiler/grainformat/reformat.re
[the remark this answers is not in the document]
marcusroberts: Now I have the original source I can replace the AST with its original source code (much like a // formatter-ignore) which I think is a better approach?

compiler/test/formatter_outputs/variants.gr
marcusroberts: I think I am too.  I took this short cut of always adding a trailing comma because of adding trailing comments (easier to always add the comma as I can then always safely add a comment after) but I think the printer has support for that.

compiler/grainformat/grainformat.re
```diff
@@ -0,0 +1,157 @@
...
+    switch (filename) {
+    | None =>
+      // force from stdin for now
+
+      /* read from stdin until we get end of buffer */
+      try(
+        while (true) {
+          let line = read_line();
+          program_str := program_str^ ++ line ++ "\n";
+          lines := lines^ @ [line];
```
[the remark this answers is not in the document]
marcusroberts: Nice catch.  I only started collecting the lines to implement ignore-formatter, this is a great optimisation, thanks!

compiler/grainformat/grainformat.re
```diff
@@ -0,0 +1,157 @@
...
+        }
+      ) {
+      | exn => ()
+      };
+
+      Grain_utils.Config.base_path := dirname(filenm);
+      Compile.compile_file(~hook=stop_after_parse, filenm);
+    };
+
+  // switch (compile_state) {
```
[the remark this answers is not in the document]
marcusroberts: Cleaned up.

compiler/grainformat/walktree.re
```diff
@@ -0,0 +1,280 @@
+open Grain_parsing;
```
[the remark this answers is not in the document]
marcusroberts: Great idea, done!

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+
+let get_original_code =
+    (location: Grain_parsing.Location.t, source: list(string)) => {
+  let (_, startline, startc, _) = get_raw_pos_info(location.loc_start);
+  let (_, endline, endc, _) = get_raw_pos_info(location.loc_end);
+
+  let text = ref("");
+  if (List.length(source) > endline - 1) {
+    if (startline == endline) {
+      let full_line = List.nth(source, startline - 1);
```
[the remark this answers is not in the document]
marcusroberts: Great idea, done.

compiler/grainformat/reformat.re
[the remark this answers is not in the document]
marcusroberts: Function renamed and if removed.

compiler/grainformat/reformat.re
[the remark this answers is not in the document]
marcusroberts: I've put a test on the number of expressions just to be sure.  The expressions come through as a list from the parser, so I think I'll leave as a list for now as I'm not sure of the cost of converting to an array vs this iteration anyway.

compiler/grainformat/reformat.re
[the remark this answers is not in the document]
marcusroberts: Resolved by replacing the pattern with the original source - the file will then just format without updating the unsupported syntax.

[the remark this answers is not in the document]
peblair: @marcusroberts I like that solution :+1:

compiler/test/formatter_outputs/ignores.gr
marcusroberts: :-D

compiler/test/formatter_outputs/list_sugar.gr
marcusroberts: Agreed, fixed.

compiler/test/formatter_outputs/variants.gr
marcusroberts: Fixed.

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+    (
+      ~func: Parsetree.expression,
+      ~expressions: list(Parsetree.expression),
+      ~parent_loc: Location.t,
+      ~original_source: list(string),
+    ) => {
+  let function_name = get_function_name(func);
+
+  if (prefixop(function_name)) {
+    if (List.length(expressions) == 1) {
```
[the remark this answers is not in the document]
marcusroberts: Could you expand a little - I agree on the  unguarded List. hd, so switch on the function name or the expression lengths?

compiler/grainformat/reformat.re
marcusroberts: Confirmed hex values print through and I've added tests for all the other value types too

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+                Doc.rbrace,
+              ]),
+            ),
+          ])
+        };
+
+      if (parentIsArrow) {
+        Doc.group(
+          Doc.concat([
+            Doc.text("if "),
```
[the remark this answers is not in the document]
marcusroberts: Thanks, fixed.

compiler/grainformat/reformat.re
marcusroberts: Reformatting of numbers is a good question re leading spaces. I  think this happens as they'll have been handled by the lexer

marcusroberts:
(superceded by 829)

compiler/grainformat/reformat.re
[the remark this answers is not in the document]
peblair: I'd say you could do both:
```reason
switch(expressions) {
  | [first] when prefixop(function_name) => {...}
  | _ when prefixop(function_name) => {formatter error}
  | [first, second] when infixop(function_name) => {...}
  | _ when infixop(function_name) => {formatter error}
  | _ => {...}
}
```
if you're willing to sacrifice a bit of precision in the error messages, you could even do
```reason
switch(expressions) {
  | [first] when prefixop(function_name) => {...}
  | [first, second] when infixop(function_name) => {...}
  | _ when (prefixop(function_name) || infixop(function_name)) => {formatter error}
  | _ => {...}
}
```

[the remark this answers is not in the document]
marcusroberts: Wow!  Clever code!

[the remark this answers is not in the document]
marcusroberts: Sweet, Worked a treat, thanks!

compiler/grainformat/walktree.re
```diff
@@ -0,0 +1,280 @@
...
+  //   n =>
+  //     switch (n) {
+  //     | Code(_) => Debug.print_loc("code", get_node_loc(n))
+  //     | Comment(_) => Debug.print_loc("comment", get_node_loc(n))
+  //     },
+  //   all_locations^,
+  // );
+};
+
+let partition_comments =
```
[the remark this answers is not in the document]
marcusroberts: Comments added to the new interface file.

compiler/grainformat/reformat.re
marcusroberts: Nope, redundant and removed.

compiler/grainformat/reformat.re
marcusroberts: Rather than halt, I've replaced the original source instead.  This will let the formatter continue on code it doesn't understand.

compiler/grainformat/reformat.re
marcusroberts: Done

compiler/grainformat/reformat.re
```diff
@@ -0,0 +1,2753 @@
...
+      if (get_loc_line(get_comment_loc(c)) == line) {
+        let (thisline, below) = acc;
+        (thisline @ [c], below);
+      } else {
+        let (thisline, below) = acc;
+        (thisline, below @ [c]);
+      },
+    ([], []),
+    comments,
+  );
```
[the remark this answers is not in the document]
marcusroberts: Nice, thanks!

compiler/grainformat/reformat.re
marcusroberts: Good catch, done

compiler/grainformat/reformat.re
marcusroberts: Hey @ospencer   not sure what you mean here.  I have an infix list but I thought these were the only ones sugared.  Should I make a sublist?

marcusroberts: Thanks for the conversation, all good.

peblair:
Closing in favor of #829. @marcusroberts reopen if needed.

PR #781, opened 2021-07-21
fix(stdlib): String.writeUtf8CodePoint should write 3 bytes for code point 0xFFFF.

cician:


peblair (approved):
Good catch!

phated (approved):
I don't know code points, but I trust @peblair's approval. 👍

phated:
Oh, is there a test we can add for this to avoid regression?

peblair:
For reference: http://www.ltg.ed.ac.uk/~richard/utf-8.cgi?input=FFFF&mode=hex

ospencer (approved):
Thanks, and sorry for the delay!

cician:
No problem and you're welcome.

PR #782, opened 2021-07-21
feat(stdlib): Optimized coerceNumberToWasmI32.

cician:
Avoid unnecessary and slow conversion to WasmI64 and back in case of simple and boxed int32 numbers.


peblair (approved):
Thanks!

phated (approved):
LGTM 👍

PR #784, opened 2021-07-23
fix(graindoc): Correctly parse attributes, no matter the order

phated:
While implementing `@since` I ran into an issue where as soon as the `Str.global_replace` didn't match the group `([ ]+([^:]+):)?`, it stopped trying to match that group at all for future attributes. This meant that `@since` wouldn't be parsed after `@returns` attribute.

I fixed this by changing the way we parse attributes to parse **just** the attribute name and content into 2 groups, and then we use a custom regexp parser for each attribute's contents. This solves this issue I encountered, but allows us to more easily allow custom formats for attribute contents.

### DIFF
--- a/compiler/src/diagnostics/comments.re
+++ b/compiler/src/diagnostics/comments.re
@@ -33,80 +33,116 @@ module Attribute = {
       })
     | Deprecated({attr_desc});
 
-  let try_or_malformed = (~attr, ~hint, fn) =>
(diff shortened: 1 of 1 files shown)


ospencer (approved):
I think this is a lot cleaner!

marcusroberts (approved):
LGTM!

PR #785, opened 2021-07-23
feat(graindoc): Support `@since` and `@history` attributes

phated:
Blocked on #784

### DIFF
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -180,13 +180,14 @@ let to_markdown = (~current_version, docblock) => {
        });
   if (Option.is_some(since_attr) || List.length(history_attrs) > 0) {
     let summary = Option.value(~default="History", since_attr);
+    let disabled = List.length(history_attrs) == 0 ? true : false;
     let details =
       if (List.length(history_attrs) == 0) {
         "No other changes yet.";
       } else {
         Html.table(~headers=["version", "changes"], history_attrs);
       };
-    Buffer.add_string(buf, Html.details(~summary, details));
+    Buffer.add_string(buf, Html.details(~disabled, ~summary, details));
   };
   Buffer.add_string(buf, Markdown.code_block(docblock.type_sig));
   if (String.length(docblock.description) > 0) {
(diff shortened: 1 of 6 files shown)


phated:
I want to reword/remove some of the commits, but I think this is basically ready for review

compiler/graindoc/docblock.re
```diff
@@ -145,6 +145,49 @@ let to_markdown = docblock => {
       deprecations,
     );
   };
+  // TODO: Should we fail if more than one `@since` attribute?
```
ospencer, line 148: Probably.

compiler/src/utils/version.re
```diff
@@ -0,0 +1,72 @@
+// A really bad version parsing library
```
ospencer, line 1: 😤

ospencer: But honestly after reading through it, I really don't think it's all that bad! The only feature that I see is missing is support for alphas/betas/rcs. Otherwise looks fairly feature complete to me.

ospencer (approved):
Excellent!

compiler/graindoc/docblock.re
phated: @ospencer I was just trying to implement this and I actually want to do it in a separate PR because I want to raise a good exn and it would have to exist in 2 places. I'm going to plan to refactor the module docblock stuff and the per-export docblock into deduped methods. I'll create an issue for this.

compiler/src/utils/version.re
phated: I changed the comment to "moderately bad" 😛 I actually wrote the comment when it was a "Really bad semver library" but then realized that I didn't need semver and just needed version compares.

compiler/graindoc/docblock.re
phated: #787

marcusroberts (commented):
LGTM!

marcusroberts (approved):

phated:
Sorting and website styling have been addressed. Merging this.

PR #789, opened 2021-08-06
chore(docs): Add `@since` and `@history` annotations for Array

phated:
This updates the Array module doc blocks with `@since` and `@history` attributes. I also noticed that I missed 1 `@returns` in my initial PR.

I also checked in the generated `array.md` file.

### DIFF
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -1,6 +1,11 @@
 /**
  * @module Array: Utilities for working with arrays.
+ *
  * @example import Array from "array"
+ *
+ * @since v0.2.0
+ * @history v0.1.0: Originally named `arrays`
+ * @history v0.2.0: Renamed to `array`
  */
 
 import WasmI32 from "runtime/unsafe/wasmi32"
@@ -32,6 +37,9 @@ let initLength = (length) => {
  * Returns the length of the input array.
  *
  * @param array: The array to inspect
+ * @returns The amount of elements in the array
+ *
+ * @since v0.1.0
  */
 @disableGC
 export let length = (array) => {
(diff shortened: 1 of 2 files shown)


stdlib/array.md
```diff
@@ -0,0 +1,1008 @@
...
+
+<details>
+<summary>Added in <code>0.2.0</code></summary>
+<table>
+<thead>
+<tr><th>version</th><th>changes</th></tr>
+</thead>
+<tbody>
+<tr><td><code>0.1.0</code></td><td>Originally named `arrays`</td></tr>
+<tr><td><code>0.2.0</code></td><td>Renamed to `array`</td></tr>
```
phated, line 15: So, there's technically edge-case bug with these: markdown inside html isn't processed as markdown. I'm not sure the best way to handle (I really don't want to parse and convert it in graindoc).

stdlib/array.gr
```diff
@@ -32,6 +37,9 @@ let initLength = (length) => {
  * Returns the length of the input array.
  *
  * @param array: The array to inspect
+ * @returns The amount of elements in the array
```
ospencer, line 40:
```suggestion
 * @returns The number of elements in the array
```

ospencer (commented):
Looks good! One small tweak but otherwise 👍 from me.

stdlib/array.md
peblair: Deferring to @ospencer on what the best option would be here

phated: I talked with @ospencer and he said we should just leave this for now.

ospencer (approved):

peblair (approved):

marcusroberts (approved):

PR #792, opened 2021-08-11
feat(stdlib)!: Convert sys functions to return Results instead of throwing errors

### LINKED ISSUE #688: Sys operations should return result type
From discord.

These currently throw but they should return a Result so consumers can handle the errors.

phated:
Fixes #688

I changed all the methods in `stdlib/sys` that throw errors into Result-returning functions. This solves the issue where you want to see if a file exists and handle that case.

I also changed `sys/process.gr`, `sys/time.gr` and `sys/random.gr` but I'm unsure how bad it'll feel to use those methods. What are y'alls opinions?

I wanted to get feedback here before I update the docs, but those will go in this same PR.

### DIFF
--- a/compiler/test/stdlib/sys.file.test.gr
+++ b/compiler/test/stdlib/sys.file.test.gr
@@ -1,7 +1,8 @@
 import Fs from "sys/file"
+import Result from "result"
 
 // fdRead
-let foo = Fs.pathOpen(
+let foo = Result.unwrap(Fs.pathOpen(
   Fs.pwdfd,
   [Fs.SymlinkFollow],
   "test/test-data/foo.txt",
@@ -9,9 +10,9 @@ let foo = Fs.pathOpen(
   [Fs.FdRead],
   [Fs.FdRead],
   []
-)
+))
 
-let (buf, nread) = Fs.fdRead(foo, 40)
+let (buf, nread) = Result.unwrap(Fs.fdRead(foo, 40))
 
 Fs.fdClose(foo)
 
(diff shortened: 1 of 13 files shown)


marcusroberts:
I really hate exceptions and prefer everything as a result :)

phated:
Couple weird things here:
1. The auto-generated wrapping on some of our signatures is really weird. We can either keep it or I can try to dive into the box stuff again to attempt to fix it. Not sure if we should block on it.
2. The `Ok` and `Err` returns are a bit weird to document (like `Option` but even weirder because `Err` has contents). Review and let me know what the pattern should be.
3. We lost all the inline comments on enum variants. I don't actually have a good solution for this because we defer to `printtyp` to generate that example code. Any ideas? If not, are we okay losing them?

phated:
Alright @grain-lang/core - this is ready for review with the caveats above, please let me know what you think. Oh, this also builds upon #808 #809 and #810 so please review those first and then filter the cherry-picked commits from here (starting at e30c5c7)

marcusroberts (approved):
I like this approach.

ospencer (approved):

peblair (approved):
🔥  Looks great! Maybe wait until #803 lands? 🤭

peblair (changes requested):
Marking as "request changes" now that #803 has landed, since these changes now need `incRef` and `decRef` calls (I marked a couple, but there are many more). Recommendation: make a function and replace the `Ok` and `Err` calls with a layer of indirection:
```grain
let wasmSafeOk = (val) => {
  Memory.incRef(Ok)
  Memory.incRef(val)
  Ok(val)
}
```

phated:
This has been updated as per the suggestion by @peblair - but in writing some smoke tests, it was found that our sys/time module isn't actually working. Also, it seems that none of the functions inside any `sys/` module are marked as recursive nor self-decRef.

peblair (approved):

marcusroberts (approved):
That's a great piece of work!

ospencer (approved):
Updates look great!

jozanza (approved):

Issue #598, opened 2021-04-19
Compiler: Rework dependency resolution to use absolute paths


ospencer:

ospencer:
See https://github.com/grain-lang/grain/pull/590

PR #797, opened 2021-08-18
fix(compiler)!: Refactor module resolution in order to properly resolve relative paths

peblair:
This PR fixes the outstanding issues surrounding relative path imports. Additionally, relative paths are now _always_ relative to the file in which they are written (one could engineer strange edge case behavior with the old semantics). The module resolution/dependency graph code could still use a bit of TLC, but this should be an improvement (and gets us closer to fully fixing #598).


compiler/src/linking/link.re
```diff
@@ -26,46 +26,46 @@ let grain_main = "_gmain";
 let grain_start = "_start";
 let function_table = "tbl";
 
-let resolve = mod_name => {
+let grain_module_name = mod_name => {
+  Str.string_after(mod_name, 13)
+}
+
+let resolve = (~base_dir=?, mod_name) => {
   // Remove GRAIN$MODULE$ and add extension
```
ospencer, line 34: Could you move the first half of this comment to the helper function?

ospencer (commented):
Looks awesome! Very clean 😎 

I would love one or two regression tests, but otherwise I'm happy with it!

compiler/test/suites/imports.re
```diff
@@ -169,16 +170,20 @@ describe("imports", ({test}) => {
     "import_same_module_unify2",
     "import *, TList from \"tlists\"; Cons(5, TList.Empty)",
   );
-  /* import filepath tests */
-  assertSnapshot("import_relative_path", "import * from \"./exportStar\"; x");
+  /* import filepath tests (test names are important to make relpaths work) */
```
ospencer, line 173: We should probably make two new files for these tests and use `assertFileshot`, since it's more realistic anyway 🙂

ospencer (approved):
Woohoo! 🎉

marcusroberts (approved):
Nice work!

compiler/src/typed/module_resolution.re
```diff
@@ -75,14 +75,18 @@ let find_ext_in_dir = (dir, name) => {
   process_ext;
 };
 
-let find_in_path_uncap = (~exts=[], path, name) => {
+let is_relpath = path =>
+  Filename.is_relative(path) && !Filename.is_implicit(path);
```
phated, line 79: Does `Filename.is_implicit` make assumptions about the ocaml compiler?

phated (approved):
🔥 Just 1 question

compiler/src/typed/module_resolution.re
peblair: @phated I'm not sure how you mean...can you elaborate? This is the documentation for the function:

> Return true if the file name is relative and does not start with an explicit reference to the current directory (./ or ../ in Unix), false if it starts with an explicit reference to the root directory or the current directory.

phated: Sorry, I didn't look up the docs and assumed it had something to do with implicit modules that ocaml knows about. Don't mind me 🙈

PR #798, opened 2021-08-19
fix(compiler): Return full string locations from parsing

### LINKED ISSUE #483: Compiler: String locations are only reported as the first character
We can use the same technique as comments to resolve this.

ospencer:
Closes #483.

### DIFF
--- a/compiler/src/parsing/lexer.mll
+++ b/compiler/src/parsing/lexer.mll
@@ -74,6 +74,12 @@
     let loc = { start_loc with Location.loc_end = end_loc.Location.loc_end } in
     comments := comment_type s loc :: !comments
 
+  let wrap_string_lexer read_string lexbuf =
+    let start_loc = lexbuf.lex_start_p  in
+    let token = read_string lexbuf in
+    let () = lexbuf.lex_start_p <- start_loc in
+    token
+
 }
 
 let dec_digit = ['0'-'9']
@@ -198,7 +204,7 @@ rule token = parse
   | "||" { PIPEPIPE }
   | "!" { NOT }
   | "@" { AT }
-  | '"'   { read_str (Buffer.create 16) lexbuf }
+  | '"'   { wrap_string_lexer (read_str (Buffer.create 16)) lexbuf }
   | '\'' { read_char (Buffer.create 4) lexbuf }
   | "_" { UNDERSCORE }
   | ident as x { ID x }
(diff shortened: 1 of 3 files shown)


compiler/src/parsing/lexer.mll
```diff
@@ -218,7 +224,8 @@ and read_str buf =
   | "\\\"" { Buffer.add_char buf '"'; read_str buf lexbuf }
   | "\\\\" { Buffer.add_char buf '\\'; read_str buf lexbuf }
   | num_esc { add_code_point buf (lexeme lexbuf) (lexbuf_loc lexbuf); read_str buf lexbuf }
-  | [^ '"' '\\']+ { process_newlines lexbuf; Buffer.add_string buf (lexeme lexbuf); read_str buf lexbuf }
+  | '\n' as char { Buffer.add_char buf char; new_line lexbuf; read_str buf lexbuf }
```
ospencer, line 227: Since we only support `\r\n` and `\n` line endings, this is cross-platform.

compiler/test/runner.re
```diff
@@ -271,12 +277,16 @@ let makeParseRunner =
           comments: List.map(comment_loc_stripper, comments),
           prog_loc: Location.dummy_loc,
         };
-      let parsed = strip_locs @@ parseString(name, input);
-      let untagged = strip_locs @@ parsed;
+      let parsed =
+        if (keep_locs) {
+          parseString(name, input);
+        } else {
+          strip_locs @@ parseString(name, input);
```
peblair, line 284: Just curious, why do we strip the locs at all here?

peblair (approved):
LGTM. Just left one question. 🚀

compiler/src/parsing/lexer.mll
```diff
@@ -218,7 +224,8 @@ and read_str buf =
...
+  | [^ '"' '\\' '\n']+ { Buffer.add_string buf (lexeme lexbuf); read_str buf lexbuf }
```
marcusroberts, line 228: Just trying to learn the lexer.. Is specifying \n here just for safety, as it should have been captured by the line above, or necessary?

marcusroberts (approved):
Look good.  Just the one question but for my learning rather than a code problem

compiler/src/parsing/lexer.mll
peblair: @marcusroberts The `^` at the start of the character class negates it, so this class includes things which are _not_ `\n`.

peblair: (Thinking about it, though, you're right; I suppose it is redundant)

compiler/test/runner.re
ospencer: For most of the parsing tests, we just care about the nodes in the AST and it'd be a huge PITA to _also_ have to figure out the locations of everything, especially since the locations are all character-based instead of line/column based. You can't even write a helper to convert a line/column pair into this kind of loc since it depends on what's come before it in the file 😩

compiler/src/parsing/lexer.mll
ospencer: It's necessary in this case! Maybe slightly more regex reasons rather than lexing reasons, but the `+` is greedy and will keep consuming until it hits a `"` or `\`, so if this rule is in effect (without the `\n` in there), it'll capture all of the newlines without stopping to let us process them.

PR #800, opened 2021-08-22
feat(compiler): Add --wasi-polyfill flag for custom implementations

ospencer:
This PR adds a `--wasi-polyfill <filename>` flag. This flag will instruct the compiler to use provided implementations for WASI calls rather than using WASI functions from the host. This is largely to support runtimes that do not support WASI, but instead provide similar functionality through other host functions. This flag would allow you to map provided host functions to their respective WASI calls.

This allows us to continue to develop as though WASI is always available. Once WASI is more widespread, we can remove this feature.

There is one limitation, however—the polyfill executes in Grain's "runtime" mode, with no access to memory allocation. Static memory buffers are available, though. This is because Grain's runtime uses WASI, so using the runtime would cause a circular dependency. This limitation generally shouldn't be an issue, as this feature is meant for very simple mapping.

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -15,6 +15,7 @@ let memory_debugging_enabled = false;
 /** Environment */
 
 type codegen_env = {
+  name: option(string),
   num_args: int,
   global_offset: int,
   stack_size,
@@ -309,7 +310,8 @@ let runtime_function_imports =
 let runtime_imports =
   List.append(runtime_global_imports, runtime_function_imports);
 
-let init_codegen_env = () => {
+let init_codegen_env = name => {
+  name,
   num_args: 0,
   global_offset: 2,
   stack_size: {
@@ -343,8 +345,6 @@ let runtime_heap_ptr = 0x400;
 let runtime_heap_start = 0x410;
 // Static pointer to runtime type information
 let runtime_type_metadata_ptr = 0x408;
-
-let global_function_table = "tbl";
 
 let get_imported_name = (mod_, name) =>
   Printf.sprintf(
@@ -3124,7 +3124,21 @@ let compile_imports = (wasm_mod, env, {imports}) => {
 
   let compile_module_name = name =>
     fun
-    | MImportWasm => Ident.name(name)
+    | MImportWasm => {
+        switch (Config.wasi_polyfill^, name) {
+        | (Some(file), {Ident.name: "wasi_snapshot_preview1"})
+            when !compiling_wasi_polyfill(env.name) =>
+          let polyfill_path =
+            Filename.remove_extension(
+              Module_resolution.relativize(
+                ~source=Config.base_path^,
+                ~dest=file,
+              ),
+            );
+          "GRAIN$MODULE$" ++ polyfill_path;
+        | _ => Ident.name(name)
+        };
+      }
     | MImportGrain => "GRAIN$MODULE$" ++ Ident.name(name);
 
   let compile_import_name = name =>
@@ -3473,7 +3487,7 @@ let compile_wasm_module = (~env=?, ~name=?, prog) => {
   reset_labels();
   let env =
     switch (env) {
-    | None => init_codegen_env()
+    | None => init_codegen_env(name)
     | Some(e) => e
     };
   let (env, prog) = prepare(env, prog);
@@ -3516,6 +3530,14 @@ let compile_wasm_module = (~env=?, ~name=?, prog) => {
     });
   } else {
     compile_all();
+  };
+
+  if (compiling_wasi_polyfill(name)) {
+    write_universal_exports(
+      wasm_mod,
+      prog.signature,
+      get_exported_names(wasm_mod),
+    );
   };
 
   let serialized_cmi = Cmi_format.serialize_cmi(prog.signature);
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -3124,21 +3124,7 @@ let compile_imports = (wasm_mod, env, {imports}) => {
 
   let compile_module_name = name =>
     fun
-    | MImportWasm => {
-        switch (Config.wasi_polyfill^, name) {
-        | (Some(file), {Ident.name: "wasi_snapshot_preview1"})
-            when !compiling_wasi_polyfill(env.name) =>
-          let polyfill_path =
-            Filename.remove_extension(
-              Module_resolution.relativize(
-                ~source=Config.base_path^,
-                ~dest=file,
-              ),
-            );
-          "GRAIN$MODULE$" ++ polyfill_path;
-        | _ => Ident.name(name)
-        };
-      }
+    | MImportWasm => Ident.name(name)
     | MImportGrain => "GRAIN$MODULE$" ++ Ident.name(name);
 
   let compile_import_name = name =>
--- a/compiler/src/compile.re
+++ b/compiler/src/compile.re
@@ -39,6 +39,22 @@ type error =
 
 exception InlineFlagsError(Location.t, error);
 
+/** `remove_extension` new enough that we should just use this */
+
+let safe_remove_extension = name =>
+  try(Filename.chop_extension(name)) {
+  | Invalid_argument(_) => name
+  };
+
+let default_output_filename = name =>
+  safe_remove_extension(name) ++ ".gr.wasm";
+
+let default_assembly_filename = name =>
+  safe_remove_extension(name) ++ ".wast";
+
+let default_mashtree_filename = name =>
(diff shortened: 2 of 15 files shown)


compiler/src/codegen/compcore.re
```diff
@@ -3518,6 +3518,14 @@ let compile_wasm_module = (~env=?, ~name=?, prog) => {
     compile_all();
   };
 
+  if (compiling_wasi_polyfill(name)) {
+    write_universal_exports(
```
ospencer, line 3522: This is something that the linker already did, but it's needed to export plain wasm functions. There's no compiler flag for this because it would break linking (since Grain modules expect to import Grain values—if it helps you think about it, a module with universal exports could only be imported from via the `foreign wasm` syntax), and fully-linked modules export plain functions.

compiler/src/compile.re
```diff
@@ -39,6 +39,22 @@ type error =
 
 exception InlineFlagsError(Location.t, error);
 
+/** `remove_extension` new enough that we should just use this */
```
ospencer, line 42: Moved here from `grainc`

compiler/src/compile.re
```diff
@@ -183,36 +199,6 @@ let rec compile_resume = (~hook=?, s: compilation_state) => {
   };
 };
 
-let reset_compiler_state = () => {
```
ospencer, line 186: This is all moved lower in this file.

compiler/src/compile.re
```diff
@@ -265,6 +251,54 @@ let stop_after_assembled =
...
+let reset_compiler_state = () => {
+  Env.clear_imports(); // TODO: (#576) reenable if necessary (makes tests super slow, but seems to be safe?)
+  // Grain_utils.Fs_access.flush_all_cached_data();
+  Grain_utils.Warnings.reset_warnings();
+};
+
+let compile_string = (~hook=?, ~name=?, ~outfile=?, ~reset=true, str) => {
+  if (reset) {
+    reset_compiler_state();
+    compile_wasi_polyfill();
```
ospencer, line 279: This is the one line that's added to this function and `compile_file`.

compiler/src/linking/link.re
```diff
@@ -316,104 +342,6 @@ let rec globalize_names = (local_names, expr) => {
   };
 };
 
-let type_of_repr = repr => {
```
ospencer, line 319: This was all moved to `comp_utils`.

compiler/test/input/wasiPolyfill.gr
```diff
@@ -0,0 +1,7 @@
+import foreign wasm fd_write : (WasmI32, WasmI32, WasmI32, WasmI32) -> WasmI32 from "wasi_snapshot_preview1"
+
+export let fd_write: (WasmI32, WasmI32, WasmI32, WasmI32) -> WasmI32 = (fd, iovs, iovs_len, nwritten) => {
+    fd_write(fd, iovs, iovs_len, nwritten)
+    fd_write(fd, iovs, iovs_len, nwritten)
```
marcusroberts, line 5: Why do you call fd_write three times?

marcusroberts: Ah, to test this is the polyfill being called, and not the real one right?

marcusroberts (approved):
A soft approval from me as some of this is still beyond my full comprehension ;)

I think I answered my own question about the pollyfill printing three times, otherwise all looks good to me.

compiler/src/linking/link.re
```diff
@@ -46,6 +45,17 @@ let is_grain_module = mod_name => {
...
+let wasi_polyfill_module = () => {
+  "GRAIN$MODULE$./"
+  ++ Filename.remove_extension(Option.get(Config.wasi_polyfill^));
+};
+
+let is_wasi_module = mod_name => {
+  mod_name == "wasi_snapshot_preview1";
+};
+
+let is_wasi_polyfill_module = mod_name => mod_name == wasi_polyfill_module();
```
peblair, line 57: 😬  I see why we have to do this, but I think we should open an issue to remove this one day...I'm concerned about fragility in the long-term, and ideally we could just have the polyfill be a `.gr.wasm` file instead of the `.gr` source (and then Silo will bring it all together smoothly)
```suggestion

// [TODO] When silo is implemented, remove this logic
let is_wasi_polyfill_module = mod_name => mod_name == wasi_polyfill_module();
```

compiler/src/linking/link.re
```diff
@@ -95,6 +106,18 @@ let rec build_dependency_graph = (~base_dir, mod_name) => {
         );
       };
       G.add_edge(dependency_graph, mod_name, imported_module);
+    } else if (has_wasi_polyfill
+               && is_wasi_module(imported_module)
+               && !is_wasi_polyfill_module(mod_name)) {
+      let imported_module = wasi_polyfill_module();
```
peblair, line 112: Just to make this a bit easier to read in the future
```suggestion
      // Perform any WASI polyfilling. Note that we skip this step if we are compiling the polyfill module itself.
      // If we are importing a foreign from WASI, then add a dependency to the polyfill instead.
      let imported_module = wasi_polyfill_module();
```

compiler/src/linking/link.re
```diff
@@ -511,6 +440,18 @@ let link_all = (linked_mod, dependencies, signature) => {
               imported_name,
             );
           Hashtbl.add(local_names, internal_name, new_name);
+        } else if (has_wasi_polyfill
+                   && is_wasi_module(imported_module)
+                   && !is_wasi_polyfill_module(dep)) {
+          let imported_name = Import.function_import_get_base(func);
```
peblair, line 446: For easier reading
```suggestion
          // Perform any WASI polyfilling. Note that we skip this step if we are compiling the polyfill module itself.
          // If we are importing a foreign from WASI, then we swap it out for the foreign from the polyfill.
          let imported_name = Import.function_import_get_base(func);
```

compiler/test/suites/linking.re
```diff
@@ -16,4 +21,17 @@ describe("linking", ({test}) => {
...
+    "wasi_polyfill",
+    {|print("foo")|},
+    "foo\nfoo\nfoo\n",
+  );
+  assertWasiPolyfillRun(
+    "test/input/wasiPolyfillNoop.gr",
+    "wasi_polyfill_noop",
+    {|print("foo")|},
+    "",
+  );
```
peblair, line 36: I think we should add a test for the situation where you use a WASI function not defined in the polyfill.

peblair (changes requested):
Looks good! There's one additional test I think we should add, and I left a couple of comment suggestions in the code for you to mull over.

compiler/test/input/wasiPolyfill.gr
ospencer: Yup!

compiler/src/linking/link.re
ospencer: Technically we don't have to do this if we don't want to support making wasi calls from the polyfills—it's _really_ only here because there isn't really a better way to write a meaningful test, haha. I feel like it'd be rare to do wasi calls from the polyfill, since if you had wasi, you wouldn't use the polyfill.

I have mixed feelings about it being a `.gr.wasm` file, since we can't have the compiler force you into runtime mode, for example. It also sort of makes it unclear what'd be allowed as a polyfill, since a fully linked module wouldn't be allowed, for example. That does bring back the idea of different file extensions for linked/unlinked modules though as a sort of solution.

compiler/test/suites/linking.re
ospencer: Will follow-up with a PR 👍

compiler/src/linking/link.re
peblair: > I feel like it'd be rare to do wasi calls from the polyfill, since if you had wasi, you wouldn't use the polyfill.

It might be something we do in fact want to support, though, for reasons similar to `LD_PRELOAD` in Linux, so I guess it makes sense...I'm not going to hold up this PR on it, but I am concerned about having conditional logic based on is-this-the-polyfill in the linker.

> That does bring back the idea of different file extensions for linked/unlinked modules though as a sort of solution.

I think this would be the most pragmatic long-term solution to this issue.

peblair (approved):

compiler/test/suites/linking.re
phated: Please create an issue @ospencer

phated (approved):
I'm happy with this. Good job with the quick ideation and patch @ospencer 🔥

compiler/test/suites/linking.re
ospencer: https://github.com/grain-lang/grain/issues/840

Issue #555, opened 2021-03-09
Runtime: Consider removing dtoa_buffered


ospencer:
We don't currently use it or really have plans to use it.

Code:
```
export let dtoa_buffered = (buffer: usize, value: f64) => {
  let mut buffer = buffer
  let str = if (WasmF64.eq(value, 0.W)) {
    WasmI32.store8(buffer, _CHAR_CODE_0, 0n)
    WasmI32.store8(buffer, _CHAR_CODE_DOT, 1n)
    WasmI32.store8(buffer, _CHAR_CODE_0,  2n)
    3n
  } else if (!isFinite(value)) {
    if (isNaN(value)) {
      WasmI32.store8(buffer, 0x4En, 0n) // N
      WasmI32.store8(buffer, 0x61n, 1n) // a
      WasmI32.store8(buffer, 0x4En, 2n) // N
      3n
    } else {
      let sign = WasmF64.lt(value, 0.W)
      if (sign) {
        WasmI32.store8(buffer, _CHAR_CODE_MINUS, 0n)
        buffer += 1n;
      }
      // Would be nice if there were a way to specify alignment here as we
      // don't know if this store is aligned. I don't know what's fastest—a
      // misaligned store or an explict alignment check + branch to store with
      // align=0 or align=3 _or_ just always align=0. I imagine just align=0?
      WasmI64.store(buffer, 0x7974696E69666E49N, 0n) // ytinifnI
      8n + (if (sign) 1n else 0n)
    }
  } else {
    dtoa_core(buffer, value)
  }
  WasmI32.toGrain(str): String
}
```

PR #803, opened 2021-08-22
feat(compiler)!: Callee-owned values

### LINKED ISSUE #774: If expression causes memory leak
The following program stops with OutOfMemory error.

The cause is the `if`-`else` expression in this very specific configuration.
The leak doesn't happen for example if `b` is replaced with a constant or if the if expression is not in the printBool function.

``` grain
let foo = (s: String) => void
let printBool = (b: Bool) => foo(if (b) "true" else "false")

let b = true
for (let mut i=0; i<100000000; i += 1) {
  printBool(true)
}
```

peblair:
This PR migrates the Grain memory model from a caller-owned reference paradigm to a callee-owned reference paradigm. This change was done for a variety of reasons, mostly pertaining to lingering bugs in the garbage collector.

## User Impact
This change makes `@disableGC` much more annoying to use. While its use is discouraged, those who do use it should take care to follow the callee-owned paradigm (and take special care to `decRef` closures). These rules will be better documented in an updated version of #768. Additionally, this change is not binary-compatible with previous versions of Grain (hence all test snapshots being updated).

## Notes to Reviewers
While I will not detail the `Memory.incRef` calling convention here (see above note about #768), do note the following:
- All `@disableGC` functions which are called from a non-`@disableGC` context (i.e. those in the runtime which are outward-facing) must be recursive so that they can `decRef` themselves (since callers will have `incRef`ed them)
- Not all internal functions contain the full set of `incRef`s and `decRef`s. The guiding rule of thumb is that if a function accepts or returns a `WasmXX` value, we assume that it can _only_ be called from a `@disableGC` context (and therefore omit the extra memory management steps).

## Known Issues and Future Changes
In a follow-up PR, we will reenable cyclic garbage collection, which should fix the lingering garbage collection issues. Additionally, the user-facing impact will be better documented in #768.

## Bugfixes
Fixes #774.


compiler/src/codegen/compcore.re
```diff
@@ -1066,48 +1009,35 @@ let tee_swap =
...
+      // [TODO] ??
+      let skip_incref =
+        switch (rtype) {
+        | [I32Type] => false
+        | _ => true
+        };
+      Expression.Block.make(wasm_mod, gensym_label("cleanup_locals")) @@
+      Concatlist.list_of_t(
+        singleton(set_swap(wasm_mod, env, 0, arg, ~skip_incref))
+        //@ singleton(tracepoint(wasm_mod, env, tp_count^))
```
ospencer, line 1033: looks like there needs to be some cleanup around this section

js-runner/src/runner.js
```diff
@@ -1,3 +1,4 @@
+import process from 'process';
```
ospencer, line 1: should probably remove this unless we actually use it

stdlib/array.gr
```diff
@@ -34,11 +34,15 @@ let initLength = (length) => {
...
+export let rec length = (array) => {
   let ptr = WasmI32.fromGrain(array: Array<a>)
-  tagSimpleNumber(WasmI32.load(ptr, _ARRAY_LENGTH_OFFSET))
+  let ret = tagSimpleNumber(WasmI32.load(ptr, _ARRAY_LENGTH_OFFSET))
+  Memory.decRef(ptr)
+  Memory.decRef(WasmI32.fromGrain(length))
+  ret
 }
 
+// [FIXME] (#793) Type annotation breaks compiler? 😰
```
ospencer, line 45:
```suggestion
// [FIXME] (#793) Type annotation breaks compiler
```

stdlib/array.gr
```diff
@@ -72,16 +81,23 @@ export let make : (Number, a) -> Array<a>  = (length, item) => {
...
   let length = initLength(length)
   let byteLength = WasmI32.mul(length, 4n)
   let array = allocateArray(length)
-  let mut index = 0
+  let mut index = 0n
   for (let mut i = 0n; WasmI32.ltS(i, byteLength); i = WasmI32.add(i, 4n)) {
-    WasmI32.store(WasmI32.add(array, i), WasmI32.fromGrain(fn(index)), _ARRAY_START_OFFSET)
-    index += 1
+    Memory.incRef(WasmI32.fromGrain(fn));
+    // [FIXME] This line fails the array/map test suite???
```
ospencer, line 91:
```suggestion
    // [FIXME] This line fails the array/map test suite
```

stdlib/map.gr
```diff
@@ -278,24 +278,31 @@ export let fromList = (list) => {
 }
 
 let setInArray = (array) => {
-  @disableGC
+  //@disableGC
```
ospencer, line 281: What's up with this?

stdlib/number.gr
```diff
@@ -112,25 +116,36 @@ export let max = (x: Number, y: Number) => if (x > y) x else y
...
-  WasmI32.toGrain(reducedInteger(ceiling)): Number
+  let ret = WasmI32.toGrain(reducedInteger(ceiling)): Number
+  Memory.decRef(WasmI32.fromGrain(x))
+  Memory.decRef(WasmI32.fromGrain(ceil))
+  ret
 }
 
 /**
  * Rounds its operand down to the largest integer less than the operand.
+<<<<<<< HEAD
```
ospencer, line 130: Conflict!

stdlib/number.gr
```diff
@@ -249,23 +273,34 @@ export let isNaN = (x: Number) => {
...
     false
   }
+  Memory.decRef(asPtr)
+  Memory.decRef(WasmI32.fromGrain(isNaN))
+  ret
 }
 
 /**
  * Checks if a number is infinite, that is either of floating point positive or negative infinity.
+<<<<<<< HEAD
```
ospencer, line 283: ⚔️

stdlib/pervasives.gr
```diff
@@ -97,6 +99,11 @@ let (isnt) : (a, a) -> Bool = (x, y) => !(x is y)
 
 export enum List<a> { [], [...](a, List<a>) }
 
+/**
+ * @deprecated This will be removed in a future release of Grain.
+ */
+let cons = (a, b) => [a, ...b] // <- workaround for (grain-lang/grain#802) [TODO] fix #802 and delete before release, ideally
```
ospencer, line 105:
```suggestion
let cons = (a, b) => [a, ...b] // <- workaround for (grain-lang/grain#802) [TODO] fix #802 and delete
```

stdlib/runtime/exception.gr
```diff
@@ -29,10 +29,18 @@ export let dangerouslyRegisterBasePrinter = (f) => {
     }
     current = next
   }
+  // Leaks memory (cyclic dep. with Memory) :(
+  //Memory.decRef(WasmI32.fromGrain(dangerouslyRegisterBasePrinter))
+  //Memory.decRef(WasmI32.fromGrain(f))
```
ospencer, line 34: Probably let's leave a comment about why we aren't doing the decref here.
```suggestion
  // We don't decRef the closure or arguments here to avoid a cyclic dep. on Memory.
  // This is fine, as this function should only be called once.
```

stdlib/runtime/exception.gr
```diff
@@ -29,10 +29,18 @@ export let dangerouslyRegisterBasePrinter = (f) => {
...
+  void
 }
 
-export let dangerouslyRegisterPrinter = (f) => {
+export let rec dangerouslyRegisterPrinter = (f) => {
   printers = WasmI32.fromGrain((f, printers))
+  // Leaks memory (cyclic dep. with Memory) :(
+  //Memory.decRef(WasmI32.fromGrain(dangerouslyRegisterPrinter))
+  //Memory.decRef(WasmI32.fromGrain(f))
```
ospencer, line 42:
```suggestion
  // We don't decRef the closure or arguments here to avoid a cyclic dep. on Memory.
  // This is fine, as this function is only called seldomly.
```

stdlib/runtime/gc.gr
```diff
@@ -28,15 +28,102 @@ import WasmI32, {
...
+let _INCREF_DEBUG_STR_3 = "; tag: "
+let _INCREF_DEBUG_STR_4 = ")\n"
+
+// decRef debug messages: "decRef: 0xNNNNNN (prev count: nn)\n"
+let _DECREF_DEBUG_STR_1 = "decRef: 0x"
+let _DECREF_DEBUG_STR_2 = " (prev count: "
+let _DECREF_DEBUG_STR_3 = "; ignoreZeros: true)\n"
+let _DECREF_DEBUG_STR_4 = "; ignoreZeros: false)\n"
+
+let logIncRef = (userPtr: WasmI32, refCount: WasmI32) => {
```
ospencer, line 63: I think for now, we should comment out the debugging stuff (probably just big block comment) until our optimizations are a tad better.

stdlib/runtime/malloc.gr
```diff
@@ -74,7 +74,8 @@ let logMallocHeaderSize = 3n
  */
 let mut heapSize = 0n
 
-export let _RESERVED_RUNTIME_SPACE = 16384n
+export let _RESERVED_RUNTIME_SPACE = 0x4000n
+//export let _RESERVED_RUNTIME_SPACE = 0x8000n
```
ospencer, line 78:
```suggestion
```

stdlib/runtime/string.gr
```diff
@@ -486,25 +498,30 @@ let rec heapValueToString = (ptr, extraIndents, toplevel) => {
...
         }
         WasmI32.store(ptr, tupleLength, 4n)
 
-        strings = ["(", ...strings]
+        Memory.incRef(WasmI32.fromGrain(strings))
+        strings = slConsDisableGc("(", strings)
 
         let string = join(strings)
-        Memory.decRef(WasmI32.fromGrain(strings))
+        //Memory.decRef(WasmI32.fromGrain(strings))
```
ospencer, line 524: Do we not want this anymore?

ospencer (changes requested):
Fantastic work here!! I know it was a huge PITA to get this done, so a huge thank you is deserved 🙏 👏 🎊 

Looks like there's still some cleanup that needs to be done, but overall I approve of this and everything makes sense to me! 😄

stdlib/number.gr
peblair: Oops, forgot to save this file

stdlib/map.gr
```diff
@@ -278,24 +278,28 @@ export let fromList = (list) => {
 }
 
 let setInArray = (array) => {
-  @disableGC
```
ospencer, line 281: Is this truly okay? The array here is uninitialized, so won't it be dangerous to set into the array, since the old item will be decRef'd? I know that decRef handles null just fine, but random values probably will cause problems.

stdlib/pervasives.gr
```diff
@@ -97,6 +99,11 @@ let (isnt) : (a, a) -> Bool = (x, y) => !(x is y)
 
 export enum List<a> { [], [...](a, List<a>) }
 
+/**
+ * @deprecated This will be removed in a future release of Grain.
+ */
+let cons = (a, b) => [a, ...b] // <- workaround for (grain-lang/grain#802) [TODO] fix #802 and delete
```
ospencer, line 105: This sorta makes me want to add `empty` as well. Feels awkward just to have `cons`.

stdlib/runtime/numberUtils.gr
```diff
@@ -895,17 +914,17 @@ export let utoa32 = (value, radix) => {
   } else if (radix == 10n) {
     let decimals = decimalCount32(value)
     let out = allocateString(decimals)
-    utoa32_dec_core(out + 8n, value, decimals)
+    utoa32Buf(out + 8n, value, decimals)
```
ospencer, line 917: This is broken because the third argument should be `radix`, but probably better to not change any of these at all so we don't have to compute the number of decimals multiple times.

stdlib/runtime/numberUtils.gr
```diff
@@ -895,17 +914,17 @@ export let utoa32 = (value, radix) => {
...
     out
   } else if (radix == 16n) {
     let decimals = WasmI32.shrU(31n - WasmI32.clz(value), 2n) + 1n
     let out = allocateString(decimals)
-    utoa32_hex_core(out + 8n, value, decimals)
+    utoa32Buf(out + 8n, value, decimals)
```
ospencer, line 922: also broken

stdlib/runtime/numberUtils.gr
```diff
@@ -886,6 +886,25 @@ let utoa64_any_core = (buffer, num, offset, radix) => {
   }
 }
 
+export let utoa32Buf = (buf, value, radix) => {
```
ospencer, line 889: Could we instead call this `utoa32Buffered`? Reminiscing https://github.com/grain-lang/grain/issues/555.

stdlib/runtime/numbers.gr
```diff
@@ -1245,21 +1296,33 @@ export let coerceNumberToInt32 = (x: Number) => {
     // can possibly fail
     newInt32(coerceNumberToWasmI32(WasmI32.toGrain(x) : Number))
   }
-  WasmI32.toGrain(result): Int32
+  let ret = WasmI32.toGrain(result): Int32
+  if (WasmI32.fromGrain(ret) != WasmI32.fromGrain(x)) {
```
ospencer, line 1300: This feels like a code smell. Shouldn't a function incRef the value it's returning (even if it's an argument)?

ospencer (changes requested):

stdlib/runtime/numbers.gr
peblair: I believe this is correct as written. If we call `coerceNumberToInt32(x)`, we expect to receive back `x` with a net refcount increase of +1 (excluding the `incRef` we do as part of the function call):
```
caller context: x <refcount: n>
function call: incRef(x) -> <refcount: n+1>
~~ in coerceNumberToInt32 ~~
let result = x <refcount: n+1> // simplified; this is the only case where this `if` becomes relevant
let ret = WasmI32.toGrain(result): Int32 == x <refcount: n+1>
```
at this point, we want to return `ret` as-is, with no decrease in refcount.

In contrast, if we call with a non-Int32 value, then we _do_ want to decRef here, since `result` is a different object.

ospencer: My bad! I guess I shouldn't review PRs late—I thought this was a consuming function of `coerceNumberToInt32` that was compensating for incorrect behavior. This _IS_ `coerceNumberToInt32`, just not being incorrect 😂  Ignore me, carry on 🙈

ospencer (approved):

marcusroberts (approved):
My comprehension is low, but LGTM!

stdlib/runtime/numbers.gr
```diff
@@ -1275,58 +1338,89 @@ export let coerceNumberToRational = (x: Number) => {
...
 }
 
 /// USER-EXPOSED CONVERSION FUNCTIONS
 
-export let convertExactToInexact = (x: Number) => {
-  x
+export let rec convertExactToInexact = (x: Number) => {
+  let ret = x
+  Memory.decRef(WasmI32.fromGrain(convertExactToInexact))
+  ret
```
phated, line 1420: This function has become funnier

stdlib/runtime/numbers.gr
```diff
@@ -1415,6 +1534,8 @@ export let numberMod = (%);
 // inc/dec
 
 export let incr = (x) => {
+  Memory.incRef(WasmI32.fromGrain((+)))
```
phated, line 1537: Why does `(+)` never get decref in here?

stdlib/runtime/string.gr
```diff
@@ -26,14 +26,28 @@ import NumberUtils from "runtime/numberUtils"
...
+let slConsDisableGc = (a, b) => {
+  Memory.incRef(WasmI32.fromGrain(SLCons))
+  SLCons(a, b)
+}
+
+let mut _RUNTIME_TYPE_METADATA_PTR = 0x01n;
+
+@disableGC
+let initPtr = () => {
+  // hack to avoid incRef on this pointer
```
phated, line 47: Do we have an issue to allow `disableGC` on `let mut`? If not, can we?

stdlib/runtime/string.gr
```diff
@@ -486,22 +498,27 @@ let rec heapValueToString = (ptr, extraIndents, toplevel) => {
...
           // Special case: unary tuple
-          strings = [",", ...strings]
+          strings = slConsDisableGc(",", strings)
         }
         for (let mut i = tupleLength * 4n - 4n; i >= 0n; i -= 4n) {
           let item = toStringHelp(WasmI32.load(ptr + i, 8n), extraIndents, false)
-          strings = [item, ...strings]
+          Memory.incRef(WasmI32.fromGrain(item))
+          Memory.incRef(WasmI32.fromGrain(strings))
+          //let item = toStringHelp((WasmI32.fromGrain(comspace) << 1n) + 1n, extraIndents, false)
```
phated, line 510: Why is this commented now?

stdlib/runtime/unsafe/printWasm.gr
```diff
@@ -1,21 +1,77 @@
 import Conv from "runtime/unsafe/conv"
+import WasmI32 from "runtime/unsafe/wasmi32"
+import Memory from "runtime/unsafe/memory"
+
+// [FIXME] These all leak ATM (grain-lang/grain#791)
```
phated, line 5: This file is quite ridiculous, haha! 🤣

stdlib/runtime/wasi.gr
```diff
@@ -26,6 +26,15 @@ import foreign wasm clock_time_get: (WasmI32, WasmI64, WasmI32) -> WasmI32 from
...
 import foreign wasm fd_pread : (WasmI32, WasmI32, WasmI32, WasmI64, WasmI32) -> WasmI32 from "wasi_snapshot_preview1"
+/**
+ * Invokes the `fd_write` system call.
+ *
+ * @param file_descriptor - The file descriptor to write to
+ * @param iovs - The pointer to the array of iovs to write
+ * @param iovs_len - The length of the array of iovs
+ * @param nwritten - Where to store the number of bytes written
+ * @return The number of bytes written
+ */
```
phated, line 37: Technically not the correct format for graindoc, but I doubt we want to run graindoc on these, yeah?

phated: Actually, I don't think it iterates foreigns anyway.

stdlib/string.gr
```diff
@@ -930,7 +1021,8 @@ export let encodeAtWithBom = (s, encoding, dest, destPos) => {
 // @param includeBom: Bool - Whether to include the byte-order marker in the encoded output
 // @returns Bytes
 @disableGC
-let encodeHelp = (s: String, encoding: Encoding, includeBom: Bool) => {
+let rec encodeHelp = (s: String, encoding: Encoding, includeBom: Bool) => {
+  Memory.incRef(WasmI32.fromGrain((+)))
```
phated, line 1025: I asked this on another chunk of code, but why is `(+)` incref'd but never decref'd in this function?

stdlib/string.gr
```diff
@@ -1399,23 +1514,28 @@ export let forEachCodePoint = (fn: (Number) -> Void, str: String) => {
     // avoid heap allocations. `getCodePoint` will throw
     // MalformedUnicode exception for values exceeding this limit.
     let codePoint = getCodePoint(ptr)
+    Memory.incRef(WasmI32.fromGrain(fn))
```
phated, line 1517: incref'd in a loop, but only decref'd once outside of the loop?

stdlib/string.gr
```diff
@@ -1452,9 +1572,14 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
```
phated, line 1575: Same question as above. I don't understand this.

stdlib/sys/file.gr
```diff
@@ -585,45 +585,20 @@ export let fdSync = (fd: FileDescriptor) => {
...
+  // Memory.incRef(WasmI32.fromGrain([...]))
+  Memory.incRef(WasmI32.fromGrain(cons))
+  Memory.incRef(WasmI32.fromGrain(a))
+  Memory.incRef(WasmI32.fromGrain(b))
+  cons(a, b)
+  // [TODO] Once grain-lang/grain#802 is fixed:
+  // [a, ...b]
+}
+
+let orderedFdflags = wasmSafeCons(Append, wasmSafeCons(Dsync, wasmSafeCons(Nonblock, wasmSafeCons(Rsync, wasmSafeCons(Sync, [])))))
```
phated, line 600: We are going to conflict on these, since I moved stuff around in this file in #792

stdlib/sys/file.gr
```diff
@@ -585,45 +585,20 @@ export let fdSync = (fd: FileDescriptor) => {
...
-  FdFilestats,
-  FdSetSize,
-  FdSetTimes,
-  PathSymlink,
-  PathRemoveDirectory,
-  PathUnlinkFile,
-  PollFdReadwrite,
-  SockShutdown
-]
+let wasmSafeCons = (a, b) => {
```
phated, line 589: This should have `@disableGC` right?

compiler/src/codegen/compcore.re
```diff
@@ -965,14 +896,24 @@ let safe_drop = (wasm_mod, env, arg) =>
     call_decref_ignore_zeros(wasm_mod, env, arg),
   );
 
-let get_swap = (~ty as typ=I32Type, wasm_mod, env, idx) =>
+let get_swap =
+    (
+      ~ty as typ=I32Type,
+      ~skip_incref=true,
+      ~skip_decref=true,
```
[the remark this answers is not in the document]
peblair: So, one of the changes I made was to really clamp down on the swap variables; since they should be treated as volatile registers and almost never interact with the refcounts. The swap variables only hold intermediate computation values, so we know their refcounts are valid during that computation.

compiler/src/codegen/compcore.re
```diff
@@ -1066,48 +1009,27 @@ let tee_swap =
...
-  | [{instr_desc: MSwitch(value, branches, default, asmty)} as i] =>
-    let branches =
-      List.map(((n, block)) => (n, finalize_gc(block)), branches);
-    let default = finalize_gc(default);
-    [{...i, instr_desc: MSwitch(value, branches, default, asmty)}];
-  | [hd, ...tl] => [hd, ...finalize_gc(tl)]
-  | [] => []
-  };
-};
+let tp_count = ref(1);
```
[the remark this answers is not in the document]
peblair: It stands for "I forgot to delete this variable" :joy:

compiler/src/codegen/compcore.re
```diff
@@ -1239,14 +1165,22 @@ let compile_array_op = (wasm_mod, env, arr_imm, op) => {
   let get_arr = () => compile_imm(wasm_mod, env, arr_imm);
   switch (op) {
   | MArrayGet(idx_imm) =>
+    // ASSUMPTION: idx is a basic (non-heap) int
     let idx = compile_imm(wasm_mod, env, idx_imm);
+    let set_idx = () => set_swap(1, untag_number(wasm_mod, idx));
     let get_idx = () => get_swap(1);
+    let set_arr = () => set_swap(2, get_arr());
```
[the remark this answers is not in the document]
peblair: Rephrased to make more clear

compiler/src/codegen/compcore.re
```diff
@@ -1380,8 +1328,52 @@ let compile_array_op = (wasm_mod, env, arr_imm, op) => {
             ),
             get_arr(),
           ),
-          /* [TODO] decref the old item */
-          call_incref(wasm_mod, env, val_),
+          /* [TODO] decref the old item more efficiently (using a swap slot, most likely) */
+          Expression.Tuple_extract.make(
```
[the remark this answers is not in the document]
peblair: Primarily because `/* [TODO] decref the old item */` turned unto `decref the old item more efficiently`. We weren't doing the reference management correctly before, but now we are (even if it could be improved in the future).

compiler/src/codegen/compcore.re
```diff
@@ -1807,21 +1818,15 @@ let allocate_closure =
...
       ~offset=0,
       wasm_mod,
-      tee_swap(
-        ~skip_incref=true,
-        wasm_mod,
-        env,
-        0,
-        heap_allocate(wasm_mod, env, closure_size),
-      ),
+      tee_swap(wasm_mod, env, 0, heap_allocate(wasm_mod, env, closure_size)),
```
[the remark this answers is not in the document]
peblair: So, `tee_swap`'s default arguments were always `skip_incref=true` and `skip_decref=true`. This diff just removes the redundant `skip_incref=true`.

compiler/src/codegen/compcore.re
```diff
@@ -2138,6 +2143,7 @@ let compile_prim1 = (wasm_mod, env, p1, arg, loc): Expression.t => {
       ],
     )
   | Throw =>
+    // [TODO] When we have exception handling, revisit whether there is any GC required here
```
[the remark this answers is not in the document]
peblair: #813

compiler/src/codegen/transl_anf.re
```diff
@@ -650,7 +650,7 @@ let compile_wrapper = (id, env, func_name, args, rets): Mashtree.closure_data =>
       stack_size_f32: 0,
       stack_size_f64: 0,
     },
-    attrs: [],
+    attrs: [Disable_gc],
```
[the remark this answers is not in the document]
peblair: These are the wrapper functions produced around Wasm functions. In order for the runtime to work correctly, those wrappers need to just be thin pass-through functions which don't do any refcount management.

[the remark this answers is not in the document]
peblair: It wasn't a mistake before because in the old calling convention there were no GC instructions to not emit

stdlib/runtime/string.gr
peblair: #814

stdlib/string.gr
peblair: We invoke `+` on the next line, which follows the normal calling convention (meaning that it will `decRef` itself, so it should be `incRef`ed before the call)

stdlib/string.gr
peblair: Same as above; it will `decRef` itself per the calling convention

stdlib/string.gr
peblair: Same as above

stdlib/sys/file.gr
peblair: This whole file is `--no-gc`

stdlib/array.gr
```diff
@@ -72,16 +81,23 @@ export let make : (Number, a) -> Array<a>  = (length, item) => {
...
   let array = allocateArray(length)
-  let mut index = 0
+  let mut index = 0n
   for (let mut i = 0n; WasmI32.ltS(i, byteLength); i = WasmI32.add(i, 4n)) {
-    WasmI32.store(WasmI32.add(array, i), WasmI32.fromGrain(fn(index)), _ARRAY_START_OFFSET)
-    index += 1
+    Memory.incRef(WasmI32.fromGrain(fn));
+    // [FIXME] This line fails the array/map test suite
+    //assert !WasmI32.eqz(WasmI32.and(WasmI32.fromGrain(index), 1n)) // must be a simple int for next line to be correct
+    //assert !WasmI32.eqz(WasmI32.clz(index)) // check we can left-shift
```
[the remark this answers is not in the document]
peblair: #815

stdlib/runtime/exception.gr
```diff
@@ -15,7 +15,7 @@ export let mut printers = 0n
 // no GC operations. As such, they should only be called by this module and/or
 // modules that understand these restrictions, namely Pervasives.
 
-export let dangerouslyRegisterBasePrinter = (f) => {
+export let rec dangerouslyRegisterBasePrinter = (f) => {
```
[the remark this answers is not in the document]
peblair: I had originally added in the decrefs before I knew about the cyclic dependency issue.

stdlib/runtime/gc.gr
```diff
@@ -28,15 +28,106 @@ import WasmI32, {
...
 primitive (||) : (Bool, Bool) -> Bool = "@or"
 primitive throw : Exception -> a = "@throw"
 primitive ignore : a -> Void = "@ignore"
+primitive box : a -> Box<a> = "@box"
+primitive unbox : Box<a> -> a = "@unbox"
 
 exception DecRefError
 
+export let decimalCount32 = box((n) => 0n)
+export let utoa32Buffered = box((a, b, c) => void)
```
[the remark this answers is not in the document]
peblair: Can't. These are updated by `numberUtils`, through the boxes being reexported by `memory`.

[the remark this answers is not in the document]
peblair: (also not sure if cross-module `let mut`s even work)

stdlib/runtime/numbers.gr
peblair: See other replies

stdlib/runtime/wasi.gr
peblair: 🤷‍♂️ 
@phated Should I change it? If so, can you suggest a change?

stdlib/runtime/gc.gr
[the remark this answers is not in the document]
ospencer: They do work, but I don't see why it wouldn't work here?

stdlib/runtime/wasi.gr
phated:
```suggestion
/**
 * Invokes the `fd_write` system call.
 *
 * @param file_descriptor: The file descriptor to write to
 * @param iovs: The pointer to the array of iovs to write
 * @param iovs_len: The length of the array of iovs
 * @param nwritten: Where to store the number of bytes written
 * @return The number of bytes written
 */
```
We use colon as a separator, but like I mentioned above, this won't be documented by graindoc because we don't iterate foreign statements afaik (and the signatures don't have parameter names to match on).

stdlib/sys/file.gr
phated: Whoops, thanks!

stdlib/runtime/gc.gr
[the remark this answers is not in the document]
peblair: The indirection (`gc -> memory -> numberUtils`) means that we still can't use `let mut`:
image

stdlib/runtime/exception.gr
```diff
@@ -29,10 +29,16 @@ export let dangerouslyRegisterBasePrinter = (f) => {
     }
     current = next
   }
+  // We don't decRef the closure or arguments here to avoid a cyclic dep. on Memory.
+  // This is fine, as this function should only be called once.
+  void
 }
 
-export let dangerouslyRegisterPrinter = (f) => {
+export let rec dangerouslyRegisterPrinter = (f) => {
```
phated, line 37: Let's remove this one too then.
```suggestion
export let dangerouslyRegisterPrinter = (f) => {
```

stdlib/runtime/gc.gr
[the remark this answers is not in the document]
phated: #818

bobzhang:
Curios why grain moved from caller owned to callee owned? It seems the callee owned convention introduced a lot more incRef/decRefs.

ospencer:
Hey @bobzhang 🙂 

Back when this PR went in it was just because it was easier for us to maintain since we weren't very experienced with managing refcounts. We've learned a lot since then and could probably implement it much better now.

The only open question now would be how to deal with tail calls—we'd need to come up with a way to transfer ownership of a value to the callee.

ospencer:
I actually just rewrote the callee-owned implementation in #1714 and reduced the number of incRefs quite a bit. There's still a few optimizations we could add to improve it further, but it was a big leap.

bobzhang:
@ospencer thanks for sharing the experience. Tailcall indeed seems to be challenging for caller owned ownership.

PR #807, opened 2021-08-23
feat(stdlib): Int32/64 clz, ctz, popcnt, rotl, rotr, eq, ne, eqz

### LINKED ISSUE #804: Int32 and Int64 Popcnt
Please add as many as possible bit operations that available on wasm. such as popcnt. and rotl rotr... etc.
They are very good in encryption operation and more.
Thanks...

ospencer:
Closes #804

### DIFF
--- a/compiler/test/stdlib/int32.test.gr
+++ b/compiler/test/stdlib/int32.test.gr
@@ -41,3 +41,16 @@ assert !gt(5l, 5l)
 assert !gte(5l, 22l)
 assert !lt(5l, -17l)
 assert !lte(5l, 4l)
+
+assert clz(0b11l) == 30l
+assert ctz(0b11000l) == 3l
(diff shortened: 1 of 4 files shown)


phated (approved):
Seems straight forward! One question though - do we want `eq` and `ne` to be exported as that or as the `==` and `!=` operators? I forget where we landed on all that.

marcusroberts (approved):

ospencer:
@phated I think we wanted the infix operators for the things that would eventually get exported from Pervasives, but for these, the alphabetical names are easier to use when importing (and to force you to be explicit if you want to shadow the operators).

peblair (changes requested):
Needs GC instructions now that #803 has landed

peblair (approved):
LGTM 👏 🚀

PR #808, opened 2021-08-23
feat(stdlib): Add Result.expect & Result.unwrap functions

phated:
This adds `Result.expect` and `Result.unwrap` functions to the stdlib. I realized that my test changes in #792 wouldn't actually work, so I'll need these methods to write those tests in a (relatively) nice fashion.


ospencer (approved):
Fantastic!

peblair (approved):

PR #809, opened 2021-08-23
feat(graindoc): Add support for extra information in module docblock

phated:
When I started writing the docs for our `stdlib/sys/file.gr` module, I noticed that I needed this extra content under the module definition:

> Many of the functions in this module are not intended to be used directly, but rather for other libraries to be built on top of them.

This is too long to inline into the `@module` attribute line, so instead I am now inserting any non-attribute content between the attr_desc and the example.


ospencer (approved):
LGTM, but won't we get an extra newline in the Markdown if there's no extra content? 😛

phated:
> won't we get an extra newline in the Markdown if there's no extra content?

Probably, let me try running this on another file 🤔

phated:
@ospencer you were right. I've updated to check for a non-empty string.

marcusroberts (approved):
Nice work

peblair (approved):
:shipit:

PR #811, opened 2021-08-24
chore(compiler): Introduce timeout for compiler tests

peblair:
This PR introduces a timeout for tests in the test suite. This is useful when developing changes to the runtime, as memory corruption can often result in hangs. When this happens, the runner tells you the test suite but not the individual test which is hanging. With this change, hanging tests will fail, and they will then be displayed on the console.

I went ahead and set the timeout to 15 seconds, but this was an arbitrary choice.

### DIFF
--- a/compiler/test/runner.re
+++ b/compiler/test/runner.re
@@ -83,7 +83,15 @@ let run = (~num_pages=?, file) => {
     Unix.open_process_full(command, Unix.environment());
 
   let pid = Unix.process_full_pid((stdout, stdin, stderr));
-  let (_, status) = Unix.waitpid([], pid);
+  let (_, status, timed_out) =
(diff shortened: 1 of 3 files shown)


compiler/test/runner.re
```diff
@@ -98,6 +106,12 @@ let run = (~num_pages=?, file) => {
     | _ => failwith("process did not exit properly")
     };
 
+  let out =
+    if (timed_out) {
+      "Timed out!\n" ++ out;
```
peblair, line 111: This is just to make the test fail. It was simpler than wiring up the `expect` values, and it has the added benefit of still displaying whatever output _did_ manage to get printed.

marcusroberts (approved):

PR #817, opened 2021-08-24
feat(compiler)!: Remove user manual memory management calls when --no-gc supplied

### LINKED ISSUE #596: Compiler: Determine if dependency cache needs to be reset when running tests

ospencer:
Closes #596


compiler/test/suites/optimizations.re
```diff
@@ -401,7 +403,7 @@ describe("optimizations", ({test}) => {
                   AExp.comp(
                     Comp.app(
                       ~allocation_type=HeapAllocated,
-                      ~tail=true,
+                      ~tail=false, // [TODO] not marked as tail-callable?
```
peblair, line 406: @ospencer Can you please investigate this?

peblair:
> LGTM 👍 I was wondering what caused all of the snapshots to change so drastically, though.

@phated I believe it was because our compiler flags were getting erroneously polluted as the test suite ran in the past. Note that one of the changes I made on one of these PRs (I forget if it was this or one of the ones we merged in) is that each test's config gets reset to `[]`.

It appears we weren't emitting the GC instructions for the basic functionality tests (for example), but we never noticed, since there was never a function call or anything heap-y. Technically, since the tests don't explicitly disable them, we should be emitting the instructions.

PR #819, opened 2021-08-25
feat(compiler): Add --no-bulk-memory flag to disable bulk memory ops

phated:
This is a derivative of @ospencer's `near-example` branch that adds a `--no-bulk-memory` CLI flag to turn off our bulk memory operations on platforms that don't support it.

There's still a bug with 2 tests that we need to debug.


compiler/test/runner.re
```diff
@@ -29,6 +29,7 @@ let read_stream = cstream => {
 let compile = (~hook=?, name, prog) => {
   Config.preserve_config(() => {
     Config.include_dirs := [test_libs_dir, ...Config.include_dirs^];
+    Config.no_bulk_memory := true;
```
phated, line 32: Just for showing this is working (or not, in the case of the 2 failing tests). Will remove before merging.

compiler/src/middle_end/analyze_inline_wasm.re
```diff
@@ -247,7 +237,22 @@ let get_primitive_i32 = get_primitive(primitive_map_i32);
...
+      List.to_seq([
+        ("grow", "@wasm.memory_grow"),
+        ("size", "@wasm.memory_size"),
+        ("compare", "@wasm.memory_compare"),
+      ]),
+    );
+  if (Config.no_bulk_memory^ == false) {
+    StringHash.add(primitive_map_memory, "copy", "@wasm.memory_copy");
+    StringHash.add(primitive_map_memory, "fill", "@wasm.memory_fill");
+  };
```
phated, line 253: I might turn this into a `switch` instead because this is a bit too clever.

phated:
I decided to update the `analyze_inline_wasm.re` file to not use the StringHash maps and instead used switch statements - this allows us to easily vary which ones are enabled using pattern guards!

compiler/src/middle_end/analyze_inline_wasm.re
```diff
@@ -32,222 +33,216 @@ let get_inline_wasm_type = (id: Ident.t): inline_type =>
...
+  get_primitive(prim);
+};
+let get_primitive_memory = id => {
+  let prim =
+    switch (id) {
+    | "grow" => Some("@wasm.memory_grow")
+    | "size" => Some("@wasm.memory_size")
+    | "compare" => Some("@wasm.memory_compare")
+    | "copy" when Config.bulk_memory^ => Some("@wasm.memory_copy")
+    | "fill" when Config.bulk_memory^ => Some("@wasm.memory_fill")
```
phated, line 240: So much nicer when using a `switch` and pattern guards!

phated:
@grain-lang/core this should be ready to go (tests and all!) 🎉

peblair (approved):
LGTM, but I will defer to @ospencer on the `analyze_inline_wasm` changes (since I'm less familiar with that part of the code).

marcusroberts (approved):
Looks good.

cli/bin/grain.js
```diff
@@ -113,6 +113,7 @@ program
   .graincOption("--lsp", "generate lsp errors and warnings only")
   .graincOption("--no-color", "disable colored output")
   .graincOption("--no-gc", "turn off reference counting garbage collection")
+  .graincOption("--no-bulk-memory", "turn off bulk memory operations")
```
ospencer, line 116: What do you think of this description?
```suggestion
  .graincOption("--no-bulk-memory", "polyfill WebAssembly bulk memory instructions")
```

phated: I like it!

PR #824, opened 2021-08-26
fix(compiler): Added the link time fixes for Windows to Graindoc

marcusroberts:
Added the missing Dune magic to make the Windows version of the Graindoc exe link properly


ospencer (approved):
@phated Potentially. I don't think we'll really gain much by splitting it up though. You're right that the tooling can depend on just some of the subpackages, but for now it's not really a huge deal.

PR #827, opened 2021-08-28
fix(compiler): Fix typechecking of single-argument tuple annotation

ospencer:
Do not @ me about if Grain should support single argument tuples or not; I just realized that the typechecker doesn't like it when you write an annotation for one 😂


peblair (approved):
@ospencer

phated (approved):
⚔️

marcusroberts (approved):
Looks good, I'll go fix the formatter.

PR #830, opened 2021-08-29
fix(compiler): Do not add transitive dependencies as imports during CRC consistency check

peblair:
I ran into this when doing some development today, and I'm not precisely sure why it isn't captured by the existing relative import tests. Basically, the CRC check introduces transitive dependencies as direct dependencies, which causes them to bubble up in the CMIs of modules in the dependency tree. This is a problem for two reasons:

1. It makes the CMIs much bigger than needed (since they don't contain just the direct dependencies)
2. Relative imports can result in CMIs containing bogus dependencies (since they are the relative path relative to _some other module_, as opposed to the module whose CMI it is placed in).

I found the source of the problem and removed it. This doesn't appear to have been a deliberate decision by us; it appears to have been a holdover from the OCaml typechecker (which doesn't need to worry about the issue, since they don't do dependency compilation). 

NB: The snapshots are changed since the CMIs are now smaller.

### DIFF
--- a/compiler/src/typed/env.re
+++ b/compiler/src/typed/env.re
@@ -747,7 +747,6 @@ let check_consistency = ps =>
         switch (crco) {
         | None => ()
         | Some(crc) =>
-          add_import(name);
           let resolved_file_name =
             Module_resolution.resolve_unit(
               ~base_dir=Filename.dirname(ps.ps_filename),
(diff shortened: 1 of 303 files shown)


ospencer (approved):
I always wondered why we did this but never had time to investigate! Thanks for taking care of this 😄

marcusroberts (approved):
Probably the most efficient PR ever?

PR #831, opened 2021-08-29
feat(stdlib): Add unsigned version of Int32/Int64 comparison operations

peblair:
`Int32` and `Int64` currently do not expose the unsigned versions of their operations, making unsigned i32/i64 values currently unsupported. I've added them in this PR, but we should discuss details such as naming before merging.


ospencer:
I've thought about this previously and wondered if we should instead provide Uint32/Uint64 libs that have the unsigned instructions for the Int32 type, but I never came to a decision. What do you think?

peblair:
I don't have a strong opinion about it, as there are pros and cons to both.

Pros for this approach: Arithmetic operations are easily facilitated between signed/unsigned values, and, at the end of the day, signed-ness is only important for comparison operations.

Pros for the other approach: More difficult to shoot yourself in the foot, thanks to the explicit type.

So, like I said, no strong opinion. If you or the others think we should make the change, we can do so. I just think we should expose this functionality _somehow_.

ospencer:
Oh, sorry if I made it sound like I wanted a `Uint32` type. My thought was to use the `Int32` type, just the unsigned instructions are in a module called `uint32.gr`. So no extra type safety 😛

ospencer:
But also maybe we should just do it that way. I _do_ like type safety.

peblair:
@ospencer I implemented what you suggested, but ended up coming up with an implementation which is sort of in the middle between the original version of this PR and what you suggest: I kept the new `ltU`/`gtU`/etc functions, but also added `Uint32`/`Uint64` modules which default to unsigned operations (and provide dual `ltS`/`gtS`/etc functions). Let me know what you think (also, apologies for the whitespace in the diff).

ospencer:
In general I'm good with it! Can give a deeper review in a bit. Only thing I wonder about is the "two ways to do the same thing" and if we should only have the unsigned ones in `uint32.gr` 🤔

peblair:
My rationale for doing it this was was the existence of `shrU`, `divU`, and `remU` in `Int32`/`Int64`. I felt it would be weird to leave those and not put in the corresponding `ltU`/etc; alternatively, we could remove those, but then this would be breaking (and I sort of like it this way, since it makes it easy to access the alternative implementation if you just need it in one spot in a file).

ospencer:
@peblair Could you convert this to `@unsafe`?

peblair:
@ospencer This should be ready for approval now (it uses `@unsafe`).

stdlib/uint32.gr
```diff
@@ -0,0 +1,411 @@
...
+/**
+ * Computes the quotient of its operands using signed division.
+ *
+ * @param x: The first operand
+ * @param y: The second operand
+ * @returns The quotient of its operands
+ *
+ * @since v0.5.0
+ */
+export let divS = Int32.div
```
phated, line 117: @ospencer was right, this feels really weird to me. Just re-exporting with the names swapped doesn't make sense to me when we already allow people to "import-as"

phated (changes requested):
So, yeah, can we drop the "re-export" modules and just have the unsigned operators in Int32 and Int64? Having these modules completely aliased but swapping the names seems confusing (not to mention all the work keeping docs in sync, etc).

phated:
I think we just have Int32 and Int64 that have the "U" operations. If someone wants to use those as their primary operations, they can `import { divU as div } from "int32" `

I believe the only changes to this PR would be removing the uint32 and uint64 modules and moving the tests?

peblair:
Got it, thanks. @ospencer can you confirm that you're cool with this change before I implement it?

ospencer:
Yep, that's fine.

phated (approved):
Looking great! Let's merge it 😄

ospencer (approved):
Excellent work!

PR #832, opened 2021-08-30
fix(js-runner): Monkeypatch fs_write in browser environment

phated:
While trying to update the Grain web example, I found out that using fd_write on `/dev/stdout` doesn't actually log in the browser.

This monkeypatches the memfs library used by wasmerjs to `console.log` or `console.error` when fd 0 or 1 is written to.

The downside to this patching is that the newline at the end of our `print` is stored in a different iov (?) so it logs separately (here's a screenshot):
Screen Shot 2021-08-29 at 5 49 33 PM

### DIFF
--- a/js-runner/src/core/grain-module.js
+++ b/js-runner/src/core/grain-module.js
@@ -6,6 +6,21 @@ let bindings;
 
 if (__RUNNER_BROWSER) {
   const wasmFs = new WasmFs();
+  const decoder = new TextDecoder("utf-8");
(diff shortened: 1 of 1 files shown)


marcusroberts (approved):
I can't see any better way to do this having looked at the wasmer code. so LGTM!

peblair (approved):
Quite sad about the newlines, but I'm glad we can at least print 🙂 
@phated Maybe make a follow-up issue about the newlines?

phated:
> Maybe make a follow-up issue about the newlines?
#849

PR #833, opened 2021-08-30
fix(js-runner): Add default object for destructuring

phated:
While trying to update the Grain web example, I found out that our js-runner crashes if you don't provide an options object to `buildGrainRunner` due to trying to destructure from undefined in `ManagedMemory`.

This just adds a default object to the constructor of `ManagedMemory` so it doesn't crash like that.


ospencer (approved):

marcusroberts (approved):
I saw the same earlier, nice to see if fixed.

peblair (approved):

PR #837, opened 2021-08-30
fix(compiler): Fix incorrect char well-formedness error

ospencer:
@jozanza noticed that 3-byte utf-8 characters trigger a well-formedness error for char literals. I think there's a bug in the utf8 library we're using (or that I just didn't understand what a few functions did), so here's an easy workaround. Issue to see what's up with the function we were using: https://github.com/reasonml/reason-native/issues/266


phated (approved):
👍

marcusroberts (approved):
LGTM!

Issue #795, opened 2021-08-18
Array.rotate


ospencer:
Implement `Array.rotate`. Moves a number of items from the start of an array to the end. A negative number moves items from the end to the start. This operation is done in place. See Ruby's [`Array#rotate!`](https://ruby-doc.org/core-2.6.1/Array.html#method-i-rotate-21).

ibdafna:
Going to give this a try!

cician:
The merged implementation fails for empty arrays and only works for arrays of Number. The intention was to accept arrays of any type of elements, right?

If so, do you want me to make the adjustments?

Cheers

cician:
I went ahead and made a PR: #1048

phated:
This was completed between #838 and #1048

PR #838, opened 2021-08-30
feat(stdlib): Add rotate function in Array module

ibdafna:
Signed-off-by: Itay Dafna <i.b.dafna@gmail.com>

Addresses #795 

Usage: 
```reason
let myarray = [> 1, 2, 3, 4, 5]
Array.rotate(myarray, 2)
print(myarray) // [> 4, 5, 1, 2, 3]

let myarray2 = [> 1, 2, 3, 4, 5]
Array.rotate(myarray2, -1)
print(myarray2) // [> 2, 3, 4, 5, 1]
```

TODO:

- [x] Documentation / Docstring
- [x] Testing

### DIFF
--- a/compiler/test/stdlib/array.test.gr
+++ b/compiler/test/stdlib/array.test.gr
@@ -360,9 +360,26 @@ assert result == void
 
 
 // Array.rotate
-assert Array.rotate([> 1, 2, 3, 4, 5], 0) == [> 1, 2, 3, 4, 5]
-assert Array.rotate([> 1, 2, 3, 4, 5], 3) == [> 3, 4, 5, 1, 2]
-assert Array.rotate([> 1, 2, 3, 4, 5], -4) == [> 5, 1, 2, 3, 4]
-assert Array.rotate([> 1, 2, 3, 4, 5], 5) == [> 1, 2, 3, 4, 5]
-assert Array.rotate([> 1, 2, 3, 4, 5], 48) == [> 3, 4, 5, 1, 2]
-assert Array.rotate([> 1, 2, 3, 4, 5], -54) == [> 5, 1, 2, 3, 4]
+let arr1 = [> 1, 2, 3, 4, 5]
+Array.rotate(0, arr1) 
+assert arr1 == [> 1, 2, 3, 4, 5]
+
+let arr2 = [> 1, 2, 3, 4, 5]
+Array.rotate(3, arr2)
+assert arr2 == [> 3, 4, 5, 1, 2]
+
+let arr3 = [> 1, 2, 3, 4, 5]
+Array.rotate(-4, arr3)
+assert arr3 == [> 5, 1, 2, 3, 4]
+
+let arr4 = [> 1, 2, 3, 4, 5]
+Array.rotate(5, arr4)
+assert arr4 == [> 1, 2, 3, 4, 5]
+
+let arr5 = [> 1, 2, 3, 4, 5]
+Array.rotate(48, arr5) 
+assert arr5 == [> 3, 4, 5, 1, 2]
+
+let arr6 = [> 1, 2, 3, 4, 5]
+Array.rotate(-54, arr6) 
+assert arr6 == [> 5, 1, 2, 3, 4]
--- a/compiler/test/stdlib/array.test.gr
+++ b/compiler/test/stdlib/array.test.gr
@@ -357,3 +357,9 @@ let result = Array.sort(comp, arr)
 assert arr == [> "a", "a", "ab", "abc", "abcd", "abcde"]
 // Returns void
 assert result == void
+
+
+// Array.rotate
+assert Array.rotate([> 1, 2, 3, 4, 5], 0) == [> 1, 2, 3, 4, 5]
+assert Array.rotate([> 1, 2, 3, 4, 5], 3) == [> 3, 4, 5, 1, 2]
+assert Array.rotate([> 1, 2, 3, 4, 5], -4) == [> 5, 1, 2, 3, 4]
--- a/compiler/test/stdlib/array.test.gr
+++ b/compiler/test/stdlib/array.test.gr
@@ -363,3 +363,6 @@ assert result == void
 assert Array.rotate([> 1, 2, 3, 4, 5], 0) == [> 1, 2, 3, 4, 5]
 assert Array.rotate([> 1, 2, 3, 4, 5], 3) == [> 3, 4, 5, 1, 2]
 assert Array.rotate([> 1, 2, 3, 4, 5], -4) == [> 5, 1, 2, 3, 4]
+assert Array.rotate([> 1, 2, 3, 4, 5], 5) == [> 1, 2, 3, 4, 5]
+assert Array.rotate([> 1, 2, 3, 4, 5], 48) == [> 3, 4, 5, 1, 2]
+assert Array.rotate([> 1, 2, 3, 4, 5], -54) == [> 5, 1, 2, 3, 4]
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -906,9 +906,9 @@ export let sort = (comp, array) => {
  * @param n: The number of elements to rotate by
  * @returns The rotated array
  *
- * @example: 'rotate([> 1, 2, 3, 4, 5], 2) == [> 4, 5, 1, 2, 3]'
- * @example: 'rotate([> 1, 2, 3, 4, 5], -1) == [> 2, 3, 4, 5, 1]'
- * @since v0.4.4
+ * @example: rotate([> 1, 2, 3, 4, 5], 2) == [> 4, 5, 1, 2, 3]
+ * @example: rotate([> 1, 2, 3, 4, 5], -1) == [> 2, 3, 4, 5, 1]
+ * @since v0.4.5
  */
 export let rotate = (arr, n) => {
   let rec gcd = (a, b) => {
@@ -927,13 +927,13 @@ export let rotate = (arr, n) => {
   let mut j  = 0
   for (let mut i = 0; i < gcd(arrLen, k); i += 1) {
     j = i
-    temp = get(i, arr)
+    temp = arr[i]
     while (true) {
       d = (j - k) % arrLen
       if (d == i) {
         break
       }
-      let newVal = get(d, arr)
+      let newVal = arr[d]
       arr[j] = newVal
       j = d
     }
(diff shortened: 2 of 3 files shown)


phated:
@ibdafna this is very cool! Let us know if you need help or guidance getting the 2 TODOs completed. Cheers 🍻

hhugo:
The current implementation will likely trigger a stack overflow on large array

ibdafna:
@phated I'll sort everything out today - I saw the post on `graindoc` so will try that out 😄
We can also discuss an alternative implementation if there are concerns with the recursive approach!

peblair (changes requested):
Hello, @ibdafna! Thanks for the contribution!

Taking a look at the implementation and considering @hhugo 's reply, I agree that it may make more sense to use a `for` loop here. Because the recursion is not in the tail position, the size of the call stack will grow with the size of the array. Would you mind refactoring this to use a `for` loop instead? Here is an [example](https://www.codewhoop.com/array/rotation-in-place.html) of what this might look like (while still being `O(n)`).

ibdafna:
@peblair Sure, happy to refactor!

phated:
@ibdafna Any way we can help you get this over the finish line?

ibdafna:
@phated @marcusroberts That's awesome! I'll be attending this Thursday and looking forward to meet you all 😄 

@peblair Many thanks for the link! My implementation is 99% based on the logic there - apart from a small change to make it rotate to the right and not to the left.

As I am using @phated 's logic for wrapping a negative index number (defined in Array.slice), I made that function a private helper function in https://github.com/grain-lang/grain/blob/17acd95f7b74c690483d25f51b4180227caf27d1/stdlib/array.gr#L808-L824 I wasn't sure where to put the function in the file, as it must be defined before any other function in array.gr can use it, so I just put it above the only two functions which are currently using it.

ibdafna:
@phated no worries! All done now 😄

phated (approved):
Excellent! Love this. I'll toss a request for reviews by the rest of the team 🍻

phated:
Reminder to myself: We need to regenerate the markdown file before merge.

compiler/test/stdlib/array.test.gr
```diff
@@ -286,3 +286,8 @@ assert Array.slice(2, 1, testChars) == [>]
 assert Array.slice(-1, -2, testChars) == [>]
 // Empty
 assert Array.slice(1, 4, [>]) == [>]
+
+// Array.rotate
+assert Array.rotate([> 1, 2, 3, 4, 5], 0) == [> 1, 2, 3, 4, 5]
+assert Array.rotate([> 1, 2, 3, 4, 5], 3) == [> 3, 4, 5, 1, 2]
+assert Array.rotate([> 1, 2, 3, 4, 5], -4) == [> 5, 1, 2, 3, 4]
```
ospencer, line 293: Going to add a couple more tests here to test the gcd behavior!

```suggestion
assert Array.rotate([> 1, 2, 3, 4, 5], -4) == [> 5, 1, 2, 3, 4]
assert Array.rotate([> 1, 2, 3, 4, 5], 5) == [> 1, 2, 3, 4, 5]
assert Array.rotate([> 1, 2, 3, 4, 5], 48) == [> 3, 4, 5, 1, 2]
assert Array.rotate([> 1, 2, 3, 4, 5], -54) == [> 5, 1, 2, 3, 4]
```

ospencer:
@ibdafna The tests I added failed, and (after double-checking that my tests were correct) I realized there's a small bug. `wrapNegativeIndex` only works if the index is properly in the range of the array, so asking to rotate a length 5 array by `-54` was treating the `-54` as `-49` rather than the value `1`. Conveniently, Grain's modulo operator is proper modulo (not remainder), so `n % arrLen` provides the behavior we want here. (`-4 % 5 == 1`, `-54 % 5 == 1`, etc.)

Hope that makes sense! (And let me know if it doesn't 😄)

ospencer:
@phated Random other thought: this implementation returns the array. Should it?

phated:
> Random other thought: this implementation returns the array. Should it?

I was thinking about this in relation to the Array.sort PR... I think we should be consistent between the 2, but I'm not sure which we would prefer. I like that sort returns the array in JS but it isn't clear that it is a mutating operation, so maybe any mutating operations in Grain should always return Void. WDYT?

ospencer:
I'm torn for the same reason. Returning the array is really convenient, but it's not obvious that it mutates. I might lean more towards doing the obvious thing?

phated:
> I might lean more towards doing the obvious thing?

Same, I think obvious is more important than convenient.

spotandjake:
> I'm torn for the same reason. Returning the array is really convenient, but it's not obvious that it mutates. I might lean more towards doing the obvious thing?

could good documentation make up for this. the convenience is really nice as long as it isnt overly missleading. Those are just my though so either way is good i am just wondering if good documentation could be a solution here while allowing the convience.

ibdafna:
> 
> 
> @ibdafna The tests I added failed, and (after double-checking that my tests were correct) I realized there's a small bug. `wrapNegativeIndex` only works if the index is properly in the range of the array, so asking to rotate a length 5 array by `-54` was treating the `-54` as `-49` rather than the value `1`. Conveniently, Grain's modulo operator is proper modulo (not remainder), so `n % arrLen` provides the behavior we want here. (`-4 % 5 == 1`, `-54 % 5 == 1`, etc.)
> 
> Hope that makes sense! (And let me know if it doesn't 😄)

@ospencer Thanks for adding these, good catch! I remember getting a notification from GitHub with the new test cases comment and thinking to myself "this case is actually not handled and will break the function"

ibdafna:
Rebased on main. I did notice unrelated tests (formatter) failing on both this and the main branch.

stdlib/array.gr
```diff
@@ -892,3 +895,49 @@ export let sort = (comp, array) => {
...
+    if (b == 0) {
+      a
+    }
+    else{
+      gcd(b, a % b)
+    }
+  }
+
+  let arrLen = length(arr)
+  let k = wrapNegativeIndex(arrLen, n % arrLen)
```
ospencer, line 924: This isn't necessary anymore, since `n % arrLen` will always be positive.

stdlib/array.gr
```diff
@@ -892,3 +895,49 @@ export let sort = (comp, array) => {
...
+      if (d == i) {
+        break
+      }
+      let newVal = arr[d]
+      arr[j] = newVal
+      j = d
+    }
+    arr[j] = temp
+  }
+  arr
```
phated, line 942: I believe we are going to remove the return value since this is a mutating function
```suggestion
```

stdlib/array.gr
```diff
@@ -892,3 +895,49 @@ export let sort = (comp, array) => {
...
+ *
+ * @param arr: The array to be rotated
+ * @param n: The number of elements to rotate by
+ * @returns The rotated array
+ *
+ * @example: rotate([> 1, 2, 3, 4, 5], 2) == [> 4, 5, 1, 2, 3]
+ * @example: rotate([> 1, 2, 3, 4, 5], -1) == [> 2, 3, 4, 5, 1]
+ * @since v0.4.5
+ */
+export let rotate = (arr, n) => {
```
phated, line 913:
```suggestion
export let rotate = (n, arr) => {
```
Swapping the array argument because we are making everything data-last right now (this might change in the future, but this is our blanket rule).

stdlib/array.gr
```diff
@@ -892,3 +895,49 @@ export let sort = (comp, array) => {
...
+
+/**
+ * Rotates an array by n elements to the right, in place.
+ *
+ * If n is negative, the array will be rotated by n elements
+ * to the left. See examples.
+ *
+ * @param arr: The array to be rotated
+ * @param n: The number of elements to rotate by
+ * @returns The rotated array
```
phated, line 907:
```suggestion
 * @param n: The number of elements to rotate by
 * @param arr: The array to be rotated
```
Removed the returns and updated the param order.

stdlib/array.gr
```diff
@@ -892,3 +895,49 @@ export let sort = (comp, array) => {
...
+ *
+ * @example: rotate([> 1, 2, 3, 4, 5], 2) == [> 4, 5, 1, 2, 3]
+ * @example: rotate([> 1, 2, 3, 4, 5], -1) == [> 2, 3, 4, 5, 1]
```
phated, line 910:
```suggestion
 * @example let array = [> 1, 2, 3, 4, 5]; rotate(2, arr); arr == [> 4, 5, 1, 2, 3]
 * @example let array = [> 1, 2, 3, 4, 5]; rotate(-1, arr); arr == [> 2, 3, 4, 5, 1]
```
A little messy on these examples, but that's the only way to do it if we don't return an array.

compiler/test/stdlib/array.test.gr
```diff
@@ -357,3 +357,12 @@ let result = Array.sort(comp, arr)
...
 assert result == void
+
+
+// Array.rotate
+assert Array.rotate([> 1, 2, 3, 4, 5], 0) == [> 1, 2, 3, 4, 5]
+assert Array.rotate([> 1, 2, 3, 4, 5], 3) == [> 3, 4, 5, 1, 2]
+assert Array.rotate([> 1, 2, 3, 4, 5], -4) == [> 5, 1, 2, 3, 4]
+assert Array.rotate([> 1, 2, 3, 4, 5], 5) == [> 1, 2, 3, 4, 5]
+assert Array.rotate([> 1, 2, 3, 4, 5], 48) == [> 3, 4, 5, 1, 2]
+assert Array.rotate([> 1, 2, 3, 4, 5], -54) == [> 5, 1, 2, 3, 4]
```
phated, line 368: there will all need to be updated since the argument order will change and the function won't return an array.

phated (changes requested):
Talked with the team and we want a couple things.
1. All stdlib functions are data-last, so we'd want the array being rotated to be the 2nd param
2. Mutating functions shouldn't return the mutated value, otherwise the API might be confusing (like array.sort in JS).

ibdafna:
> Talked with the team and we want a couple things.
> 
>     1. All stdlib functions are data-last, so we'd want the array being rotated to be the 2nd param
> 
>     2. Mutating functions shouldn't return the mutated value, otherwise the API might be confusing (like array.sort in JS).

All done! 😄

phated (approved):
Looks **great** to me!

ospencer (approved):
Amazing job, and thanks for your patience 🙏

PR #839, opened 2021-08-31
fix(linker): Split inits from starts

phated:
This is a bit of a thought exercise, but also a potential fix for the issue we have where a start function breaks nodejs (and browser?) compatibility. I've done it by separating the things needed to initialize main being configured in an "init" function which is set as the start function and then only calling the main function inside the start (which happens after initialization).


phated:
Closing, as this only works as a happenstance and wouldn't work across the board.

PR #844, opened 2021-09-01
feat(linker): Add --use-start-section flag to use start section in output

phated:
This adds a config (and corresponding flag) to the linker that allows people to disable the WASI convention of exporting a `_start` function and instead set a start section in the linked output. This is needed for NEAR because they don't call the `_start` function and rely on the start section during instantiation.

I want to make 1 more improvement to the added tests after @ospencer takes a stab at parsing the exports in our Wasm_utils.

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -121,6 +121,10 @@ program
     "--wasi-polyfill <filename>",
     "path to custom WASI implementation"
   )
+  .graincOption(
+    "--use-start-section",
+    "replaces the _start export with a start section during linking"
+  )
   .graincOption("--no-link", "disable static linking")
   .graincOption(
     "--no-pervasives",
--- a/compiler/test/suites/linking.re
+++ b/compiler/test/suites/linking.re
@@ -34,4 +34,73 @@ describe("linking", ({test}) => {
     {|print("foo")|},
     "",
   );
+
+  let tuple_equal = ((a1, a2), (b1, b2)) => a1 == b1 && a2 == b2;
+  test("no_start_section", ({expect}) => {
+    let name = "no_start_section";
+    let outfile = wasmfile(name);
+    compile(name, {|print("Hello, world!")|});
+    let sections =
+      Grain_utils.Wasm_utils.get_wasm_sections(open_in_bin(outfile));
+    let export_sections =
+      List.find_map(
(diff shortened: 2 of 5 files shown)


compiler/test/suites/linking.re
```diff
@@ -16,4 +16,46 @@ describe("linking", ({test}) => {
...
+    let start_section =
+      List.find_opt(
+        (sec: Grain_utils.Wasm_utils.wasm_bin_section) =>
+          switch (sec) {
+          | {sec_type: Start} => true
+          | _ => false
+          },
+        sections,
+      );
+    //TODO: Check that _start is not exported, waiting on Oscar's Wasm_utils update
```
peblair, line 58: Maybe we should link this to an issue?

compiler/test/suites/linking.re
```diff
@@ -16,4 +16,46 @@ describe("linking", ({test}) => {
...
+    let start_section =
+      List.find_opt(
+        (sec: Grain_utils.Wasm_utils.wasm_bin_section) =>
+          switch (sec) {
+          | {sec_type: Start} => true
+          | _ => false
+          },
+        sections,
+      );
+    //TODO: Check that _start is exported, waiting on Oscar's Wasm_utils update
```
peblair, line 35: See below

peblair (approved):
See comments, but looks good 👍

compiler/test/suites/linking.re
phated: I'm hoping to have it done before I merge this 🤞 - otherwise I'll open an issue.

phated:
@grain-lang/core this is now done (with export tests, too, thanks to @ospencer's #845!)

ospencer (approved):

phated:
I just realized that I didn't propagate this to the grain CLI. On it!

cli/bin/grain.js
```diff
@@ -121,6 +121,10 @@ program
     "--wasi-polyfill <filename>",
     "path to custom WASI implementation"
   )
+  .graincOption(
+    "--use-start-section",
+    "replaces the _start export with a start section during linking"
```
phated, line 126: @ospencer are you happy with this description?

phated: Checked in discord and he said good enough 😄

phated:
Opened this #850 about incompatible flag combinations.

PR #846, opened 2021-09-02
fix(stdlib): Remove fd_sync calls that are not used

phated:
Between #800 being submitted and merged, we also landed some changes to the stdlib that changed the calling convention in non-GC contexts. This required adding some `fd_sync` imports for debugging and they didn't get removed, which caused the linker to fail on our wasi polyfill tests.

Here, I've added a better error message for this case and then removed the 2 references to fd_sync in our stdlib, so the tests should pass again.

### DIFF
--- a/compiler/src/linking/link.re
+++ b/compiler/src/linking/link.re
@@ -451,10 +451,22 @@ let link_all = (linked_mod, dependencies, signature) => {
           let internal_name = Function.get_name(func);
           let wasi_polyfill = wasi_polyfill_module();
           let new_name =
-            Hashtbl.find(
+            Hashtbl.find_opt(
               Hashtbl.find(exported_names, wasi_polyfill),
               imported_name,
             );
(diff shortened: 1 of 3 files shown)


peblair (approved):
Thanks for the quick fix!

compiler/src/linking/link.re
```diff
@@ -451,10 +451,22 @@ let link_all = (linked_mod, dependencies, signature) => {
...
               Hashtbl.find(exported_names, wasi_polyfill),
               imported_name,
             );
+          let new_name =
+            switch (new_name) {
+            | Some(new_name) => new_name
+            | None =>
+              failwith(
+                Printf.sprintf(
+                  "Unable to locate `%s` in your polyfill. Required by `%s`",
```
ospencer, line 464: 👏

ospencer (approved):

PR #852, opened 2021-09-04
fix(cli): Support NodeJS installed into a path with spaces

marcusroberts:
When my NodeJS was in C:\Program Files\... the exec would try and run C:\Program

Fix: Wrap the node executable in quotes


peblair (approved):

phated (approved):
Wow, great catch! I assume you'll do this to your formatter command too?

ospencer (approved):

PR #854, opened 2021-09-04
fix(stdlib)!: Use random seed for hash module

ospencer:


phated (approved):
I think this is a breaking fix because there would be a tiny chance that someone was reliant on consistent hashes between runs (They shouldn't have been, but just maybe?)

ospencer:
Maybe? But we can mark it breaking just in case :)

marcusroberts (approved):
Looks good!

peblair (approved):

jozanza (approved):

PR #857, opened 2021-09-04
feat(compiler)!: Don't use import keyword with foreign

ospencer:


ospencer:
We're swapping parsers, so if we still want this it'll need to be reimplemented.

PR #860, opened 2021-09-05
fix(grainfmt): Handle trailing block comments better

### LINKED ISSUE #856: Formatter: Line comments above binding gets extra space
I'm not sure if the title is the correct description...

Given this code:

```grain
// Tell the host where the key and value are located and their lengths.
// If there is an old value it is written to _REG.
let _KEY_LEN = WasmI64.extendI32S(_KEY_LEN)
```

The formatter produces:

```grain
// Tell the host where the key and value are located and their lengths.
// If there is an old value it is written to _REG.

let _KEY_LEN = WasmI64.extendI32S(_KEY_LEN)
```

But I think it should probably keep the comments directly above the code it describes.

marcusroberts:
Improvements to comment handling - handle trailing comments on the same line as the expression only, and then look for any remaining comments at the end of blocks.

Fixes #856

### DIFF
--- a/compiler/grainformat/reformat.re
+++ b/compiler/grainformat/reformat.re
@@ -1179,11 +1179,18 @@ and print_expression =
       expr: Parsetree.expression,
     ) => {
   let (leading_comments, trailing_comments) =
-    Walktree.partition_comments(expr.pexp_loc, Some(parent_loc));
-
-  Walktree.remove_used_comments(leading_comments, trailing_comments);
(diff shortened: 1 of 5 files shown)


phated (approved):
Looks excellent! 💪

phated:
I changed the title a little. Let's use `grainfmt` as the specifier for formatter stuff.

ospencer (approved):

PR #861, opened 2021-09-05
fix(compiler): Don't report comment locations when disabled

marcusroberts:
Noticed when comparing ASTs for the formatter that comments still had their locations shown when location reporting was disabled.


ospencer (approved):

phated (approved):
Great catch!

PR #863, opened 2021-09-05
fix(stdlib): Premature free in toString

ospencer:
As an optimization in the runtime string module, we don't follow the regular calling convention to avoid some unnecessary incRefs/decRefs. Before the calling convention changed, an enum constructor would always incRef its arguments. With the new convention, it still does, however it also decRefs the arguments, expecting that they were incRef'd before being passed.

The tl;dr is that `SLEmpty` needs an extra incRef. We should maybe consider reworking the lists in this module to follow the convention, but I think this is a fine fix for now.

### DIFF
--- a/compiler/test/suites/print.re
+++ b/compiler/test/suites/print.re
@@ -24,4 +24,9 @@ describe("print", ({test}) => {
     "record Foo { foo: Number }; record Bar { bar: Foo }; print({ bar: { foo: 1 } })",
     "{\n  bar: {\n    foo: 1\n  }\n}\n",
   );
+  assertRun(
+    "print_nested_records_multiple",
+    "record Foo { foo: Number }; record Bar { bar: Foo }; print({ bar: { foo: 1 } }); print({ bar: { foo: 1 } }); print({ bar: { foo: 1 } })",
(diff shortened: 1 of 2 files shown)


marcusroberts (approved):
Great job in tracking this down!

phated (approved):
Makes sense! 👍

Why wasn't this caught by the test suite? Does it only surface in certain scenarios? If so, we probably want a regression test.

peblair (commented):
@ospencer I'm a bit confused -- why is the fix here not adding the `incRef` calls to `a` and `b` (or maybe just `b`, with a comment, in order to avoid needing to `decRef` `a` after the call) in `slConsDisableGc`?

ospencer:
@peblair That was sort of my point about maybe refactoring this so it's more understandable. (Also, if you add an incRef to `b`, those cons cells will never get freed since they'd have refcounts of +2. `slConsDisableGC` would need to incRef and then decRef again since it was an argument.)

peblair (approved):
I wish there was a safer way 😢. Maybe we should write a garbage collector for the `@disableGC` code 😂

PR #864, opened 2021-09-06
feat(compiler): Support "or" pattern matching

### LINKED ISSUE #264: Pattern Matching: Support "or" patterns
This is the ability to match variants within a subpattern, like the "or" pattern `P | Q` in this example:

```grain
data Foo = Foo(Bar)
data Bar = P | Q

match(...) {
  | Foo(P | Q) => ...
}
```

phated:
This implements "or" pattern matches using `|` between patterns.

Closes #264

I'm not happy with the way they line-wrap, so I need some help from @marcusroberts to get that working right.

I'll also add some tests here shortly.

### DIFF
--- a/compiler/src/middle_end/matchcomp.re
+++ b/compiler/src/middle_end/matchcomp.re
@@ -317,7 +317,17 @@ module MatchTreeCompiler = {
       };
     };
     let res = extract_bindings(patt, expr);
-    /*Printf.eprintf "Bindings:\n%s\n" (Sexplib.Sexp.to_string_hum (sexp_of_list (sexp_of_pair sexp_of_string (fun x -> (sexp_of_string (Pretty.string_of_cexpr x)))) res));*/
+    /*
+       Printf.eprintf(
+         "Bindings:\n%s\n",
+         Sexplib.Sexp.to_string_hum(
+           sexp_of_list(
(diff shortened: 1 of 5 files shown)


compiler/src/middle_end/matchcomp.re
```diff
@@ -305,11 +305,29 @@ module MatchTreeCompiler = {
...
+       Printf.eprintf(
+         "Bindings:\n%s\n",
+         Sexplib.Sexp.to_string_hum(
+           sexp_of_list(
+             sexp_of_pair(Ident.sexp_of_t, sexp_of_comp_expression),
+             res,
+           ),
+         ),
+       );
+     */
```
jozanza, line 330: What's this? 👀

phated: A utility for printing bindings. It already existed in the code but it was in ocaml since comments don't get converted. I updated it to Reason and made it work with the current compiler.

phated:
This pattern fails with `Internal error: compile_matrix: non-wildcard column returned no constructors`:

```grain
match (1) { 1 | 2 | 3 => true }
```

Need to ask @ospencer or @peblair why...

phated:
Not as simple as this, it seems.

phated:
Going to close this out, but we can keep the branch around for a little while.

PR #865, opened 2021-09-06
fix(grainfmt): Correctly add space between arguments in enum pattern

phated:
The formatter was removing the space between arguments, which I ran into this issue when writing some enum pattern matches. This fixes that so there is always a space after a comma.


ospencer (approved):

marcusroberts (approved):
Nice one!

Issue #783, opened 2021-07-22
Parse error when using a generic type parameter to annotate a function call.


phated:
You get a parse error if you remove the parens around `Array<a>` in
```grain
let array = WasmI32.toGrain(allocateArray(length)): (Array<a>)
```

_Originally posted by @phated in https://github.com/grain-lang/grain/pull/757#r675075568_

peblair:
@phated This program builds on `main`:
```grain
let array = [>1, 2]: Array<a>
```
Are we good to close this?

marcusroberts:
If we are let me know so the formatter can stop adding parens.

phated:
> Are we good to close this?

I don't remember any parser code that was changed to "fix" this. Do you?

phated:
The plan was to change our annotation syntax to `::` instead of `:`

ospencer:
This is still a bug, that example doesn't quite trigger it. Here's one:
```
let foo = [] : List<a>
6
```

ospencer:
Or just 
```
[] : List<a>
6
```

Bug is this is ambiguous between statements `[]: List<a>`/`6`, and `([]: List < a) > 6`.

PR #866, opened 2021-09-06
fix(grainfmt): Use Doc.space and surround trailing type annotations with parens

marcusroberts:
The compiler currently needs parens around the trailing type annotation:

`let ret = WasmI32.toGrain(array): (Array<a>)`

so the formatter adds these now.  Also used Doc.space near colons as better form.

### DIFF
--- a/compiler/grainformat/reformat.re
+++ b/compiler/grainformat/reformat.re
@@ -1806,7 +1806,7 @@ and print_expression =
           ),
           Doc.text(":"),
           Doc.space,
-          Doc.lparen, // needed to fix compiler bug (trailing type annotation needs paren)
+          Doc.lparen, // needed to fix compiler bug (trailing type annotation needs paren, #866)
           print_type(parsed_type, original_source),
           Doc.rparen,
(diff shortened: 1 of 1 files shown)


peblair (approved):
LGTM :+1:

compiler/grainformat/reformat.re
```diff
@@ -1794,8 +1804,11 @@ and print_expression =
             ~parent_loc,
             expression,
           ),
-          Doc.text(": "),
+          Doc.text(":"),
+          Doc.space,
+          Doc.lparen, // needed to fix compiler bug
```
ospencer, line 1809: Maybe reference the issue here?

ospencer (approved):

compiler/grainformat/reformat.re
marcusroberts: Fixed.

phated (approved):
We should really get this parser bug fixed 🙈

ng-marcus:
For future referece #783

PR #868, opened 2021-09-06
feat(grainfmt): Add a mode to check the formatter's AST matches the original

marcusroberts:
This is an optional (default off for performance reasons) check that the AST produced by the formatter matches the original.
It also disables re-writing single expressions inside if statements into blocks as this intentionally modifies the AST.

Draft as I need some help with the command line arguments

### DIFF
--- a/compiler/grainformat/grainformat.re
+++ b/compiler/grainformat/grainformat.re
@@ -83,7 +83,7 @@ let format_code =
   // HELP NEEDED
   // I need to get this value from the command line and
 
-  let check_format = false;
+  let check_format = true;
 
   if (check_format) {
     Grain_utils.Config.formatter_maintain_ast := true;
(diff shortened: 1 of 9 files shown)


compiler/grainformat/grainformat.re
```diff
@@ -80,14 +80,58 @@ let compile_parsed = (filename: option(string)) => {
 
 let format_code =
     (program: Parsetree.parsed_program, original_source: array(string)) => {
-  Reformat.reformat_ast(program, original_source);
-  `Ok();
+  // HELP NEEDED
+  // I need to get this value from the command line and
+
+  let check_format = true;
```
peblair, line 86: @marcusroberts I'm pretty sure it should just be
```suggestion
  let check_format = Grain_utils.Config.formatter_maintain_ast^;
```
, no?

PR #869, opened 2021-09-06
fix(compiler): Stop line comment lex absorbing the newlines after

marcusroberts:
I believe a line comment should only ever be a single line.

The lexer was including any following newlines

// this is a comment\n
\n
\n

Change to add only the newline at the end of the comment to the comment source


peblair (approved):
Nice catch!

ospencer (approved):

PR #875, opened 2021-09-07
fix(stdlib): Correct fdPwrite return value

phated:
When regenerating these docs, I noticed that fdPwrite had the wrong return value.


ospencer (approved):
We should maybe get this out soon.

marcusroberts (approved):

PR #876, opened 2021-09-07
fix(stdlib): Annotate generic types in Buffer

phated:
I generated the Buffer docs and found out that these 2 functions have their parameters inferred as generic types, so we need to annotate them.


ospencer (approved):
Nice catch!

marcusroberts (approved):

PR #877, opened 2021-09-07
fix(grainfmt): Remove extra spacing from type signatures

### LINKED ISSUE #870: Formatter: Extra spaces added in some function signatures
Given this code:

```grain
export let fd_write: (WasmI32, WasmI32, WasmI32, WasmI32) -> WasmI32 =
  (
    fd,
    iovs,
    iovs_len,
    nwritten,
  ) => {
  0n
}
```

The formatter produces:

```grain
export let fd_write: ( WasmI32, WasmI32, WasmI32, WasmI32 ) -> WasmI32 =
  (
    fd,
    iovs,
    iovs_len,
    nwritten,
  ) => {
  0n
}
```

You can see the diff at https://github.com/grain-lang/near-simple-counter/pull/2/commits/ca86fd733f4f7ae775c39dca131a0db5cbfc09fe

marcusroberts:
Fixes #870 

Don't add spaces before/after types in function signature

### DIFF
--- a/compiler/test/formatter_outputs/function_params.gr
+++ b/compiler/test/formatter_outputs/function_params.gr
@@ -3,3 +3,13 @@ let single_arg = x => x
 let unit_arg = () => 3
 
 let two_args = (x, y) => 4
+
+export let fd_write: (WasmI32, WasmI32, WasmI32, WasmI32) -> WasmI32 =
+  (
+    fd,
+    iovs,
+    iovs_len,
+    nwritten,
+  ) => {
+  0n
+}
--- a/compiler/grainformat/reformat.re
+++ b/compiler/grainformat/reformat.re
@@ -923,14 +923,14 @@ and print_type =
             Doc.lparen,
             Doc.indent(
               Doc.concat([
-                Doc.line,
+                Doc.softLine,
                 Doc.join(
                   Doc.concat([Doc.comma, Doc.line]),
                   List.map(t => print_type(t, original_source), types),
                 ),
               ]),
             ),
-            Doc.line,
+            Doc.softLine,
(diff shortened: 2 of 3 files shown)


phated (approved):
Approved! But let's add a test

marcusroberts:
Oops, had a local test, added it to the suite

compiler/test/formatter_outputs/function_params.gr
```diff
@@ -3,3 +3,13 @@ let single_arg = x => x
...
+
+export let fd_write: (WasmI32, WasmI32, WasmI32, WasmI32) -> WasmI32 =
+  (
+    fd,
+    iovs,
+    iovs_len,
+    nwritten,
+  ) => {
+  0n
+}
```
ospencer, line 15: Formatter doesn't put a newline at eof?

ospencer (approved):

compiler/test/formatter_outputs/function_params.gr
phated: I remember when reading the tests that all line endings are removed because rely doesn't like them. So this isn't (can't be?) tested.

However, I believe this is already a setting in things like vscode, so it would be weird if we'd do it. It could cause constant conflicts if our decision was different than the user's editor settings.

marcusroberts: @phated is correct there is a newline at the end, but there's a bug in a Reason/Jest interaction, I can't where I raised the issue, that needs fixing.

But hexdump of the formatter output:

0000080 7b 0a 20 20 30 6e 0a 7d 0a

so it ends in a newline

PR #881, opened 2021-09-07
chore(stdlib): Reorganize buffer library for better docs

phated:
I wasn't happy with the way the docs for Buffer are organized on the website, so I reorganized the file and added a new section.

### DIFF
--- a/stdlib/buffer.gr
+++ b/stdlib/buffer.gr
@@ -504,7 +504,7 @@ export let addBufferSlice = (start, length, srcBuffer, dstBuffer) => {
 }
 
 /**
- * @section Raw bytes: Functions for working with the raw bytes in a buffer.
+ * @section Binary operations on integers: Functions for encoding and decoding integers stored in a buffer.
  */
 
 
--- a/stdlib/buffer.gr
+++ b/stdlib/buffer.gr
@@ -96,295 +96,30 @@ let checkIsIndexInBounds = (i, len, buf) => {
   if (i >= buf.len || i + len > buf.len) throw Exception.IndexOutOfBounds
 }
 
-/**
- * @section Values: Functions for working with the Buffer data type.
- */
-
-/**
- * Gets a signed 8-bit integer starting at the given byte index.
- *
- * @param index: The byte index to access
- * @param buffer: The buffer to access
- * @returns A 32-bit integer representing a signed 8-bit integer that starts at the given index
- *
- * @since v0.4.0
- */
-export let getInt8S = (index, buffer) => {
-  checkIsIndexInBounds(index, _8BIT_LEN, buffer)
-  Bytes.getInt8S(index, buffer.data)
-}
-
-/**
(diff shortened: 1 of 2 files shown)


stdlib/buffer.gr
```diff
@@ -96,674 +96,691 @@ let checkIsIndexInBounds = (i, len, buf) => {
...
+  buffer.len = buffer.len + _16BIT_LEN
+  Bytes.setInt16(index, value, buffer.data)
+}
+
+let addInt32help = (value, buffer) => {
+  autogrow(_32BIT_LEN, buffer)
+  let index = buffer.len
+  buffer.len = buffer.len + _32BIT_LEN
+  Bytes.setInt32(index, value, buffer.data)
+}
```
phated, line 118: I added these help functions which are needed by `addChar` - in addition, they no longer call the `set*` functions that also do a index assertion. It's probably a tiny performance improvement, but I don't think they need to do the assertion since they grow the buffer - right @jozanza?

stdlib/buffer.gr
```diff
@@ -96,674 +96,691 @@ let checkIsIndexInBounds = (i, len, buf) => {
...
-export let setInt64 = (index, value, buffer) => {
-  checkIsIndexInBounds(index, _64BIT_LEN, buffer)
-  Bytes.setInt64(index, value, buffer.data)
+export let addBufferSlice = (start, length, srcBuffer, dstBuffer) => {
+  addBytesSlice(start, length, srcBuffer.data, dstBuffer)
 }
 
 /**
- * Appends a signed 64-bit integer to a buffer.
+ * @section Raw bytes: Functions for working with the raw bytes in a buffer.
```
phated, line 507: @jozanza @ospencer this `@section` probably needs a better title and description. Thoughts?

stdlib/buffer.gr
```diff
@@ -96,674 +96,691 @@ let checkIsIndexInBounds = (i, len, buf) => {
...
  *
  * @since v0.4.0
  */
-export let addFloat64 = (value, buffer) => {
-  autogrow(_64BIT_LEN, buffer)
-  let index = buffer.len
-  buffer.len = buffer.len + _64BIT_LEN
-  setFloat64(index, value, buffer)
+export let addInt8 = (value, buffer) => {
+  addInt8help(value, buffer)
```
phated, line 562: These were updated to use the `help` functions that I refactored to the top of the file.

marcusroberts (approved):
LGTM, the Github comparison is way off so I went off the file itself

stdlib/buffer.gr
ospencer: OCaml calls this section `Binary encoding/decoding of integers`. I think that's pretty fair!

phated:
```suggestion
 * @section Binary operations on integers: Functions for encoding and decoding integers stored in a buffer.
```
@ospencer how about this for title and description?

ospencer (approved):

PR #888, opened 2021-09-12
fix(grainfmt): Resolve issues with comments within braces

### LINKED ISSUE #884: Formattter introduces syntax error
I tried the new formatter and I like it, but I found that it fails for match statements with a comment on the same line like this:
```grain
match (1) { // fails
    1 => "1",
	_ => "*"
  }
```

The resulting code:
```grain
match (1 // fails) {
  1 => "1",
  _ => "*",
}
```

Cheers.

marcusroberts:
Fixes #884 and other similar code cases where comments follow an opening brace.

Improved record and record pattern printing

### DIFF
--- a/compiler/grainformat/reformat.re
+++ b/compiler/grainformat/reformat.re
@@ -82,7 +82,23 @@ let make_line_start = (loc: Grain_parsing.Location.t) => {
   let merged_loc: Grain_parsing.Location.t = {
     loc_start: startpos,
     loc_end: loc.loc_start,
-    loc_ghost: loc.loc_ghost,
+    loc_ghost: true,
+  };
+  merged_loc;
+};
+
+let make_to_line_end = (loc: Grain_parsing.Location.t) => {
+  let endpos: Stdlib__lexing.position = {
+    pos_fname: "",
+    pos_lnum: loc.loc_end.pos_lnum,
+    pos_cnum: max_int,
+    pos_bol: 0,
+  };
+
+  let merged_loc: Grain_parsing.Location.t = {
+    loc_start: loc.loc_end,
+    loc_end: endpos,
+    loc_ghost: true,
   };
   merged_loc;
 };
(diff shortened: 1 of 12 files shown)


compiler/grainformat/reformat.re
```diff
@@ -1949,6 +2048,12 @@ and print_expression =
                     get_original_code(e.pexp_loc, original_source);
                   Walktree.remove_comments_in_ignore_block(e.pexp_loc);
 
+                  // get any trailing comments as the find doesn't look past the end of the
+                  // statement itself
+                  // we throw these away
+                  let _after_brace_comments =
```
ospencer, line 2054: Why is it thrown away?

compiler/test/formatter_inputs/function_params.gr
```diff
@@ -4,10 +4,13 @@ let unit_arg = () => 3
 
 let two_args = (x,y) => 4
 
+@disableGC
```
ospencer, line 7: This is maybe somewhat meta, but could we instead swap the signature of this function to use non-unsafe values? People often look at tests for examples, and I don't want them getting ideas 😛

compiler/grainformat/reformat.re
marcusroberts: We get the list of comments from the parser, but in the ignore case, we just use the original source code instead, which can contain some of these comments.   So we find any comments in the non-formatted range and remove them from the comment list.  This has changed because of the way I now find comments (after braces for example) where the comment finder is more restrictive, so we look for any final ones left.

(If we didn't remove the comment from the list, it would be picked up as a preceding comment by the next node)

compiler/test/formatter_inputs/function_params.gr
marcusroberts: Good point !

marcusroberts: Fixed.

compiler/grainformat/reformat.re
ospencer: Gotcha!

ospencer (approved):

phated (approved):
Very cool!

Issue #681, opened 2021-05-30
Resolve type checker warnings in `regex.test.gr`


peblair:
When running `regex.test.gr`, we currently receive a plethora of warnings like this:
```
File "compiler/test/stdlib/regex.test.gr", line 15, characters 34-48:
Warning 4: matchNumGroups was selected from type %Regex.MatchResult.
It is not visible in the current scope, and will not 
be selected if the type becomes unknown.
File "compiler/test/stdlib/regex.test.gr", line 15, characters 34-48:
Warning 2: this type-based field disambiguation is not principal.
```
These warnings are coming from the type checker and are likely due to us violating some assumptions which were made in the original OCaml type checker. We should determine how to resolve these, since the code in `regex.test.gr` is written using idioms we would like to have in Grain.

PR #898, opened 2021-09-14
chore(compiler): Disable by default principal type and type name scope compiler warnings

peblair:
This PR will allow us to land Regex without exposing users to those compiler warnings. This PR buys us more time to properly investigate #681.


ospencer (approved):
Looks good, but I'd probably adjust the title to not be about Regex since it affects all modules.

marcusroberts (approved):

ospencer (approved):

Issue #114, opened 2020-05-08
Remove --experimental-wasm-bigint flag when wasm-bigint is no longer experimental


ospencer:
Once this feature in Node.js is no longer experimental, we can remove this flag.

https://github.com/grain-lang/grain/blob/master/cli/bin/grain.js#L2

phated:
Just ran into this. We need to specify an engine because I was using node 10 LTS and it failed

ospencer:
Made #120 to address. I'll probably get around to it tomorrow.

ospencer:
Node 16 becomes LTS on October 26. I think that's when we should drop support for Node 14 🙂

phated:
Since that's a breaking change, I don't expect it to happen until our next minor release, which I'm also not expecting to happen in 1.5 months 😛 But, likely in Grain 0.5, yeah.

PR #899, opened 2021-09-14
fix(cli): Avoid --experimental-wasm-bigint flag on node 16

### LINKED ISSUE #883: Flag : experimental-wasm-bigint
Hello, I just built grain from source on Arch Linux using Node v16.8.0 and yarn v1.22.11. 
I have a warning when I run the `grain` command
```
Error: unrecognized flag --experimental-wasm-bigint
```
Everything seems to work fine, but this message is a bit annoying. I removed the flag [here](https://github.com/grain-lang/grain/blob/2a5ec3db84be3533ea707c4b5118f81b4f0c1be7/cli/bin/grain.js#L13), and it resolved the problem. 

So, not a big deal for me, but I post it here in case you want to remove that in future releases. Also asking in case that flag is essential, and I broke something by removing it.

phated:
This adds an conditional around the `--experimental-wasm-bigint` flag so we don't get a warning in node 16+. I also added node 16 to the test matrix in CI.

Closes #883 (but we should leave #114 open until we don't support node < 16 anymore).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,12 +10,13 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
+        node-version: ["14", "16"]
 
     steps:
       - name: Setup Node.js
         uses: actions/setup-node@v2.1.2
         with:
-          node-version: '14'
+          node-version: ${{ matrix.node-version }}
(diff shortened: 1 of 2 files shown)


ospencer (approved):
Nice!

marcusroberts (approved):
LGTM!

PR #902, opened 2021-09-18
fix(grainfmt): Remove extraneous parens around infix function application

### LINKED ISSUE #859: Formatter: adds redundant parenthesis around expressions
The formatter is cautious and adds parenthesis around expressions, particularly for infix operators, but this can lead to code like

`setCounter((loadCounter()) + 1l)`

instead of

`setCounter(loadCounter() + 1l)`

marcusroberts:
Fixes #859 

The formatter was cautiously adding parens around too many infix function application parameters.
This has been fixed.   The formatter still adds some redundant parens, e.g.

3 + (4 * 5) can be simplified to 3 + 4 * 5 

which I will fix in another PR where we take operator precedence into account


compiler/grainformat/reformat.re
```diff
@@ -1045,78 +1055,87 @@ and print_application =
...
-    Doc.concat([
-      first_brackets,
-      Doc.space,
-      Doc.text(function_name),
-      Doc.space,
-      second_brackets,
-    ]);
+
+    if (level > 0 && !has_standard_precedence(function_name)) {
+      // only wrap maths functions (where there's a precedence)
```
ospencer, line 1120: This logic isn't _quite_ right, since you're welcome to shadow any of the operators but they keep their precedence (`++` has the same precedence as `+`):
```
let (++) = (a, b) => a + a + b + b
```

marcusroberts: Ah, good point.

marcusroberts: I was going to do proper precedence comparison anyway to determine if the brackets were needed.   Any  thoughts on keeping track of that for shadowed operators?

ospencer: No need to keep track! The precedence remains the same regardless of whether you shadow or not.

marcusroberts: Converting this to Draft whilst I implement this fix

marcusroberts: I added a naieve implementation if you could take a look?

compiler/grainformat/reformat.re
```diff
@@ -1045,78 +1055,136 @@ and print_application =
...
+      | "&" => 70
+      | "^" => 60
+      | "|" => 50
+      | "&&" => 40
+      | "||" => 30
+      | "_" => 10
+      | _ => 0
+      };
+
+    let override_precedence = (fn2, fn1) => {
```
marcusroberts, line 1140: How does this look?  I've only coded in the infix operators

compiler/grainformat/reformat.re
```diff
@@ -1045,78 +1048,136 @@ and print_application =
...
+      let left_prec = op_precedence(fn1);
+      let right_prec = op_precedence(fn2);
+
+      if (left_prec < right_prec) {
+        true;
+      } else {
+        false;
+      };
```
ospencer, line 1148:
```suggestion
      left_prec < right_prec
```

marcusroberts: Oh dear god.   Yes, yours is much better.

marcusroberts: (That was left over from some debugging)

marcusroberts:
I have much better confidence in this version as I'm using operator precedence and making the comparison in the right place.  Quite a few tests too.

compiler/grainformat/reformat.re
```diff
@@ -1035,88 +1063,91 @@ and print_application =
...
+      | _ => false
+      };
+
+    let right_is_if =
+      switch (second.pexp_desc) {
+      | PExpIf(_) => true
+      | _ => false
+      };
+
+    let (leftparen, rightparen) =
```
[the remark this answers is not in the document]
marcusroberts: I like how you tried to trick me to use camel case there :-D.   Fixed.

ospencer (approved):
Love it!

phated (approved):
Excellent! I really love how we write formatter tests.

compiler/grainformat/reformat.re
```diff
@@ -997,10 +1025,10 @@ and print_type =
 }
 and print_application =
     (
-      ~func: Parsetree.expression,
       ~expressions: list(Parsetree.expression),
       ~parent_loc: Location.t,
       ~original_source: array(string),
+      func: Parsetree.expression,
```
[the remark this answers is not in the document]
marcusroberts: Great advice on not making them all labelled!

PR #904, opened 2021-09-18
feat(grainfmt): Add format in place or output to a new file

marcusroberts:
Added support for

--format-in-place to overwrite the contents of the file with the formatted output
-o <filename>. output formatted results to a different file

Because the formatter can run on stdin without a source filename specified it will ignore --format-in-place if no input filename is given.

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -75,11 +75,18 @@ class ForwardOption extends program.Option {
   }
 }
 
-program.forwardOption = function (flags, description, parser, defaultValue) {
+// Adds .forwardOption to commands. Similar to Commander's native .option,
+// but will forward the flag to the underlying program.
+program.Command.prototype.forwardOption = function (
+  flags,
+  description,
+  parser,
+  defaultValue
+) {
   const option = new ForwardOption(flags, description);
   if (parser) option.argParser(parser);
   if (typeof defaultValue !== "undefined") option.default(defaultValue);
-  return program.addOption(option);
+  return this.addOption(option);
 };
 
 program
@@ -194,7 +201,9 @@ program
   .command("lsp <file>")
   .description("check a grain file for LSP")
   .action(
-    wrapAction(function (file, options, program) {
+    wrapAction(function (file) {
+      // The lsp subcommand inherits all options of the
+      // top level grain command
       lsp(file, program);
     })
   );
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -143,6 +143,7 @@ program
     "--verbose",
     "print critical information at various stages of compilation"
   )
+  .graincOption('--format-in-place', 'reformat in place')
   // The root command that compiles & runs
   .arguments("<file>")
   .action(function (file) {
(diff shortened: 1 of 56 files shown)


cli/bin/grain.js
```diff
@@ -143,6 +143,7 @@ program
     "--verbose",
     "print critical information at various stages of compilation"
   )
+  .graincOption("--format-in-place", "reformat in place")
```
ospencer, line 146: I think you can move the option to the `format` command so it only applies when that subcommand is active.

ospencer (commented):
Any thoughts on `--write`? I've liked that name a lot when using Prettier!

cli/bin/grain.js
```diff
@@ -211,6 +211,7 @@ program
 program
   .command("format [file]")
   .description("format a grain file")
+  .addOption(new ForwardOption("--in-place", "reformat in place"))
```
ospencer, line 214:
```suggestion
  .addOption(new ForwardOption("--in-place", "format in place"))
```

ospencer (approved):

compiler/src/utils/config.re
```diff
@@ -624,3 +624,6 @@ let get_implicit_opens = () => {
     ]);
   };
 };
+
+let format_in_place =
```
peblair, line 628: @marcusroberts @ospencer Should we make an issue to split out the grainformat config items into a module that's not in the compiler? It feels odd to have it here

ospencer: Somewhat related—I hadn't realized that the tools inherit all of the compiler config. We should probably also add to that issue (or make an issue for that too). It seems like they only need the stdlib path and include paths to be able to enter the typechecking phase, but I think they should just take those flags and ignore the rest.

peblair: #917 created. Feel free to edit with any clarifications you think are necessary

cli/bin/grain.js
```diff
@@ -211,6 +211,7 @@ program
 program
   .command("format [file]")
   .description("format a grain file")
+  .addOption(new ForwardOption("--in-place", "format in place"))
```
peblair, line 214: Should be reconciled with the `.forwardOption` change in #916 (if that is merged before this PR).

compiler/src/utils/config.rei
```diff
@@ -193,3 +193,7 @@ type implicit_opens =
   | Gc_mod;
 
 let get_implicit_opens: unit => list(implicit_opens);
+
+/** Formatter rewrites file in place */
+
+let format_in_place: ref(bool);
```
peblair, line 199: @ospencer So do we want this moved into `grainformat.re`? Or is that left to #917?

ospencer: Yeah, let's get that moved before this is merged.

peblair: @marcusroberts If you could move this flag out of `config.re` and into `grainformat.re` then that'd be :100:

marcusroberts: That's done

peblair (approved):
Looks good!

ospencer (commented):
Looks great! I merged #916 so you can use `.forwardOption` now :)

PR #907, opened 2021-09-19
fix(grainfmt): handle empty type for foreign exports

marcusroberts:
Fixes 

export foreign wasm promise_results_count: () -> WasmI64 as promiseResultsCount from "env"

to not break on the empty param types


peblair (approved):
🙌

ospencer (approved):

PR #908, opened 2021-09-19
feat(cli): Refactor forwardable options

ospencer:
This PR refactors `GraincOption` into `ForwardOption`. The major change is that cli subcommands will receive the options for their subcommand rather than the top-level `grain` command (except for `compile` and `run`).

Now, for example, for a subcommand you can add `.addOption(new ForwardOption("--write", "my description"))` and the `--write` flag will be forwarded to the underlying command.

The `--graceful` flag was really only meant to be used with `grain run`, but it's sort of made its way to being a catch-all. To keep backward compatibility I added a `wrapAction` function which will preserve the graceful behavior for all of the subcommands, though we should consider removing that behavior in a breaking version.

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -35,11 +35,27 @@ function graincVersion() {
   return exec.grainc("--version", program).toString().trim();
 }
 
-class GraincOption extends program.Option {
-  grainc = true;
+function wrapAction(action, logError = false) {
+  return (...args) => {
+    try {
+      return action(...args);
+    } catch (e) {
+      if (logError) console.error(e);
+      if (program.opts().graceful) {
+        process.exit();
+      } else {
+        process.exit(1);
+      }
+    }
+  };
+}
 
-  toFlag() {
-    const value = program.opts()[this.attributeName()];
+class ForwardOption extends program.Option {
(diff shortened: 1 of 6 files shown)


marcusroberts (approved):
Thanks for doing this, it makes the writing flags much easier now!

cli/bin/grain.js
```diff
@@ -35,11 +35,27 @@ function graincVersion() {
...
 
-class GraincOption extends program.Option {
-  grainc = true;
+function wrapAction(action, logError = false) {
+  return (...args) => {
+    try {
+      return action(...args);
+    } catch (e) {
+      if (logError) console.error(e);
+      if (program.opts().graceful) {
```
phated, line 44: Seems like program is undefined here.

phated (changes requested):
On a cursory glance, just 1 thing that looks broken.

cli/bin/grain.js
ospencer: It's defined above, on line 16.

phated: 🤦 I was having flashbacks to when we introduced this bug in v0.3

phated (approved):
LGTM

PR #909, opened 2021-09-19
fix(compiler): Do not inline aliased mutables

peblair:


ospencer (approved):
Fantastic, thanks!

marcusroberts (approved):
Good stuff!

PR #911, opened 2021-09-19
fix(grainfmt): Fix the parens around single arg function params

marcusroberts:
Make sure we keep the parens for single param functions like this:

export let batchActionCreateAccount = ({index}) => {
  Native.promiseBatchActionCreateAccount(Conv.fromInt64(index))
}


ospencer (approved):
🎉

phated (approved):
Looking good!

PR #913, opened 2021-09-20
fix(compiler): Don't warn on exhaustive boolean matches

ospencer:
Most of the code here is really new testing support around warnings. Added custom matchers so we can do `expect.ext.warning.toHaveTriggered` and `expect.ext.warning.toHaveTriggeredNoWarnings`, and added some tests around the warning generation for boolean matches.

### DIFF
--- a/compiler/src/typed/parmatch.re
+++ b/compiler/src/typed/parmatch.re
@@ -699,6 +699,8 @@ let full_match = (closing, env) =>
       /* extensions */
       List.length(env) == c.cstr_consts + c.cstr_nonconsts;
     }
+  | [({pat_desc: TPatConstant(Const_bool(_))}, _), ..._] =>
+    List.length(env) == 2
   | [({pat_desc: TPatArray(_)}, _), ..._]
   | [({pat_desc: TPatConstant(_)}, _), ..._] => false
   | [({pat_desc: TPatTuple(_)}, _), ..._]
@@ -967,11 +969,18 @@ let build_other = (ext, env) =>
       p,
       env,
     )
-  /*| ({pat_desc=(TPatConstant (Const_bool _))} as p,_) :: _ ->
-    build_other_constant
-      (function TPatConstant(Const_bool i) -> i | _ -> assert false)
-      (function i -> TPatConstant(Const_bool i))
-      ??? p env*/
+  | [({pat_desc: TPatConstant(Const_bool(_))} as p, _), ..._] =>
+    build_other_constant(
+      fun
+      | TPatConstant(Const_bool(b)) => b
+      | _ => assert(false),
+      fun
+      | b => TPatConstant(Const_bool(b)),
+      false,
+      (!),
+      p,
+      env,
+    )
   | [({pat_desc: TPatConstant(Const_string(_))} as p, _), ..._] =>
     build_other_constant(
       fun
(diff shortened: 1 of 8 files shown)


compiler/test/suites/pattern_matching.re
```diff
@@ -197,4 +203,49 @@ describe("pattern matching", ({test}) => {
...
+       true => print(5),
+     }",
+    Warnings.PartialMatch("false"),
+  );
+  assertNoWarning(
+    "bool_exhaustiveness4",
+    "match (true) {
+       false => print(5),
+       true => print(5),
+     }",
```
phated, line 233: Is there anything weird with `when` clause on a boolean constant? Can you add a test?

compiler/src/typed/parmatch.re
```diff
@@ -699,6 +699,8 @@ let full_match = (closing, env) =>
       /* extensions */
       List.length(env) == c.cstr_consts + c.cstr_nonconsts;
     }
+  | [({pat_desc: TPatConstant(Const_bool(_))}, _), ..._] =>
+    List.length(env) == 2
```
phated, line 703: Why is it okay that we just check the env has a length of 2?

compiler/src/typed/parmatch.re
```diff
@@ -967,11 +969,18 @@ let build_other = (ext, env) =>
       p,
       env,
     )
-  /*| ({pat_desc=(TPatConstant (Const_bool _))} as p,_) :: _ ->
-    build_other_constant
-      (function TPatConstant(Const_bool i) -> i | _ -> assert false)
-      (function i -> TPatConstant(Const_bool i))
-      ??? p env*/
+  | [({pat_desc: TPatConstant(Const_bool(_))} as p, _), ..._] =>
```
phated, line 972: the ol' uncomment and fix an issue! 🔥

phated (changes requested):
Couple questions and a request for another test.

compiler/src/typed/parmatch.re
ospencer: The description of this function explains the first part better than I can, so let's start there:
> Check whether the first column of env makes up a complete signature or not. We work on the discriminating patterns of each sub-matrix: they are simplified, and are not omega/Tpat_any. 

Essentially `env` contains all of the unique non-"any " patterns for a section of the total match. Since we know that the type is boolean, there aren't any "any" patterns, and they're all unique, if the env contains two patterns, they must be the patterns `true` and `false`.

A similar technique is used above for variants—as long as the number of patterns in the env is equal to the total number of constructors, since they're simplified and unique, that means each constructor appears exactly once.

Hope that makes sense!

compiler/src/typed/parmatch.re
ospencer: Haha, this one was clear that we didn't really know how to implement it 😛 

But we do now!

compiler/test/suites/pattern_matching.re
ospencer: Not really! Guarded clauses aren't considered when determining exhaustiveness, since the guard only makes the pattern match sometimes. You'll still get a "this isn't exhaustive" warning, but you get an extra message telling you that it's because that branch is guarded. Test incoming 🙂

compiler/src/typed/parmatch.re
phated: 💯

phated (approved):
⚔️

marcusroberts (approved):
Nice improvements!

PR #916, opened 2021-09-22
fix(cli): Forward stdlib option to lsp and graindoc

ospencer:
Just need to specify that this needs to be forwarded.

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -201,7 +201,9 @@ program
   .command("lsp <file>")
   .description("check a grain file for LSP")
   .action(
-    wrapAction(function (file, options, program) {
+    wrapAction(function (file) {
+      // The lsp subcommand inherits all options of the
+      // top level grain command
       lsp(file, program);
     })
   );
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -193,6 +193,12 @@ program
 program
   .command("lsp <file>")
   .description("check a grain file for LSP")
+  .addOption(
+    new ForwardOption(
+      "-S, --stdlib <path>",
+      "override the standard libary with your own"
+    ).default(stdlibPath)
+  )
   .action(
     wrapAction(function (file, options, program) {
       lsp(file, program);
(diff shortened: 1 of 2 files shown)


marcusroberts (approved):

marcusroberts (approved):
I like this even better!

cli/bin/grain.js
```diff
@@ -75,11 +75,18 @@ class ForwardOption extends program.Option {
   }
 }
 
-program.forwardOption = function (flags, description, parser, defaultValue) {
+// Adds .forwardOption to commands. Similar to Commander's native .option,
+// but will forward the flag to the underlying program.
+program.Command.prototype.forwardOption = function (
```
peblair, line 80: Out of curiosity, why was this change needed?

peblair (approved):
One question, but looks good.

cli/bin/grain.js
```diff
@@ -193,6 +200,12 @@ program
 program
   .command("lsp <file>")
   .description("check a grain file for LSP")
+  .forwardOption(
+    "-S, --stdlib <path>",
+    "override the standard libary with your own",
+    null,
+    stdlibPath
+  )
```
phated, line 208: If these are specified separately from the main program, then we need to add others. such as `--output`. I'm not sure all of them that'll need to be duplicated

phated (changes requested):
One (big?) thing

cli/bin/grain.js
ospencer: It's not needed to fix the issue, but it's needed to be able to `.forwardOption` on the subcommand. We just decided it was cleaner to be consistent around `.forwardOption` versus `.addOption(new ForwardOption())`!

cli/bin/grain.js
ospencer: Yup yup. Having a realization now. I'll make these inherit all of the compiler config for now until we get it truly separated.

ospencer:
@phated should be all set.

cli/bin/grain.js
```diff
@@ -194,7 +201,9 @@ program
   .command("lsp <file>")
   .description("check a grain file for LSP")
   .action(
-    wrapAction(function (file, options, program) {
+    wrapAction(function (file) {
```
phated, line 204: I can't comment on the lines above the diff, but does this mean that the `run` command needs to be updated too?

phated: Nevermind, it looks like that isn't using `wrapAction`

phated (approved):
I quickly tossed some commands I use at this and they seem to work 👍

Issue #918, opened 2021-09-23
Compiler: Function exports in unlinked mode not always written


ospencer:
When compiling `export let add = (a, b) => a + b`, the compiler should generate both a wasm function export for the `add` function and a global export for `add`'s closure.

It appears that sometimes only the closure is exported.

PR #919, opened 2021-09-23
fix(compiler): Remove existing exports when writing universal exports

ospencer:
This never was an issue with the linker since it generates a fresh module (which has no defined exports). It's an issue for the polyfill since it writes exports on an already-exiting module.

Somewhat related: #918

### DIFF
--- a/compiler/src/codegen/comp_utils.re
+++ b/compiler/src/codegen/comp_utils.re
@@ -191,6 +191,8 @@ let write_universal_exports =
               [||],
               function_body,
             );
+            // Remove existing Grain export (if any)
+            Export.remove_export(wasm_mod, name);
             ignore @@ Export.add_function_export(wasm_mod, name, name);
           | TSigValue(_)
           | TSigType(_)
--- a/compiler/test/input/wasiPolyfill.gr
+++ b/compiler/test/input/wasiPolyfill.gr
@@ -1,7 +1,22 @@
-import foreign wasm fd_write : (WasmI32, WasmI32, WasmI32, WasmI32) -> WasmI32 from "wasi_snapshot_preview1"
+import foreign wasm fd_write: (
+  WasmI32,
+  WasmI32,
+  WasmI32,
+  WasmI32
+) -> WasmI32 from "wasi_snapshot_preview1"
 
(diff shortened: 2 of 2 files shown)


compiler/src/codegen/comp_utils.re
```diff
@@ -191,6 +191,8 @@ let write_universal_exports =
               [||],
               function_body,
             );
+            // Remove existing Grain export (if any)
+            Export.remove_export(wasm_mod, name);
```
phated, line 195: How can we add a test for this? Might be a good use case for a snapshot 😉

ospencer: I modified our existing test to test this, though the issue in #918 was what kept this test from failing before. What were you thinking of for a snapshot?

phated: Ah, so that test would fail on `main` but then this would make it green again?

ospencer: Indeed!

phated (approved):
LGTM!

compiler/src/codegen/comp_utils.re
ospencer: Though actually I just double-checked to make sure that this fails on main, and it's not. #918 is a real bugger.

peblair: @ospencer There's no way to check if the export is in there before we add it? Or that wouldn't fix it anyway?

ospencer: It's a different export. Essentially we're overwriting the Grain one (which takes the closure as an argument and always returns an i32) with a universal one (doesn't take a closure and doesn't necessarily return an i32).

ospencer: @phated just updated the test to one that's actually failing. If you look at the diff you'll see that I've figured out what's wrong with #918 :)

phated: So having a type signature avoids 918, but removing it causes it to not be written?

ospencer: The opposite—a function with a type signature doesn't get written as a wasm function export, which avoids the bug that causes what this PR fixes. Removing the type signature in the test _causes_ the export to get written, which triggers the bug to make this test fail on `main` (but not on this branch because this PR fixes it).

peblair (approved):

PR #920, opened 2021-09-23
fix(compiler): Export functions with aliased names as wasm functions

### LINKED ISSUE #918: Compiler: Function exports in unlinked mode not always written

ospencer:
Fixes #918


compiler/test/suites/exports.re
```diff
@@ -41,4 +41,12 @@ describe("exports", ({test}) => {
   );
 
   assertSnapshot("let_rec_export", "export let rec foo = () => 5");
+  assertSnapshot(
```
peblair, line 44: I'm worried about the long-term maintenance of these tests, since we do have mass snapshot changes periodically...I feel like it'd be easy to accidentally cause these to fail without us really knowing. Maybe we should do the `assertSnapshot` and also do some sort of string search on the snapshot contents? 🤔

ospencer: That seems like it might be difficult to write, but I can write it without using a snapshot.

ospencer: @peblair done.

marcusroberts (approved):
LGTM!

peblair (approved):
🙌

PR #921, opened 2021-09-23
feat(stdlib): Add library for pseudo-random number generation

peblair:
This PR adds support for pseudo-random number generation in Grain. I'm marking it as a draft so that we can discuss what other APIs would be good to have in here.

### DIFF
--- a/compiler/test/stdlib/random.test.gr
+++ b/compiler/test/stdlib/random.test.gr
@@ -1,0 +1,25 @@
+import Random from "random"
+
+let rng1 = Random.make(0xf00L)
+let rng2 = Random.make(0xf00L)
+
+assert Random.nextInt32(rng1) == 833494444l
+assert Random.nextInt32(rng2) == 833494444l
+assert Random.nextInt32(rng1) == -1746193004l
+assert Random.nextInt32(rng2) == -1746193004l
+
+assert Random.nextInt32(rng1) == Random.nextInt32(rng2)
+assert Random.nextInt64(rng1) == Random.nextInt64(rng2)
+assert Random.nextInt64(rng1) == Random.nextInt64(rng2)
+
+assert Random.nextInt64(rng1) == -2063148927841070477L
+assert Random.nextInt64(rng2) == -2063148927841070477L
+assert Random.nextInt64(rng1) != -2063148927841070477L
+
+assert Random.nextInt32InRange(rng1, 5l, 20l) == 5l
+assert Random.nextInt32InRange(rng1, 5l, 20l) == 18l
+assert Random.nextInt32InRange(rng1, 5l, 20l) == 19l
+
+assert Random.nextInt64InRange(rng1, 5L, 20L) == 18L
+assert Random.nextInt64InRange(rng1, 5L, 20L) == 10L
+assert Random.nextInt64InRange(rng1, 5L, 20L) == 11L
--- a/stdlib/random.gr
+++ b/stdlib/random.gr
@@ -20,8 +20,8 @@ let incCounter = random => {
 // https://arxiv.org/pdf/2004.06278v3.pdf
 @unsafe
 let squares = (ctr: Int64, key: Int64) => {
-  // Need to implement with @unsafe because
-  // overflows are part of this algorithm
+  // Implemented with @unsafe to boost efficiency
+  // and have fine-grained control over overflow semantics
   let ctr = WasmI64.load(WasmI32.fromGrain(ctr), 8n)
   let key = WasmI64.load(WasmI32.fromGrain(key), 8n)
   let mut x = WasmI64.mul(ctr, key)
@@ -55,7 +55,7 @@ export let make = seed => {
 /**
  * Creates a new pseudo-random number generator with a random seed.
  *
- * @returns The pseudo-random number generator
+ * @returns `Ok(generator)` of a pseudo-random number generator if successful or `Err(exception)` otherwise
  */
 export let makeUnseeded = () => {
   // TODO: Should we just .expect this result for UX's sake?
@@ -114,18 +114,6 @@ export let nextInt64 = (random: Random) => {
   )
   incCounter(random)
   Int64.lor(Int64.shl(ret1, 32L), ret2)
-}
-
-// (can be removed after #831 lands)
-@disableGC
-let rec gtU = (x: Int32, y: Int32) => {
-  let xv = WasmI32.load(WasmI32.fromGrain(x), 8n)
-  let yv = WasmI32.load(WasmI32.fromGrain(y), 8n)
-  let ret = WasmI32.gtU(xv, yv)
-  Memory.decRef(WasmI32.fromGrain(x))
-  Memory.decRef(WasmI32.fromGrain(y))
-  Memory.decRef(WasmI32.fromGrain(gtU))
-  ret
 }
 
 /**
--- a/stdlib/random.gr
+++ b/stdlib/random.gr
@@ -1,3 +1,8 @@
+/**
+ * @module Random: A pseudo-random number generator implementation.
+ * @example import Random from "random"
+ * @since v0.5.0
+ */
 import WasiRandom from "sys/random"
 import Result from "result"
 import Int32 from "int32"
@@ -7,11 +12,19 @@ import WasmI64 from "runtime/unsafe/wasmi64"
 import Memory from "runtime/unsafe/memory"
 import DS from "runtime/dataStructures"
 
+/**
+ * @section Types: Type declarations included in the Random module.
+ */
+
 record Random {
   seed: Int64,
   mut counter: Int64,
   mut initialized: Bool,
 }
+
+/**
+ * @section Values: Functions for working with pseudo-random number generators.
+ */
 
 let incCounter = random => {
   random.counter = Int64.incr(random.counter)
@@ -47,6 +60,8 @@ let squares = (ctr: Int64, key: Int64) => {
  *
  * @param seed: The seed for the pseudo-random number generator
  * @returns The pseudo-random number generator
+ *
+ * @since v0.5.0
  */
 export let make = seed => {
   { seed, counter: 0L, initialized: false }
@@ -56,6 +71,8 @@ export let make = seed => {
  * Creates a new pseudo-random number generator with a random seed.
  *
  * @returns `Ok(generator)` of a pseudo-random number generator if successful or `Err(exception)` otherwise
+ *
+ * @since v0.5.0
  */
 export let makeUnseeded = () => {
   // TODO: Should we just .expect this result for UX's sake?
@@ -89,6 +106,8 @@ let checkInitialized = (random: Random) => {
  *
  * @param random: The pseudo-random number generator to use
  * @returns The randomly generated number
+ *
+ * @since v0.5.0
  */
 export let nextInt32 = (random: Random) => {
   checkInitialized(random)
@@ -102,6 +121,8 @@ export let nextInt32 = (random: Random) => {
  *
  * @param random: The pseudo-random number generator to use
  * @returns The randomly generated number
+ *
+ * @since v0.5.0
  */
 export let nextInt64 = (random: Random) => {
   checkInitialized(random)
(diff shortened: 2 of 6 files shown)


stdlib/random.gr
```diff
@@ -0,0 +1,128 @@
...
+
+let incCounter = random => {
+  random.counter = Int64.incr(random.counter)
+}
+
+// https://arxiv.org/pdf/2004.06278v3.pdf
+@disableGC
+let rec squares = (ctr: Int64, key: Int64) => {
+  // Need to implement with @disableGC because
+  // overflows are part of this algorithm
```
ospencer, line 20: All of the Int64 ops support overflows, though.

ospencer (commented):
I've been waiting for this one!! This is awesome!

ospencer:
Dare I say I prefer the global state versions where if I need reproducibility I just `Random.setSeed(seed)` but otherwise I don't need to pass a random instance??

ospencer:
(Sorta like Python's `random` module)

ospencer:
Oh wow, even OCaml uses global state! https://ocaml.org/api/Random.html

peblair:
Yeah, I feel that. There certainly are a lot of use cases for isolated random number generators like the API here, but maybe we should make the main (maybe only?) API just based on a global seed. I'd love to have both, but a concise API becomes difficult to design with both.

stdlib/random.gr
peblair: I'm not sure why (possibly due to the equivalent of `WasmI32.wrapI64`?), but I was getting overflow errors when using the Grain functions.

Either way, there is an efficiency argument to be made as well, since our boxed numeric operations are not as efficient as working with the raw WASM types.

cician:
I'd like to implement pseudo-random shuffling of arrays, for which I'd need a function to generate integers within a specific range.

If I understand correctly it can be tricky to implement in regards to uniform distribution guarantees. Quite a bit beyond my computer science knowledge.

My use case for array shuffling is running micro benchmarks. I want to compare multiple implementations of a function. I want to avoid bias of a specific execution order, while avoiding true randomness.

It would be nice to avoid per element allocations if possible, though not important for this specific use case. I think a good compromise can be achieved by simply returning a Number, which would allow to avoid allocations at least for the Simple Number range. Otherwise the only way would be to expose functions that return unsafe WASM numbers.

ps: I don't like global state. It's OK as an additional option for ergonomics, but important aspects of a PRNG are determinism and reproducibility. Not representing it explicitly in APIs would lead to sloppy usage because we programmers are too lazy to think about these details until we're forced to. Just like Option vs null. In the same vein I don't like the Hash having a shared and random seed. Just my opinionated opinion.

peblair:
I've marked this PR as ready for review for 0.5, and I've added `nextInt32InRange`/`nextInt64InRange` functions, per @cician's request. We can workshop their API a bit; I'm not to picky about how their named/if they take one or two arguments/etc. Note that this is now based off of the #831 branch, since the two new functions need the unsigned greater than implementations.

stdlib/random.gr
```diff
@@ -0,0 +1,176 @@
...
+  let xv = WasmI32.load(WasmI32.fromGrain(x), 8n)
+  let yv = WasmI32.load(WasmI32.fromGrain(y), 8n)
+  let ret = WasmI32.gtU(xv, yv)
+  Memory.decRef(WasmI32.fromGrain(x))
+  Memory.decRef(WasmI32.fromGrain(y))
+  Memory.decRef(WasmI32.fromGrain(gtU))
+  ret
+}
+
+export let nextInt32InRange = (random: Random, low: Int32, high: Int32) => {
```
ospencer, line 142: Probably weird to hear me say this, but it could be worth it to implement this via low-level instructions.

ospencer (commented):
Overall looks good. I think I'd like `next()` and `nextInRange` that work on integers of Grain's `Number` type as well. One other thing to consider is that we do have a `Range` type from the range stdlib which may or may not make sense here.

stdlib/random.gr
```diff
@@ -0,0 +1,176 @@
...
+  x = WasmI64.add(WasmI64.mul(x, x), z)
+  x = WasmI64.or(WasmI64.shrU(x, 32N), WasmI64.shl(x, 32N))
+  // round 3
+  x = WasmI64.add(WasmI64.mul(x, x), y)
+  x = WasmI64.or(WasmI64.shrU(x, 32N), WasmI64.shl(x, 32N))
+  let ret = WasmI32.wrapI64(
+    WasmI64.shrU(WasmI64.add(WasmI64.mul(x, x), z), 32N)
+  )
+  let ret = WasmI32.toGrain(DS.newInt32(ret)): Int32
+  Memory.decRef(WasmI32.fromGrain(ctrArg))
```
phated, line 45: Why decref'ing `ctrArg` and not `ctr`?

stdlib/random.gr
```diff
@@ -0,0 +1,176 @@
...
+  Memory.decRef(WasmI32.fromGrain(keyArg))
+  Memory.decRef(WasmI32.fromGrain(squares))
+  ret
+}
+
+// [NOTE] The PRNG algorithm seems to not work as well with small seeds, but
+// users are more likely to use small numbers for their seeds. Consequently,
+// we produce an actual seed xoring feeding the given seed with a
```
phated, line 53: Typo here? "Xoring feeding"

stdlib/random.gr
```diff
@@ -0,0 +1,176 @@
...
+// constant (large) number.
+
+/**
+ * Creates a new random number generator with the given seed.
```
phated, line 57: Do we want to be clear this is pseudo-random in all of the docs?

stdlib/random.gr
```diff
@@ -0,0 +1,176 @@
...
+}
+
+/**
+ * Creates a new random number generator with a random seed.
+ *
+ * @param seed: The seed for the random number generator
+ * @returns The random number generator
+ */
+export let makeUnseeded = () => {
+  // TODO: Should we just .expect this result for UX's sake?
```
phated, line 73: This is hard to say. Personally, I think it's "easier" for us to force the `.expect` calls onto the user. Then, they can also audit their code for crashes more easily.

stdlib/random.gr
```diff
@@ -0,0 +1,176 @@
...
+ * @returns The random number generator
+ */
+export let make = seed => {
+  { seed, counter: 0L, initialized: false }
+}
+
+/**
+ * Creates a new random number generator with a random seed.
+ *
+ * @param seed: The seed for the random number generator
```
phated, line 69:
```suggestion
```
No arguments to this

stdlib/sys/random.gr
```diff
@@ -26,6 +27,48 @@ let wasmSafeErr = err => {
   Err(err)
 }
 
+/**
+ * Produce a random 32-bit integer. This function can be slow, so it's best to seed a generator if lots of random data is needed.
+ *
+ * @returns `Ok(num)` of a random number if successful or `Err(exception)` otherwise
```
phated, line 33:
```suggestion
 * @returns `Ok(num)` of a random Int32 if successful or `Err(exception)` otherwise
```
I could go either way on this.

stdlib/sys/random.gr
```diff
@@ -26,6 +27,48 @@ let wasmSafeErr = err => {
...
+    wasmSafeOk(WasmI32.toGrain(newInt32(rand)): Int32)
+  }
+  Memory.decRef(WasmI32.fromGrain(randomInt32))
+  ret
+}
+
+/**
+ * Produce a random 64-bit integer. This function can be slow, so it's best to seed a generator if lots of random data is needed.
+ *
+ * @returns `Ok(num)` of a random number if successful or `Err(exception)` otherwise
```
phated, line 54:
```suggestion
 * @returns `Ok(num)` of a random Int64 if successful or `Err(exception)` otherwise
```
I could go either way on this.

stdlib/random.gr
```diff
@@ -0,0 +1,176 @@
...
+  incCounter(random)
+  let ret2 = Int64.fromNumber(
+    Int32.toNumber(squares(random.counter, random.seed))
+  )
+  incCounter(random)
+  Int64.lor(Int64.shl(ret1, 32L), ret2)
+}
+
+@disableGC
+export let rec gtU = (x: Int32, y: Int32) => {
```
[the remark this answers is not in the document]
phated: Let's delete this. It's showing up in the docs too.

compiler/test/stdlib/random.test.gr
```diff
@@ -0,0 +1,25 @@
+import Random from "random"
+
+let rng1 = Random.make(0xf00L)
+let rng2 = Random.make(0xf00L)
+
+assert Random.nextInt32(rng1) == 833494444l
```
phated, line 6: Do we like this API instead of `Int32.random()` and `Int32.nextRandom(rng)`?

phated (changes requested):
A few things and API discussion.

stdlib/random.gr
peblair: Because `ctrArg` is the pointer to the (wrapped i64) argument, and `ctr` is the actual i64 it holds. We want to decref the pointer. I'm updating this now to use `@unsafe` anyway, though, so it's moot.

compiler/test/stdlib/random.test.gr
peblair: No strong opinion. @ospencer Thoughts?

phated:
I think this needs a rebase on main (because it added `uint*.gr` files that were removed from #831)

compiler/test/stdlib/random.test.gr
ospencer: Yeah, I prefer it all existing in the random module.

stdlib/random.gr
phated: @peblair @ospencer did you ever decide on what you wanted? Should we just open an issue to evaluate/re-evaluate?

stdlib/random.gr
```diff
@@ -0,0 +1,183 @@
...
+  )
+  incCounter(random)
+  let ret2 = Int64.fromNumber(
+    Int32.toNumber(squares(random.counter, random.seed))
+  )
+  incCounter(random)
+  Int64.lor(Int64.shl(ret1, 32L), ret2)
+}
+
+// (can be removed after #831 lands)
```
phated, line 119: This has landed, so this should be removed?

stdlib/random.gr
```diff
@@ -0,0 +1,183 @@
...
+ * @returns The pseudo-random number generator
+ */
+export let make = seed => {
+  { seed, counter: 0L, initialized: false }
+}
+
+/**
+ * Creates a new pseudo-random number generator with a random seed.
+ *
+ * @returns The pseudo-random number generator
```
phated, line 58:
```suggestion
 * @returns A Result containing the pseudo-random number generator or a low-level exception
```
If we keep this returning a result, we'll want to update the docs to something like this

stdlib/random.gr
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * Generates a random 32-bit integer from the given pseudo-random number generator
+ * from a uniform distribution in the given range.
+ *
+ * @param random: The pseudo-random number generator to use
+ * @param low: The lower bound of the range (inclusive)
+ * @param high: The upper bound of the range (exclusive)
+ * @returns The randomly generated number
+ */
+export let nextInt32InRange = (random: Random, low: Int32, high: Int32) => {
```
phated, line 140: I wish we could use a Range here but I think those only allow `Number` type. 😞

stdlib/random.gr
phated:
```suggestion
 * @returns `Ok(generator)` of a pseudo-random number generator if successful or `Err(exception)` otherwise
```
Or, actually, this is more aligned with our other docs.

stdlib/random.gr
ospencer: Yeah, that's fine. Let's update the comment to just say that's why, though.

stdlib/random.gr
ospencer: I could go either way. I don't really want to make an issue since I feel like it'll kinda just sit around. For that reason I lean more towards doing the "correct" thing by just handing over a result and letting them decide. That way just avoids any problem when someone actually cares about what happens from WASI.

ospencer (approved):

phated (approved):
Approved!

phated (changes requested):
Whoops, just realized this doesn't have `@since` attributes, which is why it was docgen'ing without the required param.

phated (approved):
Had @ospencer look at the additional docs I added. This should be gtg now.

PR #922, opened 2021-09-24
chore: Leverage upstreamed js_of_ocaml patches

phated:
This is a draft against my current PR for JSOO that removes our custom runtime logic in favor of the upstreamed implementations.

There looks to be 1 more patch we'll need so we can mount the C: drive on Windows without adding a jsoo_hacks file, but I think that should be a fairly easy addition.


compiler/graindoc/jsoo_hacks.js
```diff
@@ -1,28 +1,11 @@
-//Provides: caml_make_path
-//Requires: caml_current_dir
-//Requires: caml_jsstring_of_string
-//Requires: caml_root, jsoo_mount_point, MlNodeDevice
-if (caml_root.charCodeAt(1) == 58 && caml_root[0] != 'C') {
-  // This ensure we can find our C:/snapshot/ stuff even when running in a diff drive
-  // Not the ideal place to do this but it is the easiest to avoid free variables
-  jsoo_mount_point.push({ path: 'C:/', device: new MlNodeDevice('C:/') });
-}
```
hhugo, line 9: Why not keep the 5 lines above, they seem much simpler than the new logic split between js and ocaml.

phated: @hhugo they get removed from the final JSOO code. Am I doing something wrong? It'd be much nicer to keep these if they weren't removed.

hhugo: Code without a `//Provides` should no be dropped.

phated: 🤔 In my testing, it seems like any `//Provides` that is not used as an `external` in the OCaml code gets dropped.

hhugo: Don't include the provide !
just put
```
//Requires: caml_root, jsoo_mount_point, MlNodeDevice
if (caml_root.charCodeAt(1) == 58 && caml_root[0] != 'C') {
  // This ensure we can find our C:/snapshot/ stuff even when running in a diff drive
  // Not the ideal place to do this but it is the easiest to avoid free variables
  jsoo_mount_point.push({ path: 'C:/', device: new MlNodeDevice('C:/') });
}
```
nothing else

phated: I'll give that a try!

phated: @hhugo Trying your example, the JS crashes with `caml_root is not defined`

hhugo: We should fix this. In the meantime, you could just keep theses lines under a `init_jsoo_for_grain` function (with a //Provides) and declare and call the external on the ml side.

phated: I was hoping I could do the `external` trick, but it seems that unless I provide a C-stub that it won't actually be detected (as per https://github.com/ocsigen/js_of_ocaml/issues/804#issuecomment-495826520). I tried providing the C-stub but it doesn't work because we have separate compilation targets for the JS vs native builds.

phated: Would you be open to adding an optional parameter to `Js_of_ocaml.Sys_js.mount` that specifies which filesystem type to mount? We would no longer need the separate JS file if that could be specified in OCaml.

hhugo: > I was hoping I could do the `external` trick, but it seems that unless I provide a C-stub that it won't actually be detected (as per [ocsigen/js_of_ocaml#804 (comment)](https://github.com/ocsigen/js_of_ocaml/issues/804#issuecomment-495826520)). I tried providing the C-stub but it doesn't work because we have separate compilation targets for the JS vs native builds.

I don't understand why this is an issue. Can't you have a c stub that does nothing ?

phated: This is the dune error that I got when trying to add the empty C-stub:
```
Error: Pure bytecode executables cannot contain foreign stubs.
Hint: If you only need to build a native executable use "(modes exe)".
```

phated: I've updated this PR to show the usage if we add the optional `fs_type` argument to `Sys_fs.mount`

hhugo: > We should fix this. In the meantime, you could just keep theses lines under a `init_jsoo_for_grain` function (with a //Provides) and declare and call the external on the ml side.

https://github.com/ocsigen/js_of_ocaml/issues/1141

compiler/grainc/dune
```diff
@@ -32,5 +26,4 @@
  (modes exe js)
  (libraries grain grain_diagnostics binaryen.js dune-build-info js_of_ocaml)
```
hhugo, line 27: Do you still need the js_of_ocaml dependency ?

hhugo:
Note that the following comment won't be true once you upgrade to next version of jsoo https://github.com/grain-lang/grain/blob/9bdd53e438937f3ddd265bca23875ef0e3d7b0e0/compiler/src/utils/files.re#L19

compiler/grainc/dune
phated: I think I can remove that now! Thanks for pointing it out

hhugo:
I've submitted a new version of js_of_ocaml to opam-repository https://github.com/ocaml/opam-repository/pull/19704

phated:
@hhugo excellent! I'll look into getting this updated and promoted from Draft once the opam PR is accepted. 🎉 Thanks for helping me get all those changes upstreamed 🙏

phated:
This should be ready for review! 🎉

PR #923, opened 2021-09-25
fix(grainfmt): Removed a trailing comma after a spread in a list, and fixed escaped single quote

marcusroberts:
Adding a comma after a spread in a list was a syntax error

[a,...b,]

Keep array set assignments on the same line as the =

Use constants direct from source to avoid losing any programmer intention regarding structure/format e.g. keep chars specified in the same format, numbers in the same base, etc.

### DIFF
--- a/compiler/grainformat/reformat.re
+++ b/compiler/grainformat/reformat.re
@@ -857,7 +857,7 @@ and print_constant = (c: Parsetree.constant) => {
       Printf.sprintf("%s/%s", n, d)
     | PConstBytes(b) => Printf.sprintf("%s", b)
     | PConstChar(c) =>
-      // String.escaped doesn't escape the single quote but we need it to
+      // Special case as a single quote isn't escaped in a string
       if (c == "'") {
         Printf.sprintf("'\\''");
       } else {
--- a/compiler/grainformat/reformat.re
+++ b/compiler/grainformat/reformat.re
@@ -1071,18 +1071,32 @@ and print_application =
   | [first] when prefixop(function_name) =>
     switch (first.pexp_desc) {
     | PExpApp(fn, _) =>
-      Doc.concat([
-        Doc.text(function_name),
-        // Doc.lparen,
-        print_expression(
-          ~parentIsArrow=false,
-          ~endChar=None,
-          ~original_source,
-          ~parent_loc,
-          first,
-        ),
-        //  Doc.rparen,
-      ])
+      let inner_fn = get_function_name(fn);
+      if (infixop(inner_fn)) {
+        Doc.concat([
+          Doc.text(function_name),
+          Doc.lparen,
+          print_expression(
+            ~parentIsArrow=false,
+            ~endChar=None,
+            ~original_source,
+            ~parent_loc,
+            first,
+          ),
+          Doc.rparen,
+        ]);
+      } else {
+        Doc.concat([
+          Doc.text(function_name),
+          print_expression(
+            ~parentIsArrow=false,
+            ~endChar=None,
+            ~original_source,
+            ~parent_loc,
+            first,
+          ),
+        ]);
+      };
 
     | _ =>
       Doc.concat([
--- a/compiler/grainformat/reformat.re
+++ b/compiler/grainformat/reformat.re
@@ -514,8 +514,6 @@ and resugar_list =
 
   let last_item_was_spread = ref(false);
 
-
-
   let items =
     List.map(
       i =>
@@ -531,7 +529,7 @@ and resugar_list =
               ~parent_loc,
               e,
             ),
-          )
+          );
         | Spread(e) =>
           last_item_was_spread := true;
           Doc.group(
@@ -545,7 +543,7 @@ and resugar_list =
                 e,
               ),
             ]),
-          )
+          );
         },
       processed_list,
     );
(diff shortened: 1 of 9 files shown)


compiler/grainformat/reformat.re
```diff
@@ -847,7 +856,14 @@ and print_constant = (c: Parsetree.constant) => {
     | PConstNumber(PConstNumberRational(n, d)) =>
       Printf.sprintf("%s/%s", n, d)
     | PConstBytes(b) => Printf.sprintf("%s", b)
-    | PConstChar(c) => Printf.sprintf("'%s'", String.escaped(c))
+    | PConstChar(c) =>
+      // String.escaped doesn't escape the single quote but we need it to
```
phated, line 860: Can you turn `String.escaped` into our own helper (as `Grain_utils.String_utils.escaped`) that does this logic? Any time we encounter ocaml stdlib functionality that we need to behave differently, we try to create a Grain_utils helper.

phated (commented):
Will this change after #926 ?

marcusroberts:
> Will this change after #926 ?

I think it looks better on the same line anyway, so I'd leave it as is.

compiler/grainformat/reformat.re
marcusroberts: I think my comment here is wrong.  A single quote doesn't need to be escaped in a string.  This is a very special case when printing a char.  What do you think?

marcusroberts: (comment updated)

phated: That's a good point. But do we use `String.escaped` anywhere else? Our parser wouldn't handle strings that weren't escaped correctly, if I'm not mistaken.

marcusroberts: The only other place is when formatting a string which is when it works properly, as an escaped \\' isn't legal

print("\'")
compiles to
Error: Illegal string character: "\\"

phated: Sounds good

marcusroberts: Oh, there's a Char.escaped!

marcusroberts: Fixed by using the proper function!

compiler/grainformat/reformat.re
```diff
@@ -847,7 +856,12 @@ and print_constant = (c: Parsetree.constant) => {
     | PConstNumber(PConstNumberRational(n, d)) =>
       Printf.sprintf("%s/%s", n, d)
     | PConstBytes(b) => Printf.sprintf("%s", b)
-    | PConstChar(c) => Printf.sprintf("'%s'", String.escaped(c))
+    | PConstChar(c) =>
+      if (String.length(c) > 0) {
+        Printf.sprintf("\'%s\'", Char.escaped(c.[0]));
```
phated, line 861: I don't understand string encodings, but can the Char in the `c` string take up more than 1 index? (e.g. will an emoji split between index 0 and 1 in the string?)

marcusroberts: I don't know.   I guess we're into the world of unicode now and what the String function returns...

marcusroberts: Resolved by simply reproducing the original source code for constants

marcusroberts:
This won't work until #930 is fixed.

compiler/grainformat/reformat.re
marcusroberts: But will fail on char constants until #930 is resolved

compiler/grainformat/reformat.re
```diff
@@ -512,11 +539,15 @@ and resugar_list =
...
   let processed_list = resugar_list_inner(expressions, parent_loc);
 
+  let last_item_was_spread = ref(false);
+
   let items =
     List.map(
       i =>
         switch (i) {
         | Regular(e) =>
+          last_item_was_spread := false;
```
peblair, line 549: I wish we didn't need to use a mutable variable here, but I can't think of a way to nicely clean it up without one :/

compiler/grainformat/reformat.re
```diff
@@ -839,30 +878,16 @@ and print_pattern =
...
-    };
-  Doc.text(print_c);
+
+and print_constant =
+    (
+      ~original_source: array(string),
+      ~loc: Grain_parsing__Location.t,
+      c: Parsetree.constant,
+    ) => {
+  Doc.text(get_original_code_snippet(loc, original_source));
```
peblair, line 888: Might warrant some sort of comment along the lines of "we just grab the original code for constants since it must be well-formatted"

compiler/grainformat/reformat.re
marcusroberts: Me too, I hate myself each time I use a mutable variable.

compiler/grainformat/reformat.re
marcusroberts: Done.

compiler/grainformat/reformat.re
marcusroberts: Taken a different approach now.

compiler/grainformat/reformat.re
peblair: 🥲

peblair (approved):
Looks good to me!

ospencer (approved):
Let it rock!

PR #928, opened 2021-09-25
fix(stdlib): Fix Char.code memory issue

### LINKED ISSUE #927: Infinite loop inside malloc.free
The below program causes an infinite loop inside `malloc.free`. It finishes without issues if executed with `--no-gc`.

I tried to narrow it down further, but gave up. Slight changes like the following make it **not** hang anymore:
- Changing the chars to simple ASCII values like 'a' and 'b'.
- Moving the if statement to the end of the loop.
- Changing the loop to `while(true)` or `for(;;)`.
- Not calling `Char.code`. Though I doubt it's got much to do with Char.code function itself, even if it's not fully conformant with the new convention for manual memory management.

``` grain
import Char from "char"

let chars = [> '\u{1F3F4}', '\u{E0067}' ]
let mut charPosition = 0
let len = 2
let mut done = false

while (!done) {
	if (charPosition >= len) {
		break
	}
	
	Char.code(chars[charPosition])
	charPosition = charPosition + 1
}
```

ospencer:
Fixes #927

### DIFF
--- a/compiler/test/stdlib/char.test.gr
+++ b/compiler/test/stdlib/char.test.gr
@@ -37,3 +37,19 @@ assert Char.code(Char.pred(Char.fromCode(0xE000))) == 0xD7FF
 // toString
 assert Char.toString('f') == "f"
 assert Char.toString('💯') == "💯"
+
+
+// issue #927
+let chars = [> '\u{1F3F4}', '\u{E0067}' ]
+let mut charPosition = 0
+let len = 2
+let mut done = false
+
+while (!done) {
+	if (charPosition >= len) {
+		break
+	}
+	
+	Char.code(chars[charPosition])
+	charPosition = charPosition + 1
+}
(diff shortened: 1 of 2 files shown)


phated (approved):
🐞 @cician puts them up and @ospencer knocks them down 🔨

marcusroberts (approved):
!!

cician:
> lady_beetle @cician puts them up and @ospencer knocks them down hammer

I thought the Wales flag emoji was cursed, but it was just bugs.  🐛 🦋 🪲🏴󠁧󠁢󠁷󠁬󠁳󠁿🐞🦗🪳

cician:
Since we're at it. Shouldn't it also decRef itself and the passed Char?

ospencer:
You are absolutely right!

ospencer:
I'll resolve those in a separate PR, though. Looks like this module was missed.

PR #931, opened 2021-09-26
fix: Correctly preserve start position for character literals

### LINKED ISSUE #930: Compiler: incorrect parse location reported for character constants
The parser seems to not include the opening quote for character constants (but it correct for the trailing quote)

for
print('a')

the location is reported as (1:7 - 1:9) 

print("a") has a location of (1:6-1:9)   

which is what I expected.

peblair:
Fixes #930. Same story as the fencepost error we had with string literals.

### DIFF
--- a/compiler/test/suites/chars.re
+++ b/compiler/test/suites/chars.re
@@ -1,10 +1,37 @@
 open Grain_tests.TestFramework;
 open Grain_tests.Runner;
+open Grain_parsing.Location;
+open Grain_middle_end.Anf_helper;
 
 describe("chars", ({test}) => {
   let assertSnapshot = makeSnapshotRunner(test);
   let assertCompileError = makeCompileErrorRunner(test);
   let assertRun = makeRunner(test);
+  let assertParseWithLocs = makeParseRunner(~keep_locs=true, test);
+  open Grain_parsing;
+  open Ast_helper;
+  let mk_loc =
+      (
+        file,
+        (start_line, start_col, start_bol),
+        (end_line, end_col, end_bol),
+      ) => {
+    loc_start: {
+      pos_fname: file,
+      pos_lnum: start_line,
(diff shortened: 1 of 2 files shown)


ospencer (approved):
Haha, I was just opening this PR!

ospencer:
If you want to add a test, we have examples for strings in the string suite at the bottom!

peblair:
@ospencer Test added

compiler/test/suites/chars.re
```diff
@@ -56,4 +83,28 @@ describe("chars", ({test}) => {
...
+    "'a'",
+    {
+      statements: [
+        char(~loc=mk_loc("char_loc_simple", (1, 0, 0), (1, 3, 0)), "a"),
+      ],
+      comments: [],
+      prog_loc: mk_loc("char_loc_simple", (1, 0, 0), (1, 3, 0)),
+    },
+  );
+  // note that the char length is calculated as having 4 bytes, not 1 codepoint
```
ospencer, line 98: This actually reminds me of that question I had around this—how do tools like vscode handle byte-based locs rather than unicode chars? 🤔

peblair: 🤷‍♂️ no clue. Want to make a follow-up issue?

ospencer: No 😛 

Sorta like a "if/when we see it's a problem..." kinda feeling.

ospencer: But! That does make me feel like we should have a test for a char like `'\u{1F3F4}'` too.

peblair: Added.

ospencer: Nice.

marcusroberts (approved):
Thanks for the quick fix!

PR #932, opened 2021-09-26
chore(compiler): Clear compiler warnings in linking tests

ospencer:
I changed the type of `compile` that's exposed to the tests, and missed that there were tests already using it directly.


peblair (approved):

phated (approved):
👍

PR #933, opened 2021-09-26
chore(stdlib): format the stdlib with the formatter 

marcusroberts:
Formatted the stdlib files with the formatter as it is in #923 - changed to reformat.re will go once #923 is merged ahead of this PR

Clean set of test results afterwards:

Test Suites: 0 failed, 1 passed, 1 total
Tests:       0 failed, 56 passed, 56 total
Time:        58.146s


stdlib/buffer.gr
```diff
@@ -45,27 +41,30 @@ let _64BIT_LEN = 8
...
 
 /* Doubles the size of buffer's underlying byte sequence, if the given size is larger than the size of a buffer's underlying byte sequence */
 let autogrow = (len, buf) => {
   while (buf.len + len > Bytes.length(buf.data)) {
     let mut n = Bytes.length(buf.data)
-    if (n == 0) n = 4 // Make sure bytes of 0 length grow too
+    if (n == 0) {
+      n = 4
+    }
+    // Make sure bytes of 0 length grow too
```
ospencer, line 53: A little sad this comment got moved.

stdlib/buffer.gr
```diff
@@ -300,8 +303,8 @@ export let rec addChar = (char: Char, buffer: Buffer) => {
       let (*) = WasmI32.mul
       let (&) = WasmI32.and
       let (>>) = WasmI32.shrU
-      for (let mut i = 0n; i < 3n; i = i + 1n) {
-        let c = Conv.toInt32(n >> (i * 8n) & 0xffn)
+      for (let mut i = 0n; i < 3n; i += 1n) {
+        let c = Conv.toInt32(n >> i * 8n & 0xffn)
```
ospencer, line 307: Hah, this totally gives me second thoughts about the parens since even I don't remember the precedence here 😂

stdlib/bytes.gr
```diff
@@ -605,5 +625,4 @@ export let rec make = (n: Number) => {
 /**
  * An empty byte sequence
  */
-export let empty = make(0)
-
+export let empty = make(0)
```
ospencer, line 628: Did it remove the newline at the end of the file?

stdlib/buffer.gr
marcusroberts: Hmmm, me too.  It shouldn't have got moved there at all.  I'll take a look.

stdlib/char.gr
```diff
@@ -196,8 +208,8 @@ export let rec toString = (c: Char) => {
   }
   let str = allocateString(n)
   Memory.copy(str + 8n, c + 4n, n)
-  let ret = WasmI32.toGrain(str): String
+  let ret = WasmI32.toGrain(str): (String)
   Memory.decRef(WasmI32.fromGrain(c))
   Memory.decRef(WasmI32.fromGrain(toString))
   ret
-}
+}
```
ospencer, line 215: Yeah, looks like it's that on all the files.

stdlib/bytes.gr
marcusroberts: Hmmm, I've seen this a couple of times now.   It may have and I think it needs to keep it.

stdlib/list.gr
```diff
@@ -149,21 +171,26 @@ let filteri = (fn, list) => {
 let rec reject = (fn, list) => {
   match (list) {
     [] => [],
-    [first, ...rest] => if (!fn(first)) [first, ...reject(fn, rest)] else reject(fn, rest)
+    [first, ...rest] =>
+      if (!(fn(first))) {
```
ospencer, line 175: Interesting that we got parens around the function call here.

stdlib/list.gr
```diff
@@ -262,42 +333,63 @@ let rec take = (n, list) => {
...
+    aItem => {
+      forEach(
+        bItem => {
+          list = [(aItem, bItem), ...list]
+        },
+        b,
+      )
+    },
+    a,
+  )
```
ospencer, line 392: This looks like it broke just about everywhere it could, but I don't think it needed to?

stdlib/map.gr
```diff
@@ -8,115 +8,111 @@ import WasmI32 from "runtime/unsafe/wasmi32"
...
-        Some(node) => {
-          node.next = None;
+    let tailNodes = Array.make(nextSize, None)
+    map.buckets = nextBuckets
+    Array.forEach(
+      old => {
+        copyNodeWithNewHash(old, nextBuckets, tailNodes)
+      },
+      currentBuckets,
+    )
```
ospencer, line 68: Yeah it seems like arguments that are functions always get pushed. This is fairly common, so do you think there's anything we can do about it?

stdlib/regex.gr
```diff
@@ -380,19 +400,39 @@ enum ParsedRegularExpression {
   REAlts(ParsedRegularExpression, ParsedRegularExpression),
   RESequence(List<ParsedRegularExpression>, Bool), // seq elts, needs backtrack
   REGroup(ParsedRegularExpression, Number), // regex, group ID
-  RERepeat(ParsedRegularExpression, Number, Option<Number>, Bool), // regex, min, max (None for infinity), true=non-greedy
+  RERepeat(ParsedRegularExpression,
+  Number,
+  Option<Number>,
+  Bool), // regex, min, max (None for infinity), true=non-greedy
```
ospencer, line 406: Looks like we're missing some indentation?

stdlib/regex.gr
```diff
@@ -611,135 +665,253 @@ parseClass = (buf: RegExBuf) => {
...
+              match (eat(buf, ']')) {
                 Err(_) => Err(parseErr(buf, "Missing closing `]`", 0)),
-                Ok(_) => Ok(List.join("", List.reverse(acc)))
+                Ok(_) => Ok(List.join("", List.reverse(acc))),
               }
             },
-            Ok(c) when (Char.code('a') <= Char.code(c) && Char.code(c) <= Char.code('z')) => {
+            Ok(c) when Char.code('a') <= Char.code(c) && Char.code(
+              c,
+            ) <= Char.code('z') => {
```
ospencer, line 688: This feels unfortunate! Could the breaks maybe happen if the infix operands can't fit on the same line?
```
Ok(c) when Char.code('a') <= Char.code(c) &&
Char.code(c) <= Char.code('z')
```

stdlib/regex.gr
```diff
@@ -1921,155 +2375,418 @@ let subArraysEqual = (arr1, start1, arr2, start2, length) => {
...
+    }
+  }
 
-let anyMatcherIterated = (max) => (buf: MatchBuf, pos: Number, start: Number, limit: Number, end: Number, state, stack) => {
-  let n = match(max) {
-    None => limit - pos,
-    Some(max) => min(max, limit - pos),
+let anyMatcherIterated = max =>
+
+    (
```
ospencer, line 2513: Looks like we got an extra space here.

stdlib/bytes.gr
marcusroberts: This is fixed in the next push

stdlib/char.gr
marcusroberts: Yup, and fixed.

stdlib/list.gr
marcusroberts: Hmmm, again I did that for a reason in some case but I can't remember, so I can remove it and run the formatter again

stdlib/buffer.gr
marcusroberts: Fixed in lated commit

stdlib/list.gr
marcusroberts: Removed for now, let's see if we get any problems

marcusroberts: Remembered it was needed around infix expressions, but fixed.

stdlib/list.gr
marcusroberts: It doesn't need to but because the block broke it propagates up the parent chain.  I need to apply some grouping to stop this happening.

marcusroberts: Looks like this is a common formatting approach, here's how it looks for reasonml:

image

stdlib/map.gr
marcusroberts: The rules of the formatter engine we are using is that once one thing breaks, everything should break.  And the pexpblock forces a break for the new lines inside the braces block.    The reason reformatter seems to do the same.

stdlib/regex.gr
marcusroberts: Good catch, fixed in next push

stdlib/regex.gr
marcusroberts: Fixed in next push

stdlib/regex.gr
marcusroberts: Nice catch - it's when a function returns a lambda.  Fixed.

Issue #841, opened 2021-09-01
Compiler: Support memory allocation in the wasi polyfill


ospencer:
Use of the `Memory` module is disallowed in the polyfill because of a cyclic dependency. This cyclic dependency can largely be ignored, since a problem would only occur if an error happened during the startup of the memory/gc modules. If that were to happen, you'd have bigger problems 😅 

This would require an update to the linker to properly link cyclic module dependencies.

PR #935, opened 2021-09-28
feat(compiler)!: Block off 1KiB of heap space as generic swap

peblair:
Currently, there is no way to allocate heap space in WASI polyfills (since they cannot import the `dataStructures.gr` module). To support cases in which some minimal heap use is desired (e.g. the NEAR SDK), this PR blocks off 1KiB of memory (`0x400`-`0x800`) which WASI polyfills can freely write to.

(Technically breaking, since the location of the runtime memory is moved)

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -350,12 +350,14 @@ let lookup_ext_func = (env, modname, itemname) =>
 /** Static runtime values */
 
 // Static pointer to the runtime heap
-// Leaves low 1000 memory unused for Binaryen optimizations
-let runtime_heap_ptr = 0x400;
+// Leaves low 1000 memory unused for Binaryen optimizations,
+// and 0x400-0x800 is reserved for WASI polyfills (which cannot
+// allocate memory)
+let runtime_heap_ptr = 0x800;
 // Start pointer for the runtime heap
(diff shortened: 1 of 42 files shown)


phated:
This is a workaround for #841, right? Did @ospencer have any ideas on how to allow memory allocation in the polyfills to avoid this breaking change?

peblair:
It does appear to be the same issue, but I had come up with this independently. This PR is a workaround for that issue. If @ospencer has a better idea, we can do that, but I need some sort of solution for the NEAR SDK.

ospencer:
Not sure how I feel about this yet, but in terms of something that'd work immediately, you can use a string or an array to block off some swap space.

ospencer:
@peblair I was chatting with Blaine about primitives and I had an idea about this:

> Right now all of our primitives are functions, but I think we could have some values as well! I was thinking through that wasm4 issue and philip's PR for reserving a chunk of heap space for something that needs it—I think it's a good use case for a `@heap.base` primitive that you can populate with like a `--memory-base` to scoot up where we start the heap. Then you can reserve whatever amount of space you need for whatever your use case is, would set us up to support something like AssemblyScript's `memory.data` (which would be so nice) etc.

He liked that idea. If you think that sounds reasonable too, I can make an issue for it!

Issue #940, opened 2021-09-30
Formatter: retain CRLF encoding


marcusroberts:
The formatter outputs all files as LF encoded, regardless of the source formatting

Detect the source format and output the formatted file in the same format.

PR #941, opened 2021-09-30
fix(grainfmt): Format CRLF files without error

marcusroberts:
The reading in of files with CRLF  line endings that left a stray \r character which caused a syntax error
This fixes that issue.   The file is left as a LF file so we need to fix to output in CRLF (see #940)

### DIFF
--- a/compiler/grainformat/grainformat.re
+++ b/compiler/grainformat/grainformat.re
@@ -38,14 +38,16 @@ let compile_parsed = (filename: option(string)) => {
         while (true) {
           let line = input_line(ic);
 
-          // TODO temporarily trim off the CRLF, but need to output in original format issue #940
-          linesList := linesList^ @ [String.trim(line)];
+          linesList := linesList^ @ [line];
         }
       ) {
       | exn => ()
       };
 
       program_str := String.concat("\n", linesList^);
+
+      // TODO add a new line to the end for where it's in CRLF more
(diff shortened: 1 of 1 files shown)


compiler/grainformat/grainformat.re
```diff
@@ -46,6 +46,9 @@ let compile_parsed = (filename: option(string)) => {
 
       program_str := String.concat("\n", linesList^);
 
+      // TODO add a new line to the end for where it's in CRLF more
```
ospencer, line 49: Could you add the issue to this TODO?

marcusroberts: Done

phated (approved):
👍  At least we have a workaround for now

compiler/grainformat/grainformat.re
```diff
@@ -46,6 +46,11 @@ let compile_parsed = (filename: option(string)) => {
 
       program_str := String.concat("\n", linesList^);
 
+      // add a new line to the end for where it's in CRLF more
+      // TODO Issue 940 handle CRLF properly
+
+      program_str := program_str^ ++ "\n";
```
phated, line 52:
```suggestion
      // add a new line to the end for where it's in CRLF more
      // TODO(#940): Handle CRLF properly
      program_str := program_str^ ++ "\n";
```
Light cleanup maybe?

PR #948, opened 2021-10-01
feat: Add npm releases to workflow

phated:
This adds additional jobs to our release workflow to publish `@grain/stdlib` and `@grain/js-runner` to npm if they were bumped as part of a release.

I actually needed to split the logic into 2 parts:
1. While we are building the compiler, we will also now run `npm pack` and uploaded the packed artifacts to the appropriate GitHub release. For example, the `js-runner.tgz` file will be uploaded to the `js-runner-vX.X.X` release on GitHub.
2. After **all** the artifacts are built and uploaded, those artifacts are used to publish to npm in their own steps.

This process is similar to what we do with our binaryen.ml package, but that is packed different due to esy.

Everything added for these npm releases need to be guarded on the "release_created" output for that library, so you'll notice that all of them have something like `if: ${{ needs.release-please.outputs.js-runner_release_created }}` guarding it from running if no release was created.

You can see these different workflows being run on my fork:
* https://github.com/phated/grain/actions/runs/1293191632 published both packages because I made changes in each.
    * Their artifacts were uploaded to respective releases at https://github.com/phated/grain/releases/tag/js-runner-v0.4.2 and https://github.com/phated/grain/releases/tag/stdlib-v0.4.4
    Screen Shot 2021-09-30 at 6 52 19 PM

* https://github.com/phated/grain/actions/runs/1293243003 only published the `stdlib` package because no changes were made to `js-runner`
    Screen Shot 2021-09-30 at 6 52 09 PM

* https://github.com/phated/grain/actions/runs/1293285946 only published the `js-runner` package because no changes were made to `stdlib`
    Screen Shot 2021-09-30 at 6 51 56 PM

* https://github.com/phated/grain/actions/runs/1293323788 published neither npm package because no changes were made to either.
    Screen Shot 2021-09-30 at 6 51 42 PM

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,10 @@ jobs:
       release_created: ${{ steps.release.outputs.release_created }}
       upload_url: ${{ steps.release.outputs.upload_url }}
       tag_name: ${{ steps.release.outputs.tag_name }}
+      stdlib_upload_url: ${{ steps.release.outputs.stdlib--upload_url }}
+      stdlib_release_created: ${{ steps.release.outputs.stdlib--release_created }}
+      js-runner_upload_url: ${{ steps.release.outputs.js-runner--upload_url }}
+      js-runner_release_created: ${{ steps.release.outputs.js-runner--release_created }}
     steps:
       - uses: GoogleCloudPlatform/release-please-action@v2.33.1
         id: release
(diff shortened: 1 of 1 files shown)


ospencer (approved):
Amazing work!!

peblair (approved):
🙌

marcusroberts (approved):

PR #950, opened 2021-10-01
fix(compiler): Resolve all modules fully during linking

ospencer:


compiler/src/linking/link.re
```diff
@@ -559,6 +545,13 @@ let link_all = (linked_mod, dependencies, signature) => {
     };
   };
   List.iter(link_one, dependencies);
+
+  Comp_utils.write_universal_exports(
+    linked_mod,
+    signature,
+    Hashtbl.find(exported_names, Module_resolution.current_filename^()),
+  );
```
phated, line 553: In our `linking` tests, we don't have a test to check the exports of a linked module. Can you add one to show that this didn't change anything?

ospencer: Our `wasm_utils` are still a little primitive and can't check the types of the exports, just that they exist (and our polyfill tests show that `write_universal_exports` does what we expect). I can make an issue to get types implemented in `wasm_utils`, though. (Or if you just want an extra smoke test on the exports existing I can do that too.)

phated: Yeah, let's open an issue to track.

phated (approved):
🔥 Love it!

PR #951, opened 2021-10-02
feat(stdlib)-Add-string-trim-commands-to-string-library

spotandjake:
Add's Trim Functions, trimStart, trimEnd & trim to the String Libary along with their appropriate tests.

### DIFF
--- a/compiler/test/stdlib/string.test.gr
+++ b/compiler/test/stdlib/string.test.gr
@@ -258,3 +258,28 @@ assert String.decode(bytes, String.UTF16_LE) == "¢"
   }, emojis)
   assert Array.reverse(Array.fromList(tmp)) == Array.mapi((c,i) => (c,i), codes)
 }
+
+// TrimStart
+assert String.trimStart("t   test") == "t   test"
+assert String.trimStart("   test") == "test"
+assert String.trimStart("   test   ") == "test   "
+assert String.trimStart("test   ") == "test   "
+assert String.trimStart("test") == "test"
+assert String.trimStart("") == ""
+assert String.trimStart("  ") == ""
+// TrimEnd
+assert String.trimEnd("test   t") == "test   t"
+assert String.trimEnd("test   ") == "test"
+assert String.trimEnd("   test   ") == "   test"
+assert String.trimEnd("   test") == "   test"
+assert String.trimEnd("test") == "test"
+assert String.trimEnd("") == ""
+assert String.trimEnd("  ") == ""
+// Trim
+assert String.trim("t   test   t") == "t   test   t"
+assert String.trim("test   ") == "test"
+assert String.trim("   test   ") == "test"
+assert String.trim("   test") == "test"
+assert String.trim("test") == "test"
+assert String.trim("") == ""
+assert String.trim("  ") == ""
--- a/compiler/test/stdlib/string.test.gr
+++ b/compiler/test/stdlib/string.test.gr
@@ -259,7 +259,7 @@ assert String.decode(bytes, String.UTF16_LE) == "¢"
   assert Array.reverse(Array.fromList(tmp)) == Array.mapi((c,i) => (c,i), codes)
 }
 
-// TrimStart
+// TrimStart Tests
 assert String.trimStart("t   test") == "t   test"
 assert String.trimStart("   test") == "test"
 assert String.trimStart("   test   ") == "test   "
@@ -267,7 +267,7 @@ assert String.trimStart("test   ") == "test   "
 assert String.trimStart("test") == "test"
 assert String.trimStart("") == ""
 assert String.trimStart("  ") == ""
-// TrimEnd
+// TrimEnd Tests
 assert String.trimEnd("test   t") == "test   t"
 assert String.trimEnd("test   ") == "test"
 assert String.trimEnd("   test   ") == "   test"
@@ -275,7 +275,7 @@ assert String.trimEnd("   test") == "   test"
 assert String.trimEnd("test") == "test"
 assert String.trimEnd("") == ""
 assert String.trimEnd("  ") == ""
-// Trim
+// Trim Tests
 assert String.trim("t   test   t") == "t   test   t"
 assert String.trim("test   ") == "test"
 assert String.trim("   test   ") == "test"
--- a/stdlib/string.gr
+++ b/stdlib/string.gr
@@ -1662,7 +1662,7 @@ export let trimStart = (str: String) => trimString(str, false)
  */
 export let trimEnd = (str: String) => trimString(str, true)
 /**
- * Trims a string removing any leading or following white spaces.
+ * Trims a string removing leading and following white spaces.
  *
  * @param str: The string to be trimmed.
  * @returns The trimmed string.
--- a/stdlib/string.gr
+++ b/stdlib/string.gr
@@ -1609,3 +1609,64 @@ export let rec forEachCodePointi = (fn: (Number, Number) -> Void, str: String) =
   Memory.decRef(WasmI32.fromGrain(forEachCodePointi))
   void
 }
+
+/**
+ * Trims a string removing any leading or following white spaces based on the end param.
+ *
+ * @param str: The string to be trimmed.
+ * @param end: If true we are trimming the following spaces, otherwise trim leading spaces.
+ * @returns The trimmed string.
+ * 
+ * @example String.trimString("   Hello World   ", false)
+ */
+let trimString = (str: String, end: Bool) => {
+  let chars = explode(str), charsLength = length(str)
+  let mut i = 0, offset = 1
+  if (end) {
+    i = charsLength-1
+    offset = -1
+  }
+  for (; i < charsLength && i > -1; i += offset) {
+    if (
+      // Spacing
+      chars[i] != '	'  && // Tab
+      chars[i] != ''  && // LINE TABULATION
+      chars[i] != ''  && // FORM FEED (FF)
+      chars[i] != ' '  && // Space
+      chars[i] != ' '  && // No Break Space
+      chars[i] != '﻿' && // 	ZERO WIDTH NO-BREAK SPACE
+      // Line Terminator
+      chars[i] != '\n' && // LF
+      chars[i] != '\r' // CR
+    ) break
+  }
+  if (end) slice(0, i+1, str)
+  else slice(i, charsLength, str)
+}
+/**
+ * Trims the `start` of a string removing any leading white spaces.
+ *
+ * @param str: The string to be trimmed.
+ * @returns The trimmed string.
+ * 
+ * @example String.trimStart("   Hello World")
+ */
+export let trimStart = (str: String) => trimString(str, false)
+/**
+ * Trims the `end` of a string removing any following white spaces.
+ *
+ * @param str: The string to be trimmed.
+ * @returns The trimmed string.
+ * 
+ * @example String.trimEnd("Hello World   ")
+ */
+export let trimEnd = (str: String) => trimString(str, true)
+/**
+ * Trims a string removing any leading or following white spaces.
+ *
+ * @param str: The string to be trimmed.
+ * @returns The trimmed string.
+ * 
+ * @example String.trim("   Hello World   ")
+ */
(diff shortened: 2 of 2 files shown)


stdlib/string.gr
```diff
@@ -1609,3 +1609,64 @@ export let rec forEachCodePointi = (fn: (Number, Number) -> Void, str: String) =
...
+  let mut i = 0, offset = 1
+  if (end) {
+    i = charsLength-1
+    offset = -1
+  }
+  for (; i < charsLength && i > -1; i += offset) {
+    if (
+      // Spacing
+      chars[i] != '	'  && // Tab
+      chars[i] != ''  && // LINE TABULATION
```
peblair, line 1633: Maybe we should use `\u` escapes for these constants? Also, there should be a TODO to use the Unicode whitespace property once #661 is completed.

stdlib/string.gr
```diff
@@ -1609,3 +1609,64 @@ export let rec forEachCodePointi = (fn: (Number, Number) -> Void, str: String) =
...
+  let chars = explode(str), charsLength = length(str)
+  let mut i = 0, offset = 1
+  if (end) {
+    i = charsLength-1
+    offset = -1
+  }
+  for (; i < charsLength && i > -1; i += offset) {
+    if (
+      // Spacing
+      chars[i] != '	'  && // Tab
```
peblair, line 1632: @ospencer Will all of these `chars[i]` get optimized into a single lookup? I think not, but I want to double-check before requesting the change.

stdlib/string.gr
```diff
@@ -1609,3 +1609,64 @@ export let rec forEachCodePointi = (fn: (Number, Number) -> Void, str: String) =
...
+/**
+ * Trims a string removing any leading or following white spaces based on the end param.
+ *
+ * @param str: The string to be trimmed.
+ * @param end: If true we are trimming the following spaces, otherwise trim leading spaces.
+ * @returns The trimmed string.
+ * 
+ * @example String.trimString("   Hello World   ", false)
+ */
+let trimString = (str: String, end: Bool) => {
```
peblair, line 1622: @grain-lang/core Thoughts on `trim` vs `strip`?

ospencer: I think Python and Ruby call it `strip` and just about everyone else calls it `trim`. I personally prefer `trim`, since when I see `strip` I think of it as being possible throughout the string. Sorta like `deburr`, honestly. Or that I get to choose character classes of what I want to strip. Whereas I like `trim` since it feels more bounded in what it could possibly do.

Maybe doing a ton of JS dev has just molded me, idk.

stdlib/string.gr
ospencer: No, they don't.

stdlib/string.gr
spotandjake: I don't know how much of a difference my input makes here but personally, I agree with @ospencer,  strip sounds a lot more like you are removing a value from the string no matter the position whereas trim sounds is more like you are trimming off extra stuff be it the start or the beginning.

stdlib/string.gr
spotandjake: Fixed in latest commit

stdlib/string.gr
spotandjake: Fixed in latest commit.

compiler/test/stdlib/string.test.gr
```diff
@@ -258,3 +258,28 @@ assert String.decode(bytes, String.UTF16_LE) == "¢"
   }, emojis)
   assert Array.reverse(Array.fromList(tmp)) == Array.mapi((c,i) => (c,i), codes)
 }
+
+// TrimStart Tests
```
phated, line 262:
```suggestion
// String.trimStart
```
nit

compiler/test/stdlib/string.test.gr
```diff
@@ -258,3 +258,28 @@ assert String.decode(bytes, String.UTF16_LE) == "¢"
...
+assert String.trimStart("t   test") == "t   test"
+assert String.trimStart("   test") == "test"
+assert String.trimStart("   test   ") == "test   "
+assert String.trimStart("test   ") == "test   "
+assert String.trimStart("test") == "test"
+assert String.trimStart("") == ""
+assert String.trimStart("  ") == ""
+// TrimEnd Tests
```
phated, line 270:
```suggestion
// Sting.trimEnd
```
nit

compiler/test/stdlib/string.test.gr
```diff
@@ -258,3 +258,28 @@ assert String.decode(bytes, String.UTF16_LE) == "¢"
...
+assert String.trimEnd("test   t") == "test   t"
+assert String.trimEnd("test   ") == "test"
+assert String.trimEnd("   test   ") == "   test"
+assert String.trimEnd("   test") == "   test"
+assert String.trimEnd("test") == "test"
+assert String.trimEnd("") == ""
+assert String.trimEnd("  ") == ""
+// Trim Tests
```
phated, line 278:
```suggestion
// String.trim
```
nit

stdlib/string.gr
```diff
@@ -1609,3 +1609,66 @@ export let rec forEachCodePointi = (fn: (Number, Number) -> Void, str: String) =
...
+
+/**
+ * Trims a string removing any leading or following white spaces based on the end param.
+ *
+ * @param str: The string to be trimmed.
+ * @param end: If true we are trimming the following spaces, otherwise trim leading spaces.
+ * @returns The trimmed string.
+ * 
+ * @example String.trimString("   Hello World   ", false)
+ */
```
phated, line 1621: No graindoc on non-exported member, please. It gives the false impression that docs will be generated for internal functions.

stdlib/string.gr
```diff
@@ -1609,3 +1609,66 @@ export let rec forEachCodePointi = (fn: (Number, Number) -> Void, str: String) =
...
+      // Spacing
+      currentChar != '\u{0009}'  && // Tab
+      currentChar != '\u{000B}'  && // LINE TABULATION
+      currentChar != '\u{000C}'  && // FORM FEED (FF)
+      currentChar != '\u{0020}'  && // Space
+      currentChar != '\u{00A0}'  && // No Break Space
+      currentChar != '\u{FEFF}' && // 	ZERO WIDTH NO-BREAK SPACE
+      // Line Terminator
+      currentChar != '\n' && // LF
+      currentChar != '\r' // CR
```
phated, line 1642:
```suggestion
      currentChar != '\u{0009}'  && // Tab
      currentChar != '\u{000B}'  && // LINE TABULATION
      currentChar != '\u{000C}'  && // FORM FEED (FF)
      currentChar != '\u{0020}'  && // Space
      currentChar != '\u{00A0}'  && // No Break Space
      currentChar != '\u{FEFF}'  && // ZERO WIDTH NO-BREAK SPACE
      // Line Terminators
      currentChar != '\n' && // LF
      currentChar != '\r' // CR
```
comment alignment nits

stdlib/string.gr
```diff
@@ -1609,3 +1609,66 @@ export let rec forEachCodePointi = (fn: (Number, Number) -> Void, str: String) =
...
+}
+/**
+ * Trims the `start` of a string removing any leading white spaces.
+ *
+ * @param str: The string to be trimmed.
+ * @returns The trimmed string.
+ * 
+ * @example String.trimStart("   Hello World")
+ */
+export let trimStart = (str: String) => trimString(str, false)
```
phated, line 1656:
```suggestion
/**
 * Trims the beginning of a string—removing any leading whitespace characters.
 *
 * @param string: The string to be trimmed
 * @returns The trimmed string
 * 
 * @example String.trimStart("   Hello World") == "Hello World"
 */
export let trimStart = (string: String) => trimString(string, false)
```

stdlib/string.gr
```diff
@@ -1609,3 +1609,66 @@ export let rec forEachCodePointi = (fn: (Number, Number) -> Void, str: String) =
...
+/**
+ * Trims the `end` of a string removing any following white spaces.
+ *
+ * @param str: The string to be trimmed.
+ * @returns The trimmed string.
+ * 
+ * @example String.trimEnd("Hello World   ")
+ */
+export let trimEnd = (str: String) => trimString(str, true)
```
phated, line 1665:
```suggestion
/**
 * Trims the end of a string—removing any trailing whitespace characters.
 *
 * @param string: The string to be trimmed
 * @returns The trimmed string
 * 
 * @example String.trimEnd("Hello World   ") == "Hello World"
 */
export let trimEnd = (string: String) => trimString(string, true)
```

stdlib/string.gr
```diff
@@ -1609,3 +1609,66 @@ export let rec forEachCodePointi = (fn: (Number, Number) -> Void, str: String) =
...
+/**
+ * Trims a string removing leading and following white spaces.
+ *
+ * @param str: The string to be trimmed.
+ * @returns The trimmed string.
+ * 
+ * @example String.trim("   Hello World   ")
+ */
+export let trim = (str: String) => trimEnd(trimStart(str))
```
phated, line 1674:
```suggestion
/**
 * Trims a string—removing all leading and trailing whitespace characters.
 *
 * @param string: The string to be trimmed
 * @returns The trimmed string
 * 
 * @example String.trim("   Hello World   ") == "Hello World"
 */
export let trim = (string: String) => trimEnd(trimStart(string))
```

phated, line 1674: An additional question on the implementation: do we always want to call 2 functions to trim a string from both ends? Is there a more efficient way to do it?

phated (changes requested):
This is looking awesome! I have a few nits around the documentation and one question about the performance on `trim`

stdlib/string.gr
spotandjake: I could use an enum as the input for the internal function with Start, End, Both as values and then just only have one function call. I cant think of and couldn't find any faster implementation but if you know of one I would love to take a crack at implementing it.

phated: I wasn't thinking about an enum, but more concerned that you are doing 2 loops on the string. Let me poke around a bit.

spotandjake: That's what I figured, I could possibly do it in one loop, and loop through both sides at once and then just break on the longer side marking the lengths of each side, I figure that would not increase speed by that much though anyways and may make the code a little a messy.

phated: It seems that Ruby implements it as 2 separate loops (actually a **lot** of loops, but let's call it 2) while lodash relies on using RegExps. I think I'm fine with the 2-loop approach here unless @peblair or @ospencer has a better solution.

stdlib/string.gr
phated: I agree with `trim`

phated (approved):
Looking good to me! Awesome work @spotandjake 🎉 

Just waiting to hear if the others have any performance improvement suggestions.

stdlib/string.gr
spotandjake: I dont know how much aot optimization the compiler does on regex i assumed it wasnt much and figured the loop would be more efficent. if the regex method is more efficent i would be happy implementing that.

phated: @spotandjake I don't believe that a Regex would be more efficient than the current implementation (and would add a large dependency to the String stdlib module), so we probably don't want to use it.

phated: My thought was to add 2 loops inside your helper function depending on the start/end flags, which would avoid an extra `explode` and `length` call.

spotandjake: To optimize allocations for large strings it may make sense to use .charAt over chars[i] with the String.explode if the string is over a certain length. this way if people have massive data a big array isnt being allocated and if they have smaller data they are getting the fastest method available.

peblair: The only problem I see with `charAt` is that it is `O(n)` because Grain strings are UTF-8 (meaning that the worst-case complexity of this function (called with a string of spaces) would be `O(n^2)`). I would agree that some sort of heuristic based on the length to decide whether to explode would make sense. Additionally, I might throw in some sort of basic `if (!isWhitespace(String.charAt(0))) {...}` / `if (!isWhitespace(String.charAt(-1))) {...}` check at the beginning to bail out quickly. Finally, I'm +1 on the two-loop solution.

Issue #702, opened 2021-06-06
Document all stdlib exports


peblair:
All functions which are `export`ed in the `stdlib` folder should have a documentation comment, including those under `stdlib/runtime`.

spotandjake:
List of all completed and uncompleted files
+ [ ] /runtime/
  + [ ] /unsafe/
    + [ ] conv
    + [ ] errors
    + [ ] memory
    + [ ] printWasm
    + [ ] tags
    + [ ] wasmf32
    + [ ] wasmf64
    + [ ] wasmi32
    + [ ] wasmi64
  + [ ] dataStructure
    + [ ] needs generation
  + [ ] debug
  + [ ] equal
  + [ ] exception
  + [ ] gc
  + [ ] malloc
  + [ ] numbers
  + [ ] numberUtils
  + [ ] string
  + [ ] wasi
+ [x] /sys/
   + [x] file
   + [x] process
   + [x] random
   + [x] time
+ [x] array
+ [x] buffer
+ [ ] bytes
  + [ ] needs generation
+ [ ] char
  + [ ] Waiting on pr
+ [ ] exception
+ [ ] float32
+ [ ] float64
+ [ ] hash
+ [ ] int32
+ [ ] int64
+ [ ] list
+ [ ] map
+ [x] number
+ [ ] option
+ [ ] pervasives
+ [ ] queue
+ [x] range
+ [x] regex
+ [ ] result
+ [ ] set
+ [ ] stack
  + [ ] Not yet Generated
+ [ ] string
  + [ ] Waiting on pr

phated:
Thanks to the excellent effort by @spotandjake, we now have all *exports* from stdlib documented! Nothing in `runtime/` is considered to be an export, so I don't think we need an issue open to add graindoc to it.

Still, we should try to update those files whenever we are working on them!

PR #956, opened 2021-10-02
chore(stdlib): Add GrainDoc to float32 Library

spotandjake:
Add GrainDoc to float32 Library.
related: #702.


stdlib/float32.gr
```diff
@@ -1,3 +1,7 @@
+/**
+ * @module Float32: Utilities for working with the Float32 type.
+ * @example import Float32 from "float32"
```
phated, line 3: Crud, I just realized that we didn't put a `@since` attribute on the Float64 module definition. Can you add it here and probably open another PR for adding it to Float64?

spotandjake: Sure

phated (approved):
Excellent! Great work 🎉

ospencer (approved):
This looks really good!

PR #959, opened 2021-10-02
fix(grainfmt): Improve formatter performance

marcusroberts:
One trick was simple, use cons instead of appending to an array.  I reverse it in the hope that helps the sort be faster.

Instead of folding over the entire list each time (and skipping) by using recursion I can stop early which speeds up partitions of the AST in the earlier phases when there are lots of non-relevant items in the node list still

### DIFF
--- a/compiler/grainformat/walktree.re
+++ b/compiler/grainformat/walktree.re
@@ -162,14 +162,14 @@ let walktree =
 
 let rec partition_comments_internal =
         (
+          ~range: option(Grain_parsing.Location.t),
+          ~locations: list(node_t),
+          ~pre_comments: list(Grain_parsing.Parsetree.comment),
+          ~leading_only: bool,
           loc: Grain_parsing.Location.t,
-          range: option(Grain_parsing.Location.t),
-          locations: list(node_t),
-          preComments: list(Grain_parsing.Parsetree.comment),
-          leadingOnly: bool,
         ) => {
   switch (locations) {
-  | [] => (preComments, [])
+  | [] => (pre_comments, [])
   | [first, ...second] =>
     let nodeLoc = get_node_loc(first);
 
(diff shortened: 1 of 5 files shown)


compiler/grainformat/walktree.re
```diff
@@ -121,9 +121,13 @@ let walktree =
 
   let iter_location = (self, location) =>
     if (!List.mem(Code(location), all_locations^)) {
-      all_locations := List.append(all_locations^, [Code(location)]);
+      all_locations := List.cons(Code(location), all_locations^);
     };
 
+  // put it in the right order ready for the final sort as hopefully
+  // almost the right order is faster to sort
+  all_locations := List.rev(all_locations^);
```
ospencer, line 129: IIRC OCaml default sort implementation is mergesort, so this wouldn't help.

compiler/grainformat/walktree.re
```diff
@@ -121,9 +121,13 @@ let walktree =
 
   let iter_location = (self, location) =>
     if (!List.mem(Code(location), all_locations^)) {
```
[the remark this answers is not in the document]
marcusroberts: Yes, I think I found that on the timings I was doing.  OK, hashtbl it is

compiler/grainformat/walktree.re
marcusroberts: removed it

compiler/grainformat/walktree.re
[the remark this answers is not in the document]
marcusroberts: Bam!   Down from 10 seconds to 4 seconds.  (The hashtbl is just for de-duping the locations added, as I need an ordered list of actual locations)

compiler/grainformat/walktree.re
```diff
@@ -148,61 +160,99 @@ let walktree =
   // );
 };
 
+let rec partition_comments_internal =
+        (
+          loc: Grain_parsing.Location.t,
+          range: option(Grain_parsing.Location.t),
+          locations: list(node_t),
+          preComments: list(Grain_parsing.Parsetree.comment),
+          leadingOnly: bool,
```
phated, line 169: Please make this a named parameter.

compiler/grainformat/walktree.re
```diff
@@ -148,61 +160,99 @@ let walktree =
...
+    res;
+  };
+};
+
 let partition_comments =
-    (loc: Grain_parsing.Location.t, range: option(Grain_parsing.Location.t))
+    (
+      loc: Grain_parsing.Location.t,
+      range: option(Grain_parsing.Location.t),
+      leadingOnly: bool,
```
phated, line 249: Please make this a named parameter

compiler/grainformat/walktree.re
marcusroberts: Thank you, done.

compiler/grainformat/walktree.re
marcusroberts: Yep!  Done.

compiler/grainformat/walktree.re
```diff
@@ -119,10 +119,22 @@ let walktree =
 
   all_locations := comment_locations;
 
-  let iter_location = (self, location) =>
-    if (!List.mem(Code(location), all_locations^)) {
-      all_locations := List.append(all_locations^, [Code(location)]);
+  // use a hash table to de-deupe adding locations to the list
```
ospencer, line 122: Any reason not to form a list from the hash when you need it?

compiler/grainformat/walktree.re
```diff
@@ -119,10 +119,22 @@ let walktree =
...
+  // because the iterator returns the same location multiple times
+
+  // best guess of size, 20 nodes per top level stmt
```
ospencer, line 125: Probably more efficient to just guess a total size. Something like a static 4096 probably works fine here.

compiler/grainformat/walktree.re
```diff
@@ -148,61 +160,105 @@ let walktree =
...
+        | Comment((l, c)) =>
+          let (pre, post) =
+            partition_comments_internal(
+              ~range,
+              ~locations=second,
+              ~pre_comments,
+              ~leading_only,
+              loc,
+            );
+          (pre, [c] @ post);
```
ospencer, line 237:
```suggestion
          (pre, [c, ...post]);
```

compiler/grainformat/walktree.re
marcusroberts: Oh, no reason!  I didn't realise you could get an ordered list out!

ospencer (approved):

compiler/grainformat/walktree.re
```diff
@@ -148,61 +156,105 @@ let walktree =
   // );
 };
 
+let rec partition_comments_internal =
```
peblair, line 159: Would maybe like to see a doc comment on either this or `partition_comments`, just so it's clear what knobs these parameters are turning.

marcusroberts: Great idea, that's in!

PR #947, opened 2021-10-01
Chore(Stdlib)-Generate-Grain-Doc-For-Stdlibs-that-already-have-it

spotandjake:
Chore(Stdlib)-Generate-Grain-Doc-For-Stdlibs-that-already-have-it

Related to issue #702 .

PR #964, opened 2021-10-05
chore(stdlib): Update docs for Bytes module

phated:
I noticed in #947 that the Bytes module was using the outdated syntax, and upon further investigation, it had the same ordering issues that I removed from the Buffer module.

This PR updates the docblock, parameter names, and re-organizes the file. I also formatted the module with grainformat and generated the markdown docs.

Diff is terrible because I reorganized the file 😭

### DIFF
--- a/stdlib/bytes.gr
+++ b/stdlib/bytes.gr
@@ -260,7 +260,6 @@ export let rec resize = (left: Number, right: Number, bytes: Bytes) => {
     size - srcOffset
   }
   if (len > 0n) {
-    // formatter-ignore
     Memory.copy(
       dst + _VALUE_OFFSET + dstOffset,
       src + _VALUE_OFFSET + srcOffset,
@@ -315,7 +314,6 @@ export let rec move =
     throw Exception.InvalidArgument("Invalid destination bytes range")
   }
   let end = srcIndex + length
-  // formatter-ignore
   let ret = Memory.copy(
     dst + _VALUE_OFFSET + dstIndex,
     src + _VALUE_OFFSET + srcIndex,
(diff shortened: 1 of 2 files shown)


stdlib/bytes.gr
```diff
@@ -49,561 +61,656 @@ let checkIndexIsInBounds = (i, byteSize, max) => {
...
+  if (dstIndex + length > dstSize) {
+    throw Exception.InvalidArgument("Invalid destination bytes range")
+  }
+  let end = srcIndex + length
+  // formatter-ignore
+  let ret = Memory.copy(
+    dst + _VALUE_OFFSET + dstIndex,
+    src + _VALUE_OFFSET + srcIndex,
+    length,
+  )
```
phated, line 323: cc @marcusroberts - this broke crazily without the comment:
```
let ret = Memory.copy(
  dst + _VALUE_OFFSET +
  dstIndex,
  src + _VALUE_OFFSET +
  srcIndex,
  length,
)
```

marcusroberts: I think that's fixed in #959 

marcus@Marcuss-Air grain % grain format simple.gr
let ret = Memory.copy(
  dst + _VALUE_OFFSET +
  dstIndex,
  src + _VALUE_OFFSET +
  srcIndex,
  length,
)

assuming that's what you needed?

marcusroberts: Ignore my last comment, yes that's still there in the latest formatter,  so raising issue #965

marcusroberts: Fixed in #966

peblair (approved):
Modulo the formatter issue, this looks good. Thanks!

marcusroberts (approved):
LGTM!

PR #967, opened 2021-10-06
fix(compiler): Don't over-optimize mutable imports

ospencer:


marcusroberts (approved):
Looks good!

peblair (approved):
Good catch 👍

Issue #773, opened 2021-07-07
Allocating a recursive function leaks memory


cician:
Allocating a recursive function leaks memory, even if it's never executed. This simple program stops with an OutOfMemory error. Here foo() is allocated for each run of the loop and never freed. If it doesn't call itself then there's no leak.

``` grain
for (let mut i=0; i<10000000; i += 1) {
	for (let mut j=0; j<10000000; j += 1) {
		let rec foo = () => {
			print("never executed")
			foo()
		}
		void
	}
}
```

phated:
This was solved in https://github.com/grain-lang/grain/pull/1152

PR #984, opened 2021-10-12
feat(stdlib): Memory leak checker

cician:
This PR adds a tool that is intended to help solve issues in code with manual memory management.

I've added a new module called `runtimeDebug`, but I'm open to suggestions if another name or placing it in an existing one is more appropriate. For example there's a `debug` module, but I don't know what its purpose is.

## How it works
The changes to the `gc` module has added support to setup hooks called on each incRef, decRef, malloc and free. Otherwise the changes here are limited to expose these hooks and remove them.

The `runtimeDebug` exposes two functions:
1. `ensureNoLeaks`, intended for leak checks in unit testing. It only prints anything if leaks are actually found.
2. `leakDebugBlock`, intended to be used when debugging gets tricky and more details are needed. It can print every incRef, decRef, malloc and free with addresses, tags and even object contents as strings (when safe-_ish_ to do so). This part is inspired by gc.setDebug functionally that is currently commented out.

When the hooks are connected, internally there's a bit of a delicate dance on hot ash, since `runtimeDebug` is using managed memory and even GC, while GC itself calls into incRef/decRef/malloc/free. To accomplish this, the hooks are inhibited temporarily.

## A few details
I've used raw Wasmi32 pointers for function hooks. Maybe I should use something else? `box`? Or actual function type?

While the hook functions use WasmI32 pointers for tracked objects, I convert them internally to Number for use in GC land. May have used Int32 instead, but I chose Number hoping to reduce allocations at least for programs with small enough heap.

## Limitations and drawbacks
Nesting `ensureNoLeaks`/`leakDebugBlock` is not supported, but not prevented.

Per function/code block based testing may not work for debugging some cases covered by the old gc.setDebug functionality. In particular when runtime modules themselves are starting up.

`ensureNoLeaks` is obviously useful for detecting memory leaks, but may not be all that helpful for other issues like _use after free_, overwritten object headers or infinite loops in malloc.free (all things experience by yours truly). It **may** help to run `leakDebugBlock` with printIncrefsAndDecRefs=true, printMallocAndFree=true and printObjectContents=false.

It adds a slight overhead to memory management for all programs, even if the hooks aren't hooked up. I haven't done any comparisons yet. Hopefully not much, just checking if the hook function is set in a variable. Some sort of conditional compilation would be nice in this case.

Finally, it's not able to leak check the leak checker, so I don't guarantee it doesn't leak memory.

### DIFF
--- a/stdlib/runtime/memoryDebug.gr
+++ b/stdlib/runtime/memoryDebug.gr
@@ -202,6 +202,14 @@ let rec isSimpleObjectType = (tag: Number) => {
   result
 }
 
+@disableGC
+let rec forceDeallocate = x => {
+  Memory.decRef(WasmI32.fromGrain(x))
+  Memory.decRef(WasmI32.fromGrain(x))
+  Memory.decRef(WasmI32.fromGrain(forceDeallocate))
+  void
+}
+
 let leakDebugBlockImpl = (options: LeakDebugBlockOptions, f: () -> Void) => {
   // The objective of this function is to find and report any memory leaks in
   // the function f to be examined.
@@ -579,6 +587,14 @@ let leakDebugBlockImpl = (options: LeakDebugBlockOptions, f: () -> Void) => {
   f()
 
   GC.removeDebugHooks()
+
+  // The above hook functions don't get deallocated because @disableGC applies not
+  // just to their internals, but also to the lifetime of the closures allocated
+  // locally here, so they need to be cleaned up manually.
+  forceDeallocate(incRefHook)
+  forceDeallocate(decRefHook)
+  forceDeallocate(mallocHook)
+  forceDeallocate(freeHook)
 
   // Determine objects that existed before executing the function that have been
   // messed up with unmaching incRefs/decRefs.


stdlib/runtime/gc.gr
```diff
@@ -172,11 +283,25 @@ let rec decRef = (userPtr: WasmI32, ignoreZeros: Bool) => {
...
       if (ignoreZeros) {
         userPtr
       } else {
-        throw DecRefError
+        // Throwing causes last logged print statements to not be printed,
+        // so it becomes very difficult to debug memory problems. Maybe a
+        // buffer is not being flushed? If the debugging hooks are active then
+        // behave ad if ignoreZeros was true to continue executing and give the
+        // debugging hooks an opportunity to report the issue.
+        if (WasmI32.ne(_DEBUG_HOOK_DECREF, 0n)) {
```
peblair, line 300: Very interesting. I wonder if we could get around this by adding a call to the flush syscall in the `if (WasmI32.ne(_DEBUG_HOOK_DECREF, 0n))` expression? 🤔

stdlib/runtime/gc.gr
```diff
@@ -250,3 +375,35 @@ export let decRef = (userPtr) => decRef(userPtr, false)
...
+  onMalloc: (WasmI32 -> Void),
+  onFree: (WasmI32 -> Void)
+) => {
+  incRefNoDebug(WasmI32.fromGrain(onIncRef))
+  incRefNoDebug(WasmI32.fromGrain(onDecRef))
+  incRefNoDebug(WasmI32.fromGrain(onMalloc))
+  incRefNoDebug(WasmI32.fromGrain(onFree))
+
+  // decRef(WasmI32.fromGrain(f))
+
```
peblair, line 391:
```suggestion
```

stdlib/runtime/runtimeDebug.gr
```diff
@@ -0,0 +1,712 @@
...
+
+let leakDebugBlockImpl = (options: LeakDebugBlockOptions, f: () -> Void) => {
+  // The main objective is to find and print objects allocated
+  // in the function, but not deallocated at the end.
+
+  let allocations = Set.make(): (Set.Set<Number>)
+
+  let tagToTypeName = (tag: Number) => {
+    match (tag) {
+      1 => "string",
```
peblair, line 71: Maybe we should use the constants from `runtime/unsafe/tags`?

stdlib/runtime/runtimeDebug.gr
```diff
@@ -0,0 +1,712 @@
...
+
+  let incRefsByTag = Map.make(): (Map.Map<Number, Number>)
+  let decRefsByTag = Map.make(): (Map.Map<Number, Number>)
+
+  let inintialRefCountsOnPreExistingObjects = Map.make(): (Map.Map<Number, Number>)
+  let incRefsOnPreExistingObjects = Map.make(): (Map.Map<Number, Number>)
+  let decRefsOnPreExistingObjects = Map.make(): (Map.Map<Number, Number>)
+
+  // Since objects created by GC.malloc have a starting refcount of 1,
+  // the total incRefsByTag would not match decRefsByTag. I'm compensating it
```
peblair, line 99:
```suggestion
  // the total incRefsByTag would not match decRefsByTag. We compensate for this
```
minor nit

stdlib/runtime/runtimeDebug.gr
```diff
@@ -0,0 +1,712 @@
...
+  // by keeping track of number of objects that have gone malloc+free cycle.
```
peblair, line 100:
```suggestion
  // by keeping track of number of objects that have gone through the malloc+free cycle.
```

stdlib/runtime/runtimeDebug.gr
```diff
@@ -0,0 +1,712 @@
...
+  // This isn't perfect because in case of leaks free isn't called and the
```
peblair, line 101:
```suggestion
  // This isn't perfect because, in the case of leaks, free isn't called and the
```

stdlib/runtime/runtimeDebug.gr
```diff
@@ -0,0 +1,712 @@
...
+  // totals can create confusion. The simplest solution would be to get the tag
+  // of mallocated objects, but they don't yet have a tag associated. So I
+  // have to compensate again by adding up the objects that have been
+  // mallocated but never freed. Which simply means the ones in "allocations"
+  // Map.
```
peblair, line 106: This part loses me a little. Can you maybe elaborate a bit?

stdlib/runtime/runtimeDebug.gr
```diff
@@ -0,0 +1,712 @@
...
+  let mallocPlusFreeByTag = Map.make(): (Map.Map<Number, Number>)
+
+  @disableGC
+  let rec readTag = (userPtrAsNumber: Number) => {
```
peblair, line 110: I can imagine why this is the case, but it may be worth a comment explaining why these functions use `Number` instead of `WasmI32`.

stdlib/runtime/runtimeDebug.gr
```diff
@@ -0,0 +1,712 @@
...
+  }
+
+  let onMalloc = (userPtr: Number) => {
+    totalMallocs += 1
+
+    // Tracking objects allocated in the debug block.
+    if (Set.contains(userPtr, allocations)) {
+      // A new allocation, but for an object already allocated inside the debug block, but never removed.
+      print(
+        "⚠ PROBLEM DETECTED: unexpected malloc at address 0x" ++
```
peblair, line 339: Sort of wish we had colored output here hehe
(not _really_ requesting this, but just a thought)

stdlib/runtime/runtimeDebug.gr
```diff
@@ -0,0 +1,712 @@
...
+          " " ++
+          tagToTypeName(tag) ++
+          ")" ++
+          maybeObjectContentsAppendix(userPtr),
+        )
+      }
+    }
+
+    if (Map.contains(userPtr, incRefsOnPreExistingObjects) ||
+    Map.contains(userPtr, decRefsOnPreExistingObjects)) {
```
peblair, line 394: Is this the formatter output? If so, cc @marcusroberts (I think that `Map` should be indented here)

stdlib/runtime/runtimeDebug.gr
```diff
@@ -0,0 +1,712 @@
...
+  let incRefHook = (userPtr: WasmI32) => {
+    Memory.incRef(WasmI32.fromGrain(Conv.wasmI32ToNumber))
+    let ptrAsNumber = Conv.wasmI32ToNumber(userPtr)
+
+    // Skip the incRef made when calling the actual function to test.
+    if (WasmI32.ne(userPtr, WasmI32.fromGrain(f)) &&
+    WasmI32.ne(userPtr, WasmI32.fromGrain(GC.removeDebugHooks))) {
+      Memory.incRef(WasmI32.fromGrain(onIncRef))
+      onIncRef(ptrAsNumber)
+    }
```
peblair, line 434: We don't need a `decRef(<self>)` postamble at the end of these functions?

peblair (commented):
I finally had some time to give this a look. @cician this is incredible stuff! I've left some first-pass comments, and I'm curious to see what the other core team members have to say.

stdlib/runtime/runtimeDebug.gr
cician: I didn't use them to avoid `@disableGC`. Should I convert `tagToTypeName` to no gc?

peblair: I hadn't considered that! In that case, I think a comment is fine for now (maybe also add a comment to `tags.gr` so we can remember to keep this in sync?).

stdlib/runtime/runtimeDebug.gr
cician: Yes, it is.

stdlib/runtime/gc.gr
```diff
@@ -45,6 +45,12 @@ export let utoa32Buffered = box((a: WasmI32, b: WasmI32, c: WasmI32) => void)
 
 let mut _DEBUG = false
 
+let mut _DEBUG_HOOK_INCREF = 0n
+let mut _DEBUG_HOOK_DECREF = 0n
+let mut _DEBUG_HOOK_MALLOC = 0n
+let mut _DEBUG_HOOK_FREE = 0n
+let mut _DEBUG_HOOK_RUNNING = false
```
ospencer, line 52: It's not obvious to me that the `_DEBUG` value above doesn't have anything to do with these values. I wonder if there's any renaming we can do to make that more clear.

stdlib/runtime/runtimeDebug.gr
```diff
@@ -0,0 +1,712 @@
+import WasmI32 from "runtime/unsafe/wasmi32"
```
ospencer, line 1: This filename is a little confusing vs the current `runtime/unsafe/debug` file. We can't combine them because `runtime/debug` needs to run in `runtime` mode—maybe we rename this file to `memoryDebug.gr` or something similar?

ospencer (commented):
Very cool! And awesome that we've already seen this in action 🙂 

A couple of thoughts from me but no big blockers :)

One thing to consider— @phated when we split the runtime from the stdlib, will this be annoying to refactor? I think this makes sense to live in the runtime, but since it uses stdlibs it might be hard to break out, and I'm not sure where it really should live then. I feel like we can worry about it when we actually make that change, but figured I'd grab your thoughts.

stdlib/runtime/runtimeDebug.gr
cician: I agree this comment is pretty confusing. I'll rewrite it.

What I mean is that to correctly present the summary of incRefs/decRefs by tag, it's not enough to just track incRefs/decRefs themselves because incRef doesn't get called for newly allocated objects. They start with ref. count of 1 set directly in `GC.malloc`. But it also isn't as simple as adding some code in the hook for `GC.malloc` because freshly allocated objects don't have a tag yet. The tag is normally set after (like in `dataStructures.allocateString` and similar functions). So what I actually do is track the lifetime of single objects in the `allocations` map. Only when the object is freed I add to the `mallocPlusFreeByTag` map to present along with incRefs/decRefs by tag. If an object is not freed then it's left in the `allocations` map. So I group these leaked objects by tag as well and present the totals in the summary as "malloc without free".

stdlib/runtime/gc.gr
cician: Or maybe just move it to the commented code below? Currently it's never used, its for the debugging functions that are all commented out. I agree though that leak checker code can get mixed up with these.

stdlib/runtime/runtimeDebug.gr
cician: `memoryDebug.gr` sounds good to me.

stdlib/runtime/runtimeDebug.gr
cician: You're right. Also for some reason I was incRefing them twice from the GC module.

Since I settled for the hook functions to accept pointers as WasmI32, I think it actually makes more sense now to remove the incRefs on them rather than having them decRef themselves.

If the hook functions were to receive Numbers or Int32s then of course it would make sense for them to use proper GC conventions.

cician:
I checked if the leak checker leaks memory, and it does.

At least in part it's due to #1003 and #773 (these functions are recursive in order to decRef themselves). I can work around #773 by moving some utility functions to the top level, but its difficult to diagnose further until #1003 is fixed.

stdlib/runtime/runtimeDebug.gr
cician: I changed this function to use `@disableGC`. It's not pretty, but probably still better than not using the constants.

stdlib/runtime/runtimeDebug.gr
cician: I've added a comment explaining the choice for Number.

stdlib/runtime/runtimeDebug.gr
cician: Should I move it outside of the runtime folder as well?

cician:
Probably not a big deal, but the leak checker currently reports leaks when converting numbers to strings due to buffers allocated and intentionally never released in the numberUtils module (_POWERS10, _DIGITS, _HEX_DIGITS etc).

It can be a bit confusing when you get a resulting like this:
```
⚠ 1 malloc/free leaks detected:
- 0x3e9c80 (ref. count: 1; tag 825241648 unknown): <unknown heap tag type: 0x31303030 | value: 0x3e9c80>
```

One solution would be to simply call a few functions preemptively to force initialize these buffers, but maybe we want to give them a proper reserved heap tag? Something like "cache" or "lookup table". Otherwise, if it used malloc directly instead of `Memory.malloc` it wouldn't get reported, but I don't know if we want that.

cician:
> I checked if the leak checker leaks memory, and it does.
> 
> At least in part it's due to #1003 and #773 (these functions are recursive in order to decRef themselves). I can work around #773 by moving some utility functions to the top level, but its difficult to diagnose further until #1003 is fixed.

With #1003 fixed and a fix in the checker itself it's now mostly leak free. There are still leaks elsewhere in the stdlib that make it leak indirectly when working on Sets, Maps and Arrays to print the summary.

If you're wondering how I leak check the leak checker, I'm just running it in an infinite loop and watching if the memory continues to rise indefinitely in the task manager.

ospencer:
We'll want to land https://github.com/grain-lang/grain/issues/1410 before we land this to minimize the effect on compiled modules.

PR #988, opened 2021-10-19
fix(compiler): Properly handle value restriction on function application & mutable vars

### LINKED ISSUE #990: Typechecker: let-mut doesn't get proper value restriction
While writing regression tests for #988, it was discovered that `let mut` doesn't get properly value restricted when you assign a value to it.

Here is the smallest reproduction I could find:

```grain
enum Baz<a> { Foo(a), Bar }
let mut a = Bar

a = Foo(1)
a = Foo("a")
```

EduardoRFS:
## Context

Currently Grain value restriction is broken, as may be seen by the reproduction below.

This happens because `generalize_expansive` is not called at https://github.com/grain-lang/grain/blob/main/compiler/src/typed/typecore.re#L1943. `generalize_expansive` is not called because `is_nonexpansive` is returning `true` for an expansive operation(applying a function).

And `is_nonexpansive` returns true because `TExpApp` is possibly a non expansive operation, which is wrong as in Grain any function call can trigger a side effect. This can be checked at https://github.com/grain-lang/grain/blob/main/compiler/src/typed/typecore.re#L363.

## History

This bug is probably exists because in OCaml not all applications are expansive, applying to a function without applying the first parameter in the case of named parameters is not an expansive operation and it will never trigger side effect.

```reason
/* example of the behavior above */
let f = (~x, ~y, x) => x;
let a = f(~y=1);
/* can always be replaced by */
let a = (~x) => f(~x, ~y=1);
```

So Grain copied the code from https://github.com/ocaml/ocaml/blob/4.04/typing/typecore.ml#L1566, but actually normal Texp_apply was hitting the catchall case below https://github.com/ocaml/ocaml/blob/4.04/typing/typecore.ml#L1618.

This can be more easily noted in modern OCaml at https://github.com/ocaml/ocaml/blob/4.12/typing/typecore.ml#L2154.

## Glossary

- expansive can be read as "a side effect may happens", so non expansive means a pure operation

## Warning

I'm not sure this is the only source of bugs in the value restriction, but it's definitely the most obvious one.

## Reproduction

```reason
import String from "string"

let f = () => {
  let mut x_box = None
  let looks_id = x => {
    let old = match (x_box) {
      None => x,
      Some(x) => x,
    }
    x_box = Some(x)
    old
  }
  (looks_id, 1)
}
let (id, _) = f()
let a = id(97)
let b = id("a")

print(String.concat("b", b))
```

Fixes #990

### DIFF
--- a/compiler/src/typed/typecore.re
+++ b/compiler/src/typed/typecore.re
@@ -350,7 +350,7 @@ let rec is_nonexpansive = exp =>
   | TExpLambda(_)
   | TExpNull => true
   | TExpTuple(es) => List.for_all(is_nonexpansive, es)
-  | TExpLet(rec_flag, mut_flag, binds) =>
+  | TExpLet(rec_flag, Immutable, binds) =>
     List.for_all(vb => is_nonexpansive(vb.vb_expr), binds)
   | TExpMatch(e, cases, _) =>
     is_nonexpansive(e)


phated:
Had to push a fix to our GC to avoid a weak type error in our GC's boxed functions that are re-assigned in other modules to handle cyclical dep problems.

phated:
I've opened #989 about the test failures that were happening and removed the 2 chunks of code that made the test suite fail (correctly!) due to this fix.

I think we should review and merge this by itself and then open an issue to investigate the mutable variable problem.

ospencer:
@phated I added a commit that fixes #990. If it makes more sense to do it as a separate PR we can do that too, but I thought I made sense to add it here.

ospencer:
I don't see why not—we can double-check to see if anyone needs anything else in, but otherwise, I say let's release it tomorrow morning!

marcusroberts:
> I don't see why not—we can double-check to see if anyone needs anything else in, but otherwise, I say let's release it tomorrow morning!

I have one small formatter fix inbound it would be great to get in, otherwise let's do it!

Issue #994, opened 2021-10-21
Linker: Hashtbl not_found error on some imports without use


phated:
When I `import *` from a module, but then don't use any of the values, the linker crashes with a `Hashtbl.find` not_found.

Here's the minimal repo:
> foo.gr
```grain
export let foo = 1
```
> main.gr
```grain
import * from "./foo"
let bar = 2
```

Running `grain main.gr` against this will crash.

I'm considering this a blocker on type aliases because I defined a file called `aliases.gr` that contained only aliases and then the linker crashed when I imported and used them with `import * from "./aliases"`

PR #995, opened 2021-10-21
feat(compiler): Type aliases

### LINKED ISSUE #269: Compiler: Type Aliases
This adds the ability to create name aliases for types. The main thoughts right now are to use one of three keywords:

```grain
data alias Foo = Bar
```
```grain
alias Foo = Bar
```
```grain
type Foo = Bar
```

phated:
Turns out that we can't share push access on #991 because I opened the PR based on @EduardoRFS's fork, so I'm re-re-recreating this with a branch on the main org.

This rolls back some of the debugging from #512 and applies a 1-line patch by @EduardoRFS (absolutely amazing debugging stream ‼️ ❤️ 🙇)

It's possible that the middle_end code needs to be fixed and we need to add a bunch more tests.

Closes #512 
Closes #269

### DIFF
--- a/compiler/src/typed/typemod.re
+++ b/compiler/src/typed/typemod.re
@@ -508,8 +508,8 @@ let type_module = (~toplevel=false, funct_body, anchor, env, sstr /*scope*/) => 
               {
                 ...info.data_type,
                 type_kind: TDataAbstract,
-                // Removing the manifest makes this type unable to be
-                // used as the aliased type.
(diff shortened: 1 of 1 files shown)


phated:
Additional tests currently blocked on #994

phated:
Another note: @ospencer wants type aliases not `export`ed to become fully abstract to consumers. This isn't the case as the code currently exists.

PR #1012, opened 2021-10-29
feat(stdlib): Add sort function to Array module (#125)

timgestson:
Implement Array.sort via the quicksort algorithm as talked about here: #125

### DIFF
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -857,6 +857,14 @@ export let slice = (startIndex, endIndex, array) => {
   }
 }
 
+/**
+ * Sorts an array in-place based on the quicksort algorithm.
+ * 
+ * @param comp: The comparative function that we are using to sort
+ * @param array: The array to be sorted
+ * @returns Void
+ */
+
 export let sort = (comp, array) => {
   let partition = (low, high) => {
     let pivot = array[high]
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -860,8 +860,8 @@ export let slice = (startIndex, endIndex, array) => {
 /**
  * Sorts an array in-place using the quicksort algorithm.
  * 
- * @param comp: The function that takes two arguments and returns true if the first argument preceeds the second
- * @param array: The array to be sorted
+ * @param comp: (a, a) -> Number - A function that takes two arguments. It must return 0 if its arguments compare as equal, a positive number if the first is greater, and a negative number if the first is smaller
+ * @param array: Array<a> - The array to be sorted
  * @returns void
  */
 export let sort = (comp, array) => {
@@ -869,7 +869,7 @@ export let sort = (comp, array) => {
     let pivot = array[high]
     let mut i = low - 1
     for(let mut j = low; j < high; j += 1) {
-      if(comp(array[j], pivot)){
+      if(comp(array[j], pivot) < 0){
         i += 1
         let temp = array[i]
         set(i, array[j], array)
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -856,3 +856,31 @@ export let slice = (startIndex, endIndex, array) => {
     init(newLength, n => array[startIndex + n])
   }
 }
+
+export let sort = (comp, array) => {
+  let partition = (low, high) => {
+    let pivot = array[high]
+    let mut i = low - 1
+    for(let mut j = low; j < high; j += 1) {
+      if(comp(array[j], pivot)){
+        i += 1
+        let temp = array[i]
+        set(i, array[j], array)
+        set(j, temp, array)
+      }
+    }
+    let temp = array[i+1]
+    set(i+1, array[high], array)
+    set(high, temp, array)
+    i + 1
+  }
+  let rec quicksort = (low, high) => {
+    if(low < high){
+      let partition_index = partition(low, high)
+      quicksort(partition_index + 1, high)
+      quicksort(low, partition_index - 1)
+    }
+  }
+  let len = length(array)
+  quicksort(0, len - 1)
+}
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -858,13 +858,12 @@ export let slice = (startIndex, endIndex, array) => {
 }
 
 /**
- * Sorts an array in-place based on the quicksort algorithm.
+ * Sorts an array in-place using the quicksort algorithm.
  * 
- * @param comp: The comparative function that we are using to sort
+ * @param comp: The function that takes two arguments and returns true if the first argument preceeds the second
  * @param array: The array to be sorted
- * @returns Void
+ * @returns void
  */
-
 export let sort = (comp, array) => {
   let partition = (low, high) => {
     let pivot = array[high]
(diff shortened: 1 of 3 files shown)


marcusroberts (commented):
Thanks for this. Could you add some tests for non-numeric searches so that we have some sample code for people learning how to write a comparator function?

Great contribution though, cheers!

ospencer (changes requested):
Looks great overall @timgestson! Only thing: the signature for the `compare` function should be `(a, a) -> Number` rather than `(a, a) -> Bool`, returning positive if the first value is larger, negative for the second, and `0` if they're equal.

This actually reminded me of why I had originally written `sortBy` in the issue—ideally we'd implement `Pervasives.compare`, a generic comparator, and then we could then have `Array.sort: Array<a> -> Void` using `Pervasives.compare`, and `Array.sortBy: (Comparator<a>, Array<a>) -> Void` which allows for custom comparators.

@grain-lang/core thoughts?

phated:
@ospencer I've always felt like OCaml's polymorphic compare has felt like magic. It's not really easy to understand how it acts with different types of data. I think I'd rather that the sort function be explicit, like implemented here. 🤔

ospencer:
I feel like it'd keep with the spirit of keeping things simple! Especially for simple types (lists/arrays of strings, numbers), it feels really natural to just do `Array.sort(arr)` versus having to understand that there are different compare functions for all of the types, and having to do `Array.sort(String.compare, arr)`.

I agree that it can feel like magic, certainly for variants and records. I think we can handle that through good documentation that explains exactly how it works, and when you should write your own comparator. WDYT?

phated:
@ospencer I really like the `Array.sort(String.compare, arr)` example! This allows us to be explicit with the comparators that we provide in the standard library—and choose not to provide others, like comparators for enums or records, since they often need a custom comparator anyway.

The thing that I really like with this approach is the refactoring story. For example:
```grain
let arr = [> "foo", "bar"]
Array.sort(String.compare, arr)
```
But then if I refactor to using Option, I'll get a type error and be alerted to the need of updating my sort:
```grain
let arr = [> None, Some("foo"), None, Some("bar")]
Array.sort(String.compare, arr) // <- I'll get a type error, thus avoiding a potential logic bug
```

You don't get this type checker warning if sort used a polymorphic compare.

stdlib/array.gr
```diff
@@ -871,3 +871,38 @@ export let slice = (startIndex, endIndex, array) => {
     init(newLength, n => array[startIndex + n])
   }
 }
+
+/**
+ * Sorts an array in-place using the quicksort algorithm.
```
phated, line 876: Question for the team. Do we want to specify which algorithm is used in the description of this function?

stdlib/array.gr
```diff
@@ -871,3 +871,38 @@ export let slice = (startIndex, endIndex, array) => {
...
+ * 
+ * @param comp: (a, a) -> Number - A function that takes two arguments. It must return 0 if its arguments compare as equal, a positive number if the first is greater, and a negative number if the first is smaller
```
phated, line 878:
```suggestion
 * @param comp: The comparator function used to indicate sort order
```
Our graindoc tool actually uses the inferred type, so we don't specify it in the docs. Also, I think the description of the function should discuss the comparator functions instead of the argument.

stdlib/array.gr
```diff
@@ -871,3 +871,38 @@ export let slice = (startIndex, endIndex, array) => {
     init(newLength, n => array[startIndex + n])
   }
 }
+
+/**
+ * Sorts an array in-place using the quicksort algorithm.
+ *
```
phated, line 877:
```suggestion
 *
 * Ordering is calculated using a comparator function—a function which takes two array elements and must return 0 if both are equal, a positive number if the first is greater, and a negative number if the first is smaller.
```

stdlib/array.gr
```diff
@@ -871,3 +871,38 @@ export let slice = (startIndex, endIndex, array) => {
     init(newLength, n => array[startIndex + n])
   }
 }
+
+/**
+ * Sorts an array in-place using the quicksort algorithm.
+ * 
+ * @param comp: (a, a) -> Number - A function that takes two arguments. It must return 0 if its arguments compare as equal, a positive number if the first is greater, and a negative number if the first is smaller
+ * @param array: Array<a> - The array to be sorted
```
phated, line 879:
```suggestion
 * @param array: The array to be sorted
```
Type isn't needed

stdlib/array.gr
```diff
@@ -871,3 +871,38 @@ export let slice = (startIndex, endIndex, array) => {
...
+ * @returns void
```
phated, line 880:
```suggestion
```
Returns isn't needed if there is none.

stdlib/array.gr
```diff
@@ -871,3 +871,38 @@ export let slice = (startIndex, endIndex, array) => {
...
+ */
```
phated, line 881:
```suggestion
 * @since v0.4.5
 */
```
Hopefully this will be released with v0.4.5!

stdlib/array.gr
```diff
@@ -871,3 +871,38 @@ export let slice = (startIndex, endIndex, array) => {
...
+export let sort = (comp, array) => {
+  let partition = (low, high) => {
+    let pivot = array[high]
+    let mut i = low - 1
+    for(let mut j = low; j < high; j += 1) {
+      if(comp(array[j], pivot) < 0){
+        i += 1
+        let temp = array[i]
+        set(i, array[j], array)
+        set(j, temp, array)
```
phated, line 891:
```suggestion
        array[i] = array[j]
        array[j] = temp
```
We can assign into the arrays

stdlib/array.gr
```diff
@@ -871,3 +871,38 @@ export let slice = (startIndex, endIndex, array) => {
...
+      }
+    }
+    let temp = array[i + 1]
+    set(i + 1, array[high], array)
+    set(high, temp, array)
```
phated, line 896:
```suggestion
    array[i + 1] = array[high]
    array[high] = temp
```

stdlib/array.gr
```diff
@@ -871,3 +871,38 @@ export let slice = (startIndex, endIndex, array) => {
...
+    i + 1
+  }
+  let rec quicksort = (low, high) => {
+    if(low < high){
+      let partition_index = partition(low, high)
+      quicksort(partition_index + 1, high)
+      quicksort(low, partition_index - 1)
```
phated, line 903:
```suggestion
      let partitionIndex = partition(low, high)
      quicksort(partitionIndex + 1, high)
      quicksort(low, partitionIndex - 1)
```
Small style nit—we generally don't use snake_case in our grain code.

phated (changes requested):
This is looking awesome! Had a bit of feedback.

stdlib/array.gr
ospencer: Not here, at least.

phated (approved):
Excellent! Thanks!

ospencer (approved):
Fantastic! Thank you!

phated:
I rebased and regenerated the array docs. I'll merge this as soon as the tests finish running.

PR #1014, opened 2021-10-30
feat(stdlib): List.sort via a mergesort (#125)

timgestson:
Implement List.sort with a mergesort algorithm.  Result returns a new list.

### DIFF
--- a/compiler/test/stdlib/list.test.gr
+++ b/compiler/test/stdlib/list.test.gr
@@ -1,4 +1,5 @@
 import * from "list"
+import String from "string"
 
 let list = [1, 2, 3]
 
@@ -231,3 +232,12 @@ let comp = (left, right) => {
 }
 assert sort(comp, [3,5,2,4,1]) == [1,2,3,4,5]
 assert sort(comp, []) == []
+let comp = (left, right) => {
+    match((String.length(left), String.length(right))) {
+        (left, right) when left > right => 1,
+        (left, right) when left == right => 0,
+        _ => -1
+    }
+}
+let list = ["a", "abcde", "abc", "ab", "abcd", "a"]
+assert sort(comp, list) == ["a", "a", "ab", "abc", "abcd", "abcde"]
--- a/compiler/test/stdlib/list.test.gr
+++ b/compiler/test/stdlib/list.test.gr
@@ -241,3 +241,4 @@ let comp = (left, right) => {
 }
 let list = ["a", "abcde", "abc", "ab", "abcd", "a"]
 assert sort(comp, list) == ["a", "a", "ab", "abc", "abcd", "abcde"]
+assert sort(comp, ["a", "a", "a", "a"]) == ["a", "a", "a", "a"]
--- a/compiler/test/stdlib/list.test.gr
+++ b/compiler/test/stdlib/list.test.gr
@@ -219,3 +219,15 @@ assert init(-10, idx => fail "Shouldn't be called") == []
 // List.join
 assert join(", ", ["a", "b", "c"]) == "a, b, c"
 assert join(", ", []) == ""
+
+
+// List.sort
+let comp = (left, right) => {
+    match((left, right)) {
+        (left, right) when left > right => 1,
+        (left, right) when left == right => 0,
+        _ => -1
+    }
+}
+assert sort(comp, [3,5,2,4,1]) == [1,2,3,4,5]
+assert sort(comp, []) == []
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -331,11 +331,12 @@ let join = (separator: String, items: List<String>) => {
 }
 
 /**
- * Sorts the given list based on a given comparator function via a mergesort algorithm. The resulting list is sorted in increasing order.
+ * Sorts the given list based on a given comparator function. The resulting list is sorted in increasing order.
  * 
- * @param comp: (a, a) -> Number - A function that takes two arguments. It must return 0 if its arguments compare as equal, a positive Number if the first is greater, and a negative Number if the first is smaller
- * @param list: List<a> - The list to be sorted
- * @returns List<a>
+ * Ordering is calculated using a comparator function which takes two list elements and must return 0 if both are equal, a positive number if the first is greater, and a negative number if the first is smaller.
+ * @param comp: The comparator function used to indicate sort order
+ * @param list: The list to be sorted
+ * @since v0.4.5
  */
 let sort = (comp, list) => { 
   let rec merge = (left, right, list) => {
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -1,5 +1,4 @@
 // Standard library for list functionality
-
 export *
 
 let init = (length, fn) => {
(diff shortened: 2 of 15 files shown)


timgestson:
hmm, not quite sure what the failed test here is related to, but can dig deeper

spotandjake:
From what i can tell it looks like this test [here](https://github.com/grain-lang/grain/blob/471df49b2e31a64d08e38bd2ec61744235f14adf/compiler/test/suites/stdlib.re#L12) failed: I think I failed because it looks like they are doing snapshot testing on the generated code. Not 100% sure though hope that helps you, and hopefully a team member can confirm. maybe @ospencer can confirm.

phated:
@spotandjake is correct. Sometimes changes in the stdlib cause snapshots to be updated. You can run the tests with -u to update them

stdlib/list.gr
```diff
@@ -329,3 +329,42 @@ let join = (separator: String, items: List<String>) => {
     Some(s) => s,
   }
 }
+
+/**
+ * Sorts the given list based on a given comparator function via a mergesort algorithm. The resulting list is sorted in increasing order.
+ * 
+ * @param comp: (a, a) -> Number - A function that takes two arguments. It must return 0 if its arguments compare as equal, a positive Number if the first is greater, and a negative Number if the first is smaller
+ * @param list: List<a> - The list to be sorted
+ * @returns List<a>
```
phated, line 338: @timgestson can you update these the same way that you updated Array.sort? 🙏

compiler/test/stdlib/list.test.gr
```diff
@@ -219,3 +219,15 @@ assert init(-10, idx => fail "Shouldn't be called") == []
...
+// List.sort
+let comp = (left, right) => {
+    match((left, right)) {
+        (left, right) when left > right => 1,
+        (left, right) when left == right => 0,
+        _ => -1
+    }
+}
+assert sort(comp, [3,5,2,4,1]) == [1,2,3,4,5]
+assert sort(comp, []) == []
```
phated, line 233: Do we want any other tests, like the strings test in Array.sort?

phated (commented):
A couple of things

compiler/test/stdlib/list.test.gr
```diff
@@ -219,3 +220,24 @@ assert init(-10, idx => fail "Shouldn't be called") == []
...
+let comp = (left, right) => {
+    match((String.length(left), String.length(right))) {
+        (left, right) when left > right => 1,
+        (left, right) when left == right => 0,
+        _ => -1
+    }
+}
+let list = ["a", "abcde", "abc", "ab", "abcd", "a"]
+assert sort(comp, list) == ["a", "a", "ab", "abc", "abcd", "abcde"]
```
marcusroberts, line 243: Looks good!  Just nitpicking, Could we test a couple of edge cases - an empty list, a list with values all the same etc?

timgestson: What do you have in mind?  There is an empty list test on line 234 and this test has duplicate items in it ("a").  Will be happy to add more if you have a specific case in mind

timgestson: I'll add 1 more with all identical items

ospencer (commented):
This looks great! I have two requests of you—could you add a `List.revAppend` (though check with @phated for naming, he might prefer `reverseAppend`) so we can avoid the `reverse(append(` calls? That'd make this a bit more efficient.

The second one is a bit of a bigger ask—if you take a look at OCaml's implementation, they have a neat trick to avoid the calls to `part`—they pass offsets around. That's a bit more complicated though, so it's fine if we want to hold off on that part and open an issue to implement that 🙂

timgestson:
Yeah let's separate out that into another issue.  I added revAppend for now.  Is this what you are referring to? https://github.com/ocaml/ocaml/blob/trunk/stdlib/list.ml#L340

stdlib/list.gr
```diff
@@ -329,3 +329,54 @@ let join = (separator: String, items: List<String>) => {
     Some(s) => s,
   }
 }
+
+/**
+ * Reverses the list and appends items to the end
```
ospencer, line 334:
```suggestion
 * Reverses the first list and appends the second list to the end.
```

@phated what do you think about mentioning that it's for efficiency? OCaml describes it as
```
`rev_append l1 l2` reverses `l1` and concatenates it with `l2`. This is equivalent to `(List.rev l1) @ l2`, but `rev_append` is tail-recursive and more efficient.
```

stdlib/list.gr
```diff
@@ -329,3 +329,54 @@ let join = (separator: String, items: List<String>) => {
...
+ * 
+ * @param items: Items to be appended to the reversed list
+ * @param list: The list that is being reversed and appended to
+ * @since v0.4.5
+ */
+let revAppend = (items, list) => { 
+  append(reverse(list), items)
+}
```
ospencer, line 342: Sorry if I was a bit unclear. I wanted this so we could have an efficient implementation. Usually the main list goes last, but for `append`-like functions it goes in the order they get appended:

```suggestion
 * @param list1: The list to reverse
 * @param list2: The list to append
 * @since v0.4.5
 */
let rec revAppend = (list1, list2) => { 
  match (list1) {
    [hd, ...tl] => revAppend(tl, [hd, ...list2]),
    [] => list2
  }
}
```

We'll have to update the call sites to reverse the arguments as well.

ospencer:
> Yeah let's separate out that into another issue. I added revAppend for now. Is this what you are referring to? https://github.com/ocaml/ocaml/blob/trunk/stdlib/list.ml#L340

Yeah, exactly that! But no worries, we can make it a separate issue 🙂

stdlib/list.gr
```diff
@@ -329,3 +329,57 @@ let join = (separator: String, items: List<String>) => {
...
 }
+
+/**
+ * Reverses the first list and appends the second list to the end.
+ * 
+ * @param list1: The list to reverse
+ * @param list2: The list to append
+ * @since v0.4.5
+ */
+let rec revAppend = (list1, list2) => {
```
ospencer, line 340: @phated Did you have any thoughts on `revAppend` vs `reverseAppend`? I personally like `revAppend` but `reverseAppend` might be more clear. (But I really like `revAppend`.)

ospencer (approved):
Thanks @timgestson for making this happen and your patience with us! 🙇

stdlib/list.gr
phated: No opinion really. You know I like verbosity for clarity, but I don't think `rev` gets confused with anything else (can you think of anything?). If you are **really** like `revAppend` then let's stick with it for now.

ospencer: Works for me!

phated (approved):
Approved! (pending @ospencer doesn't come up with something `rev` can be confused with and we regenerate the docs in this PR).

ospencer:
We realized that we don't want to generate the docs just yet! And I looked up all the words that start with `rev`, and I don't think any of them will get confused with reverse. 🚀

PR #1016, opened 2021-11-01
chore(compiler): Update vscode script to use a custom template on non-Windows

phated:
In vscode-ocaml-platform v1.0.0, they added a configuration option to specify a "custom" template for finding the ocamllsp in your sandbox. Before this, we needed to have a special script that found the absolute path to `grain/compiler` and labeled that as the esy sandbox. However, with this "custom" template, we can actually defer the lookup to `yarn` and our local `esy` command.

I've removed the "vscode" script and the file it executed and then committed a workspace `settings.json` file that implements the custom command that I got working. We need to test this on windows still 😭

cc @cician - Can you try this out in your environments (without a global esy installed)?

### DIFF
--- a/.vscode/setup-ocaml-lsp.js
+++ b/.vscode/setup-ocaml-lsp.js
@@ -21,7 +21,7 @@ if (process.platform === 'win32') {
   console.log("NOTE: It looks like you are using Windows. You'll need to install esy globally using `npm i -g esy`");
 
   output[ocamlsandbox] = {
-    "root": "${compilerPath}",
+    "root": `${compilerPath}`,
     "kind": "esy"
   };
 } else {
--- a/.vscode/setup-ocaml-lsp.js
+++ b/.vscode/setup-ocaml-lsp.js
@@ -3,16 +3,32 @@ const path = require('path');
 
 const settingsPath = path.join(__dirname, 'settings.json');
 
-const compilerPath = path.join(__dirname, '../compiler');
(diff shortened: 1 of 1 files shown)


phated:
> Should this be a chore?

I mean....probably? But **all** of my work is always a chore and it makes me sad.

ospencer:
😭

marcusroberts (approved):
Nice fix!

phated:
It seems that this breaks ocamllsp support in Windows because extensions don't run in the cwd of the workspace.

phated:
And they don't seem to care: https://github.com/microsoft/vscode/issues/43237

phated:
Looks like we still need the script. I also opened an issue on the vscode extension: https://github.com/ocamllabs/vscode-ocaml-platform/issues/760

phated:
While not the most ideal situation on Windows, at least it works (and running the script notifies you to install esy globally).

ospencer (approved):
Excellent!

phated:
> While that is a little clunky, your situation isn't all that common so I deem this workaround sufficient.
> 
> I'd prefer not to create an additional layer of knobs for customizing binary locations. Especially if this layer is going to be only usable for exactly 3 binaries inside one extension. Therefore, I will reject this feature request. Sorry about that.

As expected, they are telling us that our script here is the extension actually "working as intended", so we're screwed and just need to ship this. Pretty lame.

PR #1019, opened 2021-11-05
fix(compiler): Allow modules to re-export imported types containing generics

### LINKED ISSUE #665: Can't reference Map record from my own types
I stumbled upon this error to use the Map module from Standard Library: "Missing file for module Map".

I initially thought it's just the fact that the Map record isn't exported, but I'm not sure. I tried modifying the Map module to export the record and recompiling the compiler, but I get the same error. I can reference without problems both List and other non pervasive types from the standard library like File.Stats. Creating it with Map.make() does work, but I can't use it in my own type like this:

``` Grain
import List from "list"
import Map from "map"
import Float64 from "float64"

export enum JSON {
	JSONNull,
	JSONBoolean(Bool),
	JSONNumber(Float64),
	JSONString(String),
	JSONArray(List<JSON>),
	JSONObject(Map.Map<String,JSON>)
}
```

phated:
Fixes #665

It seems that when we didn't have a filename for a module, we were using it's identifier as a filepath. However, if we raise `Not_found`, the compiler can generally sort out the type you want.

### DIFF
--- a/compiler/src/typed/env.re
+++ b/compiler/src/typed/env.re
@@ -968,11 +968,11 @@ let rec find_module_descr = (path, filename, env) => {
     try(IdTbl.find_same(id, env.components)) {
     | Not_found =>
       let (_, unit_source, _) = get_unit();
-      let filename = Option.value(~default=Ident.name(id), filename);
-      if (Ident.persistent(id) && !(filename == unit_source)) {
-        find_pers_struct(~loc=Location.dummy_loc, filename).ps_comps;
-      } else {
-        raise(Not_found);
+      switch (filename) {
+      | Some(filename)
+          when Ident.persistent(id) && !(filename == unit_source) =>
+        find_pers_struct(~loc=Location.dummy_loc, filename).ps_comps
+      | _ => raise(Not_found)
       };
     }
   | PExternal(m, s, pos) =>
--- a/compiler/test/suites/records.re
+++ b/compiler/test/suites/records.re
@@ -161,4 +161,13 @@ describe("records", ({test}) => {
       bar.foo = Some(foo)
     |},
   );
+  assertRun(
+    "export_import_record_issue_665",
+    {|
+      import { Foo } from "data"
+      export enum Bar { Baz(Foo<Number>) }
+      print(Baz({ bar: 1 }))
+    |},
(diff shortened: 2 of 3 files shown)


phated:
I'm not sure if this is the best semantic commit description.

compiler/test/suites/records.re
```diff
@@ -161,4 +161,13 @@ describe("records", ({test}) => {
...
     |},
   );
+  assertRun(
+    "export_import_record_issue_665",
+    {|
+      import { Foo } from "data"
+      export enum Bar { Baz(Foo<Number>) }
+      print(Baz({ bar: 1 }))
+    |},
+    "Baz(<record value>)\n",
```
phated, line 171: @ospencer any idea why the record value is opaque to the printer? I exported the record from the imported module.

compiler/src/typed/env.re
```diff
@@ -968,11 +968,11 @@ let rec find_module_descr = (path, filename, env) => {
     try(IdTbl.find_same(id, env.components)) {
     | Not_found =>
       let (_, unit_source, _) = get_unit();
-      let filename = Option.value(~default=Ident.name(id), filename);
```
phated, line 971: There's a similar chunk of code in the `find_module` function inside this file. I'm wondering if that needs to be adjusted like this too. (I changed it locally and it didn't affect any tests at all).

compiler/test/suites/records.re
ospencer: No, that's very odd. Very concerning. Shouldn't be related to this change though...

ospencer: Realized what's wrong here. Definitely unrelated.

ospencer (approved):
Very nice!

ospencer:
> I'm not sure if this is the best semantic commit description.

I'd do `fix(compiler)` but otherwise I think it's fine.

marcusroberts (approved):
Looks good

compiler/test/suites/records.re
phated: #1021

compiler/src/typed/env.re
phated: I created a branch for this in case we want it in the future.

PR #1026, opened 2021-11-08
fix(grainfmt): Resugar match lists properly

### LINKED ISSUE #1024: grain format fails to format list matching correctly when dealing with multiple items
grain formatting:
```
let testFormatting = charlist => {
    match (charlist) {
      ['f', 'o', 'r', 'm', 'a', 't', ...rest] => true,
    }
}
```
results in:
```
let testFormatting = charlist => {
  match (charlist) {
    ['f', ...['o', ...['r', ...['m', ...['a', ...['t', ...rest]]]]]] => true,
  }
}
```

### LINKED ISSUE #1025: Formatter: Dependent data types don't format correctly
Given:

```grain
record Node { node: AstValue },

enum AstValue {
  NodeValue(Node),
  StringValue(String),
}
```

The formatter (incorrectly) turns it into this code:

```grain
record Node { node: AstValue },enum AstValue {
  NodeValue(Node),
  StringValue(String),
}
```

marcusroberts:
Fixes #1025 
Fixes #1024

Resugars pattern match strings properly
Prints recursive enum declarations properly

Extra tests added

### DIFF
--- a/compiler/grainformat/reformat.re
+++ b/compiler/grainformat/reformat.re
@@ -2987,7 +2987,7 @@ let data_print =
       original_source: array(string),
     ) => {
   Doc.join(
-    Doc.concat([Doc.comma, Doc.hardLine, Doc.hardLine]),
+    Doc.concat([Doc.comma, Doc.hardLine]),
     List.map(
       data => {
         let (expt, decl) = data;
--- a/compiler/grainformat/reformat.re
+++ b/compiler/grainformat/reformat.re
@@ -515,27 +515,6 @@ let rec resugar_list_patterns =
       Doc.rbracket,
     ]),
   );
-  // let items =
-  //   List.map(
-  //     i =>
-  //       switch (i) {
-  //       | RegularPattern(e) =>
-  //         print_pattern(~pat=e, ~parent_loc, ~original_source)
-  //       | SpreadPattern(e) =>
-  //         Doc.concat([
-  //           Doc.text("..."),
-  //           print_pattern(~pat=e, ~parent_loc, ~original_source),
-  //         ])
-  //       },
-  //     processed_list,
-  //   );
-  // Doc.group(
-  //   Doc.concat([
-  //     Doc.lbracket,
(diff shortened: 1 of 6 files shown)


compiler/grainformat/reformat.re
```diff
@@ -2962,7 +2987,7 @@ let data_print =
       original_source: array(string),
     ) => {
   Doc.join(
-    Doc.comma,
+    Doc.concat([Doc.comma, Doc.hardLine, Doc.hardLine]),
```
phated, line 2990: Why 2 hardlines here?

marcusroberts: I think they look better with a clear line between them, as in your failing example. Just a style choice really, but they seem squashed with no blank line between them

marcusroberts: Also, because the formatter usually adds a line between top level statements but these are seen as the same top level statement in the AST

phated: We might want to discuss with @ospencer and @peblair. I think it might be common for these to only have 1 newline between, but it'd be good to get more opinions. e.g.
```grain
enum Foo { Baz },
record Bar { a: Foo }
```

ospencer: I would personally vote for only one line, since in my mind I think of them as the same statement. The extra line adds some dissonance for me. Makes me wonder if we should go back to the `and` keyword, much more clear...

marcusroberts: I'll set it to one line with no gap, but also a vote for and!

phated (approved):
Looking good to me! Something that I'd like to mention is the PR name in relation to semantic commits. Generally, we'd want a semantic commit for each fix, and this can be solved when you squash commit. You'll change the main commit to something like `fix(grainfmt): Resugar match lists properly` and then in the body you can add `fix(grainfmt): Handle recursive enums`. Semantic commits are read out of the header and body, so this should work.

ospencer (approved):

PR #1027, opened 2021-11-09
feat(stdlib): Add reverse function to String module

technosophos:
Adds a new reverse() function to the String standard library.

Signed-off-by: Matt Butcher <matt.butcher@fermyon.com>

### DIFF
--- a/stdlib/string.gr
+++ b/stdlib/string.gr
@@ -382,7 +382,7 @@ export let rec implode = (arr: Array<Char>) => {
   ret
 }
 
-// Helper to get the length in constant time
+// Helper to get the length in constant time without depending on Array
 primitive arrayLength : Array<a> -> Number = "@array.length"
 
 /**
--- a/stdlib/string.gr
+++ b/stdlib/string.gr
@@ -390,20 +390,18 @@ export let rec implode = (arr: Array<Char>) => {
  *
  * @example String.reverse("olleH") == "Hello"
  *
- * @since v0.5.0
+ * @since v0.4.5
  */
 export let reverse = (string) => {
-  let len = wasmSafeLength(string)
+  let len = length(string)
   let halfLen = len / 2
   let mut arr = explode(string)
-  let mut i = 0
-  while (i < halfLen) {
+  for (let mut i = 0; i < halfLen; i += 1) {
     let lastIdx = len - i - 1
     let last = arr[lastIdx]
     let first = arr[i]
     arr[i] = last
     arr[lastIdx] = first
-    i = i + 1
   }
   implode(arr)
 }
--- a/stdlib/string.gr
+++ b/stdlib/string.gr
@@ -383,6 +383,32 @@ export let rec implode = (arr: Array<Char>) => {
 }
 
 /**
+ * Create a string that is the given string reversed.
+ *
+ * @param string: The string to reverse
+ * @returns A string whose characters are in the reverse order of the given string
+ *
+ * @example String.reverse("olleH") == "Hello"
+ *
+ * @since v0.5.0
+ */
+export let reverse = (string) => {
+  let len = wasmSafeLength(string)
+  let halfLen = len / 2
+  let mut arr = explode(string)
+  let mut i = 0
+  while (i < halfLen) {
+    let lastIdx = len - i - 1
+    let last = arr[lastIdx]
+    let first = arr[i]
+    arr[i] = last
+    arr[lastIdx] = first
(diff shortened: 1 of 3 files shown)


stdlib/string.gr
```diff
@@ -382,6 +382,32 @@ export let rec implode = (arr: Array<Char>) => {
...
+ * @example String.reverse("olleH") == "Hello"
+ *
+ * @since v0.5.0
+ */
+export let reverse = (string) => {
+  let len = wasmSafeLength(string)
+  let halfLen = len / 2
+  let mut arr = explode(string)
+  let mut i = 0
+  while (i < halfLen) {
```
phated, line 400: Should we use a `for` loop? I know a lot of the stdlib uses `while` because we wrote them before we had a for-loop, but I think we've been wanting to convert stuff over.

ospencer: Yeah, let's just use a for loop 🙂

phated: @technosophos we are hoping to get v0.4.5 released very soon. Any chance we could have this updated to land? Sorry to be a pain during the holiday week.

phated, line 400:
```suggestion
  for (let mut i = 0; i < halfLen; i += 1) {
```

stdlib/string.gr
```diff
@@ -382,6 +382,32 @@ export let rec implode = (arr: Array<Char>) => {
...
+    let lastIdx = len - i - 1
+    let last = arr[lastIdx]
+    let first = arr[i]
+    arr[i] = last
+    arr[lastIdx] = first
+    i = i + 1
```
phated, line 406:
```suggestion
```

stdlib/string.gr
```diff
@@ -382,6 +382,32 @@ export let rec implode = (arr: Array<Char>) => {
...
 
+/**
+ * Create a string that is the given string reversed.
+ *
+ * @param string: The string to reverse
+ * @returns A string whose characters are in the reverse order of the given string
+ *
+ * @example String.reverse("olleH") == "Hello"
+ *
+ * @since v0.5.0
```
phated, line 393:
```suggestion
 * @since v0.4.5
```

stdlib/string.gr
```diff
@@ -382,6 +382,32 @@ export let rec implode = (arr: Array<Char>) => {
...
+ */
+export let reverse = (string) => {
+  let len = wasmSafeLength(string)
```
phated, line 396: Since this isn't a disableGC context, it's very bad to use this function. I believe it should be reporting a warning.
```suggestion
  let len = length(string)
```

phated: @ospencer is looking into a constant time lookup with the array length instead

phated (changes requested):

stdlib/string.gr
```diff
@@ -382,6 +382,30 @@ export let rec implode = (arr: Array<Char>) => {
   ret
 }
 
+/**
```
phated, line 385:
```suggestion
// Helper to get the length in constant time
primitive arrayLength : Array<a> -> Number = "@array.length"

/**
```

stdlib/string.gr
```diff
@@ -382,6 +382,30 @@ export let rec implode = (arr: Array<Char>) => {
...
+ * @returns A string whose characters are in the reverse order of the given string
+ *
+ * @example String.reverse("olleH") == "Hello"
+ *
+ * @since v0.4.5
+ */
+export let reverse = (string) => {
+  let len = length(string)
+  let halfLen = len / 2
+  let mut arr = explode(string)
```
phated, line 398:
```suggestion
  let mut arr = explode(string)
  let len = arrayLength(arr)
  let halfLen = len / 2
```

phated (changes requested):
I believe these are the changes that'd make the length check constant time.

stdlib/string.gr
```diff
@@ -382,6 +382,33 @@ export let rec implode = (arr: Array<Char>) => {
   ret
 }
 
+// Helper to get the length in constant time
```
ospencer, line 385:
```suggestion
// Helper to get the length in constant time without depending on Array
```

ospencer (commented):
One suggestion but otherwise looks great to me!

phated (approved):
Looks great to me! Thanks for getting this started @technosophos—we'll get it into 0.4.5

ospencer (approved):
Thank you!

PR #1032, opened 2021-11-17
chore: Add -slim Dockerfile and update Docker workflows

### LINKED ISSUE #1007: Reimplement Docker Hub Github Actions

### LINKED ISSUE #1010: Docker images don't publish on our release tags
Since we've added Release Please, a workflow is creating our tags, which can't trigger another workflow to run (to avoid the chance that it causes a circular trigger). However, we can trigger it with workflow_dispatch to create a tagged release.

ospencer:
Closes #1007
Closes #1010 

This PR updates the `docker/build-push-action` GitHub action to v2 (which allows us to publish to DockerHub again) and adds a new `Dockerfile-slim` which provides a minimal environment that can compile Grain programs. The slim version is about 10x smaller!

image

You can see a [successful run of the GitHub workflow](https://github.com/grain-lang/grain/runs/4243823185?check_suite_focus=true), [images published on DockerHub](https://hub.docker.com/repository/docker/grainlang/grain/tags?page=1&ordering=last_updated&name=oscar-docker), and [on the GitHub Container Registry](https://github.com/grain-lang/grain/pkgs/container/grain). Note that we previously were (accidentally) publishing on GHCR as `grain-lang/grain/grain` rather than `grain-lang/grain`.

In addition to the push on `main`, there's also a `workflow_dispatch` event that gets triggered on release. Shoutout to @phated for making it easy for me to figure out how to do that 😤 

You can see that the tag is handled properly via [this workflow](https://github.com/ospencer/grain/runs/4244964750?check_suite_focus=true) (and that it's ignored on `main` via [this workflow](https://github.com/ospencer/grain/runs/4244950280?check_suite_focus=true)), but I've also got some screenshots here to save you some clicks:

For the standard image:
image

For the slim image:
image

When just running on a push to `main`:
image

The docker metadata action is quite nice, as you can see that it will also handle having a release tag (i.e. `grain:0.4.4`) and also a tag for the minor for automatic patch updates (i.e. `grain:0.4`).

Besides that, the main Dockerfile was simplified quite a bit. The slim version is similar, but it has to manually install some packages and then does a bunch of cleanup.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -21,6 +21,7 @@ script/node_modules
 **/*.log
 _build/
 */_build/
+cli/bin/*.exe
 .installed-pkgs
 notes.org
 node_modules/
--- a/.dockerignore
+++ b/.dockerignore
@@ -22,6 +22,7 @@ script/node_modules
 _build/
 */_build/
 cli/bin/*.exe
+cli/bin/*.bc.js
 .installed-pkgs
 notes.org
 node_modules/
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,30 +1,105 @@
-name: Publish Docker image (HEAD)
+name: Publish Docker images
 on:
   push:
-    branches: [main]
-    tags: ["v*.*.*"]
+    branches: main
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "The tag of release"
+        required: true
 jobs:
-  push_to_registries:
-    name: Push Docker image to multiple registries
+  push_base_image:
+    name: Push base Docker image to multiple registries
     runs-on: ubuntu-latest
     steps:
       - name: Check out the repo
         uses: actions/checkout@v2
 
-      # TODO: Re-implement with v2 Docker GitHub Actions
-      # - name: Push to Docker Hub
-      #   uses: docker/build-push-action@v1
-      #   with:
-      #     username: ${{ secrets.DOCKER_USERNAME }}
-      #     password: ${{ secrets.DOCKER_PASSWORD }}
-      #     repository: grainlang/grain
-      #     tag_with_ref: true
+      - name: Parse tag
+        # This step converts Grain tags into standard semver, i.e. grain-v1.2.3 -> v1.2.3
+        id: vars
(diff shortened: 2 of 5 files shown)


marcusroberts (approved):
That is sweet!

.github/workflows/docker-publish.yml
```diff
@@ -1,30 +1,105 @@
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: Dockerfile
+          push: ${{ github.event_name != 'pull_request' }}
```
phated, line 54: If we only run this on `main`, why do we need to filter on `!= pull_request`?

.dockerignore
```diff
@@ -21,6 +21,7 @@ script/node_modules
 **/*.log
 _build/
 */_build/
+cli/bin/*.exe
```
phated, line 24: Should also exclude the bc.js stuff
```suggestion
cli/bin/*.exe
cli/bin/*.bc.js
```

.github/workflows/docker-publish.yml
```diff
@@ -1,30 +1,105 @@
...
-          registry: docker.pkg.github.com
-          repository: grain-lang/grain/grain
-          tag_with_ref: true
+
+      - name: Build and push slim image
+        uses: docker/build-push-action@v2.7.0
+        with:
+          context: .
+          file: Dockerfile-slim
+          push: ${{ github.event_name != 'pull_request' }}
```
phated, line 103: same question

phated (commented):
This looks really solid to me! There are a couple things I've noted, and I believe you should add "Closes X" for possibly 2-3 issues in the tracker (?)

.github/workflows/docker-publish.yml
ospencer: copypasta 🙈

ospencer:
@phated Should be all set here, and added some issues to close!

phated (approved):
:shipit: Excellent work!

PR #1033, opened 2021-11-19
feat(compiler)!: Re-implement Grain parser

### LINKED ISSUE #323: Lang: Define whitespace syntax for binary operators
Currently, Grain disallows newline characters between the left side argument and the operator, and allows newline characters between the operator and the right side argument.

In addition to having no line breaks, e.g.
```grain
1 + 1
```
This is **allowed**:
```grain
1 +
1
```
And this is **disallowed**:
```grain
1
+ 1
```

## The Question

Are we comfortable with the current syntax, should we disallow newlines completely (which is how Python does it), or should we allow newlines on both sides (which is how JavaScript does it)?

### 1. Statement terminators

Grain recognizes `EOL` and `;` as valid statement terminators. If we want to take this very strictly, that would mean that we completely disallow newline characters (a la Python).

### 2. Ambiguities

There are no ambiguities with allowing newline characters on the right side. 
```grain
1 +
1
```
would always mean `1 + 1` if this were allowed.

There are, however, ambiguities if newlines are allowed on the left side of the operator, particularly with unary negation:
```grain
1
- 1
```
could either mean `1; -1` or `1 - 1`. This ambiguity would need to be resolved. Our [operator precedence document](https://github.com/grain-lang/grain/blob/master/docs/contributor/operator_precedence.md) doesn't quite define how to resolve this, but my interpretation is that it should be resolved as `1 - 1` since subtraction takes precedence over unary negation. Performing this resolution would be difficult and could potentially have performance issues, but that should not be a factor when deciding the best syntax here for Grain.

Since Grain has implicit returns, writing a program like this would be incorrect:
```grain
let f = () => {
  performSomeAction()
  -1 # return negative exit code
}
```
This would parse as `minus(performSomeAction(), 1)`, which would almost always not be what the user wanted. However, writing something like this
```grain
let value = 1
- 2
- 3
- 4
```
would work as expected.

### LINKED ISSUE #573: Compiler: Find better solution for parsing of one-sided `if`s
The solution introduced in https://github.com/grain-lang/grain/pull/572 isn't great, but I couldn't think of anything better at the time.

### LINKED ISSUE #679: Error referencing variables in record construction without a trailing comma
I encountered a compilation error for something I intuitively think would be correct syntax. I can't simply assign an existing variable in a record constructor, yet I can pass constants or function calls with closures referencing same variables. Then I managed to make it work by randomly appending a trailing comma after the entry.

``` grain
import List from "list"
import Map from "map"
import Option from "option"

record MyRecord<k, v> {
	entries: String
}

let s = "hello"

let a = {
import List from "list"
import Map from "map"
import Option from "option"

record MyRecord {
	entry: String
}

let s = "hello"

let a = {
	// this doesn't work
	entry: s

	// ...yet this does!
	//entry: s,

	// as well as this
	//entry: "hello"

	// or this:
	//entry: Option.unwrap(Option.map((id) => s, Some("ignore me")))
}

print(a)
```

### LINKED ISSUE #709: Compiler: Support calling function within array
This would add the ability to do `arr[0]()`.

See original discussion:

> This doesn't work and I don't understand why it doesn't work. Uncommenting this line makes the parser give really cryptic error messages. I think it's fine to move forward without this for now (since it's not a super common use case) and if this PR is approved, I'll make an issue for this and reference it here.

_Originally posted by @ospencer in https://github.com/grain-lang/grain/pull/701#discussion_r646041182_

### LINKED ISSUE #783: Parse error when using a generic type parameter to annotate a function call.

### LINKED ISSUE #1062: Bad range with import foreign
Problem replication:

```rs
import foreign wasm to_string: (WasmI64, WasmI32) -> void from "lunatic::error"
```

Looks like it reports `(WasmI64, WasmI32` as the range, when the problem is the `Void` keyword.

ospencer:
Closes #323 
Closes #573 
Closes #679 
Closes #709 
Closes #783 
Closes #1062 

# Parse Grain programs using Menhir 🗿 
[Menhir](http://gallium.inria.fr/~fpottier/menhir/) is an LR(1) parser generator for OCaml. It generates parsers that are fast and gives us tools to help maintain the parser. The main motivation for this change is efficiency—this implementation parses the standard library's hefty `regex.gr` **25X faster** than the current implementation. In addition to the speed improvements, syntax error messages have better location information and there are now nearly **200 custom error messages** for all possible parser error states.

## Syntax changes
There are some minor syntax changes that come along with this change. Grain ends expressions with a newline character— most of the changes are around where line breaks are allowed. These changes shouldn't affect how a majority of users use Grain. In fact, most code that is formatted with Grain's code formatter from Grain v0.4 will continue to parse with Grain v0.5.

### Binary operators
Grain is taking Python's approach to binary operators—newline characters are no longer allowed before a binary operator.

✅ Accepted
```
foo /
bar
```
```
foo &&
bar
```
```
foo.
bar
```

❌ Not accepted
```
foo
/ bar
```
```
foo
&& bar
```
```
foo
.bar
```

### Records
`{ x }` will no longer parse as a single-argument record, and instead will parse as a block with the identifier `x`. Similar to a tuple, a record with a single punned argument can be made by adding a trailing comma:

✅ Accepted
```
{ x, }
```
❌ Not accepted
```
{ x }
```

Conversely, a block that starts with `x:` will parse as a record rather than an expression with a type annotation. For example, `() => { foo: Bar }` will parse as a function that returns a record with key `foo` and value `Bar` rather than a block with value `foo` annotated with `Bar`.

## Syntax error messages
Menhir's default parser generation generates a parser that is very fast, but can't provide useful error messages. Menhir can produce a slower, more compact parser that _is_ capable of producing good error messages, so for that reason, we actually generate two parsers—a really fast parser that can generate a Grain AST, and a slower parser that doesn't generate anything, but can give good error messages. We try to parse a program using the fast parser, and if it fails, we reparse it using the slow parser and try to provide a good error message.

Menhir provides the `.messages` file format to maintain parser error messages. [You can read all about it here](http://gallium.inria.fr/~fpottier/menhir/manual.html#sec68). Our file is called `parser.messages`. It's large, but the file is mostly comments!

I added some new yarn commands for working with the messages:

`yarn compiler parser:list-errors`
This command will generate a file, `parser.messages.generated`, which contains all of the possible error states that exist in the parser, with default error messages. You can use this to view all of the states and cross-check with `parser.messages`.

`yarn compiler parser:check-errors`
This command will generate `parser.messages.generated` (as above) and verify that there is an error message defined in `parser.messages` for each error state in `parser.messages.generated`. (This is also run in CI.)

`yarn compiler parser:interpret`
This command starts an interpreter for the parser. You can type tokens and check if they're accepted by the parser:

image

I think if we define aliases for all of the tokens, you can type regular Grain code and it'd work. Maybe something for the future if people find the interpreter useful!

`yarn compiler parser:interpret-error`
This is similar to the regular interpret, except that it expects a syntax error to occur on the last token. It'll then print information about the error state, in the same format as the `.messages` file.

image

## Info for parsing nerds
Menhir is an LR(1) parser generator, and Dypgen (our current parser generator) is a GLR parser. GLR parsers have a worst-case runtime complexity of O(n^3), while LR(k) parsers are O(n). We can't completely blame the performance on the algorithm, though. GLR parsers are also O(n) for unambiguous grammars—we could take all of the work that was done to make the Grain grammar work with Menhir and patch it back into the Dypgen parser and it would run significantly faster. However, Menhir forces us to keep our grammar unambiguous and gives us tools to maintain things like error messages, so overall it makes sense to keep the Menhir implementation.

### Caveats
LR parsers can only accept a subset of the languages that GLR parsers can. This is somewhat of a problem, as Grain's grammar is not actually LR—Grain has arrow functions. After encountering an open paren, an LR parser doesn't know if it should be parsing a tuple or a function. This is an important distinction since tuples contain expressions and function arguments contain destructuring patterns. In fact, an LR parser wouldn't be able to tell which it should be parsing until it encountered (or didn't encounter) an arrow after the closing paren. The 1 in LR(1) (or the k in LR(k)) represents the number of tokens the parser will look ahead, in this case, just one. To check for that arrow, the parser would need to look ahead a potentially infinite number of tokens, which is decidedly more than one.

The tools that Menhir gives us are so good that it's worth implementing a [lexer hack](https://en.wikipedia.org/wiki/Lexer_hack) to solve this one problem. The simple solution to the tuple/arrow function issue is to just tell the parser when it's parsing a function. We borrow this solution from ReasonML's lexer. In the lexer, when an open paren is encountered, we scan ahead to the matching closing paren and check if the next token is an arrow. If so, we inject a special `FUN` token right before the open paren. To avoid a slowdown, all of the tokens that are seen during this process are cached—work is never duplicated.


compiler/src/parsing/ast_helper.re
```diff
@@ -17,32 +17,20 @@
...
+
+type listitem('a) =
+  | ListItem('a)
+  | ListSpread('a, Location.t);
+
 type id = loc(Identifier.t);
 type str = loc(string);
 type loc = Location.t;
 
-let default_loc_src = ref(() => Location.dummy_loc);
```
peblair, line 24: 🙌

compiler/src/parsing/lexer.mll
```diff
@@ -197,13 +202,19 @@ rule token = parse
   | "*" { STAR }
   | "/" { SLASH }
   | "%" { PERCENT }
+  | "+=" { PLUSEQ }
```
peblair, line 205: ‼️ 

How did we lex this stuff before? Or was it part of the parser?

compiler/test/formatter_outputs/aliases.gr
```diff
@@ -18,18 +18,15 @@ type AReallyReallyReallyReallyReallyReallyReallyReallyReallyLongLineBreakingType
 type AReallyReallyReallyReallyReallyReallyReallyReallyReallyLongLineBreakingType<
   a,
   b
-> = String // Test comment
```
peblair, line 21: Why does this disappear?

compiler/src/parsing/parser.mly
```diff
@@ -0,0 +1,707 @@
...
+
+eols:
+  | eol+ {}
+
+%inline opt_eols:
+  | ioption(eols) {}
+
+eos:
+  | eols {}
+  | SEMI opt_eols %prec SEMI {}
```
peblair, line 145: Any tips for understanding what these `%prec`s do? I looked at the manual entry, but I'm struggling with the intuition.

peblair (commented):
Another beautiful pull request from our very own @ospencer. I haven't read through the `messages` file yet, but the rest of this looks awesome! I left a couple of comments with questions, but they are very minor.

compiler/src/parsing/lexer.mll
ospencer: Just a part of the parser! Having it lexed as a single token makes the error handlers much nicer.

compiler/test/formatter_outputs/aliases.gr
ospencer: Basically, the locations of the type parameters were just completely wrong before 😶

They're correct now, and the comments were only kept coincidentally before. @marcusroberts is already working on a fix for me!

compiler/src/parsing/parser.mly
ospencer: `%prec expr` makes the reduction happen with the precedence of `expr`, whether that's a real token or a fake one that exists just for precedence. This one should be unnecessary—I had a nonterminal for SEMI but removed it (so I needed the annotation to give this the same precedence as SEMI).

They're mostly not necessary, but they resolve shift/reduce conflicts in some cases. The major one we have is one for `if`, to handle the famous "dangling else" problem. There's a section of the Menhir manual about conflicts that's really helpful!

compiler/src/parsing/parser.mly
```diff
@@ -0,0 +1,706 @@
+%{
+(* These opens are used inside the actual parser *)
+open Parsetree
+open Ast_helper
+open Asttypes
+
+(* Including the Parser_extra file allows it to be written in Reason and have editor tooling *)
+include Parser_header
+
+(* https://github.com/ocaml/dune/issues/2450 *)
```
phated, line 10: :sob: dune maintainers hurt my head

compiler/src/parsing/parser.mly
```diff
@@ -0,0 +1,706 @@
...
+%token LET MUT REC IF WHEN ELSE MATCH WHILE FOR CONTINUE BREAK
+%token AMPAMP PIPEPIPE NOT AT
+%token AMP PIPE
+
+%token ENUM RECORD TYPE IMPORT EXPORT FOREIGN WASM PRIMITIVE
+%token EXCEPT FROM
+%token EOL EOF
+
+// reserved tokens
+%token TRY CATCH COLONCOLON
```
phated, line 54: Was `::` reserved in the old parser?

compiler/src/parsing/parser.mly
```diff
@@ -0,0 +1,706 @@
...
+%left AMP
+%left EQEQ NOTEQ IS ISNT
+%left LCARET LESSEQ RCARET GREATEREQ
+%left LCARETLCARET
+%left PLUS DASH PLUSPLUS
+%left STAR SLASH PERCENT
+
+%right SEMI EOL COMMA DOT COLON
+
+%nonassoc _if
```
phated, line 76: Why is this `_if` when everything else is capitalized?

compiler/src/parsing/parser.mly
```diff
@@ -0,0 +1,706 @@
...
+  type_id
+  value_binds
+
+%%
+
+// List helpers. These will parse a left-associated list.
+// Solves the shift/reduce conflicts you get from Menhir's
+// built-in lists.
+
+lseparated_nonempty_list_inner(sep, X):
```
phated, line 121: How'd you decide this didn't need `%inline` but the 2 rules below needed it?

compiler/src/parsing/parser.mly
```diff
@@ -0,0 +1,706 @@
...
+
+// reserved tokens
+%token TRY CATCH COLONCOLON
+
+// Not a real token, this is injected by the lexer
+%token FUN
+
+/* Operator precedence may be found in /docs/contributor/operator_precedence.md */
+
+%nonassoc _below_infix
```
phated, line 61: Can you explain this?

compiler/src/parsing/parser.mly
```diff
@@ -0,0 +1,706 @@
...
+non_stmt_expr:
+  | binop_expr { $1 }
+  | annotated_expr { $1 }
+
+%inline opt_annotation:
+  | %prec COLON { None }
+  | colon typ { Some $2 }
+
+annotated_expr:
+  | non_binop_expr opt_annotation { Option.fold ~none:$1 ~some:(fun ann -> Exp.constraint_ ~loc:(to_loc $loc) $1 ann) $2 }
```
phated, line 229: Any reason to not have a helper for this? I kinda tend towards having extra helpers written in Reason that we can use very simply in the ocaml-syntax of the parser.

compiler/src/parsing/parser.mly
```diff
@@ -0,0 +1,706 @@
...
+binop_expr:
+  | non_stmt_expr infix_op opt_eols non_stmt_expr { Exp.binop ~loc:(to_loc $loc) (mkid_expr $loc($2) [$2]) [$1; $4] }
+
+ellipsis_prefix(X):
+  | ELLIPSIS X {$2}
+
+pattern:
+  | pattern colon typ { Pat.constraint_ ~loc:(to_loc $loc) $1 $3 }
+  | UNDERSCORE { Pat.any ~loc:(to_loc $loc) () }
+  | const { let (pat, loc) = $1 in Pat.constant ~loc:(to_loc loc) pat }
```
phated, line 240: This is *so* weird. It took me like 3 reads to understand what was happening. Any way to tighten that up?

compiler/src/parsing/parser.mly
```diff
@@ -0,0 +1,706 @@
...
+  // Allow rational numbers in patterns
+  | dash_op? NUMBER_INT slash_op dash_op? NUMBER_INT { Pat.constant ~loc:(to_loc $sloc) @@ Const.number (PConstNumberRational ((if Option.is_some $1 then "-" ^ $2 else $2), (if Option.is_some $4 then "-" ^ $5 else $5))) }
```
phated, line 242: Is there a reason you don't have a rule for rationals instead of using this in multiple places?

compiler/src/parsing/parser.mly
```diff
@@ -0,0 +1,706 @@
...
+import_shape:
+  | id { PImportModule $1 }
+  | STAR import_exception? { PImportAllExcept (Option.value ~default:[] $2) }
+  | lbrace import_ids? rbrace { PImportValues (Option.value ~default:[] $2) }
+
+import_stmt:
+  | IMPORT lseparated_nonempty_list(comma, import_shape) comma? FROM file_path { Imp.mk ~loc:(to_loc $loc) $2 $5 }
+
+data_declaration_stmt:
+  // TODO: Attach attributes to the node
```
phated, line 327: What happens now if they are not?

compiler/src/parsing/parser.mly
```diff
@@ -0,0 +1,706 @@
...
+
+type_id:
+  | lseparated_nonempty_list(dot, TYPEID) { (mkid $1) (to_loc $loc) }
+
+id_expr:
+  // Force any following colon to cause a shift
+  | id %prec COLON { Exp.ident ~loc:(to_loc $loc) $1 }
+
+simple_expr:
+  | const { let (exp, loc) = $1 in Exp.constant ~loc:(to_loc loc) exp }
```
phated, line 502: Another one of these strange looking ones. I understood it more this 2nd time I encountered it.

phated (commented):
I love this. So excited for the new parser. 🔥  I reviewed everything except the messages file because safari can't load it, but I wanted to submit my initial comments before I review the messages.

compiler/test/suites/types.re
```diff
@@ -43,9 +43,8 @@ describe("aliased types", ({test}) => {
     "type_alias_5",
     {|
       type Foo<a> = (String, List<a>)
-      let foo = (x: Foo<Number>) => {
+      let foo = (x: Foo<Number>) =>
         x: (String, List<Number>)
-      }
```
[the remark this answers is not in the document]
ospencer: Yes, that's why. If you wanted to keep the block but have that annotation, you could write it like this:
```
let foo = x => {
  (x: (String, List<Number>))
}
```
or like this:
```
let foo = x => {
  (x): (String, List<Number>)
}
```

If we ever decide to change our annotation syntax to use `::` (and I don't think anyone objected to it, it just needed to be implemented, so we could just do it now), then it wouldn't be a problem, and it would look like this:
```
let foo = x => {
  x :: (String, List<Number>)
}
```

stdlib/array.gr
```diff
@@ -913,8 +913,7 @@ export let rotate = (n, arr) => {
   let rec gcd = (a, b) => {
     if (b == 0) {
       a
-    }
-    else{
+    } else {
```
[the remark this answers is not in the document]
ospencer: It's possible, just difficult because of our `EOL` tokens (for the same reason why we made the change to the infix operators). I can try to play around with it some more to see if I can get it to work.

stdlib/string.gr
```diff
@@ -1761,8 +1761,11 @@ let trimString = (str: String, end: Bool) => {
       currentChar != '\r' // CR
     ) break
   }
-  if (end) slice(0, i+1, str)
-  else slice(i, charsLength, str)
+  if (end) {
+    slice(0, i+1, str)
+  } else { 
+    slice(i, charsLength, str)
+  }
```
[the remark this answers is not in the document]
ospencer: Yup! That's still supported.

compiler/test/formatter_outputs/aliases.gr
marcusroberts: I think this is working now!

marcusroberts (approved):
Early approval from me.  I've read the code which increased my understanding a lot, and I like the performance and errors we get from this new parser.   Fantastic job!

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## id
+##
+## WARNING: This example involves spurious reductions.
+## This implies that, although the LR(1) items shown above provide an
+## accurate view of the past (what has been recognized so far), they
+## may provide an INCOMPLETE view of the future (what was expected next).
+## In state 122, spurious reduction of production id -> lseparated_nonempty_list_inner(dot,TYPEID)
+##
+
+Expected a colon followed by an expression, a comma followed by more record fields, or an immediate '}' to complete the record.
```
peblair:
```suggestion
Expected a colon followed by an expression, a comma followed by more record fields, or just a '}' to complete the record.
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+##
+## Ends in an error in state: 352.
+##
+## pattern -> lparen special_op . rparen [ WHEN THICKARROW RPAREN RBRACK RBRACE EQUAL EOL COMMA COLON ]
+##
+## The known suffix of the stack is as follows:
+## lparen special_op
+##
+
+Expected a closing paren.
```
peblair:
```suggestion
Expected a closing parenthesis.
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## WARNING: This example involves spurious reductions.
+## This implies that, although the LR(1) items shown above provide an
+## accurate view of the past (what has been recognized so far), they
+## may provide an INCOMPLETE view of the future (what was expected next).
+## In state 3, spurious reduction of production nonempty_list(eol) -> EOL
+## In state 5, spurious reduction of production eols -> nonempty_list(eol)
+## In state 89, spurious reduction of production comma -> COMMA eols
+##
+
+Expected a pattern (or a closing paren to complete the tuple pattern).
```
peblair:
```suggestion
Expected a pattern (or a closing parenthesis to complete the tuple pattern).
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## pattern COMMA eols lseparated_nonempty_list_inner(comma,pattern)
+##
+## WARNING: This example involves spurious reductions.
+## This implies that, although the LR(1) items shown above provide an
+## accurate view of the past (what has been recognized so far), they
+## may provide an INCOMPLETE view of the future (what was expected next).
+## In state 334, spurious reduction of production lseparated_nonempty_list_inner(comma,pattern) -> pattern
+##
+
+Expected a comma followed by another pattern (or a closing paren to complete the tuple pattern).
```
peblair:
```suggestion
Expected a comma followed by another pattern (or a closing parenthesis to complete the tuple pattern).
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## WARNING: This example involves spurious reductions.
+## This implies that, although the LR(1) items shown above provide an
+## accurate view of the past (what has been recognized so far), they
+## may provide an INCOMPLETE view of the future (what was expected next).
+## In state 3, spurious reduction of production nonempty_list(eol) -> EOL
+## In state 5, spurious reduction of production eols -> nonempty_list(eol)
+## In state 89, spurious reduction of production comma -> COMMA eols
+##
+
+Expected another pattern (or a closing paren to complete the tuple pattern).
```
peblair:
```suggestion
Expected another pattern (or a closing parenthesis to complete the tuple pattern).
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## pattern COMMA lseparated_nonempty_list_inner(comma,pattern)
+##
+## WARNING: This example involves spurious reductions.
+## This implies that, although the LR(1) items shown above provide an
+## accurate view of the past (what has been recognized so far), they
+## may provide an INCOMPLETE view of the future (what was expected next).
+## In state 334, spurious reduction of production lseparated_nonempty_list_inner(comma,pattern) -> pattern
+##
+
+Expected a comma followed by another pattern (or a closing paren to complete the tuple pattern).
```
peblair:
```suggestion
Expected a comma followed by another pattern (or a closing parenthesis to complete the tuple pattern).
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## tuple_patterns -> pattern COMMA . lseparated_nonempty_list_inner(comma,pattern) [ RPAREN EOL ]
+## tuple_patterns -> pattern COMMA . lseparated_nonempty_list_inner(comma,pattern) comma [ RPAREN EOL ]
+## tuple_patterns -> pattern COMMA . eols lseparated_nonempty_list_inner(comma,pattern) [ RPAREN EOL ]
+## tuple_patterns -> pattern COMMA . eols lseparated_nonempty_list_inner(comma,pattern) comma [ RPAREN EOL ]
+##
+## The known suffix of the stack is as follows:
+## pattern COMMA
+##
+
+Expected a pattern (or an immediate closing paren for a single-argument tuple pattern).
```
peblair:
```suggestion
Expected a pattern (or just a closing parenthesis for a single-argument tuple pattern).
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## tuple_patterns -> pattern . COMMA lseparated_nonempty_list_inner(comma,pattern) [ RPAREN EOL ]
+## tuple_patterns -> pattern . COMMA lseparated_nonempty_list_inner(comma,pattern) comma [ RPAREN EOL ]
+## tuple_patterns -> pattern . COMMA eols lseparated_nonempty_list_inner(comma,pattern) [ RPAREN EOL ]
+## tuple_patterns -> pattern . COMMA eols lseparated_nonempty_list_inner(comma,pattern) comma [ RPAREN EOL ]
+##
+## The known suffix of the stack is as follows:
+## lparen pattern
+##
+
+Expected a type annotation, a closing paren (for a grouped pattern), or a comma (for a tuple pattern).
```
peblair:
```suggestion
Expected a type annotation, a closing parenthesis (for a grouped pattern), or a comma (for a tuple pattern).
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## WARNING: This example involves spurious reductions.
+## This implies that, although the LR(1) items shown above provide an
+## accurate view of the past (what has been recognized so far), they
+## may provide an INCOMPLETE view of the future (what was expected next).
+## In state 334, spurious reduction of production lseparated_nonempty_list_inner(comma,pattern) -> pattern
+## In state 344, spurious reduction of production option(comma) ->
+## In state 345, spurious reduction of production patterns -> lseparated_nonempty_list_inner(comma,pattern) option(comma)
+##
+
+Expected a closing paren to complete the variant pattern.
```
peblair:
```suggestion
Expected a closing parenthesis to complete the variant pattern.
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## may provide an INCOMPLETE view of the future (what was expected next).
+## In state 41, spurious reduction of production non_assign_expr -> simple_expr
+## In state 120, spurious reduction of production non_binop_expr -> non_assign_expr
+## In state 64, spurious reduction of production annotated_expr -> non_binop_expr
+## In state 238, spurious reduction of production non_stmt_expr -> annotated_expr
+## In state 60, spurious reduction of production expr -> non_stmt_expr
+## In state 253, spurious reduction of production lseparated_nonempty_list_inner(comma,expr) -> expr
+##
+
+Expected a comma followed by an expression (or a closing paren to complete the function call).
```
peblair:
```suggestion
Expected a comma followed by an expression (or a closing parenthesis to complete the function call).
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## Ends in an error in state: 349.
+##
+## option(dash_op) -> DASH . [ WASMI64 WASMI32 WASMF64 WASMF32 NUMBER_INT NUMBER_FLOAT INT64 INT32 FLOAT64 FLOAT32 ]
+## special_op -> DASH . [ RPAREN EOL ]
+##
+## The known suffix of the stack is as follows:
+## DASH
+##
+
+Expected a number (to create a negative number) or a closing paren (to reference the minus function).
```
peblair:
```suggestion
Expected a number (to create a negative number) or a closing parenthesis (to reference the minus function).
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## Ends in an error in state: 155.
+##
+## prefix_op -> NOT . [ RPAREN EOL ]
+## prim1_expr -> NOT . non_assign_expr [ STAR SLASH RPAREN RCARET PLUSPLUS PLUS PIPEPIPE PIPE PERCENT NOTEQ LESSEQ LCARETLCARET LCARET ISNT IS GREATEREQ EQEQ EOL DASH COMMA COLON CARET AMPAMP AMP ]
+##
+## The known suffix of the stack is as follows:
+## NOT
+##
+
+Expected an expression (or a closing paren to reference the not operator).
```
peblair:
```suggestion
Expected an expression (or a closing parenthesis to reference the not operator).
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+##
+## Ends in an error in state: 149.
+##
+## non_modid -> lparen special_op . rparen [ THICKARROW STAREQ STAR SLASHEQ SLASH SEMI RPAREN RCARET RBRACK RBRACE PLUSPLUS PLUSEQ PLUS PIPEPIPE PIPE PERCENTEQ PERCENT NOTEQ LPAREN LESSEQ LCARETLCARET LCARET LBRACK ISNT IS GREATEREQ GETS FROM EQUAL EQEQ EOL EOF ELSE DOT DASHEQ DASH COMMA COLON CARET AS AMPAMP AMP ]
+##
+## The known suffix of the stack is as follows:
+## lparen special_op
+##
+
+Expected a closing paren.
```
peblair:
```suggestion
Expected a closing parenthesis.
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## WARNING: This example involves spurious reductions.
+## This implies that, although the LR(1) items shown above provide an
+## accurate view of the past (what has been recognized so far), they
+## may provide an INCOMPLETE view of the future (what was expected next).
+## In state 88, spurious reduction of production comma -> COMMA
+## In state 168, spurious reduction of production option(comma) -> comma
+## In state 485, spurious reduction of production tuple_exprs -> expr COMMA lseparated_nonempty_list_inner(comma,expr) option(comma)
+##
+
+Expected a closing paren to complete the tuple expression.
```
peblair:
```suggestion
Expected a closing parenthesis to complete the tuple expression.
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## may provide an INCOMPLETE view of the future (what was expected next).
+## In state 41, spurious reduction of production non_assign_expr -> simple_expr
+## In state 120, spurious reduction of production non_binop_expr -> non_assign_expr
+## In state 64, spurious reduction of production annotated_expr -> non_binop_expr
+## In state 238, spurious reduction of production non_stmt_expr -> annotated_expr
+## In state 60, spurious reduction of production expr -> non_stmt_expr
+## In state 253, spurious reduction of production lseparated_nonempty_list_inner(comma,expr) -> expr
+##
+
+Expected a comma followed by an expression or a closing paren to complete the tuple expression.
```
peblair:
```suggestion
Expected a comma followed by an expression or a closing parenthesis to complete the tuple expression.
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+##
+## tuple_exprs -> expr COMMA . [ RPAREN EOL ]
+## tuple_exprs -> expr COMMA . lseparated_nonempty_list_inner(comma,expr) option(comma) [ RPAREN EOL ]
+## tuple_exprs -> expr COMMA . eols lseparated_nonempty_list_inner(comma,expr) option(comma) [ RPAREN EOL ]
+##
+## The known suffix of the stack is as follows:
+## expr COMMA
+##
+
+Expected an expression (or a closing paren for a single-argument tuple).
```
peblair:
```suggestion
Expected an expression (or a closing parenthesis for a single-argument tuple).
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+##
+## WARNING: This example involves spurious reductions.
+## This implies that, although the LR(1) items shown above provide an
+## accurate view of the past (what has been recognized so far), they
+## may provide an INCOMPLETE view of the future (what was expected next).
+## In state 3, spurious reduction of production nonempty_list(eol) -> EOL
+## In state 5, spurious reduction of production eols -> nonempty_list(eol)
+##
+
+Expected a closing paren.
```
peblair:
```suggestion
Expected a closing parenthesis.
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## accurate view of the past (what has been recognized so far), they
+## may provide an INCOMPLETE view of the future (what was expected next).
+## In state 41, spurious reduction of production non_assign_expr -> simple_expr
+## In state 120, spurious reduction of production non_binop_expr -> non_assign_expr
+## In state 64, spurious reduction of production annotated_expr -> non_binop_expr
+## In state 238, spurious reduction of production non_stmt_expr -> annotated_expr
+## In state 60, spurious reduction of production expr -> non_stmt_expr
+##
+
+Expected a closing paren (for a grouped expression) or a comma (for a tuple expression).
```
peblair:
```suggestion
Expected a closing parenthesis (for a grouped expression) or a comma (for a tuple expression).
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## In state 738, spurious reduction of production option(when_guard) -> when_guard
+##
+
+Expected an arrow (=>) followed by an expression.
+
+program: MATCH LPAREN WASMI64 RPAREN LBRACE WASMI64 WHILE
+##
+## Ends in an error in state: 735.
+##
+## match_branch -> patte
```
peblair:
```suggestion
Expected a closing parenthesis after the match value.
```

peblair:
```suggestion
Expected an open parenthesis after the 'match' keyword.
```

peblair: Maybe should add this parenthetical to all other instances of "type annotation" in this file

peblair:
```suggestion
Expected a closing parenthesis to complete the infix operator reference.
```

peblair:
```suggestion
Expected another record field definition (or just a '}' to complete the record declaration).
```

peblair:
```suggestion
Expected a comma followed by another record field definition (or just a '}' to complete the record declaration).
```

peblair:
```suggestion
Expected another type variable (or a closing angle bracket to complete the type arguments).
```

peblair:
```suggestion
Expected a comma followed by more type variables or a closing angle bracket to complete the type arguments.
```

peblair:
```suggestion
Expected a closing angle bracket.
```

peblair:
```suggestion
Expected a comma-separated list of type variables surrounded by '<>', or a '{' to list record fields.
```

peblair: See above about "type expression", though this one bothers me less for some reason

peblair: "type expression"

peblair: "type expression"

(in spite of disclaimer on previous line)

peblair:
```suggestion
Expected a closing parenthesis to complete the function type parameters.
```

peblair:
```suggestion
Expected an identifier for a single-argument function, or a comma-separated list of types surrounded by parentheses.
```

peblair:
```suggestion
Expected a type expression (or an immediate closing parenthesis to complete the tuple type).
```
(also, "type expression")

peblair:
```suggestion
Expected a comma followed by another type expression or a closing parenthesis to complete the tuple type.
```
(also, "type expression")

peblair: "type expressions"

peblair:
```suggestion
Expected a closing parenthesis (for a grouped type) or a comma followed by more types (for a tuple type).
```

peblair: "type expression"

peblair: "type expressions"

peblair:
```suggestion
Expected a closing angle bracket to complete the type expression.
```
also, "type expression"

peblair:
```suggestion
Expected a closing parenthesis to complete the function call.
```

peblair:
```suggestion
Expected a comma-separated list of expressions or a closing parenthesis to complete the function call.
```

peblair:
```suggestion
Expected a comma followed by more expressions or a closing parenthesis to complete the function call.
```

peblair:
```suggestion
Expected a closing parenthesis after the while loop condition.
```

peblair:
```suggestion
Expected an open parenthesis followed by an expression for the while loop condition.
```

peblair (commented):
Did a review of the error messages. Great work! See review comments, but one general comment: I think we should refrain from having "paren" in error messages in place of "parenthesis". I don't see a reason to abbreviate.

peblair:
The latter bit of my review comments got broken? GitHub is showing the diffs based on the wrong line (at least locally for me)...the line numbers are correct, but the diffs are not

compiler/dune-project
```diff
@@ -110,7 +107,7 @@
     (reason (>= 3.6.0))
     (ppx_sexp_conv (>= 0.14.0))
     (sexplib (>= 0.14.0))
-    (grain_dypgen (= 0.1))
+    (menhir (= 20211012))
```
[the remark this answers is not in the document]
phated: Still need to accept this

compiler/package.json
```diff
@@ -23,6 +23,10 @@
     "test": "esy test",
     "format": "esy format",
     "check-format": "esy check-format",
+    "parser:interpret": "esy b menhir src/parsing/parser.mly --unused-tokens --interpret",
+    "parser:interpret-error": "esy b menhir src/parsing/parser.mly --unused-tokens --interpret-error",
+    "parser:list-errors": "esy b menhir src/parsing/parser.mly --unused-tokens --list-errors > src/parsing/parser.messages.generated",
+    "parser:check-errors": "yarn parser:list-errors && esy b menhir src/parsing/parser.mly --unused-tokens --compare-errors src/parsing/parser.messages.generated --compare-errors src/parsing/parser.messages",
```
phated, line 29: Question about this. Should we move these commands into the esy.json file or keep them here?

compiler/src/parsing/ast_helper.re
```diff
@@ -64,7 +52,7 @@ module Const = {
 
 module Typ = {
   let mk = (~loc=?, d) => {
-    let loc = Option.value(~default=default_loc_src^(), loc);
+    let loc = Option.value(~default=Location.dummy_loc, loc);
```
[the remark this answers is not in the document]
phated: Did you decide if you'd do that here or not?

compiler/src/parsing/ast_helper.re
```diff
@@ -220,17 +229,54 @@ module Exp = {
     mk(~loc?, ~attributes?, PExpLambda(a, b));
   let apply = (~loc=?, ~attributes=?, a, b) =>
     mk(~loc?, ~attributes?, PExpApp(a, b));
+  let binop = (~loc=?, ~attributes=?, a, b) => {
```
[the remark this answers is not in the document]
phated: Comment still please 🙏

compiler/src/parsing/driver.re
```diff
@@ -2,79 +2,120 @@
 open Lexing;
 open Location;
 
-type error =
-  | AmbiguousParse(list(Parsetree.parsed_program))
-  | NoValidParse;
+let apply_filename_to_lexbuf = (name, lexbuf) => {
```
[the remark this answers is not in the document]
phated: Still need this change, I believe.

compiler/src/parsing/driver.re
```diff
@@ -2,79 +2,120 @@
...
+  | Some(I.Element(s, _, _, _)) => I.number(s)
+  | None =>
+    // The parser hasn't left its initial state. The state number is probably
+    // zero, but it's not guaranteed. A future version of Menhir is supposed
+    // to provide a real solution for this.
+    0
+  };
+
+let show = (text, positions) =>
+  E.extract(text, positions) |> E.sanitize |> E.compress |> E.shorten(20);
```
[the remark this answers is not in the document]
phated: What do we want it to be? Why shorten at all?

compiler/src/parsing/driver.re
```diff
@@ -2,79 +2,120 @@
...
+
+let get = (text, checkpoint, i) =>
+  switch (I.get(i, env(checkpoint))) {
+  | Some(I.Element(_, _, pos1, pos2)) => show(text, (pos1, pos2))
+  | None => assert(false)
+  };
+
+let succeed = _v => assert(false);
```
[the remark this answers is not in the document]
phated: Interesting. Should we at least add a `failwith "Impossible by:"`?

compiler/src/parsing/driver.re
```diff
@@ -2,79 +2,120 @@
...
+      Lexing.from_string(source),
+      source,
     );
-  let loc_end = lexbuf.lex_curr_p;
-  let startpos = {loc_start, loc_end, loc_ghost: true};
-  switch (Parser.parse_program(Lexer.token, lexbuf)) {
-  | [] => raise(Error(startpos, NoValidParse))
-  | [(x, _)] => {...x, comments: Lexer.consume_comments()}
-  | parses => raise(Error(startpos, AmbiguousParse(List.map(fst, parses))))
+    assert(false);
```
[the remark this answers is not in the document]
phated: Should we have a `failwith "Impossible by:"`?

compiler/src/parsing/parser.mly
phated: I like that suggestion! I think the issue was that we were introducing a new let-binding in the block

compiler/package.json
ospencer: I would say you know best! If you think we should move it then I will.

compiler/src/parsing/parser.mly
phated: I have no idea what I was seeing then

compiler/src/parsing/ast_helper.re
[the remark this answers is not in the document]
ospencer: Yeah, decided to do it in another PR. Should I make an issue?

compiler/src/parsing/parser.mly
phated: Let's do the same thing as above where we don't introduce a let-binding

compiler/test/suites/lists.re
```diff
@@ -19,9 +19,4 @@ describe("lists", ({test}) => {
   assertSnapshot("list1_trailing", "[1, 2, 3,]");
   assertSnapshot("list1_trailing_space", "[1, 2, 3, ]");
   assertCompileError("invalid_empty_trailing", "[,]", "Error: Syntax error");
-  assertCompileError(
-    "invalid_list_spread_trailing",
-    "let a = [3, 4]; [1, 2, ...a,]",
-    "Error: Syntax error",
-  );
```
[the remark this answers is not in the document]
phated: What does it look like as a positive test?

compiler/test/suites/types.re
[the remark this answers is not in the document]
phated: Can we keep the curlies using the 2nd example and add a TODO (issue?) about updating the test when we change to double colons (we should really change to double colons in 0.5...)

compiler/test/suites/types.re
```diff
@@ -54,9 +53,8 @@ describe("aliased types", ({test}) => {
     "type_alias_6",
     {|
       type Foo<a> = (String, List<a>)
-      let foo = (x: (String, List<Number>)) => {
+      let foo = (x: (String, List<Number>)) =>
         x: Foo<Number>
-      }
```
phated, line 59: Same as above

compiler/test/suites/types.re
```diff
@@ -66,9 +64,8 @@ describe("aliased types", ({test}) => {
     {|
       type Bar = Number
       type Foo = Bar
-      let foo = (x: Foo) => {
+      let foo = (x: Foo) =>
         x: Bar
-      }
```
phated, line 71: Same as above

compiler/src/parsing/ast_helper.re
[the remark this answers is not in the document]
phated: Yes, issue please 🙏

compiler/package.json
phated: I don't have a strong feeling about it. I think all of the other ones just proxy to the esy.json scripts

compiler/src/parsing/ast_helper.re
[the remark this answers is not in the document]
ospencer: https://github.com/grain-lang/grain/issues/1106

compiler/src/parsing/driver.re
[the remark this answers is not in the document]
ospencer: I'm not sure what we want to it be. We can play around with it though... generally we'd want to shorten because I believe this is token-based, and it could get awkward if your token was very long (like a long string, etc.)

compiler/test/suites/lists.re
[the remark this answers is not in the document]
ospencer: Like this:
```
assertSnapshot("list_spread_trailing", "let a = [3, 4]; [1, 2, ...a, ]");
```

compiler/src/parsing/parser.mly
ospencer:
```suggestion
  | const { Exp.constant ~loc:(to_loc (snd $1)) (fst $1) }
```

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## In state 41, spurious reduction of production non_assign_expr -> simple_expr
+## In state 120, spurious reduction of production non_binop_expr -> non_assign_expr
+## In state 64, spurious reduction of production annotated_expr -> non_binop_expr
+## In state 238, spurious reduction of production non_stmt_expr -> annotated_expr
+## In state 60, spurious reduction of production expr -> non_stmt_expr
+## In state 839, spurious reduction of production toplevel_stmt -> expr
+## In state 950, spurious reduction of production toplevel_stmts -> toplevel_stmt
+##
+
+Expected an infix operator to continue the expression or a newline character to terminate it.
```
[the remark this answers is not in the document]
ospencer: This one is kind of weird. It's really a matter of those "spurious reductions". We could instead just be like "this doesn't make sense" or try an %on_error_reduce here.

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## WARNING: This example involves spurious reductions.
+## This implies that, although the LR(1) items shown above provide an
+## accurate view of the past (what has been recognized so far), they
+## may provide an INCOMPLETE view of the future (what was expected next).
+## In state 3, spurious reduction of production nonempty_list(eol) -> EOL
+## In state 5, spurious reduction of production eols -> nonempty_list(eol)
+## In state 89, spurious reduction of production comma -> COMMA eols
+##
+
+Expected an identifier to exclude from the export.
```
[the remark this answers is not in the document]
ospencer: That's correct, yeah.

compiler/src/parsing/parser.messages
```diff
@@ -0,0 +1,7448 @@
...
+## IF lparen expr rparen eols expr ELSE eols
+##
+## WARNING: This example involves spurious reductions.
+## This implies that, although the LR(1) items shown above provide an
+## accurate view of the past (what has been recognized so far), they
+## may provide an INCOMPLETE view of the future (what was expected next).
+## In state 3, spurious reduction of production nonempty_list(eol) -> EOL
+## In state 5, spurious reduction of production eols -> nonempty_list(eol)
+##
+program: IF LPAREN WASMI64 RPAREN EOL WASMI64 ELSE WHEN
```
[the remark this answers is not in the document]
ospencer: Not sure if you're joking or not 😅 

Yeah, `when` is only allowed in pattern matches.

compiler/src/parsing/driver.re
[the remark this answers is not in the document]
phated: Alright. I'm fine leaving it as 20 for now. Do you want to open a tracking issue mentioning that it might be adjusted?

phated (approved):
Absolutely heroic effort on this @ospencer! The review process has also been great (amazing description, error messages in gdoc, etc). Thank you 🙇

phated:
@ospencer One additional thing I'd like is for you to grep for any TODOs with the issue numbers you are closing and either fix or create a new issue to update that code path. For example, the formatter has a workaround for trailing type annotations (though it looks like it references #866 which then references the #783 you are closing).

peblair (approved):
Fantastic work!

compiler/src/parsing/driver.re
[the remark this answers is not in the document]
ospencer: Do you think it's worth it? I think's probably fine as-is unless we decide that we want to mess around with it.

ospencer:
@phated greps are done and all references to those issues are complete, going to do https://github.com/grain-lang/grain/issues/1107 right after this lands

Issue #692, opened 2021-06-04
Use a type alias for `RERange` in regular expression module


peblair:
At the moment, we denote the contents of `RERange` as a `List<(Number, Number)>`. Because this type is used in many places throughout the `regex.gr` module, it would be better to use a type alias here and add more annotations as appropriate.
```grain
type CharRangeElt = Number
type CharRange = List<(CharRangeElt, CharRangeElt)>

// later in the file
let rec rangeAdd = (rng: CharRange, v: CharRangeElt) => {...}
```

PR #1036, opened 2021-11-25
chore(stdlib)!: Add Types Aliases to regex lib

spotandjake:
closes: #692 .

### DIFF
--- a/stdlib/regex.gr
+++ b/stdlib/regex.gr
@@ -168,6 +168,16 @@ let moreN = (buf: RegExBuf, n) => {
 
 // END Parsing internals for recursive descent
 
+/*
+
+=================================
+REGEX TYPE DEFINITIONS
+=================================
+
+*/
+type RERange = List<(Number, Number)>
+type CharRangeElt = Number
+type CharRange = List<(CharRangeElt, CharRangeElt)>
 
 /*
 
@@ -179,9 +189,7 @@ Based on https://github.com/racket/racket/blob/0a9c70e95a69743dd5d219a395e995be4
 
 */
 
-// [TODO] alias type RERange as List<(Number, Number)>
-
-let rangeInvert = (rng, limitC) => {
+let rangeInvert = (rng: CharRange, limitC) => {
   let rec help = (rng, start) => {
     match(rng) {
       [] when start > limitC => [],
@@ -192,7 +200,7 @@ let rangeInvert = (rng, limitC) => {
   help(rng, 0)
 }
 
-let rec rangeContains = (rng, v) => {
+let rec rangeContains = (rng: CharRange, v: CharRangeElt) => {
   match(rng) {
     [] => false,
     [(start, end), ..._] when (start <= v) && (v <= end) => true,
@@ -200,7 +208,7 @@ let rec rangeContains = (rng, v) => {
   }
 }
 
-let rec rangeAdd = (rng, v) => {
+let rec rangeAdd = (rng: CharRange, v: CharRangeElt) => {
   match(rng) {
     _ when rangeContains(rng, v) => rng,
     _ => rangeUnion(rng, [(v, v)])
@@ -226,18 +234,18 @@ rangeUnion = (rng1, rng2) => {
   }
 }
 
-let rangeAddSpan = (rng, fromC, toC) => {
+let rangeAddSpan = (rng: CharRange, fromC, toC) => {
   rangeUnion(rng, [(fromC, toC)])
 }
 
-let rangeSingleton = (rng) => {
+let rangeSingleton = (rng: CharRange) => {
   match(rng) {
     [(c1, c2)] when c1 == c2 => Some(c1),
     _ => None
   }
 }
 
-let rec rangeIncludes = (rng, lo, hi) => {
+let rec rangeIncludes = (rng: CharRange, lo, hi) => {
   match(rng) {
     [] => false,
     [(c1, c2), ...tl] when lo > c2 => rangeIncludes(tl, lo, hi),
@@ -245,7 +253,7 @@ let rec rangeIncludes = (rng, lo, hi) => {
   }
 }
 
-let rec rangeWithin = (rng, lo, hi) => {
+let rec rangeWithin = (rng: CharRange, lo, hi) => {
   match(rng) {
     [] => true,
     [(c1, _), ..._] when c1 < lo => false,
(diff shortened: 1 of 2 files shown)


stdlib/regex.gr
```diff
@@ -387,7 +395,7 @@ enum ParsedRegularExpression {
   RELookbehind(ParsedRegularExpression, Bool, Box<Number>, Box<Number>, Number, Number), // regex, is-match, lb-min, lb-max, n-start, num-n (lb-xx values patched in later)
   RECut(ParsedRegularExpression, Number, Number, Bool), // regex, n-start, num-n, needs-backtrack
   REReference(Number, Bool), // n, case-sensitive
```
phated, line 397: Question for the team: Would we want things like `IsMatch` and `CaseSensitive` to be type aliases?

stdlib/regex.md
```diff
@@ -5,7 +5,7 @@ title: Regex
 Regular Expressions.
 
 <details disabled>
-<summary tabindex="-1">Added in <code>next</code></summary>
+<summary tabindex="-1">Added in <code>0.4.3</code></summary>
```
phated, line 8: @spotandjake We will want these in a separate PR

stdlib/regex.gr
ospencer: I think it'd be helpful!

spotandjake:
Alright Requested Changes Have Been Made

phated:
@spotandjake The suggestion was not only to change those 2 specifically. Those variants contain comments after each that describe what the random `Bool`/`Number`/etc types represent. We want to have type aliases for **all** of them so the comments aren't necessary.

spotandjake:
> @spotandjake The suggestion was not only to change those 2 specifically. Those variants contain comments after each that describe what the random `Bool`/`Number`/etc types represent. We want to have type aliases for **all** of them so the comments aren't necessary.

Ah think I fixed my mistake there

stdlib/regex.gr
```diff
@@ -378,17 +402,17 @@ enum ParsedRegularExpression {
...
-  RELookahead(ParsedRegularExpression, Bool, Number, Number), // regex, is-match, n-start, num-n
-  RELookbehind(ParsedRegularExpression, Bool, Box<Number>, Box<Number>, Number, Number), // regex, is-match, lb-min, lb-max, n-start, num-n (lb-xx values patched in later)
-  RECut(ParsedRegularExpression, Number, Number, Bool), // regex, n-start, num-n, needs-backtrack
-  REReference(Number, Bool), // n, case-sensitive
-  RERange(List<(Number, Number)>),
-  REUnicodeCategories(List<UnicodeCategory>, Bool) // symlist, true=match/false=does-not-match
+  RESequence(SeqElts, NeedsBackTrack), // seq elts, needs backtrack
+  REGroup(ParsedRegularExpression, GroupID),
+  RERepeat(ParsedRegularExpression, Min, Max, Greedy), // regex, min, max (None for infinity), true=non-greedy
+  REMaybe(ParsedRegularExpression, Greedy), // regex, true=non-greedy
```
phated, line 408:
```suggestion
  RERepeat(ParsedRegularExpression, Min, Max, NonGreedy), // regex, min, max (None for infinity), true=non-greedy
  REMaybe(ParsedRegularExpression, NonGreedy), // regex, true=non-greedy
```
The comment says `true=non-greedy`, so I believe this should be called NonGreedy?

phated:
I also want to get @peblair's eyes on this since he opened the issue and wrote rege

stdlib/regex.gr
peblair: I disagree. IMO, the primary point of type aliases is to make certain values more clearly non-fungible with others. For something like `type RERange = List<(Number, Number)>`, it makes total sense to use an alias, since we want these to be a list with specific semantics. For `IsMatch` and `CaseSensitive`, there are no distinct semantics that these aliases are representing. In fact, I would go as far as to say that aliasing them is _harmful_, since now if I, as a user, see an `IsMatch` type in a signature somewhere, I have to go figure out what that is, only to discover that it is a boolean. The alias sort of seems redundant to me.

peblair (changes requested):
@spotandjake Thanks for this PR! I spoke about the above conversation with the team, and I think we want to pare back down some of the suggested type aliases. If you could revert the following, then I think this is in a place to get merged:
- 001eeacc514086c72b9782a02034012944c334e1 
- 3bac01fefe73fd1208d7cb3ee6524676a03e8bc7 (**except** we _do_ want to retain the `@next` changes you made in this commit)
- 6788da6fd26aac86858e325ce0272625d15f7d45

I'm pretty sure that'd cover everything. With that, I'm happy to give this a final scan and give this a final approval.

spotandjake:
@peblair feel free to review it now i reverted those changes

peblair (approved):
Thank you!

stdlib/regex.gr
```diff
@@ -2597,7 +2604,7 @@ record RegularExpression {
  *
  * @example Regex.make("(foo|bar)[0-9]+")
  *
- * @since 0.4.3
+ * @since next
```
phated, line 2607: Mistake

spotandjake: oops fixed now

jozanza (approved):

stdlib/regex.gr
```diff
@@ -387,7 +394,7 @@ enum ParsedRegularExpression {
   RELookbehind(ParsedRegularExpression, Bool, Box<Number>, Box<Number>, Number, Number), // regex, is-match, lb-min, lb-max, n-start, num-n (lb-xx values patched in later)
   RECut(ParsedRegularExpression, Number, Number, Bool), // regex, n-start, num-n, needs-backtrack
   REReference(Number, Bool), // n, case-sensitive
-  RERange(List<(Number, Number)>),
+  RERange(RERange),
```
phated, line 397: How is `RERange(RERange)` used? Do consumers need to have access to the list?

marcusroberts (approved):
LGTM!

phated (approved):
Approving, assuming the answer to my question is "no, consumers don't need access to it"

stdlib/regex.gr
ospencer: Nah, they don't.

ospencer (approved):
Thanks again @spotandjake!

PR #1039, opened 2021-11-29
fix(compiler): Properly apply bindings when executing match guard

### LINKED ISSUE #1023: List Pattern-matching abnormalities
In the below example, the tokenizeBoolean1 will succeed and the tokenizeBoolean2 will fail based purely on the ordering of the match statements
```
let tokenizeBoolean1 = charlist => {
    match (charlist) {
      ['f', 'a', 'l', 's', 'e', ...rest] => false,
      ['t', 'r', 'u', 'e', ...rest] => true,
      _ => fail "Illegal Boolean",
    }
}
let tokenizeBoolean2 = charlist => {
    match (charlist) {
      ['t', 'r', 'u', 'e', ...rest] => true,
      ['f', 'a', 'l', 's', 'e', ...rest] => false,
      _ => fail "Illegal Boolean",
    }
}


print(tokenizeBoolean1(['f','a','l','s','e','m','o','r','e']))
print(tokenizeBoolean2(['f','a','l','s','e','m','o','r','e']))
```

ospencer:
Closes #1023 

We were over-optimizing an Explode case—the continuation for when a guard fails needs the bindings that `Explode` introduces.

### DIFF
--- a/compiler/src/middle_end/matchcomp.re
+++ b/compiler/src/middle_end/matchcomp.re
@@ -379,7 +379,7 @@ module MatchTreeCompiler = {
       (Comp.imm(~allocation_type=StackAllocated(WasmI32), Imm.trap()), [])
     /* Optimizations to avoid unneeded destructuring: */
     | Explode(_, Leaf(_) as inner)
-    | Explode(_, Guard(_) as inner)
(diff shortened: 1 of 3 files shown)


phated (approved):
Excellent! Have my enthusiastic approval! 🎊

PR #1042, opened 2021-11-30
fix(grainfmt): Refactor code formatter

marcusroberts:
I've rebuilt the formatter to use a context aware approach to determining comments to print. This saves the pre-scan of locations and comments and has much better performance.

I've also heavily refactored the code to really improve maintainability and now brings consistent formatting to anything that is iterated (e.g. top level statements, block statements, record fields etc). There are some improvements like supporting start of line comments too.


compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,519 @@
+open Grain_diagnostics;
+module Doc = Res_doc;
+
+let rec get_comments_on_line = (line: int, comments) =>
+  if (List.length(comments) == 0) {
+    [];
+  } else {
+    let c: Grain_parsing.Parsetree.comment = List.hd(comments);
```
ospencer, line 8: Maybe it's just a personal thing, but I prefer to have the type annotations on the arguments to the functions (if I have any).

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,519 @@
+open Grain_diagnostics;
+module Doc = Res_doc;
+
+let rec get_comments_on_line = (line: int, comments) =>
+  if (List.length(comments) == 0) {
```
ospencer, line 5: It looks to me like this function is begging to use a pattern match!

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,519 @@
...
+    [];
+  } else {
+    let c: Grain_parsing.Parsetree.comment = List.hd(comments);
+    let c_loc: Grain_parsing.Location.t = Locations.get_comment_loc(c);
+    let (_, cmtline, cmtchar, _) =
+      Locations.get_raw_pos_info(c_loc.loc_start);
+
+    if (cmtline == line) {
+      [c] @ get_comments_on_line(line, List.tl(comments));
```
ospencer, line 14:
```suggestion
      [c, ...get_comments_on_line(line, List.tl(comments))];
```

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,519 @@
...
+    } else {
+      []; // can stop early as there will be no more
+    };
+  };
+
+let rec get_comments_before_line = (line: int, comments) =>
+  if (List.length(comments) == 0) {
```
ospencer, line 21: Same thing with these.

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,519 @@
...
+      [c] @ get_comments_after_line(line, List.tl(comments));
+    } else {
+      get_comments_after_line(
+        line,
+        List.tl(comments) // gotta keep checking
+      );
+    };
+  };
+
+let print_comments = (comments: list(Grain_parsing.Parsetree.comment)) => {
```
ospencer, line 51: Is this used?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,519 @@
...
+  List.map(c => Comments.print_comment(c), comments);
```
ospencer, line 52:
```suggestion
  List.map(Comments.print_comment, comments);
```

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,519 @@
...
+      };
+    cmts;
+  };
+
+let rec get_comments_inside_location =
+        (
+          ~location: Grain_parsing.Location.t,
+          comments: list(Grain_parsing.Parsetree.comment),
+        ) => {
+  let (_, stmtStartLine, stmsStartChar, _) =
```
ospencer, line 91: The camel case strikes again!

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,519 @@
...
+      []; // can stop now
+    } else if (cmteline < stmtStartLine) {
+      get_comments_inside_location(location, List.tl(comments));
+    } else if
+      // need the complex comparisons here
+      (cmtsline > stmtStartLine
+       || cmtsline == stmtStartLine
+       && cmtschar >= stmsStartChar) {
+      if (cmteline < stmtEndLine
+          || cmteline == stmtEndLine
```
ospencer, line 115:
```suggestion
      if (cmteline <= stmtEndLine
```

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,519 @@
...
+  };
+
+  let location: Grain_parsing.Location.t = {
+    loc_start: start_loc,
+    loc_end: end_loc,
+    loc_ghost: true,
+  };
+  let cmts = get_comments_inside_location(~location, comments);
+
+  cmts;
```
ospencer, line 229:
```suggestion
  get_comments_inside_location(~location, comments);
```

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,519 @@
...
```
ospencer, line 263:
```suggestion
  get_comments_inside_location(~location, comments);
```

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,519 @@
...
+    pos_fname: "",
+    pos_lnum: locStartLine,
+    pos_bol: 0,
+    pos_cnum: locStartChar,
+  };
+
+  let location: Grain_parsing.Location.t = {
+    loc_start: start_loc,
+    loc_end: end_loc,
+    loc_ghost: true,
```
ospencer, line 259: Is this really a ghost? It does exist in the file, doesn't it?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,519 @@
...
+  };
+  let cmts = get_comments_inside_location(~location, comments);
+
+  cmts;
```
ospencer, line 299:
```suggestion
  get_comments_inside_location(~location, comments);
```

compiler/grainformat/reformat.re
```diff
@@ -858,30 +963,45 @@ and print_pattern =
...
-      List.append(with_leading, [trailing_comment_docs]);
+      List.append(
+        with_leading,
+        [
+          after_parens_comments_docs,
+          // if (List.length(after_parens_comments) > 0) {
+          //   Doc.Text("bb");
+          // } else {
+          //   Doc.nil;
+          // },
```
ospencer, line 998: What's up with this?

compiler/grainformat/reformat.re
```diff
@@ -948,67 +1068,98 @@ and print_record =
...
+          after_brace_comments,
         ),
+        printed_fields,
         if (List.length(fields) == 1) {
           // TODO: not needed once we annotate with ::
           Doc.comma; //  append a comma as single argument record look like block {data:val}
         } else {
-          Doc.ifBreaks(Doc.text(","), Doc.nil);
+          Doc.nil;
+          //   Doc.ifBreaks(Doc.text(","), Doc.nil);
```
ospencer, line 1162: What's up here?

compiler/grainformat/reformat.re
```diff
@@ -3286,258 +3614,193 @@ let print_primitive_value_description =
...
+    );
 
+  let final_doc = Doc.concat([top_level_stmts, Doc.hardLine]);
+  // Doc.debug(final_doc);
   Doc.toString(~width=80, final_doc);
 };
+
+//   // Use this to check the generated output
+//   //Doc.debug(final_doc);
+//   //
```
ospencer, line 3806: Maybe we can clean this up a bit?

compiler/grainformat/reformat.re
```diff
@@ -3286,258 +3614,193 @@ let print_primitive_value_description =
...
+        stmt => {
+          let attributes = stmt.ptop_attributes;
+          print_attributes(attributes);
+        },
+      ~isBlock=true,
+      ~allComments=parsed_program.comments,
+    );
 
+  let final_doc = Doc.concat([top_level_stmts, Doc.hardLine]);
+  // Doc.debug(final_doc);
```
ospencer, line 3800:
```suggestion
```

compiler/test/formatter_outputs/application.gr
```diff
@@ -6,7 +6,10 @@ let data = [>]
 
 let busNumbers = Array.map(
   ((num, idx)) => {
-    (Option.expect("Expected a number", None), idx)
+    (
```
ospencer, line 9: Did this really need to break? That looks like it's all within length.

compiler/test/formatter_outputs/comments.gr
```diff
@@ -27,7 +27,7 @@ export let slice = (startIndex, endIndex, array) => {
 
 // leading
 
-if (/*before*/ true /*whyt*/) {
+if ( /*before*/true /*whyt*/) {
```
ospencer, line 30: Hmm, this space looks wrong here.

compiler/test/formatter_outputs/comments.gr
```diff
@@ -64,7 +62,7 @@ if (a /*why*/) {
   // inner comment
   5 // trail 5
 
-  6 /* block 6a */ /* block 6 b */
+  6 /* block 6a *//* block 6 b */
```
ospencer, line 65: I think generally I'd want a space here?

compiler/grainformat/comment_utils.re
marcusroberts: Good point, adding them!

compiler/grainformat/comment_utils.re
marcusroberts: It is, and that's done

compiler/grainformat/comment_utils.re
marcusroberts: Moved to the debug utils file

compiler/grainformat/comment_utils.re
marcusroberts: Beautiful simplification!

compiler/grainformat/comment_utils.re
marcusroberts: That damn camel!   I've gone through and removed them all I think

compiler/grainformat/comment_utils.re
marcusroberts: Left over from debugging.  I'll fix

compiler/grainformat/comment_utils.re
marcusroberts: You know, I'm not sure.  The logic  I want is if cmteline < stmtEndLine I don't care about cmtechar, but if cmteline = stmtEndLine I do.   As the formatter took away my parens I think that's what the operator priority gives with the following && but I need to work that out.

compiler/grainformat/comment_utils.re
marcusroberts: It does.  I made it true because I manufactured it, but it does exist.  Switched to false everywhere

compiler/grainformat/reformat.re
marcusroberts: Well that's weird.   It was some layout debugging.

compiler/grainformat/reformat.re
marcusroberts: Cleansed

compiler/grainformat/comment_utils.re
marcusroberts: Nice, fixed.

compiler/grainformat/comment_utils.re
marcusroberts: And done throughout

compiler/test/formatter_outputs/application.gr
marcusroberts: Fixed.

compiler/test/formatter_outputs/comments.gr
marcusroberts: Fixed.

compiler/test/formatter_outputs/comments.gr
marcusroberts: Agreed, fixed.

compiler/grainformat/comment_utils.re
marcusroberts: @ospencer  I tested and I think this is right, because the && after is dependent on the first part of the or clause

In English, if the cmt ends on a before  the statement we don't take the end char into account, only if the line is the same do we (because the end char on a line before might be after the start char on a line after) if that makes sense

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,487 @@
...
+      Locations.get_raw_pos_info(c_loc.loc_start);
+
+    let (_, cmteline, cmtechar, _) =
+      Locations.get_raw_pos_info(c_loc.loc_end);
+    if (cmtsline > stmt_end_line) {
+      []; // can stop now
+    } else if (cmteline < stmt_start_line) {
+      get_comments_inside_location(location, remaining_comments);
+    } else if
+      // need the complex comparisons here
```
ospencer, line 47: Can you add why?

compiler/grainformat/reformat.re
```diff
@@ -1522,77 +1627,117 @@ and print_expression =
...
+        );
+      };
+      let (_, bracket_line, _, _) =
+        Locations.get_raw_pos_info(expr.pexp_loc.loc_end);
+      Doc.group(
+        Doc.concat([
+          Doc.lparen,
+          Doc.indent(
+            Doc.concat([
+              //  Doc.softLine,
```
ospencer, line 1652: What's up with this?

compiler/grainformat/reformat.re
```diff
@@ -1872,20 +1985,31 @@ and print_expression =
...
-      let true_false_space =
-        switch (trueExpr.pexp_desc) {
-        | PExpBlock(expressions) => Doc.space
-        | _ => if (false_is_block) {Doc.space} else {Doc.line}
-        };
+      // let true_false_space =
+      //   switch (trueExpr.pexp_desc) {
+      //   | PExpBlock(expressions) => Doc.space
+      //   | _ => if (false_is_block) {Doc.space} else {Doc.line}
+      //   };
```
ospencer, line 1992: What's up with this?

compiler/grainformat/reformat.re
```diff
@@ -2268,146 +2436,67 @@ and print_expression =
...
-        let blockEndCommentDocs =
-          if (List.length(remaining_comments_in_block) > 0) {
-            print_multi_comments(remaining_comments_in_block, previous_line^);
-          } else {
-            Doc.nil;
-          };
-
+      switch (expressions) {
+      | [] =>
+        // I think not legal syntac
```
ospencer, line 2443: Can we resolve this "I think"?

compiler/grainformat/reformat.re
```diff
@@ -2268,146 +2436,67 @@ and print_expression =
...
         Doc.breakableGroup(
           ~forceBreak=true,
           Doc.concat([
             Doc.lbrace,
-            after_brace_comments,
-            Doc.indent(Doc.concat([Doc.line, block, blockEndCommentDocs])),
-            Doc.line,
+            Doc.indent(Doc.line),
+            // Doc.line,
```
ospencer, line 2449: What's up with this?

ospencer (commented):
Code looks reasonable to me! Just a few questions about some comments.

compiler/grainformat/comment_utils.re
marcusroberts: Done

compiler/grainformat/reformat.re
marcusroberts: Removed.

compiler/grainformat/reformat.re
marcusroberts: Have done now

compiler/grainformat/reformat.re
marcusroberts: Removed.

ospencer (approved):
Beautiful! Can you update the PR title to be something like "refactor" or something like that?

phated (changes requested):
I'm going to have to block this from the release 😭 There are some formatting problems that I noticed in #1043 that seem to be caused by some deeply ingrained code and we'll need @marcusroberts around to help us solve those things.

We'll also want to tweak the internal API a little, as I had a really hard time adjusting one-off things that didn't cause cascading problems.

compiler/test/formatter_outputs/ignores.gr
```diff
@@ -11,10 +10,10 @@ let y = [3, 4, 5]
...
   let yy = [1,
-  2]  // comment
+  2] // comment
 
   let zz = [5, 6]
 }
 
 // formatter-ignore
-      let single_ast_line = "x"
+let single_ast_line = "x"
```
phated, line 19: Why did this change if it uses `// formatter-ignore`?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+          ...comments_inner(Some(cmt), bracket_line_opt, rem),
+        ]
+      };
+    }
+  };
+};
+
+let inbetween_comments_to_docs =
+    (
+      ~offset: bool,
```
phated, line 389: I don't like this `~offset` argument. Why can't you just add the space in the calling code if their `offset` variable is true?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+        && cmtschar > stmt_end_char) {
+      [cmt, ...get_comments_after_location(~location, remaining_comments)];
+    } else {
+      get_comments_after_location(~location, remaining_comments);
+    };
+  };
+};
+
+let rec trailing_comments_inner =
+        (prev: option(Parsetree.comment), comments: list(Parsetree.comment)) => {
```
phated, line 434: Why isn't this an optional, named param?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+  let comment_string = Comments.get_comment_source(comment);
+
+  Doc.text(String.trim(comment_string));
+};
+
+let get_after_brace_comments =
+    (
+      loc: Location.t,
+      comments: list(Parsetree.comment),
+      first: option(Location.t),
```
phated, line 277: Why isn't this an optional, named param?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+    if (cmtline > line) {
+      []; // can stop early as there will be no more
+    } else if (cmtline == line && cmtchar >= char) {
+      [c, ...get_comments_on_line_end(line, char, List.tl(comments))];
+    } else {
+      get_comments_on_line_end(line, char, List.tl(comments));
+    };
+  };
+
+let rec get_comments_on_line_start = (line: int, char: int, comments) =>
```
phated, line 230: Should `line` and `char` be named params so there is less chance to screw up positional order?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+  let location: Location.t = {
+    loc_start: start_loc,
+    loc_end: end_loc,
+    loc_ghost: false,
+  };
+  get_comments_inside_location(~location, comments);
+};
+
+let rec get_comments_on_line_end =
+        (line: int, char: int, comments: list(Parsetree.comment)) =>
```
phated, line 213: Should `line` and `char` be named params so there is less chance to screw up positional order?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+    loc_end: end_loc,
+    loc_ghost: false,
+  };
+
+  get_comments_inside_location(~location, comments);
+};
+
+let get_comments_from_start_of_enclosing_location =
+    (
+      ~wrapper: Location.t,
```
phated, line 152: I felt like `wrapper` was a weird name, should this be something like `outer_location` or `location_of_outer` or something?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+      ~location: Location.t,
```
phated, line 153: And then this would be `inner_location` maybe?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+  let location: Location.t = {
+    loc_start: start_loc,
+    loc_end: end_loc,
+    loc_ghost: false,
+  };
+  get_comments_inside_location(~location, comments);
+};
+
+let get_comments_between_locs =
+    (~loc1: Location.t, ~loc2: Location.t, comments: list(Parsetree.comment)) => {
```
phated, line 183: And in that vein, should these be `begin_loc` and `end_loc`?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+
+let nobreak_comment_to_doc = (comment: Parsetree.comment) => {
+  let comment_string = Comments.get_comment_source(comment);
+
+  Doc.text(String.trim(comment_string));
+};
+
+let get_after_brace_comments =
+    (
+      loc: Location.t,
```
phated, line 275: Should this be a named param?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+          cmts;
+        };
+      };
+    }
+  };
+};
+
+let rec comments_inner =
+        (
+          prev: option(Parsetree.comment),
```
phated, line 310: Why isn't this an optional, named param?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+          bracket_line_opt: option(int),
```
phated, line 311: Why isn't this an optional, named param? (and not be named "opt")

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+        ]
+      };
+    }
+  };
+};
+
+let inbetween_comments_to_docs =
+    (
+      ~offset: bool,
+      ~bracket_line: option(int),
```
phated, line 390: Why isn't this an optional param?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+      Doc.space,
+      Doc.join(
+        Doc.space,
+        List.map(c => {nobreak_comment_to_doc(c)}, comments),
+      ),
+    ])
+  };
+
+let rec new_comments_inner =
+        (prev: option(Parsetree.comment), comments: list(Parsetree.comment)) => {
```
phated, line 508: Why isn't this a named, optional param?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+    };
+  };
+};
+
+let rec trailing_comments_inner =
+        (prev: option(Parsetree.comment), comments: list(Parsetree.comment)) => {
+  switch (prev) {
+  | None =>
+    switch (comments) {
+    | [] => [Doc.nil]
```
phated, line 438: What happens if you do `Doc.concat([])`? Could this just be an empty list instead of including a `Doc.nil`?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+      ),
+    ])
+  };
+
+let rec new_comments_inner =
+        (prev: option(Parsetree.comment), comments: list(Parsetree.comment)) => {
+  switch (prev) {
+  | None =>
+    switch (comments) {
+    | [] => [Doc.nil]
```
phated, line 512: What happens if you do `Doc.concat([])`? Could this just be an empty list instead of including a `Doc.nil`?

compiler/grainformat/comment_utils.re
```diff
@@ -0,0 +1,564 @@
...
+    | [cmt, ...rem] => [
+        comment_to_doc(cmt),
+        ...new_comments_inner(Some(cmt), rem),
+      ]
+    }
+  | Some(prev_cmt) =>
+    switch (comments) {
+    | [] => [Doc.nil]
```
phated, line 520: What happens if you do `Doc.concat([])`? Could this just be an empty list instead of including a `Doc.nil`?

compiler/src/parsing/parser.dyp
```diff
@@ -163,7 +163,7 @@ pattern :
   | const { Pat.constant ~loc:(symbol_rloc dyp) $1 }
   /* If the pattern uses an external ID, we know it's a constructor, not a variable */
   | ext_constructor { Pat.construct ~loc:(symbol_rloc dyp) $1 [] }
-  | [ID | special_id | primitive] { Pat.var ~loc:(symbol_rloc dyp) (mkstr dyp $1) }
+  | [ID | special_id | primitive] { Pat.var ~loc:(symbol_rloc dyp) (mkstr (symbol_rloc dyp) $1) }
```
phated, line 166: Can we merge this as a separate PR? I believe it was based on one of @ospencer's PRs?

phated (commented):
A few things. Haven't reviewed the huge reformat.re file yet.

compiler/src/parsing/parser.dyp
marcusroberts: It was, but we decided to merge it with the formatter as it only impacts the formatter.

compiler/test/formatter_outputs/ignores.gr
marcusroberts: Keen eye!    There was a good reason for it to do with locations, I'll have another look though

compiler/grainformat/comment_utils.re
marcusroberts: It works!  Code fixed in next push

compiler/grainformat/comment_utils.re
marcusroberts: Because we only insert if when there are comments, so we'd push the check for empty comments up to the callees, which seems to lead to duplicate code.

compiler/grainformat/comment_utils.re
marcusroberts: Nice, that's changed.

compiler/grainformat/comment_utils.re
marcusroberts: Nice, fixed.

compiler/grainformat/comment_utils.re
marcusroberts: Nice.  Done.

compiler/grainformat/comment_utils.re
marcusroberts: Yup, fixed.

compiler/grainformat/comment_utils.re
marcusroberts: It's the location that encloses or wraps the code and comments we are looking for, but happy to rename

compiler/grainformat/comment_utils.re
marcusroberts: Done.

compiler/grainformat/comment_utils.re
marcusroberts: Yup fixed.

compiler/grainformat/comment_utils.re
marcusroberts: Turns out we didn't need it at all

compiler/grainformat/comment_utils.re
marcusroberts: Not needed, so removed.

compiler/grainformat/comment_utils.re
marcusroberts: It works!

compiler/grainformat/comment_utils.re
marcusroberts: Fixed.

compiler/grainformat/comment_utils.re
marcusroberts: Fixed.

compiler/grainformat/reformat.re
```diff
@@ -1490,586 +2193,692 @@ and print_attributes = attributes =>
...
+      let printed_branches = Doc.join(Doc.hardLine, items);
+
+      let printed_branches_after_brace =
+        Doc.concat([
+          force_break_if_line_comment(after_brace_comments, Doc.line),
+          printed_branches,
+        ]);
+
+      Doc.breakableGroup(
+        ~forceBreak=false,
```
phated: Pretty sure that a `breakableGroup(~forceBreak=false` is just a `group(`

compiler/test/formatter_outputs/ignores.gr
marcusroberts: Oh, because it's really hard (impossible) to add the spacing back.   We don't know how much padding the formatter is adding around where we are inserting.  So you can see it works in the line above but not here.  I can't see how to fix it presently.

compiler/grainformat/reformat.re
```diff
@@ -2114,396 +2930,300 @@ and print_expression =
...
+            | [] =>
               Doc.concat([
-                Doc.softLine,
-                Doc.lparen,
+                Doc.text(":"),
+                Doc.space,
+                print_type(
+                  ~original_source,
+                  ~comments,
+                  ~trailing_separator=false,
```
phated: I want to look at this argument—is it worth it to have this flag or can we just put that code in the paths where it is needed?

compiler/grainformat/reformat.re
```diff
@@ -2114,396 +2930,300 @@ and print_expression =
...
+            ~comments=cleaned_comments,
+            ~get_attribute_text=
+              expr => print_attributes(expr.pexp_attributes),
+            ~original_source,
+            expressions,
+          );
+
+        let start_after_brace =
+          Doc.concat([
+            // force_break_if_line_comment(after_brace_comments, Doc.line),
```
phated: Remove the comment?

compiler/grainformat/reformat.re
```diff
@@ -2114,396 +2930,300 @@ and print_expression =
...
+          remove_used_comments(~all_comments=comments, after_brace_comments);
+
+        let printed_expressions =
+          block_item_iterator(
+            ~previous=Block(expr.pexp_loc),
+            ~get_loc,
+            ~print_item,
+            ~comments=cleaned_comments,
+            ~get_attribute_text=
+              expr => print_attributes(expr.pexp_attributes),
```
phated: introduce a `let` for this func?

compiler/test/formatter_outputs/ignores.gr
marcusroberts: I figured this was the best of a poor choice

compiler/grainformat/reformat.re
```diff
@@ -2646,351 +3317,510 @@ and print_value_bind =
...
+            let function_name = get_function_name(fn);
+
+            if (infixop(function_name)) {
+              Doc.indent(Doc.concat([printed]));
+            } else {
+              printed;
+            };
+
+          | PExpIf(_) =>
+            if (Doc.willBreak(printed)) {
```
phated: I want to look into this

compiler/grainformat/reformat.re
```diff
@@ -2646,351 +3317,510 @@ and print_value_bind =
...
+      ]),
+    );
+
+  | PDataRecord(label_declarations) =>
+    let get_loc = (lbl: Parsetree.label_declaration) => {
+      lbl.pld_loc;
+    };
+
+    let print_item = (lbl: Parsetree.label_declaration, comments) => {
+      let isMutable =
```
phated:
```suggestion
      let is_mutable =
```
snake_case

compiler/grainformat/reformat.re
```diff
@@ -3018,183 +3843,158 @@ let data_print =
...
+                  ~all_comments=comments,
+                  after_brace_comments,
+                );
+
+              let exceptions =
+                block_item_iterator_line(
+                  ~previous=Block(imp.pimp_loc),
+                  ~get_loc,
+                  ~print_item,
+                  ~get_attribute_text=_ => Doc.nil,
```
phated: I wonder if `~get_attribute_text` should be an optional argument, so we don't need these kind of functions

compiler/grainformat/reformat.re
```diff
@@ -3308,258 +4127,177 @@ let print_primitive_value_description =
...
-          ~endChar=None,
+  let without_comments =
+    switch (data.ptop_desc) {
+    | PTopImport(import_declaration) =>
+      import_print(~comments, ~original_source, import_declaration)
+    | PTopForeign(export_flag, value_description) =>
+      let export =
+        switch (export_flag) {
+        | Nonexported => Doc.text("import ")
+        | Exported => Doc.text("export ")
```
phated: Wait, what?

compiler/grainformat/reformat.re
```diff
@@ -3308,258 +4127,177 @@ let print_primitive_value_description =
...
-                  (excpt: Parsetree.export_except) =>
-                    switch (excpt) {
-                    | ExportExceptData(data) => Doc.text(data.txt)
-                    | ExportExceptValue(value) => Doc.text(value.txt)
-                    },
-                  export_excepts,
-                ),
+    | PTopExportAll(export_excepts) =>
+      Doc.concat([
+        Doc.text("export * "),
```
phated: is there always a space after `*`? Seems wrong

compiler/grainformat/reformat.re
```diff
@@ -3308,258 +4127,177 @@ let print_primitive_value_description =
...
+    block_item_iterator(
+      ~previous=TopOfFile,
+      ~get_loc,
+      ~print_item,
+      ~comments=cleaned_comments,
+      ~get_attribute_text=
+        stmt => {
+          let attributes = stmt.ptop_attributes;
+          print_attributes(attributes);
+        },
```
phated: create a `let` binding for this function?

phated (commented):
@marcusroberts This was an absolute **delight** to read through. Awesome work! I had a few more pieces of advice in reformat.re

compiler/grainformat/reformat.re
```diff
@@ -110,95 +108,44 @@ let get_original_code =
...
-    loc_ghost: true,
-  };
-  merged_loc;
-};
+let force_break_if_line_comment =
+    (comments: list(Parsetree.comment), separator) => {
+  switch (comments) {
+  | [] => separator
+  | _ =>
+    let last_comment = List.nth(comments, List.length(comments) - 1);
```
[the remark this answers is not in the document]
marcusroberts: A good question, and I wondered the same.    The number of comments in the list should be short, as this is usually scoped to a statement , expression etc.  Worth changing do you think?

compiler/grainformat/reformat.re
```diff
@@ -466,147 +203,861 @@ let prefixop = (op: string) => {
...
-  let items =
-    List.map(
-      i =>
-        switch (i) {
-        | RegularPattern(e) =>
-          last_item_was_spread := false;
+let remove_used_comments =
+    (
+      ~all_comments: list(Parsetree.comment),
+      remove_comments: list(Parsetree.comment),
```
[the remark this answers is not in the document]
marcusroberts: This will usually be a small list we are manipulating, but I take your point.  It's almost always removing the comment we wrote after the opening brace from the remaining comments which I could optimise on

compiler/grainformat/reformat.re
```diff
@@ -466,147 +203,861 @@ let prefixop = (op: string) => {
...
-                ~parentIsArrow=false,
-                ~endChar=None,
-                ~original_source,
-                ~parent_loc,
-                e,
-              ),
-            ]),
+
+        if (last_comment_line == this_line) {
+          Doc.space;
```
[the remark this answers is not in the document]
marcusroberts: Usually I've used Doc.space intentionally to make sure it doesn't break for example we never want to break before a block opening, after the => etc

compiler/grainformat/reformat.re
```diff
@@ -110,95 +108,44 @@ let get_original_code =
...
-
-  let merged_loc: Grain_parsing.Location.t = {
-    loc_start: loc.loc_end,
-    loc_end: endpos,
-    loc_ghost: true,
-  };
-  merged_loc;
-};
+let force_break_if_line_comment =
+    (comments: list(Parsetree.comment), separator) => {
```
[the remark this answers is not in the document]
marcusroberts: Fixed.

compiler/grainformat/reformat.re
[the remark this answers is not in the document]
marcusroberts: Labelled comments fixed.

compiler/grainformat/reformat.re
```diff
@@ -466,147 +203,861 @@ let prefixop = (op: string) => {
...
-          },
-        ]),
-      ),
-      Doc.softLine,
-      Doc.rbracket,
-    ]),
-  );
-}
+  | PreviousItem(prev) =>
+    let (_, last_stmt_line, _, _) = Locations.get_raw_pos_info(prev.loc_end);
```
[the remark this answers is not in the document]
marcusroberts: Good call.

compiler/grainformat/reformat.re
marcusroberts: Some habits die hard :-).  Fixed.

compiler/grainformat/reformat.re
```diff
@@ -466,147 +203,861 @@ let prefixop = (op: string) => {
...
+    let this_loc = get_loc(item);
+    let (_, this_line, this_char, _) =
+      Locations.get_raw_pos_info(this_loc.loc_start);
+
+    let after_comments_break =
+      switch (leading_comments) {
+      | [] => Doc.nil
+      | in_comments =>
+        let last_comment =
+          List.nth(in_comments, List.length(in_comments) - 1);
```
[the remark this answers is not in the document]
marcusroberts: Moved to a helper.  Makes me a bit uneasy as inline it's guarded to ensure the list isn't empty, but I've put a comment to the effect of be careful when using it.

compiler/grainformat/reformat.re
```diff
@@ -466,147 +203,861 @@ let prefixop = (op: string) => {
...
+        Doc.concat([
+          bcb,
+          leading_comment_docs,
+          after_comments_break,
+          print_item(item, item_comments),
+          Doc.ifBreaks(
+            switch (separator) {
+            | None => Doc.nil
+            | Some(sp) => sp
+            },
```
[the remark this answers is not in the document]
marcusroberts: Nice!

compiler/grainformat/reformat.re
```diff
@@ -466,147 +203,861 @@ let prefixop = (op: string) => {
...
+          ),
+        ]);
+      };
+    };
+  };
+};
+
+let rec item_iterator =
+        (
+          ~previous: option(Location.t),
```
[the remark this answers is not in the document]
marcusroberts: It is now :)

compiler/grainformat/reformat.re
```diff
@@ -466,147 +203,861 @@ let prefixop = (op: string) => {
...
+          Doc.text("..."),
+          print_pattern(~original_source, ~comments, ~next_loc, e),
+        ]),
+      )
+    };
+  };
+
+  let items =
+    item_iterator(
+      ~previous=None,
```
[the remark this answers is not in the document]
marcusroberts: Done.

compiler/grainformat/reformat.re
```diff
@@ -466,147 +203,861 @@ let prefixop = (op: string) => {
...
+      Doc.ifBreaks(Doc.comma, Doc.nil),
+      Doc.softLine,
+      Doc.rbracket,
+    ]),
+  );
+}
+
+and resugar_pattern_list_inner = (patterns: list(Parsetree.pattern)) => {
+  let arg1 = List.nth(patterns, 0);
+  let arg2 = List.nth(patterns, 1);
```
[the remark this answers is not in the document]
marcusroberts: Nice, done

compiler/grainformat/reformat.re
```diff
@@ -466,147 +203,861 @@ let prefixop = (op: string) => {
...
+      switch (innerfunc.txt) {
+      | IdentName(name) => name
+      | _ => ""
+      };
+
+    if (func == "[]") {
+      [RegularPattern(arg1)];
+    } else if (func == list_cons) {
+      let inner = resugar_pattern_list_inner(innerpatterns);
+      List.append([RegularPattern(arg1)], inner);
```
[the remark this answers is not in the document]
marcusroberts: One I always forget, but definitely nicer!

compiler/grainformat/reformat.re
```diff
@@ -466,147 +203,861 @@ let prefixop = (op: string) => {
...
+          );
+        },
+      processed_list,
+    );
+
+  Doc.group(
+    Doc.concat([
+      Doc.lbracket,
+      Doc.indent(
+        Doc.concat([
```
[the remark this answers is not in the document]
marcusroberts: Nice catch!

compiler/grainformat/reformat.re
```diff
@@ -651,100 +1093,107 @@ and resugar_list_inner =
...
       }
-    };
+    }
+  | _ =>
+    // Grain syntax makes it impossible to construct a list cons without
+    // two arguments, but we'll check just to make sure
+    raise(
+      Error(Illegal_parse("List cons should always have two expressions")),
+    )
+  }
```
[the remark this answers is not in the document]
marcusroberts: Done.

compiler/grainformat/reformat.re
```diff
@@ -651,100 +1093,107 @@ and resugar_list_inner =
...
-                Grain_parsing__Location.loc(Grain_parsing__Identifier.t),
-                Grain_parsing__Parsetree.pattern,
-              ),
-            ) => {
-              let (loc, pat) = patternloc;
+  let print_item =
+      (
+        patternloc: (Location.loc(Identifier.t), Parsetree.pattern),
+        comments,
+      ) => {
```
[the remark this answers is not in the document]
marcusroberts: Yep, done

compiler/grainformat/reformat.re
```diff
@@ -1075,43 +1556,466 @@ and print_type =
...
+          Doc.indent(Doc.concat([Doc.line, body])),
+        ]);
+      };
+    };
+
+  | _ => raise(Error(Illegal_parse("Called on a non-lambda")))
+  };
+}
+
+and print_arg = (arg: Parsetree.expression, ~original_source, ~comments) => {
```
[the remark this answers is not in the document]
marcusroberts: Fixed.

compiler/grainformat/reformat.re
```diff
@@ -1075,43 +1556,466 @@ and print_type =
...
+      print_arg_lambda(~comments, ~original_source, callback);
+
+    Doc.group(printed_callback);
+
+  | [callback, expr] =>
+    let printed_callback =
+      print_arg_lambda(~comments, ~original_source, callback);
+    let printed_arg = print_arg(~comments, ~original_source, expr);
+
+    if (Doc.willBreak(printed_callback)) {
```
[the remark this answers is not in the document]
marcusroberts: I'm still convinced it's needed as the ifBreaks won't work here, but if you can work some magic, and get rid of this, that would be good.

compiler/grainformat/reformat.re
```diff
@@ -1075,43 +1556,466 @@ and print_type =
...
+      print_arg_lambda(~comments, ~original_source, last_expression);
+
+    let remainderArr =
+      Array.sub(Array.of_list(args), 0, List.length(args) - 1);
+
+    let printed_args =
+      Doc.join(
+        Doc.concat([Doc.comma, Doc.line]),
+        List.map(
+          print_arg(~comments, ~original_source),
```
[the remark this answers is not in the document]
marcusroberts: I think print_arg is consistently used isn't it, apart from where there's a single arg so I apply it directly.
Did you mean other List.maps ?

compiler/grainformat/reformat.re
```diff
@@ -1125,207 +2029,107 @@ and print_application =
...
-  | _ when prefixop(function_name) || infixop(function_name) =>
+    print_infix_application(~expressions, ~original_source, ~comments, func)
+  | _ when infixop(function_name) =>
     raise(Error(Illegal_parse("Formatter error, wrong number of args ")))
   | _ =>
     if (function_name == list_cons) {
-      resugar_list(~expressions, ~parent_loc, ~original_source);
+      resugar_list(~original_source, ~comments, expressions);
     } else if (Array.exists(fn => function_name == fn, exception_primitives)) {
+      let first_expr = List.hd(expressions);
```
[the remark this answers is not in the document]
marcusroberts: True.  I changed the pattern matching to protect against this

compiler/grainformat/reformat.re
```diff
@@ -1490,586 +2193,692 @@ and print_attributes = attributes =>
...
-  let expr_line = get_end_loc_line(expr.pexp_loc);
-
-  let trailing_comment_docs =
-    if (List.length(trailing_comments) > 0) {
-      Doc.concat([print_multi_comments(trailing_comments, expr_line)]);
-    } else {
-      Doc.nil;
-    };
+  let get_loc = (p: Parsetree.pattern) => p.ppat_loc;
+  let print_item = (p: Parsetree.pattern, comments) => {
```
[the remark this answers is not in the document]
marcusroberts: Done

compiler/grainformat/reformat.re
```diff
@@ -1490,586 +2193,692 @@ and print_attributes = attributes =>
...
+          Doc.concat([
+            Doc.group(
+              print_pattern(
+                ~original_source,
+                ~comments=branch_pattern_comments,
+                ~next_loc=
+                  switch (branch.pmb_guard) {
+                  | None => branch.pmb_body.pexp_loc
+                  | Some(b) => b.pexp_loc
+                  },
```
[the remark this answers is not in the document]
marcusroberts: Can you show me how, as I use a different value for the Some, unless I'm missing something?

compiler/grainformat/reformat.re
marcusroberts: It is indeed :-D

compiler/grainformat/reformat.re
marcusroberts: Turns out we don't need it at all!

compiler/grainformat/reformat.re
marcusroberts: I removed them where we were passing in this Doc.nil as they were unnecessary.  The get_attribute_text is always used for block_item_iterator so I've kept it (as we pass in expressions or top level statements)

compiler/grainformat/reformat.re
marcusroberts: Moved to the correct place

compiler/grainformat/reformat.re
marcusroberts: Can you explain why that's better?  Is it a performance thing?

compiler/grainformat/reformat.re
marcusroberts: Needed for this:
`import foreign wasm args_get: (WasmI32, WasmI32) -> WasmI32 from "wasi_snapshot_preview1"`

compiler/grainformat/reformat.re
marcusroberts: Can you tell me why?

compiler/grainformat/reformat.re
```diff
@@ -1075,43 +1556,466 @@ and print_type =
...
+and print_arguments_with_callback_in_first_position =
+    (~original_source, ~comments, args: list(Parsetree.expression)) => {
+  switch (args) {
+  | [] => Doc.nil
+  | [callback] =>
+    // we handle the special case of just one callback here as we call this if the first arg is a callback
+    let printed_callback =
+      print_arg_lambda(~comments, ~original_source, callback);
+
+    Doc.group(printed_callback);
```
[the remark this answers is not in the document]
marcusroberts: You are correct, I group it everywhere, so I've moved it inside.

compiler/grainformat/reformat.re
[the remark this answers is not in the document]
marcusroberts: This one I didn't need, but the other two I think we do.

compiler/grainformat/reformat.re
```diff
@@ -941,75 +1393,76 @@ and print_ident = (ident: Identifier.t) => {
...
+      recloc: Location.t,
     ) => {
-  let comments_after_brace = get_trailing_comments_to_end_of_line(recloc);
+  let get_loc = (field: (Location.loc(Identifier.t), Parsetree.expression)) => {
+    let (_, expr) = field;
+    expr.pexp_loc;
+  };
+
+  let print_item =
+      (field: (Location.loc(Identifier.t), Parsetree.expression), comments) => {
```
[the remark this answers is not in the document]
marcusroberts: Done

compiler/grainformat/reformat.re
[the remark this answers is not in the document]
marcusroberts: But line works in this instance :-D

compiler/grainformat/reformat.re
```diff
@@ -460,23 +461,39 @@ let rec block_item_iterator_line =
...
+            ~previous=PreviousItem(get_loc(item)),
+            ~get_loc,
+            ~print_item,
+            ~comments=comments_without_item_comments,
+            ~get_attribute_text,
+            ~original_source,
+            ~separator=sep,
+            remainder,
+          ),
+        ])
```
phated, line 528: These are the same code except for `~separator` right? You can just use `~separator?` to pun and unwrap the optional into the function call.

marcusroberts: Oooooh. So that's how you do it!

compiler/grainformat/reformat.re
```diff
@@ -108,12 +108,68 @@ let get_original_code = (location: Location.t, source: array(string)) => {
...
 
+// Be AWARE!  This is only to be called when you know the comments list is not empty.
+// Moved here in case we want to change the implementation in future
+let get_last_item_in_list = comments =>
+  List.nth(comments, List.length(comments) - 1);
+
+let is_disable_formatting_comment = (comment: Parsetree.comment) => {
+  switch (comment) {
+  | Line(cmt) =>
+    if (String.trim(Comments.get_comment_source(comment))
```
phated, line 119: Why are you using `get_comment_source`? The cmt already has a trimmed property.

marcusroberts: You are not wrong.  Fixed

compiler/grainformat/reformat.re
marcusroberts: Fixed

compiler/grainformat/reformat.re
```diff
@@ -2068,87 +2045,90 @@ and print_other_application =
...
-        print_expression(~original_source, ~comments, func),
-        Doc.space,
-        print_expression(~original_source, ~comments, first_expr),
-      ]);
-    } else {
-      // standard function application
-      // look out for special cases of callbacks in first or last position
+  | _ when function_name == list_cons =>
+    resugar_list(~original_source, ~comments, expressions)
+  | [first_expr, ..._]
```
phated, line 2034: Be careful with your ordering of pattern matches, since they match top-to-bottom, I'm always concerned when there's a `_` above the other matches. In this case it has a specific `when` clause, so it should only match in one case, but again, it scares me when I see it.

phated: (I like the change to `when` clauses though!)

compiler/test/formatter_outputs/application.gr
```diff
@@ -34,18 +34,13 @@ let findIndex = (x, y) => None
  * @returns The new array with only unique values
  */
 export let unique = array => {
-  filteri(
-    (el, index) =>
-      findIndex(longvalue => longvalue == el, array) == Some(index),
-    array
-  )
+  filteri((el, index) =>
+    findIndex(longvalue => longvalue == el, array) == Some(index), array)
```
phated, line 38: 💥 Fixing bugs!

compiler/grainformat/reformat.re
```diff
@@ -116,8 +116,7 @@ let get_last_item_in_list = comments =>
 let is_disable_formatting_comment = (comment: Parsetree.comment) => {
   switch (comment) {
   | Line(cmt) =>
-    if (String.trim(Comments.get_comment_source(comment))
-        == "// formatter-ignore") {
+    if (String.trim(cmt.cmt_source) == "// formatter-ignore") {
```
phated, line 119: But, why not `cmt.cmt_content == "formatter-ignore"`? There's not reason to check and trim the source.

marcusroberts: You are correct.  Fixed.

compiler/grainformat/reformat.re
phated: Nope, just a personal style nit. I don't like the look of anonymous functions passed as a named argument to a function.

compiler/grainformat/reformat.re
phated: Sorry, I totally didn't know we had an `export foreign` syntax, but grepping for it makes sense why we need/have it. Haha. 🙊

compiler/grainformat/reformat.re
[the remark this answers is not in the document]
phated: I want to defer to @ospencer because maybe he has more understanding of the implications of each.

compiler/grainformat/reformat.re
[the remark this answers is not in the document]
phated: @marcusroberts Yes, in `List.map` or `List.iter`, there are some places where the partial application is being used, vs adding an anonymous function that receives the item and passes it to the function, like `print_arg(~comments, ~original_source, item)`

compiler/grainformat/reformat.re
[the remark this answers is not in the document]
phated: Whoops, I read it in reverse. In that case, you can use `Option.fold(~none=branch.pmb_body.pexp_loc, ~some=(b) => b.pexp_loc, branch.pmb_guard)`; however, I actually think that is much harder to read, so let's not make that change.

compiler/grainformat/reformat.re
phated: Just a style nit, I don't like the look of anonymous functions passed as named arguments.

compiler/grainformat/reformat.re
marcusroberts: Changed.

compiler/grainformat/reformat.re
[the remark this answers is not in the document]
ospencer: This is probably much faster since it avoids all the allocations of creating a new list if you do `List.rev`. But of course, the fastest would be a recursive function that just pulls the last item 🙂

compiler/grainformat/reformat.re
```diff
@@ -1829,14 +1829,16 @@ and print_arg_lambda =
...
+              print_attributes(expr.pexp_attributes);
             let printed_expressions =
               block_item_iterator(
                 ~previous=Block(expression.pexp_loc),
                 ~get_loc,
                 ~print_item,
                 ~comments=cleaned_comments,
-                ~get_attribute_text=
-                  expr => print_attributes(expr.pexp_attributes),
+                ~get_attribute_text=print_attribute,
```
phated, line 1841: ooooh, as a cleanup, can we name this `print_attribute`? To me that is much more clear 🤔 What do you think?

compiler/grainformat/reformat.re
```diff
@@ -4168,17 +4172,18 @@ let reformat_ast =
       parsed_program.comments,
     );
 
+  let get_attributes = (stmt: Parsetree.toplevel_stmt) => {
```
phated, line 4175: `print_attribute`?

marcusroberts: Looks much nicer!

compiler/test/formatter_outputs/application.gr
marcusroberts: Nice work by you there!

ospencer (approved):
Fantastic work @marcusroberts!!!

phated (approved):
Absolutely fantastic! Love the work that went into this. I'll merge once CI finishes.

PR #1045, opened 2021-12-04
fix(stdlib): Fixed memory leak in hash

cician:


stdlib/hash.gr
```diff
@@ -227,13 +227,17 @@ let rec hashOne = (val, depth) => {
  *
  * @since v0.1.0
  */
-export let hash = (anything) => {
+export let rec hash = (anything) => {
```
ospencer, line 230: Looks like we should also be decRefing `anything`, right?

cician: Right.

ospencer (approved):
Perfect!

phated (approved):
🔥

PR #1048, opened 2021-12-05
fix(stdlib): Array rotate with generic elements and for empty arrays

cician:
This PR changes slightly the `Array.rotate` function to accept arrays with any types of elements. This theoretically changes the signature of the function, but I'm assuming the intention form the start was to make it generic.

Additionally calling rotate on an empty array becomes a no-op instead of failing with a runtime error.


marcusroberts:
Nice work there!

ibdafna:
Thank you!

ospencer (approved):
Looks great!

marcusroberts (approved):
Looks good!

phated (approved):
I pushed a couple commits because we needed to regenerate the docs and I wanted a test that was a completely generic array (not type cast).

PR #1055, opened 2021-12-08
fix(compiler): Fix error message when trying to export a constructor

ospencer:


phated (approved):
While "Unbound type constructor" doesn't mean much to newbs, this is inline with our current messaging. Thanks! 🎉

marcusroberts (approved):
LGTM!

PR #1056, opened 2021-12-08
fix(compiler): Remove <id></id> from default ident printer

ospencer:


phated (approved):
Most excellent!

marcusroberts (approved):
Nice!

PR #1060, opened 2021-12-08
feat(compiler): Allow @externalName attribute for foreign names

ospencer:
This PR adds the ability to specify any UTF-8 string (not just valid Grain identifiers) to use as an external identifier when importing a foreign value or exporting one:

```
@externalName("env.foo")
import foreign wasm foo: Number -> Number from "env"

@externalName("__bar%%!")
export let bar = () => foo(1)
```

This is to allow importing/exporting any foreign value WebAssembly supports.

### DIFF
--- a/compiler/src/middle_end/linearize.re
+++ b/compiler/src/middle_end/linearize.re
@@ -1421,6 +1421,18 @@ let rec transl_anf_statement =
       | Exported => Global
       | Nonexported => Nonglobal
       };
+    let external_name =
+      switch (
+        List.find_opt(
+          fun
+          | External_name(_) => true
+          | _ => false,
+          attributes,
+        )
+      ) {
+      | Some(External_name(name)) => name
+      | _ => desc.tvd_name.txt
+      };
     switch (desc.tvd_desc.ctyp_type.desc) {
     | TTyArrow(_) =>
       let (argsty, retty) =
@@ -1438,7 +1450,7 @@ let rec transl_anf_statement =
             ~global,
             desc.tvd_id,
             desc.tvd_mod.txt,
-            desc.tvd_name.txt,
+            external_name,
             FunctionShape(argsty, retty),
           ),
         ],
@@ -1452,7 +1464,7 @@ let rec transl_anf_statement =
             ~global,
             desc.tvd_id,
             desc.tvd_mod.txt,
-            desc.tvd_name.txt,
+            external_name,
             GlobalShape(ty),
           ),
         ],
--- a/compiler/src/middle_end/linearize.re
+++ b/compiler/src/middle_end/linearize.re
@@ -1422,17 +1422,14 @@ let rec transl_anf_statement =
       | Nonexported => Nonglobal
       };
     let external_name =
-      switch (
-        List.find_opt(
+      List.fold_left(
+        name =>
           fun
-          | External_name(_) => true
-          | _ => false,
-          attributes,
-        )
-      ) {
-      | Some(External_name(name)) => name
-      | _ => desc.tvd_name.txt
-      };
+          | External_name(name) => name
+          | _ => name,
+        desc.tvd_name.txt,
+        attributes,
+      );
     switch (desc.tvd_desc.ctyp_type.desc) {
     | TTyArrow(_) =>
       let (argsty, retty) =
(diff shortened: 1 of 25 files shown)


compiler/src/parsing/well_formedness.re
```diff
@@ -367,27 +377,83 @@ let no_zero_denominator_rational = (errs, super) => {
...
+              },
+            },
+          ],
+        )
+      | PTopForeign(_) => ()
+      | _ =>
+        errs :=
+          [
+            AttributeDisallowed(
+              "`externalName` is only allowed on `foreign` statements and let bindings of a single name.",
```
phated, line 450: What do you mean by "let bindings of a single name"?

compiler/src/typed/typedtree.re
```diff
@@ -21,6 +21,15 @@ open Types;
...
 
 type loc('a) = Location.loc('a);
+
+[@deriving sexp]
+type attributes = list(attribute)
+
+[@deriving sexp]
+and attribute =
+  | Disable_gc
+  | External_name(string);
```
phated, line 31: Why are these in typedtree now? Just curious

compiler/src/typed/typetexp.re
```diff
@@ -583,6 +583,18 @@ let fold_modules = fold_persistent(Env.fold_modules);
 let fold_constructors = fold_descr(Env.fold_constructors, d => d.cstr_name);
 let fold_modtypes = fold_simple(Env.fold_modtypes);
 
+let type_attributes = attrs => {
+  List.map(
+    (({txt}, args)) =>
+      switch (txt, args) {
+      | ("disableGC", []) => Disable_gc
+      | ("externalName", [{txt}]) => External_name(txt)
+      | _ => failwith("impossible by well-formedness")
```
phated, line 592:
```suggestion
      | _ => failwith("type_attributes: impossible by well-formedness")
```
Just for searching clarity 😉

phated (commented):
Looking really good. Just a couple things.

compiler/src/parsing/well_formedness.re
ospencer: It means you aren't allowed to do `@externalName let a = 1, b = 2` (or `@externalName let (a, b) = ...`). I couldn't think of a better way to say that 😢

compiler/src/typed/typedtree.re
ospencer: It felt really bad having to search through strings and assume the arguments rather than look at nice enums. I moved it here and got rid of a ton of `failwith("impossible")`s 🤘

compiler/src/parsing/well_formedness.re
phated: How about "let bindings that aren't a destructuring pattern" or something along those lines?

compiler/src/typed/typedtree.re
phated: Love it. I was just wondering why you chose the typedtree specifically? Idk how we usually determine where things should be declared.

compiler/src/parsing/well_formedness.re
ospencer: We'd still need to cover the multiple binds case. Could just turn it into two separate error messages.

compiler/src/typed/typedtree.re
ospencer: Ah. I think of the parsetree as the wild west, where anything goes but nothing is confirmed (you can write `void + true` if you want, we don't care etc., the formatter will even format it for you), but the typedtree is where things are settled—we know both sides of `+` are numbers, functions are called with the right number of arguments, etc. In a similar vein, this is sort of the typechecked version of attributes. This is the first point in time that we _know_ that the attributes will have the right number of args, etc, so we might as well encode that in here rather than waiting until later in the stages of the compiler/accidentally messing it up.

That's my thought process, at least.

ospencer:
@phated Updated the well-formedness messages!

jozanza (approved):
I don't see any red flags here!

compiler/src/middle_end/linearize.re
```diff
@@ -1421,6 +1421,18 @@ let rec transl_anf_statement =
       | Exported => Global
       | Nonexported => Nonglobal
       };
+    let external_name =
+      switch (
+        List.find_opt(
```
peblair, line 1426: Why not use `fold_left`? This access pattern seems a bit unorthodox

peblair (approved):
One comment, but looks good!

compiler/src/middle_end/linearize.re
ospencer: fold_left isn't _quite_ right here since it'll process the whole list without bailing, but I changed it anyway.

phated (approved):
Looks great! :shipit:

PR #1061, opened 2021-12-09
fix(compiler): Consider invalid CMIs as out of date and recompile

### LINKED ISSUE #708: Should a broken CMI trigger a recompilation?
Currently, when the compiler reads a file with a broken CMI, it bails out with an exception. It might be preferable to emit a warning and then recompile the file.

cc @ospencer

ospencer:
Closes #708

### DIFF
--- a/compiler/src/typed/module_resolution.re
+++ b/compiler/src/typed/module_resolution.re
@@ -365,6 +365,7 @@ module Dependency_graph =
           dn.dn_up_to_date :=
             (
               switch (read_file_cmi(objpath)) {
+              // Treat corrupted CMI as invalid
               | exception (Cmi_format.Error(_)) => false
               | cmi =>
                 config_sum == cmi.cmi_config_sum
--- a/compiler/src/typed/module_resolution.re
+++ b/compiler/src/typed/module_resolution.re
@@ -276,46 +276,48 @@ module Dependency_graph =
 
     let get_filename = dn => dn.dn_file_name;
 
-    let get_dependencies: (t, string => option(t)) => list(t) =
+    let rec get_dependencies: (t, string => option(t)) => list(t) =
       (dn, lookup) => {
         let base_dir = Filename.dirname(dn.dn_file_name);
         let active_search_path = Config.module_search_path();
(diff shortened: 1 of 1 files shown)


compiler/src/typed/module_resolution.re
```diff
@@ -359,31 +361,38 @@ module Dependency_graph =
...
+                config_sum == cmi.cmi_config_sum
+                && file_older(srcpath, objpath)
+                && List.for_all(
+                     ((name, crc)) => {
+                       let resolved = resolve_unit(~base_dir, name);
+                       let out_file_name = get_output_name(resolved);
+                       Fs_access.file_exists(out_file_name)
+                       && (
+                         switch (crc) {
+                         | None => true
```
marcusroberts, line 379: Just a question, why isn't it considered not up to state if there's no crc?

ospencer: Good question. I think that's actually a quasi-bug since this is never actually `None`, but if it ever _were_ `None`, then this should be false. Since I fixed the other case I'll fix this one too.

jozanza (approved):

compiler/src/typed/module_resolution.re
```diff
@@ -276,46 +276,48 @@ module Dependency_graph =
...
-              },
-              Grain_parsing.Driver.scan_for_imports(srcpath),
-            );
-          ret;
+        | Some(GrainModule(srcpath, None)) => from_srcpath(srcpath)
         | Some(GrainModule(srcpath, Some(objpath))) =>
-          let cmi = read_file_cmi(objpath);
-          let ret =
+          switch (read_file_cmi(objpath)) {
+          | exception (Cmi_format.Error(_)) => from_srcpath(srcpath)
```
jozanza, line 319: Just wondering should we maybe inline comments next to code like this? Normally, I would see something along these lines and think we're erroneously swallowing an error.

peblair (approved):
This looks right to me. Great work!

compiler/src/typed/module_resolution.re
ospencer: Added a comment!

PR #1065, opened 2021-12-10
fix(stdlib): Add bounds checking to Buffer addStringSlice & addBytesSlice

### LINKED ISSUE #1046: Missing bounds checking in the Buffer module
The functions `addStringSlice` and `addBytesSlice` in the Buffer module accept any interval without bounds checking. Can result in reading and/or corrupting adjacent data in linear memory.

jozanza:
Fixes #1046 

- `Buffer.addStringSlice` appends bytes from the `slice` var returned from an internal call to `String.slice` instead of the original `string`. This prevents unwanted memory access. I was originally going to guard against negative start indexes in `String.slice`, but then I realized we seem to support this behavior because I saw tests for it.
- `Buffer.addBytesSlice` prevents the usage of out-of-bounds `start` and `length` parameters entirely.

### DIFF
--- a/compiler/test/stdlib/buffer.test.gr
+++ b/compiler/test/stdlib/buffer.test.gr
@@ -197,16 +197,27 @@ Buffer.addString(str, buf)
 assert Buffer.toString(buf) == str
 
 // addStringSlice
-let slices = [>
-  ("Let's get this 🍞", 0, 16),
-  ("0123456789", -10, 10),
-  ("0123456789", -1, 10),
-];
+let slices = [> ("Let's get this 🍞", 0, 16), ("0123456789", -10, 9), ("0123456789", -1, 1), ("0123456789", 2, 1), ("0123456789", 0, 10), ("0123456789", 6, 0) ]
 for (let mut i = 0; i < Array.length(slices); i += 1) {
-  let (str, off, len) = slices[i];
+  let (str, off, len) = slices[i]
   let buf = Buffer.make(0)
   Buffer.addStringSlice(off, len, str, buf)
-  assert Buffer.toString(buf) == String.slice(off, len, str)
+  let off = if (off < 0) {
+    off + String.length(str)
+  } else {
+    off
+  }
+  assert Buffer.toString(buf) == String.slice(off, off + len, str)
+}
+
+// addBytesSlice
+let slices = [> ("Let's get this 🍞", 0, 17), ("0123456789", 2, 1), ("0123456789", 0, 10), ("0123456789", 6, 0) ]
+for (let mut i = 0; i < Array.length(slices); i += 1) {
+  let (str, off, len) = slices[i]
+  let bytes = Bytes.fromString(str)
+  let buf = Buffer.make(0)
+  Buffer.addBytesSlice(off, len, bytes, buf)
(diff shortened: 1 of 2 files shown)


compiler/test/stdlib/buffer.test.gr
```diff
@@ -196,12 +197,17 @@ Buffer.addString(str, buf)
...
-let str = "Let's get this 🍞"
-let buf = Buffer.make(0)
-Buffer.addStringSlice(0, 0, str, buf)
-assert Buffer.toString(buf) == ""
-Buffer.addStringSlice(0, String.length(str), str, buf)
-assert Buffer.toString(buf) == str
+let slices = [>
+  ("Let's get this 🍞", 0, 16),
+  ("0123456789", -10, 10),
+  ("0123456789", -1, 10),
```
phated, line 203: Do we want any tests for positive offsets?

compiler/test/stdlib/buffer.test.gr
```diff
@@ -196,12 +197,17 @@ Buffer.addString(str, buf)
...
+];
+for (let mut i = 0; i < Array.length(slices); i += 1) {
+  let (str, off, len) = slices[i];
+  let buf = Buffer.make(0)
+  Buffer.addStringSlice(off, len, str, buf)
+  assert Buffer.toString(buf) == String.slice(off, len, str)
+}
```
phated, line 210: Do we want to do this same testing for `addBytesSlice`?

phated (commented):
Couple questions about testing

stdlib/buffer.gr
```diff
@@ -414,33 +414,34 @@ export let rec addString = (string, buffer) => {
  */
 @disableGC
 export let rec addStringSlice = (start, length, string, buffer) => {
-  Memory.incRef(WasmI32.fromGrain(String.slice))
   Memory.incRef(WasmI32.fromGrain(start))
   Memory.incRef(WasmI32.fromGrain(length))
   Memory.incRef(WasmI32.fromGrain(string))
+
+  Memory.incRef(WasmI32.fromGrain(String.slice))
```
ospencer, line 421: I'd probably move this back to the top where it was—the general pattern we've been using is incRefing the function before the arguments.

stdlib/buffer.gr
```diff
@@ -462,16 +463,32 @@ export let rec addStringSlice = (start, length, string, buffer) => {
  */
 @disableGC
 export let rec addBytesSlice = (start: Number, length: Number, bytes: Bytes, buffer: Buffer) => {
+  let (<) = WasmI32.ltS
+  let (>) = WasmI32.gtS
+  let (>=) = WasmI32.geS
+
+  // bounds check start
+  let bytelen = WasmI32.load(WasmI32.fromGrain(bytes), 1n)
```
ospencer, line 471:
```suggestion
  let bytelen = WasmI32.load(WasmI32.fromGrain(bytes), 4n)
```

ospencer (changes requested):
Found one problem but other than that looks good!

peblair (approved):
Once the already-existing comments are addressed, this looks good!

ospencer (approved):

phated (approved):
Tests pass on a Friday at 7pm, so I 👍 approve

PR #1066, opened 2021-12-10
fix(compiler): Fix counting of stack size

ospencer:
@jozanza Noticed that we were incorrectly considering the arguments to functions as a factor in the stack size (in our case, the number of locals the function needs). Arguments at this stage are always constants or are pulled from a local—no new locals are needed.

We were inflating the stack with unused slots, which resulted in extra (useless) calls to `decRefIgnoreZeros`. You'll notice that every removal of a `decRefIgnoreZeros` in the snapshots is a Binaryen-optimized `decRefIgnoreZeros(0)` 🎉 

~~I believe this is why `decRefIgnoreZeros` was introduced in the first place, so now we can consider removing it.~~ The `ignoreZeros` part is about the ref count.

Great find @jozanza!

### DIFF
--- a/compiler/src/middle_end/anf_utils.re
+++ b/compiler/src/middle_end/anf_utils.re
@@ -235,8 +235,6 @@ and comp_count_vars = c =>
   | CSwitch(_, bs, _) =>
     List.fold_left(tuple_max, tuple_zero) @@
     List.map(((_, b)) => anf_count_vars(b), bs)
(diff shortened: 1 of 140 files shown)


phated (approved):
All of the snapshot changes started to look the same. I'm pre-approving this pending our discussion about decRefIgnoreZeros in discord (which might be resolved shortly).

peblair (approved):
:shipit:

PR #1070, opened 2021-12-10
fix(graindoc): Ensure value_description is resolved to outcome before printing

### LINKED ISSUE #1069: Graindoc: Incorrect generic names on params & returns
It seems that when we generate the types for our `@params` and `@returns`, they don't know about the other generics within the statement, so it turn things like `b` into `a` (which is very incorrect).

I think this bug is going to be really hard to solve.

phated:
Remember when I posted #1069 and said that I thought it was going to be hard to fix? Well, I fixed it.

Closes #1069

@ospencer do you think this is a good semantic title? I also wonder if everything should be resolved to outcome types when they are plucked out of `signature_items` 🤔

### DIFF
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -27,9 +27,16 @@ let title_for_api = (~module_name, ident: Ident.t) => {
   Format.asprintf("%s.**%a**", module_name, Printtyp.ident, ident);
 };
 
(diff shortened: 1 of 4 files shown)


ospencer (approved):
Makes sense, good sleuthing! Looks like a good title to me!

phated:
Those regenerated docs :chefskiss:

peblair (approved):
😍

PR #1075, opened 2021-12-14
chore(compiler): Stabilize snapshots

ospencer:
This PR removes the uniqueness from the internal names used for imports. This should remove changes to snapshots when imports/exports change in modules.

The uniqueness isn't needed because from the view of a single module, two different imports from the same module always come from the same module (which sounds obvious, but is important) and a module can only export a name once for each value type, so a clash is impossible.

Conflicts can arise from foreign imports, where the foreign module name has the same name as a Grain module. For this reason, I made two separate namespaces for "grain" imports and "wasm" imports.

This won't eliminate PRs with massive snapshot diffs, but those diffs should actually be significant (if this PR is accepted).

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -357,26 +357,25 @@ let runtime_heap_start = 0x410;
 // Static pointer to runtime type information
 let runtime_type_metadata_ptr = 0x408;
 
-let get_imported_name = (mod_, name) =>
-  Printf.sprintf(
-    "import_%s_%s",
-    Ident.unique_name(mod_),
-    Ident.unique_name(name),
-  );
+let get_wasm_imported_name = (mod_, name) =>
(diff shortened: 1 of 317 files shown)


phated (approved):
I don't understand the difference between "wimport" and "gimport" - is this something we should document?

peblair:
This makes sense to me, but I want to double-check something before hitting approve: if I import something from `foo/bar.gr` and something else from `baz/bar.gr`, those two `bar.gr` files will not emit the same import name, correct? I forget if these names would be just `bar` or if they would be `foo/bar`.

ospencer:
They will, but that's okay because it's relative to where it's imported.

I thought of this same thing 🙂

marcusroberts (approved):
LGTM, I had a good read through, but any subtleties are beyond my understanding.

PR #1076, opened 2021-12-14
feat(compiler): Inline all primitives

ospencer:
Right now we only inline some primitives. This PR inlines them all. This also allows us to mostly remove the `inline_wasm` optimization, but we can't completely get rid of it because of `--no-bulk-memory`—we could, but that means we'd need to write the polyfill implementations of `memory_fill` and `memory_copy` in pure wasm in the compiler (which I didn't want to do right now 😅).

### DIFF
--- a/compiler/src/middle_end/linearize.re
+++ b/compiler/src/middle_end/linearize.re
@@ -426,21 +426,19 @@ let rec transl_imm =
     ) =>
     let (ans, ans_setup) = transl_comp_expression(e);
     (Imm.trap(~loc, ~env, ()), ans_setup @ [BSeq(ans)]);
-  | TExpApp(
-      {exp_desc: TExpIdent(_, _, {val_kind: TValPrim("@and")})},
-      [arg1, arg2],
-    ) =>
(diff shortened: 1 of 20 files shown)


compiler/src/middle_end/linearize.re
```diff
@@ -1072,26 +1070,20 @@ and transl_comp_expression =
...
-      {exp_desc: TExpIdent(_, _, {val_kind: TValPrim("@or")})},
-      [arg1, arg2],
-    ) =>
-    transl_comp_expression({...e, exp_desc: TExpPrim2(Or, arg1, arg2)})
-  | TExpApp(
-      {exp_desc: TExpIdent(_, _, {val_kind: TValPrim("@not")})},
-      [arg],
-    ) =>
-    transl_comp_expression({...e, exp_desc: TExpPrim1(Not, arg)})
+  | TExpApp({exp_desc: TExpIdent(_, _, {val_kind: TValPrim(prim)})}, args) =>
```
peblair, line 1073: :fire:

peblair (approved):
Love this!

marcusroberts (approved):
LGTM but the low-level WASM I have to take on trust.

PR #1077, opened 2021-12-15
chore(stdlib): migrate string module @disableGC to @unsafe

jozanza:
I'm opening this draft PR so we can get a sense of the diff when migrating from `@disableGC` to `@unsafe` once the new pointer stack lands along with the `@unsafe` attribute.

### DIFF
--- a/stdlib/string.gr
+++ b/stdlib/string.gr
@@ -469,10 +469,7 @@ export let split = (separator: String, string: String) => {
   if (psize == 0n) {
     WasmI32.toGrain(explodeHelp(string, false)): Array<String>
   } else if (psize > size) {
-    let string = WasmI32.fromGrain(string)
-    let ptr = allocateArray(1n)
-    WasmI32.store(ptr, Memory.incRef(string), 8n)
-    WasmI32.toGrain(ptr): Array<String>
+    [> string]
   } else {
     let string = WasmI32.fromGrain(string)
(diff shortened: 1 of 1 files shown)


stdlib/string.gr
```diff
@@ -1604,18 +1489,11 @@ export let decodeRangeKeepBom = (bytes: Bytes, encoding: Encoding, start: Number
   decodeRangeHelp(bytes, encoding, false, start, size)
 }
 
-@disableGC
+@unsafe
 let rec decodeHelp = (bytes: Bytes, encoding: Encoding, skipBom: Bool) => {
   let bytesPtr = WasmI32.fromGrain(bytes)
   let bytesSize = WasmI32.load(bytesPtr, 4n)
-  Memory.incRef(WasmI32.fromGrain(decodeRangeHelp))
```
peblair, line 1611: I'm still sort of confused about how these incRefs can be dropped now. @ospencer @jozanza , maybe this will help me understand: am I correct in guessing that, with these changes, `--compilation-mode=runtime` still requires these sorts of manual incref/decref calls?

ospencer: Yup, that's correct. Happy to talk more about the implementation as well!

ospencer (approved):

marcusroberts (approved):
The code is so much cleaner now!

peblair (approved):
Not much to say, other than that this is 🔥

Issue #1017, opened 2021-11-02
Number stdlib Libary: Add following functions and values


spotandjake:
Add The Following Functions to the number Libary.
+ [ ] Number.sin()
+ [ ] Number.cos()
+ [ ] Number.tan()
+ [ ] Number.asin()
+ [ ] Number.acos()
+ [ ] Number.atan()
+ [ ] Number.factorial()
+ [ ] Number.pow()
+ [ ] Number.sign()
Along with the following values
+ [ ] Number.PI

cician:
I'd also add:
- A tau constant (https://www.khanacademy.org/math/trigonometry/unit-circle-trig-func/long-live-tau/v/tau-versus-pi) along with PI.
- `lerp` (https://en.wikipedia.org/wiki/Linear_interpolation)
- `atan2` (https://en.wikipedia.org/wiki/Atan2)
- `exp` (https://en.wikipedia.org/wiki/Exponential_function)

For logarithms it's useful to have three variants for most common bases: `log2`, `log10` and `log` for the natural logarithm. Same for `exp`.

Trigonometric functions should definitely take radians and conversion functions would be nice.

spotandjake:
#1373 has been closed as the new plan is to use The range type I am linking it because it still contains some useful stuff related to documenting the `map`, `clamp` and `lerp` functions.

Issue #693, opened 2021-06-04
Add NaN and Infinity Literal Syntax


peblair:
See title. Open questions:
- What syntax should we use?
- Should they be F32 or F64? We should probably have both, since F32/F64 NaN comparisons may not work correctly (need to check)
- Perhaps others

cician:
I think this is a duplicate of #360.

As a user I actually need to detect NaNs ind +/-infinite Numbers, but I think it needs to be done with functions like numberUtils.isNaN, which is for WASM primitives and not exposed.

peblair:
@cician This is actually talking about the _syntax_, whereas #360 is referring to the constants. As for the `isNaN` and `isFinite` predicates, we should definitely add those. I think we should open a separate issue for that.

cician:
OK, thank you. I've opened a separate issue.

ospencer:
> What syntax should we use?

I vote for `NaN` and `Infinity`.

> Should they be F32 or F64? We should probably have both, since F32/F64 NaN comparisons may not work correctly (need to check)

Great question. Float literals of the `Number` type default to f64s, so I say we remain consistent. I think questions of equality should probably be handled by the operators.

As a side note, I would probably just make `coerceFloat32ToNumber` promote to an f64 and then simplify the Number operations to only worry about f64s and avoid the equality issue altogether.

PR #1079, opened 2021-12-21
feat(stdlib): Add sign function to Number module

phated:
This adds the `Number.sign(x)` function, similar to the [`Math.sign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sign) function in JavaScript. It even accounts for the `-0` case by multiplying the `0` result by the input number (you can see it working in the tests).

Opening as a draft, just to make sure we are good with the implementation.

Ref #1017

### DIFF
--- a/compiler/test/stdlib/number.test.gr
+++ b/compiler/test/stdlib/number.test.gr
@@ -134,3 +134,9 @@ assert Result.isErr(Number.parseInt("zzzzz", 37))
 assert Result.isErr(Number.parseInt("zzzzz", 9223372036854775807))
 assert Result.isErr(Number.parseInt("10", 1.23))
 assert Result.isErr(Number.parseInt("10", 2/3))
+
+// Number.sign
+assert Number.sign(-10000) == -1
+assert Number.sign(22222) == 1
+assert Number.sign(0) == 0
+assert Number.sign(-0) == 0
--- a/compiler/test/stdlib/number.test.gr
+++ b/compiler/test/stdlib/number.test.gr
@@ -1,5 +1,6 @@
 import Number from "number"
 import Result from "result"
+import Float64 from "float64"
 
 // add
 assert Number.add(25, 5) == 30
@@ -138,5 +139,7 @@ assert Result.isErr(Number.parseInt("10", 2/3))
 // Number.sign
 assert Number.sign(-10000) == -1
 assert Number.sign(22222) == 1
-assert Number.sign(0) == 0
-assert Number.sign(-0) == 0
+// TODO(#693): Replace with Infinity literal when it exists
+assert 1 / Number.sign(0.0) == Float64.toNumber(Float64.infinity)
+// TODO(#693): Replace with -Infinity literal when it exists
+assert 1 / Number.sign(-0.0) == Number.neg(Float64.toNumber(Float64.infinity))
--- a/stdlib/number.gr
+++ b/stdlib/number.gr
@@ -94,7 +94,7 @@ export let rec sqrt = (x: Number) => {
  *
  * @example Number.sign(-10000) == -1
  * @example Number.sign(222222) == 1
- * @example Number.sign(-0.0) == -0.0
+ * @example Number.sign(0) == 0
  */
 export let sign = x => {
   match (x) {
--- a/stdlib/number.gr
+++ b/stdlib/number.gr
@@ -87,6 +87,24 @@ export let rec sqrt = (x: Number) => {
 }
 
 /**
+ * Determine the positivity or negativity of a Number.
+ *
+ * @param x: The number to inspect
+ * @returns `-1` if the number is negative, `1` if positive, or +/-`0` otherwise
+ *
+ * @example Math.sign(-10000) == -1
+ * @example Math.sign(222222) == 1
+ * @example Math.sign(-0) == -0
+ */
+export let sign = x => {
+  match (x) {
+    x when x < 0 => -1,
+    x when x > 0 => 1,
+    _ => 0 * x,
+  }
+}
+
+/**
  * Returns the smaller of its operands.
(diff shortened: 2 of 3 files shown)


peblair (approved):
Looks good to me!

phated:
@peblair Cool! I'll convert it from a draft then.

marcusroberts (approved):
It took me a second to work out why you needed _ => 0 * x, but clever!

compiler/test/stdlib/number.test.gr
```diff
@@ -134,3 +134,9 @@ assert Result.isErr(Number.parseInt("zzzzz", 37))
 assert Result.isErr(Number.parseInt("zzzzz", 9223372036854775807))
 assert Result.isErr(Number.parseInt("10", 1.23))
 assert Result.isErr(Number.parseInt("10", 2/3))
+
+// Number.sign
+assert Number.sign(-10000) == -1
+assert Number.sign(22222) == 1
+assert Number.sign(0) == 0
+assert Number.sign(-0) == 0
```
ospencer, line 142: Implementation is great, though these aren't quite checking the behavior because of int/float weirdness :/

Signedness only applies to floats, so we'll need to use `0.0` and `-0.0`. Additionally, positive/negative zero in floating point arithmetic are equal, so we'll need to check equality some other way. The best I could think of was dividing 1 by the result of `Number.sign` and checking for positive/negative infinity.

phated: We chatted about this in discord and I'm going to implement the tests we came up with, give me a bit.

phated: Tests updated. I also added TODO comments to switch to literals once #693 is done!

stdlib/number.gr
```diff
@@ -86,6 +86,24 @@ export let rec sqrt = (x: Number) => {
   ret
 }
 
+/**
+ * Determine the positivity or negativity of a Number.
+ *
+ * @param x: The number to inspect
+ * @returns `-1` if the number is negative, `1` if positive, or +/-`0` otherwise
```
ospencer, line 93: Should we revise this?

stdlib/number.gr
```diff
@@ -86,6 +86,24 @@ export let rec sqrt = (x: Number) => {
...
+ *
+ * @example Math.sign(-10000) == -1
+ * @example Math.sign(222222) == 1
+ * @example Math.sign(-0) == -0
```
phated, line 97:
```suggestion
 * @example Math.sign(-0.0) == -0.0
```
I suppose I should update this example.

stdlib/number.gr
phated: Is it worth the verbosity? How would you want to state it?

stdlib/number.gr
```diff
@@ -86,6 +86,24 @@ export let rec sqrt = (x: Number) => {
...
 
+/**
+ * Determine the positivity or negativity of a Number.
+ *
+ * @param x: The number to inspect
+ * @returns `-1` if the number is negative, `1` if positive, or +/-`0` otherwise
+ *
+ * @example Math.sign(-10000) == -1
+ * @example Math.sign(222222) == 1
+ * @example Math.sign(-0.0) == -0.0
```
ospencer, line 97: This is the Number library 😛 

```suggestion
 * @example Number.sign(-10000) == -1
 * @example Number.sign(222222) == 1
 * @example Number.sign(-0.0) == -0.0
```

Also, `Number.sign(-0.0) == 0.0`. Is worth having that example?

phated: 🙈

stdlib/number.gr
ospencer: Probably not. Two options I'll let you choose from:

```suggestion
 * @returns `-1` if the number is negative, `1` if positive, or `0` otherwise
```

```suggestion
 * @returns `-1` if the number is negative, `1` if positive, or `0` otherwise. Signedness of `-0.0` is preserved.
```

stdlib/number.gr
phated: > Also, Number.sign(-0.0) == 0.0. Is worth having that example?

So we want both `Number.sign(-0.0) == -0.0` and `Number.sign(-0.0) == 0.0` docs? It almost like we're demo'ing the equality operator at that point 😛

stdlib/number.gr
phated: I actually like that 2nd one but we should use a semicolon and no periods for the returns
```suggestion
 * @returns `-1` if the number is negative, `1` if positive, or `0` otherwise; signedness of `-0.0` is preserved
```
Thoughts?

ospencer: Works for me!

stdlib/number.gr
ospencer: Oh, I meant should we _remove_ that example because it's just checking `0 == 0`.

phated: I'll just change it to `Number.sign(0) == 0` to demonstrate the 0 behavior then.

stdlib/number.gr
```diff
@@ -86,6 +86,24 @@ export let rec sqrt = (x: Number) => {
...
 
+/**
+ * Determine the positivity or negativity of a Number.
+ *
+ * @param x: The number to inspect
+ * @returns `-1` if the number is negative, `1` if positive, or `0` otherwise; signedness of `-0.0` is preserved
+ *
+ * @example Number.sign(-10000) == -1
+ * @example Number.sign(222222) == 1
+ * @example Number.sign(-0.0) == -0.0
```
phated, line 97:
```suggestion
 * @example Number.sign(0) == 0
```

ospencer (approved):

PR #1080, opened 2021-12-25
chore(cli): Advertise the "-o" option in the help of the doc command.

cician:
The help for graindoc didn't list the `-o` option.

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -211,6 +211,7 @@ program
 program
   .command("doc <file>")
   .description("generate documentation for a grain file")
+  .option("-o <filename>", "output filename")
   .action(
     wrapAction(function (file, options, program) {
       doc(file, program);


phated (dismissed):
Thanks! I believe we never re-added this when we cleaned up the CLI option handling.

cli/bin/grain.js
```diff
@@ -211,6 +211,7 @@ program
 program
   .command("doc <file>")
   .description("generate documentation for a grain file")
+  .option("-o <filename>", "output filename")
```
phated, line 214: Oh, actually, looking at the format command, I think we need:
```suggestion
  .forwardOption("-o <filename>", "output filename")
```

cician: I tried it at first, but it fails with "option `-o' cannot be repeated".

phated: Hmm, @ospencer wrote that logic, so we'll need to ping him about it.

cician: `.forwardOption` like this seems to conflict with the `-o` from the main `grain` command.

phated: @cician correct, but those options are supposed to propagate down to the individual commands and show in the help, so this needs to actually get fixed and not worked around.

cician: I'm not sure if I understand correctly, but I think I found the root of the problem.

In exec.js the `execGraindoc` function merges the options from the main `grain` "program" and the `doc` "program" in order to inherit compiler flags. This results in the `flags` array having the `-o` twice and Cmdliner doesn't accept it. My version with `.option` only worked because this merging code skips non "forward" options, leaving just the inherited one.

I apologize if I'm off-track, but is this even necessary for the doc command to receive compilation flags? At a quick glace I don't see it actually using them, but I'm probably missing something. If it is necessary, do we want to just make an exception for the `-o` param or dynamically deduplicate them?

It was probably better if I just filed an issue. Now I'm in a new rabbit hole. 🕳

phated: Yes, since both the graindoc and grainformat tools inherit parts of the compiler, they implicitly need to accept arguments from the compiler. The option argument logic is really messy/confusing and we are hoping to clean it up for silo.

cician: I made changes to use `.forwardOption`. For now with an explicit exception for `-o`. I can change it if you think there's a better way.

ospencer:
@cician I think we'll want to just list commands for grainfmt and graindoc explicitly, but that'll take a bit of refactoring. I'll send a PR later today.

ospencer:
It looks like it's going to be a bit of a bigger refactor than I'd hoped. @phated may have a chance to take a look at this soon.

PR #1092, opened 2022-01-01
chore!: Drop node 14 support, and removes the --experimental-wasm-flag.

### LINKED ISSUE #114: Remove --experimental-wasm-bigint flag when wasm-bigint is no longer experimental

spotandjake:
Switches the engine version and github workflow to use node 16 instead of node 14, along with removes the use of the `--expermental-wasm-flag`.

closes #114.


phated:
Updated title, this is a breaking change.

.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        node-version: ["14", "16"]
+        node-version: ["16"]
```
ospencer, line 13: @phated Should we consider adding 17 here?

phated: Never consider an odd node version. Odd node versions are pre-release versions and can break at any time with no semantic meaning.

peblair (approved):

phated (approved):
Thanks @spotandjake! Sorry this took so long to land

cli/bin/grain.js
```diff
@@ -10,12 +10,6 @@ const v8 = require("v8");
  *
  * This seems to work for our needs with Node 14, but we should be cautious when updating.
```
ospencer, line 11:
```suggestion
 * This seems to work for our needs with Node 16, but we should be cautious when updating.
```

ospencer (approved):
Thank you @spotandjake!!!

PR #1098, opened 2022-01-10
fix(compiler): Move lsp flag into grainc to remove it as a recompilation flag

### LINKED ISSUE #1094: Compiler: caching of CMI files seems to refresh on every source file change
With a single source Grain file, on first compile it compiles all dependencies.  The next recompile is quick as it uses the cached dependencies and in verbose mode reports compiling only the specified source file.

If I make a change to the source file, the next compile rebuilds all dependencies.

May be related to changed made in https://github.com/grain-lang/grain/pull/1061

phated:
Closes #1094

This removes the `--lsp` flag from `Grain_utils.Config` and adds it to grainc directly. This avoids using it as a flag that triggers a full recompile of every dependency.

You can see the drastic difference if you compile the JS version and use `time grain test.gr` while adjusting code with the LSP enabled.

### DIFF
--- a/compiler/grainformat/grainformat.re
+++ b/compiler/grainformat/grainformat.re
@@ -26,7 +26,11 @@ let compile_parsed = (filename: option(string)) => {
 
       program_str := String.concat("\n", linesList^);
 
-      Compile.compile_string(~hook=stop_after_parse, ~name="", program_str^);
+      Compile.compile_string(
+        ~is_root_file=true,
+        ~hook=stop_after_parse,
+        program_str^,
+      );
     | Some(filenm) =>
       // need to read the source file in case we want to use the content
       // for formatter-ignore or decision making
@@ -52,6 +56,7 @@ let compile_parsed = (filename: option(string)) => {
 
       Grain_utils.Config.base_path := dirname(filenm);
       Compile.compile_string(
+        ~is_root_file=true,
         ~hook=stop_after_parse,
         ~name=filenm,
         program_str^,
--- a/compiler/src/utils/config.re
+++ b/compiler/src/utils/config.re
@@ -1,13 +1,19 @@
+type digestable =
+  | Digestable
+  | NotDigestable;
+
 type config_opt =
-  | Opt((ref('a), 'a)): config_opt;
+  | Opt((ref('a), 'a, digestable)): config_opt;
 
 type saved_config_opt =
-  | SavedOpt((ref('a), 'a)): saved_config_opt;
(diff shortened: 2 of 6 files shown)


marcusroberts (dismissed):
Nice fix, and will be transformational for the JS version as you say

ospencer (approved):

phated:
There were actually 3 problems here:
1. Needed to move the `--lsp` flag into grainc
2. `Grain_utils.Config.base_path` needs to always be set because it is considered in the flags digest (should it be? 🤷)
3. `is_root_file=true` wasn't set for `Compile.compile_string` in lsp mode, so `Grain_utils.Config.set_root_config();` was never called after `Parsed`

Re-requesting reviews.

ospencer (approved):

compiler/src/utils/config.re
```diff
@@ -547,17 +560,10 @@ let elide_type_info =
...
 
-let lsp_mode =
-  toggle_flag(
-    ~names=["lsp"],
-    ~doc="Generate lsp errors and warnings only",
-    false,
-  );
-
-let print_warnings = internal_opt(true);
+let print_warnings = internal_opt(true, NotDigestable);
```
phated, line 563: Do we want this to be NotDigestable?

compiler/src/utils/config.rei
```diff
@@ -130,10 +126,10 @@ let print_warnings: ref(bool);
 
 /** Type representing a saved set of configuration options */
 
-type saved_config_opt =
-  | SavedOpt((ref('a), 'a)): saved_config_opt;
+type config;
 
-type config = list(saved_config_opt);
+/** An empty config */
+let empty: config;
```
phated, line 132: export an empty list so we don't need to leak the SavedOpt impl

compiler/grainformat/grainformat.re
```diff
@@ -26,7 +26,11 @@ let compile_parsed = (filename: option(string)) => {
 
       program_str := String.concat("\n", linesList^);
 
-      Compile.compile_string(~hook=stop_after_parse, ~name="", program_str^);
```
phated, line 29: I removed the empty `~name` because it is optional

compiler/src/utils/config.re
marcusroberts: I'm not sure what that does.

marcusroberts (approved):
Tested and working fine!

ospencer (approved):

PR #1103, opened 2022-01-16
feat(compiler)!: Stack-allocated Chars

ospencer:
This PR changes the implementation of chars to use an i32 on the stack rather than a heap object.

Details can be seen in the updated data_representations doc, but the gist is that a char is now just a Unicode scalar shifted left by 3 and the last 3 bits set to `010`. Things like `Char.code`/`Char.fromCode` are much simpler, while things like printing/conversion to a string got more complex.

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -1679,59 +1679,11 @@ let allocate_bytes_uninitialized = (wasm_mod, env, size) => {
   );
 };
 
-let allocate_char = (wasm_mod, env, char) => {
-  // Copy bytes into a fresh buffer so we can guarantee a copy of a full word
-  let bytes = Bytes.make(4, Char.chr(0));
-  // OCaml String#length is byte length, not Unicode character length
-  // Guaranteed not to be longer than 4 bytes by well-formedness
-  Bytes.blit_string(char, 0, bytes, 0, String.length(char));
-  let value = Bytes.get_int32_le(bytes, 0);
-
-  let get_swap = () => get_swap(wasm_mod, env, 0);
-  let tee_swap = tee_swap(wasm_mod, env, 0);
-  Expression.Block.make(
-    wasm_mod,
-    gensym_label("allocate_char"),
-    [
-      store(
-        ~offset=0,
-        wasm_mod,
-        tee_swap(heap_allocate(wasm_mod, env, 2)),
-        Expression.Const.make(
-          wasm_mod,
-          const_int32(tag_val_of_heap_tag_type(CharType)),
-        ),
-      ),
-      store(
-        ~offset=4,
-        wasm_mod,
-        get_swap(),
-        Expression.Const.make(wasm_mod, wrap_int32(value)),
-      ),
-      get_swap(),
-    ],
-  );
-};
-
-let allocate_char_uninitialized = (wasm_mod, env) => {
-  let get_swap = () => get_swap(wasm_mod, env, 0);
-  let tee_swap = tee_swap(wasm_mod, env, 0);
-  Expression.Block.make(
-    wasm_mod,
-    gensym_label("allocate_char_uninitialized"),
-    [
-      store(
-        ~offset=0,
-        wasm_mod,
-        tee_swap(heap_allocate(wasm_mod, env, 2)),
-        Expression.Const.make(
-          wasm_mod,
-          const_int32(tag_val_of_heap_tag_type(CharType)),
-        ),
-      ),
-      get_swap(),
-    ],
-  );
+let create_char = (wasm_mod, env, char) => {
+  let uchar = List.hd @@ Utf8.decodeUtf8String(char);
+  let uchar_int: int = Utf8__Uchar.toInt(uchar);
+  let grain_char = uchar_int lsl 3 lor 0b010;
+  Expression.Const.make(wasm_mod, const_int32(grain_char));
 };
 
 let allocate_closure =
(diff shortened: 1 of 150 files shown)


stdlib/string.gr
```diff
@@ -351,17 +415,16 @@ export let rec implode = (arr: Array<Char>) => {
   let mut stringByteLength = 0n
 
   for (let mut i = 0n; i < arrLength; i += 1n) {
-    let char = WasmI32.load(arr + (i << 2n), 8n)
-    let byte = WasmI32.load8U(char, 4n)
+    let usv = WasmI32.load(arr + (i << 2n), 8n) >>> 3n
```
cician, line 418: There's a lot going on here that makes it unreadable for someone without deep knowledge of Grain's internals.
A function for untagging the char and one for reading the array element would make it self documenting. (at the price of the function call, if it's not inlined, of course)

stdlib/string.gr
```diff
@@ -351,17 +415,16 @@ export let rec implode = (arr: Array<Char>) => {
...
 
-    let n = if ((byte & 0x80n) == 0x00n) {
+    let n = if (usv <= 0x80n) {
```
cician, line 420: This should be `let n = if (usv <= 0x7Fn) {`. Code point 128 gets encoded as two bytes in UTF-8.

cician (commented):
I've ogled the stdlib part. There's one bug and a few possible minor improvements, but looks good to me overall. I don't have the knowledge to review the compiler part or the snapshots. Also I haven't looked at manual memory usage correctness aspect of it yet. I'll try to use the leak checker on it later.

stdlib/char.gr
```diff
@@ -60,69 +57,14 @@ export let isValid = charCode => {
  */
 @disableGC
 export let rec code = (char: Char) => {
-  // Algorithm from https://encoding.spec.whatwg.org/#utf-8-decoder
+  let usv = WasmI32.shrU(WasmI32.fromGrain(char), 3n)
```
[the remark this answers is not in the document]
ospencer: Yeah, feels bad since we don't have inlining for it yet. I'll add a comment.

stdlib/string.gr
ospencer: Whoops, I meant to add a comment to this.

stdlib/char.gr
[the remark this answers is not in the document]
ospencer: I went down the long rabbit hole of getting the tagging functions inlined, so now it reads much better.

stdlib/string.gr
ospencer: Should be reable now

marcusroberts (approved):
This is just a LGTM rather than a strong approval!

stdlib/string.gr
```diff
@@ -221,14 +284,17 @@ export let rec charAt = (position, string: String) => {
...
   let position = WasmI32.fromGrain(position) >>> 1n
   let string = WasmI32.fromGrain(string)
   let mut ptr = string + 8n
   let end = ptr + size
   let mut counter = 0n
-  let mut result = 0n
+  let mut result = WasmI32.toGrain(0n): Char
   while (ptr < end) {
+    if (counter == position) {
+      result = tagChar(getCodePoint(ptr))
```
peblair, line 295: This change does appear to change the semantics of this function, FYI

peblair (approved):
This is awesome! One tiny comment for you to look at, but this looks good!

compiler/src/codegen/compcore.re
```diff
@@ -1630,59 +1630,16 @@ let allocate_bytes_uninitialized = (wasm_mod, env, size) => {
...
-    ],
-  );
+let create_char = (wasm_mod, env, char) => {
+  let uchar = List.hd @@ Utf8.decodeUtf8String(char);
+  // There appears to be an issue with the distribution of the Utf8 lib that
+  // doesn't include its Uchar implementation, and the Uchar type exposed by
+  // the Utf8 module is incompatible with Stdlib.Uchar. The Uchar type is
+  // implemented as a plain int, so we do an unsafe cast. This avoids us
+  // having to reimplement a UTF8 -> int conversion.
+  let uchar_int: int = Obj.magic(uchar);
```
phated, line 1640: I believe dune have this wrapped as `Utf8__Uchar.toInt(uchar)`. Not sure why the LSP doesn't present that as one of the options

stdlib/runtime/dataStructures.gr
```diff
@@ -174,3 +166,23 @@ export primitive tagSimpleNumber: WasmI32 -> Number = "@tag.simple_number"
  */
 @unsafe
 export primitive untagSimpleNumber: Number -> WasmI32 = "@untag.simple_number"
+
+/**
+ * Tag a char.
+ *
+ * @export
```
phated, line 173:
```suggestion
```
Where the heck did these come from 🤦

stdlib/runtime/dataStructures.gr
```diff
@@ -174,3 +166,23 @@ export primitive tagSimpleNumber: WasmI32 -> Number = "@tag.simple_number"
...
+ * @param num The usv to tag
```
phated, line 174:
```suggestion
 * @param num: The usv to tag
```

stdlib/runtime/dataStructures.gr
```diff
@@ -174,3 +166,23 @@ export primitive tagSimpleNumber: WasmI32 -> Number = "@tag.simple_number"
...
+ * @returns The tagged char
+ */
+@unsafe
+export primitive tagChar: WasmI32 -> Char = "@tag.char"
+
+/**
+ * Untag a char.
+ *
+ * @export
+ * @param num The char to untag
```
phated, line 184:
```suggestion
 * @param num: The char to untag
```

phated (approved):
Absolute 🔥 

Amazing work @ospencer

stdlib/string.gr
ospencer: I think it just looks like it does—the info about number of bytes isn't needed anymore, so this just saves a load by happening first (and is a bit more correct since we don't potentially load memory we're not supposed to).

PR #1104, opened 2022-01-17
fix(graindoc): Remove special idents from receiving parens

phated:
Do any of these actually need parens when printed via `oprint`? I'm pretty sure this was just a holdover from ocaml.


ospencer (approved):
Looks like this should be good 👍🏽

ng-marcus (approved):
LGTM!

phated:
It seems that phantom @marcusroberts approved this, so I'm going to get this merged so we can finish the release.

marcusroberts (commented):
Also looks good to me!

PR #1115, opened 2022-01-22
feat(compiler)!: Add `--memory-base` flag

### LINKED ISSUE #1111: Compiler: Add heap.base primitive
@peblair I was chatting with Blaine about primitives and I had an idea about this:

> Right now all of our primitives are functions, but I think we could have some values as well! I was thinking through that wasm4 issue and philip's PR for reserving a chunk of heap space for something that needs it—I think it's a good use case for a `@heap.base` primitive that you can populate with like a `--memory-base` to scoot up where we start the heap. Then you can reserve whatever amount of space you need for whatever your use case is, would set us up to support something like AssemblyScript's `memory.data` (which would be so nice) etc.

He liked that idea. If you think that sounds reasonable too, I can make an issue for it!

_Originally posted by @ospencer in https://github.com/grain-lang/grain/issues/935#issuecomment-984737569_

peblair:
Closes #1111. The introduction of the `attrs` result to `transl_prim` felt a little awkward, but I think it was the cleanest way to slide this into the existing well-formedness checks in the compiler.

Also, I'm open to better names for this PR.

Closes #935 (supersedes it)


compiler/src/typed/translprim.re
```diff
@@ -3,7 +3,13 @@ open Ast_helper;
 open Typedtree;
 open Parsetree;
 
+type wi32_constant =
+  | HeapBase
+  | HeapStart
+  | HeapTypeMetadataPointer;
+
 type primitive =
+  | PrimitiveWasmI32(wi32_constant)
```
ospencer, line 12: Could we instead call this `PrimitiveConst` and call `wi32_constant` `prim_constant`? I definitely see us dropping in other types as well! (Want to reserve Primitive0 for zero-arg functions.)

compiler/src/typed/translprim.re
```diff
@@ -34,6 +40,12 @@ let pat_c = mkpatvar("c");
 let prim_map =
   PrimMap.of_seq(
     List.to_seq([
+      ("@heap.base", PrimitiveWasmI32(HeapBase)),
+      ("@heap.start", PrimitiveWasmI32(HeapStart)),
+      (
+        "@heap.type_metadata_pointer",
```
ospencer, line 46: I think we can drop the `_pointer` part to match the other two, with `HeapTypeMetadata`.

compiler/src/typed/translprim.re
```diff
@@ -1408,23 +1420,70 @@ let transl_prim = (env, desc) => {
...
+  // @disableGC let heapBase = 0x400n
+  //
+  // Which can pass through the rest of the compilation without issue.
+  //
+  let (value, attrs) =
     switch (prim) {
+    | PrimitiveWasmI32(const) =>
+      let value =
+        switch (const) {
+        // [NOTE] should be kept in symc with `runtime_heap_ptr` and friends in `compcore.re`
```
ospencer, line 1454:
```suggestion
        // [NOTE] should be kept in sync with `runtime_heap_ptr` and friends in `compcore.re`
```

compiler/src/typed/translprim.re
```diff
@@ -1408,23 +1420,70 @@ let transl_prim = (env, desc) => {
 
   let diable_gc = [(Location.mknoloc("disableGC"), [])];
```
ospencer, line 1421: Do you mind fixing the spelling of this while you're in here?

compiler/src/utils/config.re
```diff
@@ -403,6 +403,14 @@ let optimization_level =
     Level_three,
   );
 
+let memory_base =
+  opt(
+    ~doc="Set the start address for the Grain runtime heap.",
+    ~names=["memory-base"],
+    ~conv=Cmdliner.Arg.int,
+    0x400,
+  );
```
ospencer, line 412: Food for thought: I'm not sure if we ever expose what the default value is, but I wonder if someone seeing this will think that the low 1k is free game (and it isn't). That of course leads to the next important thing: in `compcore.re` and `link.re`, there's this Binaryen setting that'll need to be toggled off if this flag is ever set, `Settings.set_low_memory_unused(false)` to tell Binaryen to not write data to that region. That makes me wonder if this should instead be `option(int)`, with the default as `None`. If we did that, then we can choose to start the heap at 0x400 and let Binaryen own that low 1k, and when the user sets a value, we know to turn off that setting. That'd also distinguish the `--memory-base=1024` case from the default. I'm not married to that approach, but I think it could make sense here.

ospencer (changes requested):
Nice! And it makes sense that no snapshots changed since the newly generated code should be exactly the same as the old code.

The major thing we need to fix is just that one Binaryen setting so it doesn't trash data in the space that you "reserved" 😂 

A couple of tests would be good as well. The big one I'm interested in is starting with a memory base that's larger than the initial number of memory pages—I think we start with something like 64 pages. Just want to make sure that if you try to reserve more space than that everything still works.

Also, I believe you need to declare the new flag in the CLI so it can be used.

For the PR title, I think it's not bad really, but we do want to mark it as breaking in case anyone was depending on the static values, since they should use the primitives now. Maybe ```
feat(compiler)!: Add `--memory-base` flag
```
to highlight how to use it?

File: compiler/src/codegen/compcore.re [hunk 6]
```diff
@@ -1538,13 +1539,13 @@ let heap_allocate = (wasm_mod, env, num_words: int) =>
           ),
           Expression.Block.make(
             wasm_mod,
-            gensym_label("store_runtime_heap_ptr"),
+            gensym_label("store_runtime_heap_ptr^"),
```
phated, line 1542: Do we actually want `^` in this label?

phated (changes requested):
One style question, but I was also wondering if the attributes are even needed if we land #1074?

compiler/src/typed/translprim.re
peblair: So, the reason why I called it that was _because_ I expect us to drop in other types 🙂. I would imagine that we could have `PrimitiveWasmI64`, etc...I think that makes sense, no?

compiler/src/typed/translprim.re
peblair: I thought if I did you would hit me with "this should be a separate PR" 😅 

You don't think it should be a separate one?

compiler/src/utils/config.re
peblair: Instead of being an `option(int)`, why not just have it be `if memory_base < 0x400`?

compiler/src/codegen/compcore.re
peblair: Nope! Good catch

compiler/src/typed/translprim.re
ospencer: Hmm, I don't really see this as being any different than the other ones (Primitive1, etc) and we wouldn't have special versions for different typed functions. I think it might get annoying if we started to add a bunch of these for all of the different types we might have.

compiler/src/typed/translprim.re
ospencer: Nah, because you used the value in this PR. Good thinking though 😉

compiler/src/typed/translprim.re
peblair: I saw them as different because we are not wrapping them with lambdas.

compiler/src/utils/config.re
ospencer: If you supply the flag at all, we'll need to do that binaryen change because you're expecting to be able to use that memory yourself.

compiler/src/typed/translprim.re
ospencer: A value's a value!

compiler/src/utils/config.re
peblair: Ah, I understand now. If you think it's important, we can do that, but I really think that we should just document it appropriately so that users of this flag (which probably represent a very small group) know not to touch `0x0`-`0x400`. Personally, I think that would be better than adding potentially strange side effects to the flag (like, as a user, I'd be sort of surprised if `--memory-base 0x1200` caused that binaryen flag to get switched)

peblair:
@phated 

> but I was also wondering if the attributes are even needed if we land #1074?

No, they are not. If this lands before that does, then that PR should probably remove the `attrs` value added here. I'm remiss to say that this should be blocked by #1074, though, since this is such a small change (ATM).

compiler/src/utils/config.re
ospencer: Yeah, it's pretty important. Often users don't really have any control over that memory with them needing to reserve that space for the host to use it, like in Wasm-4's use case.

ospencer:
Hmm, we might still need the attributes, since the way we have it working right now would  require `@unsafe` instead of `@disableGC`. We could maybe change that though.

peblair:
That's a good point. Even less reason for this to be blocked 🙂

peblair:
Tests added.
To be super safe, I checked that the constant gets wired up correctly by `veryLarge.gr` in the `runtime/string` WAT, and it looks like it is working great:
image

compiler/test/suites/memory_base.re
```diff
@@ -0,0 +1,18 @@
+open Grain_tests.TestFramework;
+open Grain_tests.Runner;
+
+describe("memory_base", ({test}) => {
+  let assertFileRun = makeFileRunner(test);
+  // Non-snapshots for now, since we want to track functional issues.
+  // Maybe these can be migrated to snapshots once we're confident in
+  // everything being stable
```
ospencer, line 8:
```suggestion
  // Non-snapshots since we want to track functional issues.
```
Hopefully we're confident they're stable now 😅 but yeah I think it makes sense for them to always be actually run.

compiler/test/suites/memory_base.re
```diff
@@ -0,0 +1,18 @@
...
+  assertFileRun("basecase", "memoryBase/basecase", "HelloWorld\n");
+  assertFileRun("no_op", "memoryBase/noop", "HelloWorld\n");
```
ospencer, line 10: `no_op` might not be a good name since it does have an effect, turning off some class of optimizations. Not sure what to call it though 🤔

compiler/src/codegen/compcore.re
```diff
@@ -351,11 +346,23 @@ let lookup_ext_func = (env, modname, itemname) =>
 
 // Static pointer to the runtime heap
 // Leaves low 1000 memory unused for Binaryen optimizations
-let runtime_heap_ptr = 0x400;
+let runtime_heap_ptr = ref(0x400);
```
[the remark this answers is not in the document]
peblair: I should fix it, but it doesn't matter (since it gets set by `reset()`)

compiler/test/suites/memory_base.re
peblair: 😡 😅

ospencer (approved):
Perfect!

compiler/src/codegen/compcore.re
```diff
@@ -3550,7 +3566,13 @@ let compile_wasm_module = (~env=?, ~name=?, prog) => {
       default_features;
     };
   let _ = Module.set_features(wasm_mod, features);
-  let _ = Settings.set_low_memory_unused(true);
+  // we set low_memory_unused := true iff the user has not specified a memory base.
```
phated, line 3569:
```suggestion
  // we set low_memory_unused := true if the user has not specified a memory base.
```
nit

peblair: Not a typo, though?

phated: Had to google `iff` but it seems that it is a bi-directional if, which isn't true in this scenario, right? If something in the program somehow `set_low_memory_unused`, you wouldn't enable the memory base.

ospencer (approved):

compiler/src/codegen/compcore.re
peblair: Yes, `iff` is a bidirectional if ("if and only if"). It is true in this scenario, since we are the ones setting it right here. On the next line, we set the value of `low_memory_unused` to whatever the value of `Option.is_none(memory_base^)` is, meaning that `low_memory_unused` will be `true` if and only if `--memory-base` was not passed.

phated:
```suggestion
  // we set low_memory_unused := true if and only if the user has not specified a memory base.
```
I shouldn't have to google to understand comments 😛

phated:
Can you add `Closes #935 (supersedes it)` to your PR body or link it in the sidebar?

ospencer (approved):

phated (approved):
Excellent work! Love that we're knocking 0.5 things down so quickly ❤️

marcusroberts (approved):
Looks very good!

compiler/src/typed/translprim.re
```diff
@@ -1406,25 +1422,80 @@ let transl_prim = (env, desc) => {
     | Not_found => failwith("This primitive does not exist.")
     };
 
-  let diable_gc = [(Location.mknoloc("disableGC"), [])];
+  let disable_gc = [(Location.mknoloc("disableGC"), [])];
```
jozanza, line 1425: Nice

jozanza (approved):

PR #1116, opened 2022-01-22
feat(graindoc)!: Add --current-version flag, required for since/history attributes

### LINKED ISSUE #1110: Graindoc: Add required --current-version flag instead of using grain version
When comparing the `@since` attribute in graindoc, we currently use the version of grain itself to display `next` on newer versions. This doesn't work for external libraries, nor does it work for Grain itself (our docs are always 1 version behind).

We need to add a required `--current-version` flag that sets the version to compare. This is a breaking change since it is adding a required flag.

phated:
Closes #1110 

This adds a `--current-version` flag to the `grain doc` command line. If your docs are using `@since` or `@history`, this flag is required (thus breaking) because we need (want?) a version to compare so we can insert `next` into the generated markdown if the docs are being generated before a release is made.

### DIFF
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -11,6 +11,27 @@ type t = {
   attributes: list(Comments.Attribute.t),
 };
 
+exception
+  MissingFlag({
+    flag: string,
+    attr: string,
+  });
+
+let () =
+  Printexc.register_printer(exn => {
+    switch (exn) {
+    | MissingFlag({flag, attr}) =>
+      let msg =
+        Printf.sprintf(
+          "Must provide %s when generating docs with `%s` attribute.",
+          flag,
+          attr,
+        );
+      Some(msg);
+    | _ => None
+    }
+  });
+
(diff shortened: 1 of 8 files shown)


ospencer (approved):
Looks great!

compiler/graindoc/docblock.re
```diff
@@ -11,6 +11,27 @@ type t = {
   attributes: list(Comments.Attribute.t),
 };
 
+exception
+  MissingFlag({
```
peblair, line 15: I'm a bit confused -- why do we require the flag? Can we not infer a default from the compiler version?

phated: Compiler version doesn't work when external libraries authors are using Grain doc. Their `@since` and `@history` will be specific to their development workflows and library version.

ospencer:
Maybe we'll be able to do something fun in the future with the `Silo.toml` file version!

peblair (approved):
Looks good!

marcusroberts (approved):
LGTM!

PR #1117, opened 2022-01-22
chore: Remove explicit exit 0 in native executables

phated:
The process should exit successfully with code 0 upon a successful cmdliner run. Having an explicit `exit(0)` here caused weirdness in some browser hacking I was working on.


ospencer (approved):

marcusroberts (approved):
Good fix!

Issue #1072, opened 2021-12-11
Revise simple number tagging and untagging throughout the stdlib


cician:
Throughout the entire standard library we have some possibly problematic usage of simple number tagging and untagging.

1. In some if not most cases where a `WasmI32` value is "tagged" into a SImple number, we should use `conv.wasmI32ToNumber` instead of "tagging" it directly because a range of possible `WasmI32` numbers don't fit into Simple numbers. Issue #1064 is an example of this. We can review case by case if direct tagging is OK or not in order to avoid a performance regressions, but I think `conv.wasmI32ToNumber` is fast enough that we can just use it everywhere.

2. There seem to be many cases of ad-hoc tagging and untagging without even calling a function like `tagSimpleNumber`. Other than the possibility of above mentioned problems, I think it makes code too cryptic, so we may change it even if direct tagging or untagging is justified. Here's a list of places that seem to have some raw tagging/untagging logic like this. It's just a result of quick text search of shift operators, so probably not accurate or exhaustive.

- runtime/wasi.gr
stringOfSystemError

- map.gr
toArray 

- buffer.gr
addStringSlice

- array.gr
initLength

- string.gr
grainToWasmNumber
charAt
explodeHelp
split
slice
contains
utf16Length
encodeAtHelp
encodeHelp

- runtime/string.gr
getVariantName
toStringHelp
listToString
parseInt

- sys/process.gr
exit
sigRaise

- sys/file.gr
Too many to list. Just search `>> 1n`.

3. We may want to unify `runtime.dataStructures.tagSimpleNumber` and `runtime.numbers.tagSimple` functions. To avoid braking changes maybe one could reexport the other. If there are justified uses of direct untagging instead of calling `coerceNumberToWasmI32`, then we may want to export `runtime.numbers.untagSimple` and use this instead of ad-hoc conversion.

PR #1123, opened 2022-01-23
fix(stdlb): Fixed Bytes.length for size over 2GiB.

cician:
Bytes.length was always tagging the size as a Simple Number, so it was failing for sizes over 2GiB.

See also #1072.


ospencer (approved):
Looks great!

phated (approved):
LGTM 👍 At some point we should investigate adding a test for 2+ gb files 😉

PR #1124, opened 2022-01-23
feat(stdlib): Added Bytes.clear function. Avoid allocation in Buffer.clear

cician:
This PR adds a simple function `clear` to the Bytes module and uses it in `Buffer.clear` instead if `Bytes.fill`. This avoids allocating an instance of Int32 on the heap for each call to `Buffer.clear`.

### DIFF
--- a/stdlib/bytes.gr
+++ b/stdlib/bytes.gr
@@ -324,10 +324,10 @@ export let rec clear = (bytes: Bytes) => {
   let (+) = WasmI32.add
   let src = WasmI32.fromGrain(bytes)
   let size = getSize(src)
-  let ret = Memory.fill(src + _VALUE_OFFSET, 0n, size)
+  Memory.fill(src + _VALUE_OFFSET, 0n, size)
   Memory.decRef(WasmI32.fromGrain(bytes))
   Memory.decRef(WasmI32.fromGrain(clear))
-  ret
+  void
 }
 
 /**
--- a/stdlib/bytes.gr
+++ b/stdlib/bytes.gr
@@ -319,15 +319,12 @@ export let fill = (value: Int32, bytes: Bytes) => {
  *
  * @since v0.5.0
  */
-@disableGC
-export let rec clear = (bytes: Bytes) => {
+@unsafe
+export let clear = (bytes: Bytes) => {
   let (+) = WasmI32.add
   let src = WasmI32.fromGrain(bytes)
   let size = getSize(src)
   Memory.fill(src + _VALUE_OFFSET, 0n, size)
-  Memory.decRef(WasmI32.fromGrain(bytes))
-  Memory.decRef(WasmI32.fromGrain(clear))
-  void
 }
 
 /**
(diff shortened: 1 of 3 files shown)


jozanza (approved):

stdlib/bytes.gr
```diff
@@ -367,6 +367,24 @@ export let rec fill = (value: Int32, bytes: Bytes) => {
   ret
 }
 
+/**
+ * Replaces all bytes in a byte sequnce with value zero.
```
ospencer, line 371:
```suggestion
 * Replaces all bytes in a byte sequence with zeroes.
```

stdlib/bytes.gr
```diff
@@ -367,6 +367,24 @@ export let rec fill = (value: Int32, bytes: Bytes) => {
...
+ */
+@disableGC
+export let rec clear = (bytes: Bytes) => {
+  let (+) = WasmI32.add
+  let src = WasmI32.fromGrain(bytes)
+  let size = getSize(src)
+  let ret = Memory.fill(src + _VALUE_OFFSET, 0n, size)
+  Memory.decRef(WasmI32.fromGrain(bytes))
+  Memory.decRef(WasmI32.fromGrain(clear))
+  ret
```
ospencer, line 385: Since `fill` returns void, this feels a little weird. I think this is a bit more clear:
```suggestion
  Memory.fill(src + _VALUE_OFFSET, 0n, size)
  Memory.decRef(WasmI32.fromGrain(bytes))
  Memory.decRef(WasmI32.fromGrain(clear))
  void
```

cician: Oops. I blindly copy and pasted from the `fill` function which does the same thing, returning void.

peblair (approved):

stdlib/bytes.gr
```diff
@@ -367,6 +367,24 @@ export let rec fill = (value: Int32, bytes: Bytes) => {
   ret
 }
 
+/**
+ * Replaces all bytes in a byte sequence with zeroes.
+ *
+ * @param bytes: The byte sequence to clear
+ *
+ * @since v0.3.2
```
phated, line 375:
```suggestion
 * @since v0.5.0
```

marcusroberts (approved):
Looks good to me!

stdlib/bytes.gr
```diff
@@ -367,6 +367,24 @@ export let rec fill = (value: Int32, bytes: Bytes) => {
...
+@disableGC
+export let rec clear = (bytes: Bytes) => {
+  let (+) = WasmI32.add
+  let src = WasmI32.fromGrain(bytes)
+  let size = getSize(src)
+  Memory.fill(src + _VALUE_OFFSET, 0n, size)
+  Memory.decRef(WasmI32.fromGrain(bytes))
+  Memory.decRef(WasmI32.fromGrain(clear))
+  void
+}
```
ospencer, line 386:
```suggestion
@unsafe
export let clear = (bytes: Bytes) => {
  let (+) = WasmI32.add
  let src = WasmI32.fromGrain(bytes)
  let size = getSize(src)
  Memory.fill(src + _VALUE_OFFSET, 0n, size)
}
```

ospencer: Converting this over to use `@unsafe`.

PR #1125, opened 2022-01-23
fix(stdlib): Buffer.autogrow was making unnecessary intermediate resizes of the buffer.

cician:
In some cases Buffer.autogrow was making unnecessary intermediate resizes of the buffer. In particular when adding a big chunk of data at once.


jozanza (approved):
Nice. lgmt!

ospencer (approved):
Looks great!

PR #1127, opened 2022-02-02
chore: Add support for Gitpod by specifying a build image and adding a button

marcusroberts:


willemneal:
Yay!

jozanza (approved):

README.md
```diff
@@ -58,4 +58,6 @@ yarn run
 
 This will display an interactive session where you can select the project and command you want.
 
+[![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/grain-lang/grain)
```
ospencer, line 61: I think it makes a bit more sense to have this on the website's Getting Grain page than here, but I dunno what others might think.

willemneal: Why not in both places? Some people arrive here first.

peblair: +1 on putting this on the Grain site, but I'm indifferent w.r.t. whether it should go here as well.

phated: I thought there was a difference between the gitpod docker files for "using Grain" and "developing Grain"? If so, we should be advertise "using Grain" on the website and "developing Grain" on the repo.

marcusroberts: The button launches the attached repo, so the develop button goes on the main Grain repo and gives you an environment to build the compiler in.  It can go on any page on the repo.

The develop with Grain is on the new template repo and just gives you the compiler binaries

phated: Makes sense. I think maybe we should not include a badge/image and instead just have a link under its own section.

phated, line 61:
```suggestion
```

Shall we just delete this so we can get it merged and then decide where to advertise it later?

phated (changes requested):

README.md
ng-marcus: Sure.

README.md
```diff
@@ -58,4 +58,5 @@ yarn run
 
 This will display an interactive session where you can select the project and command you want.
 
+
```
phated, line 61:
```suggestion
```
Dangit Blaine

phated (approved):

ospencer (approved):

PR #1131, opened 2022-02-06
feat(compiler)!: Improved long running LSP server

marcusroberts:
The Grain LSP can now be run as a long-running server process listening for requests on stdin.

This brings big performance improvements as the current compile state is in memory ready for lookups for hovers, go to definitions, peek definition, etc.    It also hugely reduces the overhead in the information sent from the server to the client as only the information needed is returned instead of all data as before.

The LS protocol is now cleanly implemented for lenses, previously I was relying on the VSCode extension to decode which worked OK but broke other editors.

A restart LSP command has also been added.

### DIFF
--- a/cli/bin/exec.js
+++ b/cli/bin/exec.js
@@ -96,8 +96,42 @@ function execGrainformat(
   );
 }
 
+function getGrainlsp() {
+  const grainlsp = path.join(__dirname, "grainlsp.exe");
+
+  // TODO: Maybe make an installable path & check it?
+  if (process.pkg || !fs.existsSync(grainformat)) {
+    const node = process.execPath;
+    const grainlsp_js = path.join(__dirname, "grainlsp.js");
+    return `"${node}" ${grainlsp_js}`;
+  }
+
+  return `${grainlsp}`;
+}
+
+const grainlsp = getGrainlsp();
+
+function execGrainlsp(
+  program,
+  execOpts = { stdio: "pipe" }
+) {
+  const flags = [];
+  const options = program.opts();
+  program.options.forEach((option) => {
+    if (!option.forward) return;
+    const flag = option.toFlag(options);
+    if (flag) flags.push(flag);
+  });
+
+  return execSync(
+    `${grainlsp} ${flags.join(" ")} `,
+    execOpts
+  );
+}
+
 module.exports = {
   grainc: execGrainc,
   graindoc: execGraindoc,
   grainformat: execGrainformat,
+  grainlsp: execGrainlsp,
 };
--- a/cli/bin/exec.js
+++ b/cli/bin/exec.js
@@ -116,7 +116,9 @@ function execGrainlsp(
   execOpts = { stdio: "pipe" }
 ) {
   const flags = [];
-  const options = program.opts();
+
+  const options = program.parent.options.concat(program.options);
+  const opts = { ...program.parent.opts(), ...program.opts() };
   program.options.forEach((option) => {
     if (!option.forward) return;
     const flag = option.toFlag(options);
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -202,13 +202,13 @@ program
   });
 
 program
-  .command("lsp <file>")
-  .description("check a grain file for LSP")
+  .command("lsp")
+  .description("start the Grain LSP server")
   .action(
-    wrapAction(function (file) {
+    wrapAction(function (options,program) {
       // The lsp subcommand inherits all options of the
       // top level grain command
-      lsp(file, program);
+      lsp(program);
     })
   );
 
--- a/compiler/dune-project
+++ b/compiler/dune-project
@@ -1,157 +1,263 @@
 (lang dune 2.3)
+
 (name grain)
+
 (version 0.4.6)
+
 (using menhir 2.0)
 
 ; Flip this to `true` when we want to generate opam files again
+
 (generate_opam_files false)
 
 (license LGPL-3.0)
+
 (authors "Philip Blair" "Oscar Spencer")
+
 (maintainers "philip@grain-lang.org" "oscar@grain-lang.org")
-(source (github grain-lang/grain))
+
+(source
+ (github grain-lang/grain))
+
 (homepage "https://grain-lang.org")
+
 (documentation "https://grain-lang.org/docs/")
 
 ; These aren't used currently but left in for reference when we generate opam files again
-(package
-  (name grain)
-  (synopsis "The core Grain language compiler library")
-  (depends
-    (reason (>= 3.6.0))
-    (conf-openssl (>= 2))
-    (dune-build-info (>= 2.0))
-    (ppx_sexp_conv (>= 0.14.0))
-    (sexplib (>= 0.14.0))
-    (cmdliner (>= 1.0.2))
-    (grain_utils (>= 0))
-    (grain_parsing (>= 0))
-    (grain_typed (>= 0))
-    (grain_middle_end (>= 0))
-    (grain_diagnostics (>= 0))
-    (grain_codegen (>= 0))))
 
 (package
-  (name grainc)
-  (synopsis "The core Grain language compiler CLI")
-  (depends
-    (reason (>= 3.6.0))
-    (js_of_ocaml-compiler (>= 3.6.0))
-    (dune-build-info (>= 2.0))
-    (ppx_sexp_conv (>= 0.14.0))
-    (sexplib (>= 0.14.0))
-    (binaryen (= 0.9.1))
-    (cmdliner (>= 1.0.2))
-    (grain (>= 0))))
+ (name grain)
+ (synopsis "The core Grain language compiler library")
+ (depends
+  (reason
+   (>= 3.6.0))
+  (conf-openssl
+   (>= 2))
+  (dune-build-info
+   (>= 2.0))
+  (ppx_sexp_conv
+   (>= 0.14.0))
+  (sexplib
+   (>= 0.14.0))
+  (cmdliner
+   (>= 1.0.2))
+  (grain_utils
+   (>= 0))
+  (grain_parsing
+   (>= 0))
+  (grain_typed
+   (>= 0))
+  (grain_middle_end
+   (>= 0))
+  (grain_diagnostics
+   (>= 0))
+  (grain_codegen
+   (>= 0))))
 
 (package
-  (name graindoc)
-  (synopsis "Grain documentation generator")
-  (depends
-    (reason (>= 3.6.0))
-    (dune-build-info (>= 2.0))
-    (binaryen (= 0.9.1))
-    (cmdliner (>= 1.0.2))
-    (grain (>= 0))))
+ (name grainc)
+ (synopsis "The core Grain language compiler CLI")
+ (depends
+  (reason
+   (>= 3.6.0))
+  (js_of_ocaml-compiler
+   (>= 3.6.0))
+  (dune-build-info
+   (>= 2.0))
+  (ppx_sexp_conv
+   (>= 0.14.0))
+  (sexplib
+   (>= 0.14.0))
+  (binaryen
+   (= 0.9.1))
+  (cmdliner
+   (>= 1.0.2))
+  (grain
+   (>= 0))))
 
 (package
-  (name grainformat)
-  (synopsis "Grain formatter")
-  (depends
-    (reason (>= 3.6.0))
-    (dune-build-info (>= 2.0))
-    (binaryen (= 0.9.1))
-    (cmdliner (>= 1.0.2))
-    (grain (>= 0))))
+ (name graindoc)
+ (synopsis "Grain documentation generator")
+ (depends
+  (reason
+   (>= 3.6.0))
+  (dune-build-info
+   (>= 2.0))
+  (binaryen
+   (= 0.9.1))
+  (cmdliner
+   (>= 1.0.2))
+  (grain
+   (>= 0))))
 
 (package
-  (name grain_linking)
-  (synopsis "The Grain linker")
-  (depends
-    (reason (>= 3.6.0))
-    (ppx_sexp_conv (>= 0.14.0))
-    (sexplib (>= 0.14.0))
-    (binaryen (= 0.9.1))
-    (ocamlgraph (>= 2.0.0))
-    (grain_utils (>= 0))
-    (grain_codegen (>= 0))))
+ (name grainformat)
+ (synopsis "Grain formatter")
+ (depends
+  (reason
+   (>= 3.6.0))
+  (dune-build-info
+   (>= 2.0))
+  (binaryen
+   (= 0.9.1))
+  (cmdliner
+   (>= 1.0.2))
+  (grain
+   (>= 0))))
 
 (package
-  (name grain_codegen)
-  (synopsis "Grain WebAssembly code generation")
-  (depends
-    (reason (>= 3.6.0))
-    (ppx_sexp_conv (>= 0.14.0))
-    (sexplib (>= 0.14.0))
-    (cmdliner (>= 1.0.2))
-    (binaryen (= 0.9.1))
-    (grain_utils (>= 0))
-    (grain_middle_end (>= 0))))
+ (name grainlsp)
+ (synopsis "Grain LSP")
+ (depends
+  (reason
+   (>= 3.6.0))
+  (dune-build-info
+   (>= 2.0))
+  (cmdliner
+   (>= 1.0.2))
+  (grain
+   (>= 0))))
 
 (package
-  (name grain_middle_end)
-  (synopsis "Grain itermediate representations")
-  (depends
-    (reason (>= 3.6.0))
-    (ppx_sexp_conv (>= 0.14.0))
-    (sexplib (>= 0.14.0))
-    (cmdliner (>= 1.0.2))
-    (grain_utils (>= 0))
-    (grain_parsing (>= 0))
-    (grain_typed (>= 0))))
+ (name grain_linking)
+ (synopsis "The Grain linker")
+ (depends
+  (reason
+   (>= 3.6.0))
+  (ppx_sexp_conv
+   (>= 0.14.0))
+  (sexplib
+   (>= 0.14.0))
+  (binaryen
+   (= 0.9.1))
+  (ocamlgraph
+   (>= 2.0.0))
+  (grain_utils
+   (>= 0))
+  (grain_codegen
+   (>= 0))))
 
 (package
-  (name grain_parsing)
-  (synopsis "The Grain parser")
-  (depends
-    (reason (>= 3.6.0))
-    (ppx_sexp_conv (>= 0.14.0))
-    (sexplib (>= 0.14.0))
-    (menhir (= 20211125))
-    (cmdliner (>= 1.0.2))
-    (utf8 (>= 0.0.0))
-    (ppx_deriving_yojson (>= 3.5.2))
-    (yojson (>= 1.7.0))
-    (grain_utils (>= 0))))
+ (name grain_codegen)
+ (synopsis "Grain WebAssembly code generation")
+ (depends
+  (reason
+   (>= 3.6.0))
+  (ppx_sexp_conv
+   (>= 0.14.0))
+  (sexplib
+   (>= 0.14.0))
+  (cmdliner
+   (>= 1.0.2))
+  (binaryen
+   (= 0.9.1))
+  (grain_utils
+   (>= 0))
+  (grain_middle_end
+   (>= 0))))
 
 (package
-  (name grain_typed)
-  (synopsis "The Grain typechecker")
-  (depends
-    (reason (>= 3.6.0))
-    (ppx_sexp_conv (>= 0.14.0))
-    (sexplib (>= 0.14.0))
-    (ocamlgraph (>= 2.0.0))
-    (cmdliner (>= 1.0.2))
-    (ppx_deriving_yojson (>= 3.5.2))
-    (yojson (>= 1.7.0))
-    (grain_utils (>= 0))
-    (grain_parsing (>= 0))))
+ (name grain_middle_end)
+ (synopsis "Grain itermediate representations")
+ (depends
+  (reason
+   (>= 3.6.0))
+  (ppx_sexp_conv
+   (>= 0.14.0))
+  (sexplib
+   (>= 0.14.0))
+  (cmdliner
+   (>= 1.0.2))
+  (grain_utils
+   (>= 0))
+  (grain_parsing
+   (>= 0))
+  (grain_typed
+   (>= 0))))
 
 (package
-  (name grain_utils)
-  (synopsis "Various Grain utilities")
-  (depends
-    (reason (>= 3.6.0))
-    (ppx_sexp_conv (>= 0.14.0))
-    (sexplib (>= 0.14.0))
-    (fp (>= 0.0.0))
-    (fs (>= 0.0.0))
-    (cmdliner (>= 1.0.2))))
+ (name grain_parsing)
+ (synopsis "The Grain parser")
+ (depends
+  (reason
+   (>= 3.6.0))
+  (ppx_sexp_conv
+   (>= 0.14.0))
+  (sexplib
+   (>= 0.14.0))
+  (menhir
+   (= 20211125))
+  (cmdliner
+   (>= 1.0.2))
+  (utf8
+   (>= 0.0.0))
+  (ppx_deriving_yojson
+   (>= 3.5.2))
+  (yojson
+   (>= 1.7.0))
+  (grain_utils
+   (>= 0))))
+
+(package
+ (name grain_typed)
+ (synopsis "The Grain typechecker")
+ (depends
+  (reason
+   (>= 3.6.0))
+  (ppx_sexp_conv
+   (>= 0.14.0))
+  (sexplib
+   (>= 0.14.0))
+  (ocamlgraph
+   (>= 2.0.0))
+  (cmdliner
+   (>= 1.0.2))
+  (ppx_deriving_yojson
+   (>= 3.5.2))
+  (yojson
+   (>= 1.7.0))
+  (grain_utils
+   (>= 0))
+  (grain_parsing
+   (>= 0))))
+
+(package
+ (name grain_utils)
+ (synopsis "Various Grain utilities")
+ (depends
+  (reason
+   (>= 3.6.0))
+  (ppx_sexp_conv
+   (>= 0.14.0))
+  (sexplib
+   (>= 0.14.0))
+  (fp
+   (>= 0.0.0))
+  (fs
+   (>= 0.0.0))
+  (cmdliner
+   (>= 1.0.2))))
 
 (package
  (name grain_diagnostics)
  (synopsis "Diagnostics support")
  (depends
-  (reason (>= 3.6.0))
-  (ppx_deriving_yojson (>= 3.5.2))
-  (yojson (>= 1.7.0))
-  (grain_utils (>= 0))
-  (grain_parsing (>= 0))))
+  (reason
+   (>= 3.6.0))
+  (ppx_deriving_yojson
+   (>= 3.5.2))
+  (yojson
+   (>= 1.7.0))
+  (grain_utils
+   (>= 0))
+  (grain_parsing
+   (>= 0))))
 
 (package
-  (name grain-tests)
-  (synopsis "Tests")
-  (depends
-    (reason (>= 3.6.0))
-    (rely (>= 3.2.1))))
+ (name grain-tests)
+ (synopsis "Tests")
+ (depends
+  (reason
+   (>= 3.6.0))
+  (rely
+   (>= 3.2.1))))
--- a/compiler/grainlsp/config/flags.re
+++ b/compiler/grainlsp/config/flags.re
@@ -1,0 +1,20 @@
+module C = Configurator.V1;
+
+let () = {
+  C.main(~name="grainc_flags", c => {
+    let default = [];
+
+    let flags =
+      switch (C.ocaml_config_var(c, "system")) {
+      | Some("mingw64") =>
+        // MinGW needs the -static flag passed directly to the linker,
+        // to avoid needing MinGW locations in the path
+        // Ref https://github.com/grain-lang/binaryen.ml#static-linking
+        ["-ccopt", "--", "-ccopt", "-static"]
+      | Some(_) => default
+      | None => default
+      };
+
+    C.Flags.write_sexp("flags.sexp", flags);
+  });
+};
--- a/compiler/grainlsp/grainlsp.re
+++ b/compiler/grainlsp/grainlsp.re
@@ -1,0 +1,1 @@
+let _ = Lspserver.run();
--- a/compiler/grainlsp/hover.re
+++ b/compiler/grainlsp/hover.re
@@ -1,0 +1,156 @@
+open Grain_typed;
+
+let get_hover_from_statement =
+    (log, stmt: Typedtree.toplevel_stmt, line, char) =>
+  switch (stmt.ttop_desc) {
+  | TTopImport(import_declaration) => ("import declaration", None)
+  | TTopForeign(export_flag, value_description) => ("foreign", None)
+  | TTopData(data_declarations) =>
+    switch (data_declarations) {
+    | [] => ("", None)
+    | _ =>
+      let matches =
+        List.filter(
+          (dd: Typedtree.data_declaration) =>
+            Utils.is_point_inside_location(log, dd.data_loc, line, char),
+          data_declarations,
+        );
+      switch (matches) {
+      | [] => ("Internal error parsing", Some(stmt.ttop_loc))
+      | [node] =>
+        let name = node.data_name;
+        switch (node.data_kind) {
+        | TDataAbstract => ("Abstract declaration", Some(node.data_loc))
+        | TDataVariant(constrs) =>
+          let matches =
+            List.filter(
+              (cd: Typedtree.constructor_declaration) =>
+                if (Utils.is_point_inside_location(log, cd.cd_loc, line, char)) {
+                  true;
+                } else {
+                  false;
+                },
+              constrs,
+            );
+
+          switch (matches) {
+          | [decl] => (decl.cd_name.txt, Some(decl.cd_loc))
+          | _ => ("enum " ++ name.txt, Some(node.data_loc))
+          };
+
+        | TDataRecord(_) =>
+          switch (node.data_params) {
+          | [] => ("record " ++ node.data_name.txt, Some(node.data_loc))
+          | _ =>
+            let matches =
+              List.filter(
+                (dp: Typedtree.core_type) =>
+                  if (Utils.is_point_inside_location(
+                        log,
+                        dp.ctyp_loc,
+                        line,
+                        char,
+                      )) {
+                    true;
+                  } else {
+                    false;
+                  },
+                node.data_params,
+              );
+            switch (matches) {
+            | [decl] => (
+                Utils.lens_sig(decl.ctyp_type),
+                Some(decl.ctyp_loc),
+              )
+            | _ => ("record " ++ node.data_name.txt, Some(node.data_loc))
+            };
+          }
+        };
+      | _ => ("Internal error: Too many matches", None)
+      };
+    }
+
+  | TTopLet(export_flag, rec_flag, mut_flag, value_bindings) =>
+    switch (value_bindings) {
+    | [] => ("", None)
+    | _ =>
+      let matches =
+        List.map(
+          (vb: Typedtree.value_binding) =>
+            Utils.get_node_from_expression(log, vb.vb_expr, line, char),
+          value_bindings,
+        );
+      let filtered =
+        List.filter(
+          m =>
+            switch ((m: Utils.node_t)) {
+            | Error(_) => false
+            | NotInRange => false
+            | _ => true
+            },
+          matches,
+        );
+      switch (filtered) {
+      | [] =>
+        let vb = List.hd(value_bindings);
+        let expr = vb.vb_expr;
+        // return the type for the whole statement
+        (Utils.lens_sig(expr.exp_type), Some(vb.vb_loc));
+
+      | [node] =>
+        switch (node) {
+        | Error(err) => (err, None)
+        | NotInRange => ("Not in range", None)
+        | Expression(e) => (Utils.lens_sig(e.exp_type), Some(e.exp_loc))
+        | Pattern(p) => (Utils.lens_sig(p.pat_type), Some(p.pat_loc))
+        }
+      | _ => ("Internal error", None)
+      };
+    }
+
+  | TTopExpr(expression) =>
+    let node = Utils.get_node_from_expression(log, expression, line, char);
+    switch (node) {
+    | Error(err) => (err, None)
+    | NotInRange => (
+        Utils.lens_sig(expression.exp_type),
+        Some(expression.exp_loc),
+      )
+    | Expression(e) => (Utils.lens_sig(e.exp_type), Some(e.exp_loc))
+    | Pattern(p) => (Utils.lens_sig(p.pat_type), Some(p.pat_loc))
+    };
+
+  | TTopException(export_flag, type_exception) => ("exception", None)
+  | TTopExport(export_declarations) => ("export", None)
+  };
+
+let get_hover = (log, id, json, compiled_code) => {
+  switch (Utils.getTextDocumenUriAndPosition(json)) {
+  | (Some(uri), Some(line), Some(char)) =>
+    if (Hashtbl.mem(compiled_code, uri)) {
+      let ln = line + 1;
+
+      let compiled_code_opt = Hashtbl.find_opt(compiled_code, uri);
+      switch (compiled_code_opt) {
+      | None => ()
+      | Some(compiled_code) =>
+        let node = Utils.findBestMatch(compiled_code, ln, char);
+        switch (node) {
+        | Some(stmt) =>
+          let (signature, loc) =
+            get_hover_from_statement(log, stmt, ln, char);
+          let range =
+            switch (loc) {
+            | None => Utils.loc_to_range(stmt.ttop_loc)
+            | Some(l) => Utils.loc_to_range(l)
+            };
+          Rpc.send_hover(log, stdout, id, signature, range);
+        | None => ()
+        };
+      };
+    } else {
+      ();
+    }
+  | _ => ()
+  };
+};
--- a/compiler/grainlsp/lenses.re
+++ b/compiler/grainlsp/lenses.re
@@ -3,23 +3,50 @@ open Grain_typed;
 let get_signature_from_statement =
     (stmt: Grain_typed__Typedtree.toplevel_stmt) =>
   switch (stmt.ttop_desc) {
-  | TTopImport(import_declaration) => "import declaration"
-  | TTopForeign(export_flag, value_description) => "foreign"
-  | TTopData(data_declarations) => "data declaration"
-  | TTopLet(export_flag, rec_flag, mut_flag, value_bindings) =>
-    if (List.length(value_bindings) > 0) {
-      let vbs: Typedtree.value_binding = List.hd(value_bindings);
-      Utils.lens_sig(vbs.vb_expr.exp_type);
-    } else {
-      "";
-    }
+  | TTopImport(import_declaration) => None
+  | TTopForeign(export_flag, value_description) => None
+  | TTopData(data_declarations) =>
+    switch (data_declarations) {
+    | [] => None
+    | _ =>
+      let decls =
+        List.fold_left(
+          (acc, decl: Typedtree.data_declaration) =>
+            (acc == "" ? "" : acc ++ ", ")
+            ++ (
+              switch (decl.data_kind) {
+              | TDataVariant(_) => "enum " ++ decl.data_name.txt
+              | TDataRecord(_) => "record " ++ decl.data_name.txt
+              | TDataAbstract => "abstract" // not sure about this one
+              }
+            ),
+          "",
+          data_declarations,
+        );
 
+      Some(decls);
+    }
+  | TTopLet(export_flag, rec_flag, mut_flag, value_bindings) =>
+    switch (value_bindings) {
+    | [] => None
+    | _ =>
+      let vbses =
+        List.fold_left(
+          (acc, vbs: Typedtree.value_binding) =>
+            (acc == "" ? "" : acc ++ ", ")
+            ++ Utils.lens_sig(vbs.vb_expr.exp_type),
+          "",
+          value_bindings,
+        );
+      Some(vbses);
+    }
   | TTopExpr(expression) =>
     let expr_type = expression.exp_type;
-    Utils.lens_sig(expr_type);
-  | TTopException(export_flag, type_exception) => "exception"
-  | TTopExport(export_declarations) => "export"
+    Some(Utils.lens_sig(expr_type));
+  | TTopException(export_flag, type_exception) => None
+  | TTopExport(export_declarations) => None
   };
+
 let get_lenses = (typed_program: Typedtree.typed_program) => {
   List.fold_left(
     (acc, stmt: Grain_typed__Typedtree.toplevel_stmt) => {
@@ -27,9 +54,12 @@ let get_lenses = (typed_program: Typedtree.typed_program) => {
         Utils.get_raw_pos_info(stmt.ttop_loc.loc_start);
 
       let signature = get_signature_from_statement(stmt);
-      let lens: Rpc.lens_t = {line: startline, signature};
-
-      [lens, ...acc];
+      switch (signature) {
+      | None => acc
+      | Some(sigval) =>
+        let lens: Rpc.lens_t = {line: startline, signature: sigval};
+        [lens, ...acc];
+      };
     },
     [],
     typed_program.statements,
@@ -37,7 +67,6 @@ let get_lenses = (typed_program: Typedtree.typed_program) => {
 };
 
 let process_get_lenses = (log, id, json, compiled_code) => {
-  log("process_get_lenses requested");
   let params = Yojson.Safe.Util.member("params", json);
   let text_document = Yojson.Safe.Util.member("textDocument", params);
 
--- a/compiler/grainlsp/lenses.re
+++ b/compiler/grainlsp/lenses.re
@@ -1,4 +1,5 @@
 open Grain_typed;
+open Grain_diagnostics;
 
 let get_signature_from_statement =
     (stmt: Grain_typed__Typedtree.toplevel_stmt) =>
@@ -51,7 +52,7 @@ let get_lenses = (typed_program: Typedtree.typed_program) => {
   List.fold_left(
     (acc, stmt: Grain_typed__Typedtree.toplevel_stmt) => {
       let (file, startline, startchar, sbol) =
-        Utils.get_raw_pos_info(stmt.ttop_loc.loc_start);
+        Locations.get_raw_pos_info(stmt.ttop_loc.loc_start);
 
       let signature = get_signature_from_statement(stmt);
       switch (signature) {
--- a/compiler/grainlsp/lenses.re
+++ b/compiler/grainlsp/lenses.re
@@ -1,0 +1,59 @@
+open Grain_typed;
+
+let get_signature_from_statement =
+    (stmt: Grain_typed__Typedtree.toplevel_stmt) =>
+  switch (stmt.ttop_desc) {
+  | TTopImport(import_declaration) => "import declaration"
+  | TTopForeign(export_flag, value_description) => "foreign"
+  | TTopData(data_declarations) => "data declaration"
+  | TTopLet(export_flag, rec_flag, mut_flag, value_bindings) =>
+    if (List.length(value_bindings) > 0) {
+      let vbs: Typedtree.value_binding = List.hd(value_bindings);
+      Utils.lens_sig(vbs.vb_expr.exp_type);
+    } else {
+      "";
+    }
+
+  | TTopExpr(expression) =>
+    let expr_type = expression.exp_type;
+    Utils.lens_sig(expr_type);
+  | TTopException(export_flag, type_exception) => "exception"
+  | TTopExport(export_declarations) => "export"
+  };
+let get_lenses = (typed_program: Typedtree.typed_program) => {
+  List.fold_left(
+    (acc, stmt: Grain_typed__Typedtree.toplevel_stmt) => {
+      let (file, startline, startchar, sbol) =
+        Utils.get_raw_pos_info(stmt.ttop_loc.loc_start);
+
+      let signature = get_signature_from_statement(stmt);
+      let lens: Rpc.lens_t = {line: startline, signature};
+
+      [lens, ...acc];
+    },
+    [],
+    typed_program.statements,
+  );
+};
+
+let process_get_lenses = (log, id, json, compiled_code) => {
+  log("process_get_lenses requested");
+  let params = Yojson.Safe.Util.member("params", json);
+  let text_document = Yojson.Safe.Util.member("textDocument", params);
+
+  let uri =
+    Yojson.Safe.Util.member("uri", text_document)
+    |> Yojson.Safe.Util.to_string_option;
+
+  switch (uri) {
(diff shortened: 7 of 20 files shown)


marcusroberts:
This code also has the old LSP code base in place.   I've left it there as this would give us backwards compatibility.  I wonder if it's worth pushing this out in a 0.4 release, and removing the old stuff in 0.5?

compiler/grainlsp/hover.re
```diff
@@ -0,0 +1,240 @@
...
+            };
+          }
+        };
+      | _ => ("Internal error: Too many matches", None)
+      };
+    }
+
+  | TTopLet(export_flag, rec_flag, mut_flag, value_bindings) =>
+    switch (value_bindings) {
+    | [] => ("Error: empty let", None)
```
ospencer, line 100: Internal error?

compiler/grainlsp/hover.re
```diff
@@ -0,0 +1,240 @@
...
+      | _ => ("Internal error", None)
+      };
+    }
+
+  | TTopExpr(expression) =>
+    // Because of the List syntax sugar, we don't get a location
+    // But as a cover all, if we have a dummy location here,
+    // use the enclosing expression location
+
+    log("TTopExpr");
```
ospencer, line 160: Do we need this?

compiler/grainlsp/hover.re
```diff
@@ -0,0 +1,240 @@
...
+
+    | Pattern(p) => (
+        Utils.mark_down_grain(
+          Utils.lens_sig(~log, p.pat_type, ~env=compiled_code.env),
+        ),
+        Some(p.pat_loc),
+      )
+    };
+
+  | TTopException(export_flag, type_exception) => ("exception", None)
```
ospencer, line 198: Maybe report type of exception?

compiler/grainlsp/lenses.re
```diff
@@ -0,0 +1,87 @@
...
+    | _ =>
+      let decls =
+        List.fold_left(
+          (acc, decl: Typedtree.data_declaration) =>
+            (acc == "" ? "" : acc ++ ", ")
+            ++ (
+              switch (decl.data_kind) {
+              | TDataVariant(_) => "enum " ++ decl.data_name.txt
+              | TDataRecord(_) => "record " ++ decl.data_name.txt
+              | TDataAbstract => "abstract" // not sure about this one
```
ospencer, line 21: This is either a type alias or abstract type, depending on whether or not it has a type manifest

compiler/grainlsp/log.re
```diff
@@ -0,0 +1,28 @@
...
+  switch (out^) {
+  | None => ()
+  | Some(out) => close_out(out)
+  };
+  output_string(stderr, "Setting log location: " ++ location ++ "\n");
+  flush(stderr);
+  out := Some(open_out(location));
+};
+
+let spam_error = ref(false);
```
ospencer, line 16: What's this?

compiler/grainlsp/lspserver.re
```diff
@@ -0,0 +1,105 @@
...
+          is_shutting_down := true;
+        | _ => ()
+        };
+
+      | Notification("exit", _) =>
+        if (is_shutting_down^) {
+          log("Got exit! Terminating loop");
+          break := true;
+        } else {
+          log("Got exit without shutdown. Erroring out");
```
ospencer, line 66: Same questions about these logs.

compiler/grainlsp/rpc.re
```diff
@@ -0,0 +1,411 @@
...
+type completion_values = {
+  resolveProvider: bool,
+  triggerCharacters: list(string),
+};
+
+// [@deriving yojson]
+// type signature_helpers = {
+//   triggerCharacters: list(string),
+//   retriggerCharacters: list(string),
+// };
```
ospencer, line 18: What's this?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,814 @@
+open Grain;
+open Compile;
+open Grain_parsing;
+open Grain_utils;
+open Grain_typed;
+open Grain_diagnostics;
+
+let mark_down_grain = code => "```grain\n" ++ code ++ "\n```";
```
ospencer, line 8: Should probably call this `markdown_grain` 😛

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,814 @@
...
+  Format.pp_print_flush(ppf, ());
+  let sigStr = Buffer.contents(buf);
+  sigStr;
+};
+
+let rec lens_sig = (~log, ~depth=0, ~env, t: Types.type_expr) => {
+  log("lens sig at depth " ++ string_of_int(depth));
+  switch (t.desc) {
+  | TTyRecord(fields) =>
+    //  log("TTyRecord");
```
ospencer, line 154: Looks like we don't need these logs?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,814 @@
...
+        };
+
+      let (modname, typ) = parts;
+
+      if (modname == "Pervasives") {
+        typ ++ type_text;
+      } else if (typ == "Void") {
+        "Void";
+      } else {
+        "enum "
```
ospencer, line 276: Should we be using oprint or something here?

cli/bin/exec.js
```diff
@@ -96,8 +96,36 @@ function execGrainformat(
   );
 }
 
+function getGrainlsp() {
+  const grainlsp = path.join(__dirname, "grainlsp.exe");
+
+  // TODO: Maybe make an installable path & check it?
```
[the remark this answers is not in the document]
marcusroberts: Is Silo the answer?

compiler/dune-project
```diff
@@ -1,157 +1,263 @@
 (lang dune 2.3)
+
```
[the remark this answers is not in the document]
marcusroberts: My machine always formats these weirdly, but running the format command produces this output.  Not sure what I broke here :(

compiler/grainlsp/hover.re
```diff
@@ -0,0 +1,240 @@
...
+      ~uri,
+      ~line,
+      ~char,
+      ~documents,
+      ~compiled_code: Typedtree.typed_program,
+      stmt: Typedtree.toplevel_stmt,
+    ) => {
+  switch (stmt.ttop_desc) {
+  | TTopImport(import_declaration) => ("import declaration", None)
+  | TTopForeign(export_flag, value_description) => ("foreign", None)
```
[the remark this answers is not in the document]
marcusroberts: I'm not sure what that would be?

compiler/dune-project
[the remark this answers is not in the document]
marcusroberts: Fixed

compiler/grainlsp/completion.re
```diff
@@ -0,0 +1,146 @@
...
+      ~documents,
+      request,
+    ) => {
+  switch (Utils.get_text_document_uri_and_position(request)) {
+  | (Some(uri), Some(line), Some(char)) =>
+    let completable =
+      Utils.get_original_text(log, documents, uri, line, char);
+    switch (completable) {
+    | Nothing =>
+      log("Nothing to complete");
```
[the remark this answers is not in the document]
marcusroberts: Logs removed

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,814 @@
...
+                  label: ident.name,
+                  kind: 3,
+                  detail: string_of_value_description(~ident, vd),
+                  documentation: "This is some documentation",
+                };
+                Some(item);
+
+              | TSigType(ident, td, recstatus) =>
+                //log("Completing A TSigType");
+                // let string_of_type_declaration =
```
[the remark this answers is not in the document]
marcusroberts: Fixed.

compiler/grainlsp/completion.re
```diff
@@ -0,0 +1,146 @@
...
+              Env.get_all_values(compiledCode.env);
+
+            List.map(
+              ((i: Ident.t, l: Types.value_description)) => {
+                let item: Rpc.completion_item = {
+                  label: i.name,
+                  kind: get_kind(l.val_type.desc),
+                  detail:
+                    Utils.lens_sig(~log, l.val_type, ~env=compiledCode.env),
+                  documentation: "This is also some documentation",
```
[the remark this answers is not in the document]
marcusroberts: Fixed.

compiler/grainlsp/log.re
marcusroberts: Whole file deleted.  No longer needed.

compiler/grainlsp/utils.re
marcusroberts: :-D

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,814 @@
...
+    | TExpRecordGet(expr, loc, field) =>
+      if (is_point_inside_location(~line, ~char, expr.exp_loc)) {
+        lens_sig(~log, ~env=e.exp_env, expr.exp_type);
+      } else {
+        lens_sig(~log, ~env=e.exp_env, e.exp_type);
+      }
+
+    | TExpPrim1(_, exp) => lens_sig(~log, ~env=exp.exp_env, exp.exp_type)
+    | TExpPrim2(_, exp, exp2) =>
+      lens_sig(~log, ~env=exp.exp_env, exp.exp_type) // FIX me check location using get_location
```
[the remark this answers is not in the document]
marcusroberts: Yes we should!

compiler/grainlsp/hover.re
marcusroberts: Not sure how to get its type?

compiler/grainlsp/lspserver.re
marcusroberts: Killed

compiler/grainlsp/rpc.re
marcusroberts: Moved to a new PR that implements signature helpers

compiler/grainlsp/utils.re
marcusroberts: Not sure how to, need some help I think.

cli/bin/grain.js
```diff
@@ -202,13 +202,47 @@ program
   });
 
 program
-  .command("lsp <file>")
-  .description("check a grain file for LSP")
+  .command("lsp")
+  .description("start the Grain LSP server")
+  .forwardOption(
```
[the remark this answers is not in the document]
marcusroberts: @phated  how would I do that?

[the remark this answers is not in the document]
ospencer: I think it's fine to leave it for now and make an issue.

compiler/grainlsp/hover.re
```diff
@@ -0,0 +1,247 @@
...
+    | Pattern(p) => (
+        Utils.markdown_grain(
+          Utils.lens_sig(p.pat_type, ~env=compiled_code.env),
+        ),
+        Some(p.pat_loc),
+      )
+    };
+
+  | TTopException(export_flag, type_exception) =>
+    let exception_type = type_exception.ext_type; // Not sure how to get a good value here
```
ospencer, line 205: The type of all exceptions is just `Exception`, but you could show the type of this specific constructor.

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,782 @@
...
+      if (char <= raw1ce) {
+        true;
+      } else {
+        false;
+      };
+    } else if (line > raw1l && line < raw1le) {
+      true;
+    } else {
+      false;
+    };
```
ospencer, line 107: Can you clean this up to get rid of all of the `if (x) true else false`?

cli/bin/grain.js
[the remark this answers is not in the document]
marcusroberts: Made issue #1147

compiler/grainlsp/grainlsp.re
```diff
@@ -0,0 +1,48 @@
+//let _ = Lspserver.run();
```
jozanza, line 1: Should we remove this?

compiler/grainlsp/grainlsp.re
```diff
@@ -0,0 +1,48 @@
...
+
+  `Ok();
+};
+
+let lsp = opts =>
+  try(lsp_wrapper()) {
+  | exn =>
+    Format.eprintf("@[%s@]@.", Printexc.to_string(exn));
+    exit(2);
+  };
```
jozanza, line 17: Is it possible to `try(Lspserver.run()) { ... }` or do we need the `lsp_wrapper` to catch the exn here or something?

compiler/grainlsp/rpc.re
```diff
@@ -0,0 +1,424 @@
...
+    | Some(id) => Message(id, action, json)
+    };
+  } else {
+    failwith("Invalid header");
+  };
+};
+
+let send = (output, content) => {
+  let length = String.length(content);
+  let sep = Sys.os_type == "Unix" ? "\r\n\r\n" : "\n\n";
```
phated, line 231: See comment about os_type above

compiler/grainlsp/rpc.re
```diff
@@ -0,0 +1,424 @@
...
+  flush(output);
+};
+
+let send_null_message = (output, id) => {
+  let res =
+    `Assoc([
+      ("jsonrpc", `String("2.0")),
+      ("id", `Int(id)),
+      ("result", `Null),
+    ]);
```
phated, line 244: Why not define a record that derives yojson?

compiler/grainlsp/rpc.re
```diff
@@ -0,0 +1,424 @@
...
+
+  let strJson = Yojson.Safe.pretty_to_string(res);
```
phated, line 246: Do we ever want `pretty` functions?

compiler/grainlsp/rpc.re
```diff
@@ -0,0 +1,424 @@
...
+  send(output, strJson);
+};
+
+let send_capabilities = (output, id: int) => {
+  let completion_vals: completion_values = {
+    resolveProvider: true,
+    triggerCharacters: ["."],
```
phated, line 253: Missed these on the definitions, but no camelcase please. I believe you can do yojson mappings to convert between camelcase and snakecase

compiler/grainlsp/rpc.re
```diff
@@ -0,0 +1,424 @@
...
+  };
+
+  let res = capabilities_response_to_yojson(response);
+  let strJson = Yojson.Safe.to_string(res);
+
+  send(output, strJson);
+};
+
+let send_lenses = (~output, ~id: int, lenses: list(lens_t)) => {
+  let convertedLenses =
```
phated, line 289: camelcase

compiler/grainlsp/rpc.re
```diff
@@ -0,0 +1,424 @@
...
+  let res = hover_response_to_yojson(response);
+  let strJson = Yojson.Safe.pretty_to_string(res);
+  send(output, strJson);
+};
+
+let send_diagnostics =
+    (
+      ~output,
+      ~uri,
+      ~warnings: option(list(lsp_warning)),
```
phated, line 330: Why are these an option of a list? Wouldn't an empty list be sufficient?

compiler/grainlsp/rpc.re
```diff
@@ -0,0 +1,424 @@
...
+      error: option(lsp_error),
```
phated, line 331: It's weird for error to be the only positional argument here, maybe warnings and error should both be? Unsure

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+open Grain_diagnostics;
+
+//  CompletionItemKind
+let item_kind_completion_text = 1;
+let item_kind_completion_function = 3;
+let item_kind_completion_constructor = 4;
+let item_kind_completion_variable = 6;
+let item_kind_completion_struct = 22;
+
+let markdown_grain = code => "```grain\n" ++ code ++ "\n```";
```
phated, line 15: Don't I have this as a utility in `Grain_utils.Markdown` already?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+open Grain_utils;
+open Grain_typed;
+open Grain_diagnostics;
+
+//  CompletionItemKind
+let item_kind_completion_text = 1;
+let item_kind_completion_function = 3;
+let item_kind_completion_constructor = 4;
+let item_kind_completion_variable = 6;
+let item_kind_completion_struct = 22;
```
phated, line 13: Are these related to the LSP spec?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+  | Expression(Typedtree.expression)
+  | Pattern(Typedtree.pattern)
+  | NotInRange
+  | Error(string);
+
+type source_text =
+  | Nothing
+  | Lident(string);
+
+let rec print_path = (ident: Path.t) => {
```
phated, line 27: Does this not exist in oprint already?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+  switch (ident) {
+  | PIdent(name) => name.name
+  | PExternal(externalIdent, second, _) =>
+    print_path(externalIdent) ++ "." ++ second
+  };
+};
+
+let comma_join = (~sep, ~print_fn, vals) => {
```
phated, line 35: This is basically String.concat?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+  if (line == raw1l) {
+    if (char >= raw1c) {
+      if (line == raw1le) {
+        char <= raw1ce;
+      } else {
+        true;
+      };
+    } else {
+      false;
+    };
```
phated, line 89: lol, I'll trust you

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+};
+
+let print_sig = (t: Types.type_expr) => {
+  let buf = Buffer.create(64);
+  let ppf = Format.formatter_of_buffer(buf);
+  Printtyp.type_expr(ppf, t);
+  Format.pp_print_flush(ppf, ());
+  let sigStr = Buffer.contents(buf);
+  sigStr;
+};
```
phated, line 140: Is this duplicated out of graindoc? We should make it shared somehow

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+
+let string_of_type_expr = out_type => {
```
phated, line 142: Is this duplicated out of graindoc? We should make it shared somehow

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+  let printer = (ppf, out_type) => {
+    Oprint.out_type^(ppf, out_type);
+  };
+  let to_string = out_type => {
+    Format.asprintf("%a", printer, out_type);
+  };
+  Option.map(to_string, out_type);
+};
+
+let simple_sig = (t: Types.type_expr) => {
```
phated, line 152: Is this duplicated out of graindoc? We should make it shared somehow

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+  let tree = Printtyp.tree_of_typexp(true, t);
+  switch (string_of_type_expr(Some(tree))) {
+  | None => ""
+  | Some(tstr) => tstr
+  };
+};
+
+let rec lens_sig = (~depth=0, ~env, t: Types.type_expr) => {
```
phated, line 160: I'm confused by this function in general. I thought oprint did most of it?

compiler/grainlsp/utils.re
phated: Jinx

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+  | TPatTuple(args) =>
+    let pats =
+      List.filter(
+        (p: Typedtree.pattern) =>
+          is_point_inside_location(~line, ~char, p.pat_loc),
+        args,
+      );
+    switch (pats) {
+    | [p] => Pattern(p)
+    | _ => Error("")
```
phated, line 288: Why is this an error with no message?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+    List.filter(
+      (e: Typedtree.expression) =>
+        is_point_inside_location(~line, ~char, e.exp_loc),
+      expressions,
+    );
+
+  switch (exps) {
+  | [] => default
+  | [expr] => get_node_from_expression(~line, ~char, expr)
+  | _ => Error("Multiple matches, should not happen")
```
phated, line 308: Probably figure out a better way to write "should not happen"

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+  };
+}
+
+and get_node_from_expression = (~line, ~char, expr: Typedtree.expression) => {
+  let node =
+    switch (expr.exp_desc) {
+    | TExpLet(rec_flag, mut_flag, vbs) =>
+      switch (vbs) {
+      | [] => Error("")
```
phated, line 317: Why error without a message?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+          ~line,
+          ~char,
+          ~default=Expression(expr),
+          expressions,
+        )
+      }
+    | TExpApp(func, expressions) =>
+      if (is_point_inside_location(~line, ~char, func.exp_loc)) {
+        Expression(func);
+      } else {
```
phated, line 358: Use `when`?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+          let vb = List.hd(vbs);
+          let expr = vb.vb_expr;
+          Expression(expr);
+        | [hd] => hd
+        | _ => Error("Too many let matches")
+        };
+      }
+    | TExpBlock(expressions) =>
+      switch (expressions) {
+      | [] => Expression(expr)
```
phated, line 346: Separate the pattern matches instead of nesting

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+      }
+    | TExpLambda([{mb_pat: pattern, mb_body: body}], _) =>
+      let node = get_node_from_pattern(~line, ~char, pattern);
+      switch (node) {
+      | Error(_) => get_node_from_expression(~line, ~char, body)
+      | _ => node
+      };
+    | TExpIf(cond, trueexp, falseexp) =>
+      let condNode = get_node_from_expression(~line, ~char, cond);
+      let trueMatch =
```
phated, line 374: No camelcase please

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+    | TExpContinue => Expression(expr)
+    | TExpBreak => Expression(expr)
+    | TExpNull => Expression(expr)
+    | TExpIdent(_) => Expression(expr)
+    | TExpConstant(const) =>
+      if (is_point_inside_location(~line, ~char, expr.exp_loc)) {
+        Expression(expr);
+      } else {
+        NotInRange;
+      }
```
phated, line 408: Use `when`?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+      find_location_in_expressions(
+        ~line,
+        ~char,
+        ~default=Expression(expr),
+        [e1, e2],
+      )
+    | TExpMatch(cond, branches, _) =>
+      if (is_point_inside_location(~line, ~char, cond.exp_loc)) {
+        Expression(cond);
+      } else {
```
phated, line 441: Use `when`?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+        [e1, e2],
+      )
+    | TExpWhile(cond, block) =>
+      if (is_point_inside_location(~line, ~char, cond.exp_loc)) {
+        get_node_from_expression(~line, ~char, cond);
+      } else if (is_point_inside_location(~line, ~char, block.exp_loc)) {
+        get_node_from_expression(~line, ~char, block);
+      } else {
+        Expression(expr);
+      }
```
phated, line 509: Use `when`?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+
+    | TExpFor(e1, e2, e3, block) =>
+      if (is_point_inside_location(~line, ~char, block.exp_loc)) {
+        get_node_from_expression(~line, ~char, block);
+      } else {
```
phated, line 514: Use `when`?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+  | _ => node
+  };
+};
+
+let get_node_from_statement =
+    (~line, ~char, stmt: Grain_typed__Typedtree.toplevel_stmt) => {
+  switch (stmt.ttop_desc) {
+  | TTopImport(import_declaration) => Error("import declaration")
+  | TTopForeign(export_flag, value_description) => Error("foreign")
+  | TTopData(data_declarations) => Error("data declaration")
```
phated, line 557: Better error messages?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+  | TTopLet(export_flag, rec_flag, mut_flag, value_bindings) =>
+    switch (value_bindings) {
+    | [] => Error("")
```
phated, line 560: Why empty error message?

phated, line 560: Use `when` or separate pattern matches?

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+        | '0' .. '9'
+        | '.'
+        | '_' => loop(i - 1)
+        | _ =>
+          i == offset - 1
+            ? Nothing : Lident(String.sub(text, i + 1, offset - (i + 1)))
+        }
+      );
+  };
+  loop(offset - 1);
```
phated, line 612: This function hurts my head

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,816 @@
...
+
+let get_module_exports = (mod_ident, compiled_code: Typedtree.typed_program) => {
+  switch (Env.find_module(mod_ident, None, compiled_code.env)) {
+  | lookup =>
+    switch (lookup.md_filepath) {
+    | None => []
+    | Some(p) =>
+      let mtype: Grain_typed.Types.module_type = lookup.md_type;
+      switch (mtype) {
+      | TModSignature(sigs) =>
```
phated, line 635: Avoid this nested pattern match?

compiler/dune-project
```diff
@@ -64,6 +64,15 @@
...
     (grain (>= 0))))
 
+(package
+  (name grainlsp)
+  (synopsis "Grain LSP")
+  (depends
+    (reason (>= 3.6.0))
+    (dune-build-info (>= 2.0))
+    (cmdliner (>= 1.0.2))
+    (grain (>= 0))))
```
phated, line 74: Missing a bunch of the deps that you included, basically anything in libraries and ppx's need to be here

cli/bin/grain.js
```diff
@@ -201,14 +200,50 @@ program
...
+  .forwardOption("--no-gc", "turn off reference counting garbage collection")
+  .forwardOption(
+    "--no-bulk-memory",
+    "polyfill WebAssembly bulk memory instructions"
+  )
+  .forwardOption(
+    "--no-pervasives",
+    "don't automatically import the Grain Pervasives module"
+  )
+  .forwardOption("--strict-sequence", "enable strict sequencing")
```
phated, line 241: You don't have your debug flag listed here.

phated (changes requested):
The big issue here is that you used `Unix` APIs that js_of_ocaml doesn't have. The ones you tried to use would be very hard to map to nodejs APIs. Essentially, any time you reach for Unix or Sys, you need to dig into js_of_ocaml behavior (and probably add some API, like I did for our pkg usage).

I also commented a bunch of style complaints. In general, I think you should work on flattening your nested pattern matching by using `when` or more complex patterns like in other parts of the compiler. Using `when` feels a lot better than nested if/else blocks.

compiler/grainlsp/completion.re
```diff
@@ -0,0 +1,136 @@
+open Grain;
+open Compile;
+open Grain_parsing;
+open Grain_utils;
+open Grain_typed;
+
+// maps Grain types to LSP CompletionItemKind
+let rec get_kind = (desc: Types.type_desc) =>
+  switch (desc) {
+  | TTyVar(_) => Utils.item_kind_completion_variable
```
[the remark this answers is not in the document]
marcusroberts: Good catch, and all the completion related stuff has been moved out of Utils and into completion

compiler/grainlsp/completion.re
```diff
@@ -0,0 +1,136 @@
...
+  | TTyConstr(_) => Utils.item_kind_completion_constructor
+  | TTySubst(s) => get_kind(s.desc)
+  | TTyLink(t) => get_kind(t.desc)
+  | _ => Utils.item_kind_completion_text
+  };
+
+let process_resolution =
+    (
+      ~id,
+      ~compiled_code: Stdlib__hashtbl.t(string, Typedtree.typed_program),
```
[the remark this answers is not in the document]
marcusroberts: You are correct.  No idea why I did that!

compiler/grainlsp/completion.re
```diff
@@ -0,0 +1,136 @@
...
+
+let process_completion =
+    (
+      ~id,
+      ~compiled_code: Stdlib__hashtbl.t(string, Typedtree.typed_program),
+      ~cached_code: Stdlib__hashtbl.t(string, Typedtree.typed_program),
+      ~documents,
+      request,
+    ) => {
+  switch (Utils.get_text_document_uri_and_position(request)) {
```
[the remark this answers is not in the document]
marcusroberts: This is a generic function I reuse twice currently (and will reuse a lot more as I add in the source code for other LSP features).    Is it the name Utils that you don't like?

compiler/grainlsp/completion.re
```diff
@@ -0,0 +1,136 @@
...
+  | (Some(uri), Some(line), Some(char)) =>
```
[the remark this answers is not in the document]
marcusroberts: Switched to a record.

compiler/grainlsp/completion.re
```diff
@@ -0,0 +1,136 @@
...
+    switch (completable) {
+    | Nothing => Rpc.send_completion(stdout, id, [])
+    | Lident(text) =>
+      if (!Hashtbl.mem(cached_code, uri)) {
+        // no compiled code available
+        Rpc.send_completion(stdout, id, []);
+      } else {
+        let compiledCode = Hashtbl.find(cached_code, uri);
+        let modules = Env.get_all_modules(compiledCode.env);
+        let firstChar = text.[0];
```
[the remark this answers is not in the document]
marcusroberts: It won't throw as it's wrapped in a test for membership above as I really hate exceptions.
But as this could lead to problems down the line if someone changed something, I'm changing it to switch on the exception instead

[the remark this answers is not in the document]
marcusroberts: But I love camelcase :-D. Fixed.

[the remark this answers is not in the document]
marcusroberts: Or even just use the safe call that returns optional, even better.

compiler/grainlsp/completion.re
```diff
@@ -0,0 +1,136 @@
...
+
+        let completions =
+          switch (firstChar) {
+          | 'A' .. 'Z' =>
+            // autocomplete modules
+            if (String.contains(text, '.')) {
+              // find module name
+              let pos = String.rindex(text, '.');
```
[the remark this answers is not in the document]
marcusroberts: Correct, but I've wrapped it in a switch on the exception instead just in case someone changes that in the future

compiler/grainlsp/config/flags.re
```diff
@@ -0,0 +1,20 @@
...
+      switch (C.ocaml_config_var(c, "system")) {
+      | Some("mingw64") =>
+        // MinGW needs the -static flag passed directly to the linker,
+        // to avoid needing MinGW locations in the path
+        // Ref https://github.com/grain-lang/binaryen.ml#static-linking
+        ["-ccopt", "--", "-ccopt", "-static"]
+      | Some(_) => default
+      | None => default
+      };
+
```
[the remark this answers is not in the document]
marcusroberts: Fixed.

compiler/grainlsp/grainlsp.re
```diff
@@ -0,0 +1,48 @@
...
+  let _ = Lspserver.run();
+
+  `Ok();
+};
+
+let lsp = opts =>
+  try(lsp_wrapper()) {
+  | exn =>
+    Format.eprintf("@[%s@]@.", Printexc.to_string(exn));
+    exit(2);
```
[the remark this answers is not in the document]
marcusroberts: Error 2 means serious trouble

compiler/grainlsp/hover.re
```diff
@@ -0,0 +1,248 @@
...
+      Utils.print_path(import_declaration.timp_path),
+      Some(stmt.ttop_loc),
+    )
+  | TTopForeign(export_flag, value_description) =>
+    let tvd_desc = value_description.tvd_desc;
+    let type_sig = Utils.lens_sig(~env=stmt.ttop_env, tvd_desc.ctyp_type);
+    (type_sig, Some(stmt.ttop_loc));
+  | TTopData(data_declarations) =>
+    switch (data_declarations) {
+    | [] => ("empty data", None)
```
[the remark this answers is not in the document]
marcusroberts: There's not a test for empty Lists, and it seemed better to switch that do a list.length and look for 0?

compiler/grainlsp/hover.re
```diff
@@ -0,0 +1,248 @@
...
+          | _ => (
+              Utils.markdown_grain("enum " ++ name.txt),
+              Some(node.data_loc),
+            )
+          };
+
+        | TDataRecord(_) =>
+          switch (node.data_params) {
+          | [] => (
+              Utils.markdown_grain(node.data_name.txt),
```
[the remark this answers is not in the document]
marcusroberts: This is all it does

`let markdown_grain = code => "```grain\n" ++ code ++ "\n```";`

compiler/grainlsp/hover.re
```diff
@@ -0,0 +1,248 @@
...
+          }
+        | TDataVariant(constrs) =>
+          let matches =
+            List.filter(
+              (cd: Typedtree.constructor_declaration) =>
+                if (Utils.is_point_inside_location(~line, ~char, cd.cd_loc)) {
+                  true;
+                } else {
+                  false;
+                },
```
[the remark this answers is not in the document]
marcusroberts: A Marcus special :-D. These are always left over after I've optimised out debugging.

compiler/grainlsp/grainlsp.re
```diff
@@ -0,0 +1,48 @@
...
+  };
+
+[@deriving cmdliner]
+type params = {
+  [@name "debuglsp"] [@docv "DEBUGLOG"]
+  debug_log: option(string),
```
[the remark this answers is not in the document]
marcusroberts: I don't want this flag at all, but Cmdliner has confounded me again.  Can you show me how to get rid of it?

compiler/grainlsp/hover.re
```diff
@@ -0,0 +1,248 @@
...
+            };
+          }
+        };
+      | _ => ("Internal error: Too many matches", None)
+      };
+    }
+
+  | TTopLet(export_flag, rec_flag, mut_flag, value_bindings) =>
+    switch (value_bindings) {
+    | [] => ("Internal error: empty let", None)
```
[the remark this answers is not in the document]
marcusroberts: I actually like that second approach but I've gone for when

compiler/grainlsp/hover.re
```diff
@@ -0,0 +1,248 @@
...
+  };
+};
+
+let get_hover = (~id, ~compiled_code, ~documents, request) => {
+  switch (Utils.get_text_document_uri_and_position(request)) {
+  | (Some(uri), Some(line), Some(char)) =>
+    if (Hashtbl.mem(compiled_code, uri)) {
+      let ln = line + 1;
+
+      let compiled_code_opt = Hashtbl.find_opt(compiled_code, uri);
```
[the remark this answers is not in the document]
marcusroberts: Simplified to remove the check

compiler/grainlsp/lenses.re
```diff
@@ -0,0 +1,90 @@
...
+        );
+      Some(vbses);
+    }
+  | TTopExpr(expression) => Some(Utils.simple_sig(expression.exp_type))
+  | TTopException(export_flag, type_exception) => None
+  | TTopExport(export_declarations) => None
+  };
+
+let get_lenses = (typed_program: Typedtree.typed_program) => {
+  List.fold_left(
```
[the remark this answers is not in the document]
marcusroberts: Good spot!

compiler/grainlsp/lenses.re
```diff
@@ -0,0 +1,90 @@
...
+        let lens: Rpc.lens_t = {line: startline, signature: sigval};
+        [lens, ...acc];
+      };
+    },
+    [],
+    typed_program.statements,
+  );
+};
+
+let process_get_lenses = (~id, ~compiled_code, request) => {
```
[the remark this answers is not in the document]
marcusroberts: Renamed better.

compiler/grainlsp/processcode.re
```diff
@@ -0,0 +1,124 @@
+open Grain;
+open Compile;
+open Grain_parsing;
+open Grain_utils;
+open Grain_typed;
+
+let compile_source = (uri, source) => {
+  let filename = Filename.basename(uri);
```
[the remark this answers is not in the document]
marcusroberts: I was taking a shortcut to convert it from a URI, fixed now.

compiler/grainlsp/processcode.re
```diff
@@ -0,0 +1,124 @@
...
+
+let textDocument_didOpenOrChange =
+    (~documents, ~compiled_code, ~cached_code, request) => {
+  switch (get_text_document_from_params(request)) {
+  | Some((uri, text)) =>
+    if (!Hashtbl.mem(documents, uri)) {
+      Hashtbl.add(documents, uri, text);
+    } else {
+      Hashtbl.replace(documents, uri, text);
+    };
```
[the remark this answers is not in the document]
marcusroberts: Too used to C# crashing if you add an existing member.  fixed as per suggestion

compiler/grainlsp/rpc.re
marcusroberts: I'm going to push back on this :). These are only used within this file and are the LSP protocol fields.  To me it makes it clearer to future developers where this stuff is set and how it maps, rather than to a yojson mapping for the sake of it.   
If this record was accessed outside this file I'd map them.  It's almost like a lowlevel implementation that should have @unsafecamelcase tagged on it ;-)

compiler/grainlsp/utils.re
marcusroberts: Yes you do, all references switched to that

compiler/grainlsp/utils.re
marcusroberts: Yes they are

compiler/grainlsp/utils.re
marcusroberts: With a map of the args, so simplified to use that as the implementation

compiler/grainlsp/utils.re
marcusroberts: I know it's ugly, but it's more explicit than trying to use and and or operators

compiler/grainlsp/rpc.re
marcusroberts: Switching to a list, which also makes the next point clearer

compiler/grainlsp/hover.re
```diff
@@ -0,0 +1,248 @@
...
+                Some(decl.ctyp_loc),
+              )
+            | _ => (
+                Utils.markdown_grain(node.data_name.txt),
+                Some(node.data_loc),
+              )
+            };
+          }
+        };
+      | _ => ("Internal error: Too many matches", None)
```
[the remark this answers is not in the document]
marcusroberts: It's one of those "should never happen" cases where a location if found inside two top level statements. I've just gone for the first should that happen now

compiler/grainlsp/hover.re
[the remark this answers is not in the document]
marcusroberts: And yes your one done the same, so removed.

compiler/grainlsp/hover.re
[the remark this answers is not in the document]
marcusroberts: Converted it to a when

compiler/grainlsp/hover.re
```diff
@@ -0,0 +1,248 @@
...
+    ) => {
+  switch (stmt.ttop_desc) {
+  | TTopImport(import_declaration) => (
+      Utils.print_path(import_declaration.timp_path),
+      Some(stmt.ttop_loc),
+    )
+  | TTopForeign(export_flag, value_description) =>
+    let tvd_desc = value_description.tvd_desc;
+    let type_sig = Utils.lens_sig(~env=stmt.ttop_env, tvd_desc.ctyp_type);
+    (type_sig, Some(stmt.ttop_loc));
```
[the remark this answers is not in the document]
marcusroberts: Fixed to return a variant

compiler/grainlsp/completion.re
```diff
@@ -0,0 +1,136 @@
...
+    ) => {
+  switch (Utils.get_text_document_uri_and_position(request)) {
+  | (Some(uri), Some(line), Some(char)) =>
+    let completable = Utils.get_original_text(documents, uri, line, char);
+    switch (completable) {
+    | Nothing => Rpc.send_completion(stdout, id, [])
+    | Lident(text) =>
+      if (!Hashtbl.mem(cached_code, uri)) {
+        // no compiled code available
+        Rpc.send_completion(stdout, id, []);
```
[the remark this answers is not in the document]
marcusroberts: Switch away from the if to using the find_opt which is an Option which I think is better handled with a switch.

compiler/grainlsp/hover.re
```diff
@@ -0,0 +1,248 @@
...
+    | _ =>
+      let matches =
+        List.filter(
+          (dd: Typedtree.data_declaration) =>
+            Utils.is_point_inside_location(~line, ~char, dd.data_loc),
+          data_declarations,
+        );
+      switch (matches) {
+      | [] => ("Internal error parsing", Some(stmt.ttop_loc))
+      | [node] =>
```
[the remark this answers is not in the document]
marcusroberts: Sweet!  The code looks much better with that

compiler/grainlsp/utils.re
marcusroberts: Code was dead, so removed.

compiler/grainlsp/grainlsp.re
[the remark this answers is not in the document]
marcusroberts: Changed to be either on or off and actually make use of the flag

compiler/grainlsp/grainlsp.re
marcusroberts: Cleaned up to pass proper values back instead of throwing the exception

compiler/grainlsp/log.re
```diff
@@ -0,0 +1,22 @@
+let out = ref(None);
+
+let initial_dest = Filename.concat(Filename.get_temp_dir_name(), "lsp.log");
+out := Some(open_out(initial_dest));
```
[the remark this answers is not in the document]
marcusroberts: Fixed.

compiler/grainlsp/lspserver.re
```diff
@@ -0,0 +1,85 @@
...
+          )
+        | _ => ()
+        }
+      | Error(_) => is_shutting_down := true
+      };
+    };
+  };
+
+let run = () => {
+  // Will enable this again if we want LSP logging
```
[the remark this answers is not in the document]
marcusroberts: Cleaned up and use the debug flag

compiler/grainlsp/lspserver.re
```diff
@@ -0,0 +1,85 @@
...
+  // Log.set_location("lsp.log");
+
+  let initialize = () =>
+    switch (Rpc.read_message(stdin)) {
+    | Message(id, "initialize", _) =>
+      Rpc.send_capabilities(stdout, id);
+      loop();
+    | _ => failwith("Client must send 'initialize' as first event")
```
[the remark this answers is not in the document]
marcusroberts: Nope, fixed to return a proper result instead.

compiler/grainlsp/utils.re
marcusroberts: Mine too, it's taken from the Reason LSP

compiler/grainlsp/utils.re
marcusroberts: Getting the hang of this now, updated and looks much nicer

compiler/grainlsp/rpc.re
```diff
@@ -0,0 +1,424 @@
...
+
+    let idOpt =
+      Yojson.Safe.Util.member("id", json) |> Yojson.Safe.Util.to_int_option;
+
+    switch (idOpt) {
+    | None => Notification(action, json)
+    | Some(id) => Message(id, action, json)
+    };
+  } else {
+    failwith("Invalid header");
```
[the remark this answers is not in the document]
marcusroberts: Switched to an error

compiler/grainlsp/utils.re
marcusroberts: Fixed by a code fix.

compiler/grainlsp/utils.re
marcusroberts: Dead function, deleted.

compiler/grainlsp/rpc.re
```diff
@@ -0,0 +1,424 @@
...
+    /* if on windows, dont need the extra -1 */
+    let offset = Sys.os_type == "Win32" ? 0 : (-1); /* -1 for trailing \r */
+
+    let num =
+      String.sub(clength, cll, String.length(clength) - cll + offset);
+    let num = (num |> int_of_string) + (Sys.os_type == "Win32" ? 1 : 2);
+    let buffer = Buffer.create(num);
+    Buffer.add_channel(buffer, input, num);
+    let raw = Buffer.contents(buffer);
+    let json = Yojson.Safe.from_string(raw);
```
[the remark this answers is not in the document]
marcusroberts: There is, but because the protocol injects Content-Length headers into the stream, we have to read from the channel the number of characters specified.

compiler/grainlsp/lspserver.re
```diff
@@ -0,0 +1,85 @@
+let rootPath = ".";
+
+let documents = Hashtbl.create(128);
+let compiled_code = Hashtbl.create(128);
+let cached_code = Hashtbl.create(128); // we keep the last successful compile to help with completion and definitions
+
+/* Will wait up to 100ms */
+let can_read = desc => {
+  let (r, _, _) = Unix.select([desc], [], [], 0.1);
```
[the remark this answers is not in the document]
marcusroberts: We don't need it. I don't really understand why the code I borrowed this from did it, as it just busy looped rather than waiting for IO.

compiler/grainlsp/lspserver.re
```diff
@@ -0,0 +1,85 @@
...
+  r != [];
+};
+
+let break = ref(false);
+let is_shutting_down = ref(false);
+let stdin_descr = Unix.descr_of_in_channel(stdin);
```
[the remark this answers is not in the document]
marcusroberts: Don't need it so removed it.

marcusroberts:
Nice work @phated .

Not quite working on JSOO yet as it's not formatting the returned messages properly (I believe it's missing some \r) so the input problem is fixed.  I can fix the output very quickly.

Works on Windows too

compiler/grainlsp/hacks.js
```diff
@@ -0,0 +1,27 @@
...
+    var data = caml_global_data.fds[fd];
+    if(data.flags.wronly) caml_raise_sys_error("fd "+ fd + " is writeonly");
+    var refill = null;
+    if(fd == 0 && fs_node_supported()){
+      var fs = require('fs');
+      var Buffer = require('buffer').Buffer;
+      refill = function () {
+        var buf = Buffer.alloc(1024);
+        var bytesRead = fs.readSync(0, buf);
+        return caml_string_of_jsstring(buf.slice(0, bytesRead).toString('utf8'));
```
phated, line 14: I need to upstream this into JSOO

compiler/grainlsp/hacks.js
```diff
@@ -0,0 +1,57 @@
...
+      var fs = require('fs');
+      var Buffer = require('buffer').Buffer;
+      refill = function () {
+        var buf = Buffer.alloc(256);
+        var stdinFd = fs.openSync('/dev/stdin', 'rs');
+        var bytesRead = fs.readSync(fd, buf);
+        fs.closeSync(stdinFd);
+        // Ensures we don't have any trailing, empty data
+        var str = buf.slice(0, bytesRead).toString('utf8');
+        return caml_string_of_jsstring(str);
```
phated, line 18: cc @hhugo - would you be open to the changes in this file once I have everything working on Windows?

compiler/grainlsp/hacks.js
```diff
@@ -0,0 +1,57 @@
...
+function caml_ml_open_descriptor_in (fd)  {
+    var data = caml_global_data.fds[fd];
+    if(data.flags.wronly) caml_raise_sys_error("fd "+ fd + " is writeonly");
+    var refill = null;
+    if(fd == 0 && fs_node_supported()){
+      var fs = require('fs');
+      var Buffer = require('buffer').Buffer;
+      refill = function () {
+        var buf = Buffer.alloc(256);
+        var stdinFd = fs.openSync('/dev/stdin', 'rs');
```
phated, line 13: This won't work on Windows curently.

compiler/grainlsp/utils.re
marcusroberts: Yes, no longer needed.

compiler/grainlsp/utils.re
```diff
@@ -0,0 +1,814 @@
...
+    | TDataOpen =>
+      //    log("TDataOpen");
+      print_path(path) ++ type_text
+    };
+
+  | TTyArrow(args, rettype, _) =>
+    //   log("TTyArrow");
+    let arg_text = comma_join(~sep=", ", ~print_fn=print_sig, args);
+
+    "(" ++ arg_text ++ ") -> " ++ print_sig(rettype);
```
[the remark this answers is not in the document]
marcusroberts: Yes, no longer doing this.

compiler/grainlsp/log.re
```diff
@@ -0,0 +1,41 @@
+let out = ref(None);
+
+type log_level =
+  | NoLog
+  | DebugLog;
+
+//let log_dest = Filename.concat(Filename.get_temp_dir_name(), "lsp.log");
```
ospencer, line 7: comment

compiler/src/language_server/driver.re
```diff
@@ -0,0 +1,70 @@
...
+      ~cached_code,
+      ~documents,
+      params,
+    );
+    Reading;
+  | SetTrace(trace_value) =>
+    Trace.set_level(trace_value);
+    Reading;
+  | Error(msg) =>
+    prerr_endline(msg);
```
phated, line 58: What do we want to do here?

compiler/src/language_server/protocol.re
```diff
@@ -0,0 +1,203 @@
...
+      String_utils.slice(
+        ~first=String.length(header_prefix),
+        ~last=-1 /* newline? */,
+        header,
+      );
+
+    switch (int_of_string_opt(len)) {
+    | None => Error("Invalid Content-Length in header")
+    | Some(len) =>
+      // TODO: Comment about this +2
```
phated, line 155: @marcusroberts Can you add this comment?

compiler/src/language_server/protocol.re
```diff
@@ -0,0 +1,203 @@
...
+let sep = "\r\n\r\n";
+
+let request = (): result(request_message, string) => {
+  switch (input_line(stdin)) {
+  | exception exn => Error("Failed to read input")
+  | header when String_utils.starts_with(header, header_prefix) =>
+    let len =
+      String_utils.slice(
+        ~first=String.length(header_prefix),
+        ~last=-1 /* newline? */,
```
phated, line 148: @marcusroberts is this correct? We are skipping the newline? How does that work on Windows now that we are in binary mode?

compiler/src/language_server/driver.re
```diff
@@ -0,0 +1,70 @@
...
+    is_shutting_down := true;
+    Break;
+  | _ =>
+    /* TODO: What should happen here? */
```
phated, line 62: Ping @marcusroberts

compiler/src/language_server/driver.re
marcusroberts: Updated the comment,  just stopping is fine.  It signals to the client to restart, which it will try to do 5 times.

compiler/src/language_server/protocol.re
marcusroberts: Comment added with link to spec

compiler/src/language_server/protocol.re
marcusroberts: Windows in text mode used to be annoying because it stripped out the CR of the CRLF whereas Unix/Linux preserved it which is almost ironic...   Now Windows is in binary mode, we ignore the last CR (as input_line strips the LF only) on all platforms.  Comment updated.

compiler/src/language_server/driver.re
marcusroberts: Commented, this is good the LSP should quit to signal to the client to restart.

compiler/grainlsp/rpc.re
marcusroberts: All fixed by @phated !

compiler/src/language_server/driver.re
phated:
```suggestion
```
Sounds good. Let's remove this debugging print then

compiler/src/language_server/driver.re
```diff
@@ -0,0 +1,68 @@
...
+    Formatting.process(~id, ~compiled_code, ~documents, params);
+    Reading;
+  | SetTrace(trace_value) =>
+    Trace.set_level(trace_value);
+    Reading;
+  | Error(msg) =>
+    prerr_endline(msg);
+    Trace.log(msg);
+    let error: Protocol.response_error = {
+      code: Protocol.InvalidRequest,
```
phated, line 53: Not sure if this the correct enum member

compiler/src/language_server/driver.re
```diff
@@ -0,0 +1,68 @@
...
+    Reading;
+  | Formatting(id, params) when is_initialized^ =>
+    Formatting.process(~id, ~compiled_code, ~documents, params);
+    Reading;
+  | SetTrace(trace_value) =>
+    Trace.set_level(trace_value);
+    Reading;
+  | Error(msg) =>
+    prerr_endline(msg);
+    Trace.log(msg);
```
phated, line 51:
```suggestion
```
We're already sending an error response, so we don't need to trace this.

compiler/src/language_server/driver.re
phated: I think the correct error code is for `InvalidParams` because we have the message and return `Error` when the params can't parse.

phated: Also @marcusroberts - if you just inline the record into the parameter, you don't need to annotate so much. Embrace HM 😄

compiler/grainlsp/completion.re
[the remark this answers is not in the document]
marcusroberts: Moved out into a separate PR for completion now anyway

ospencer (approved):
Brilliant work! This is absolutely huge!

PR #1134, opened 2022-02-07
feat(compiler): Don't close over global values

ospencer:
This PR prevents global values from appearing in closures (in memory).

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -683,8 +683,7 @@ let appropriate_incref = (wasm_mod, env, arg, b) =>
   | MLocalBind(_, Types.HeapAllocated)
   | MSwapBind(_, Types.HeapAllocated)
   | MClosureBind(_)
-  | MGlobalBind(_, Types.HeapAllocated, true) =>
-    call_incref(wasm_mod, env, arg)
+  | MGlobalBind(_, Types.HeapAllocated) => call_incref(wasm_mod, env, arg)
   | MArgBind(_)
   | MLocalBind(_)
   | MSwapBind(_)
@@ -698,8 +697,7 @@ let appropriate_decref = (wasm_mod, env, arg, b) =>
   | MLocalBind(_, Types.HeapAllocated)
   | MSwapBind(_, Types.HeapAllocated)
   | MClosureBind(_)
-  | MGlobalBind(_, Types.HeapAllocated, true) =>
-    call_decref(wasm_mod, env, arg)
+  | MGlobalBind(_, Types.HeapAllocated) => call_decref(wasm_mod, env, arg)
   | MArgBind(_)
   | MLocalBind(_)
   | MSwapBind(_)
@@ -852,16 +850,14 @@ let compile_bind =
     | BindSet(arg) => set_slot(slot, typ, arg)
     | BindTee(arg) => tee_slot(slot, typ, arg)
     };
-  | MGlobalBind(slot, wasm_ty, gc) =>
+  | MGlobalBind(slot, wasm_ty) =>
     let typ = wasm_type(wasm_ty);
     switch (action) {
-    | BindGet when !gc => Expression.Global_get.make(wasm_mod, slot, typ)
     | BindGet =>
       appropriate_incref(
         wasm_mod,
         Expression.Global_get.make(wasm_mod, slot, typ),
       )
-    | BindSet(arg) when !gc => Expression.Global_set.make(wasm_mod, slot, arg)
     | BindSet(arg) =>
       Expression.Global_set.make(
         wasm_mod,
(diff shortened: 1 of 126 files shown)


compiler/src/codegen/compcore.re
```diff
@@ -856,16 +854,14 @@ let compile_bind =
     | BindSet(arg) => set_slot(slot, typ, arg)
     | BindTee(arg) => tee_slot(slot, typ, arg)
     };
-  | MGlobalBind(slot, wasm_ty, gc) =>
```
peblair, line 859: Why can we get rid of this? Not 100% sure I understand

stdlib/map.gr
```diff
@@ -280,6 +280,8 @@ let setInArray = array => {
     array[i] = (key, value)
     // no decRef on key and value, since they are stored in array
     Memory.decRef(WasmI32.fromGrain(iter))
+    Memory.incRef(WasmI32.fromGrain((+)))
```
peblair, line 283: Will be a conflict with #1074

peblair (approved):
Looks good, just a couple of comments

compiler/src/codegen/compcore.re
ospencer: Turns out we haven't referenced globals by slot # for ages 😅 

I took it out now to avoid making it work with closures given that it's unused.

peblair: Now I'm more confused 😅 ...the diff is showing that you dropped the `gc` argument, not `slot`

stdlib/map.gr
ospencer: Nah because this is still @disableGC code.

peblair: Ah, I guess this hasn't been migrated in a follow-on PR yet. Carry on

compiler/src/codegen/compcore.re
ospencer: Oh sorry ignore me, I was thinking about something totally different. `wasm_ty` implies GC because it tells you if it's `HeapAllocated` etc.

peblair: Okay, that was sort of what I was thinking, so I just wanted to clarify. Looks good 👍

compiler/test/suites/optimizations.re
```diff
@@ -211,79 +211,125 @@ describe("optimizations", ({test}) => {
   );
   assertAnf(
     "test_local_mutations1",
-    "let mut x = 5; x = 6",
+    "export let foo = () => {let mut x = 5; x = 6}",
```
phated, line 214: Why was this test changed?

compiler/test/suites/optimizations.re
```diff
@@ -211,79 +211,125 @@ describe("optimizations", ({test}) => {
...
+            Imm.const(Const_void),
+          ),
         ),
       );
     },
   );
   assertAnf(
     "test_no_local_mutation_optimization_of_closure_scope_mut",
-    "/* grainc-flags --experimental-wasm-tail-call */ let mut x = 5; let foo = () => x; foo()",
+    "/* grainc-flags --experimental-wasm-tail-call */ export let bar = () => { let mut x = 5; let foo = () => x; foo() }",
```
phated, line 265: Why was this test changed?

compiler/test/__snapshots__/arrays.0f9e7d37.0.snapshot
```diff
@@ -112,21 +118,21 @@ arrays › array_access
     (i32.add
      (i32.shl
       (if (result i32)
-       (i32.lt_u
-        (local.get $1)
+       (i32.lt_s
```
phated, line 121: Why did this change from `lt_u` to `lt_s`?

phated (approved):
Love the cleanup here! One thing I'm confused by is how globals work. Are they functions? Because there are incRefs and decRefs (or sometimes just one or the other) in the things getting hoisted to globals.

compiler/test/suites/optimizations.re
ospencer: Top-level mutables are now just regular wasm globals, so the optimization doesn't do anything at the top level anymore. By pushing this down into the body of a function, it makes sure that the optimization is exercised!

compiler/test/suites/optimizations.re
ospencer: Same reason.

compiler/test/suites/optimizations.re
phated: Makes sense! I wanted to make sure that this change wasn't making this breaking, but it doesn't seem like it does (just different)

compiler/test/__snapshots__/arrays.0f9e7d37.0.snapshot
ospencer: Optimizations are a bit of a mystery. This is going to be a case of Binaryen having decided that this makes more sense now vs before 🤷 

FWIW, `lt_u` of anything and 0 is always going to be false so I'm not really sure what Binaryen was doing before. I saw today actually that there's an optimization that looks at your locals and may switch between signed and unsigned instructions if it thinks it makes sense. Really wild.

ospencer:
Wasm globals are always just values, but we assign top-level closures into them, which makes it look like we're storing functions in them (and in a way we are).

Issue #552, opened 2021-03-09
Runtime: Use real list implementation in toString


ospencer:
Let's remove this once we move List into a type that is built into the compiler.

_Originally posted by @phated in https://github.com/grain-lang/grain/pull/540#discussion_r589953117_

phated:
Blocked by #1593

PR #1135, opened 2022-02-07
feat(stdlib): Convert runtime printing utils to @unsafe

ospencer:
The major thing to notice in this PR is that `runtime/numberUtils.gr` is no longer `--compilation-mode=runtime`. This means that it uses regular allocated Grain memory, which means that string literals, etc. are allowed to be used. Ideally, `malloc.gr`/`gc.gr`/`memory.gr` will be the only files in runtime mode.

### DIFF
--- a/stdlib/runtime/gc.gr
+++ b/stdlib/runtime/gc.gr
@@ -45,8 +45,18 @@ primitive unbox: Box<a> -> a = "@unbox"
 
 exception DecRefError
 
-export let decimalCount32 = box((n: WasmI32) => 0n)
-export let utoa32Buffered = box((a: WasmI32, b: WasmI32, c: WasmI32) => void)
+let decimalCount32Dummy = (n: WasmI32) => 0n
+let utoa32BufferedDummy = (a: WasmI32, b: WasmI32, c: WasmI32) => void
+
+// When these boxes are backpatched, the reference count of each function will
+// fall to zero which would cause them to be freed. We can't free anything that
+// got allocated in runtime mode (since that memory space is not managed by the
+// GC, so here we prevent that by manually setting a higher refcount.
+WasmI32.store(WasmI32.fromGrain(decimalCount32Dummy) - 8n, 2n, 0n)
+WasmI32.store(WasmI32.fromGrain(utoa32BufferedDummy) - 8n, 2n, 0n)
+
+export let decimalCount32 = box(decimalCount32Dummy)
+export let utoa32Buffered = box(utoa32BufferedDummy)
 
 let mut _DEBUG = false
 
--- a/stdlib/runtime/numberUtils.gr
+++ b/stdlib/runtime/numberUtils.gr
@@ -1,4 +1,4 @@
-/* grainc-flags --compilation-mode=runtime */
+/* grainc-flags --no-pervasives */
 
 /*
  * This file was inspired by AssemblyScript's std/assembly/util/number.ts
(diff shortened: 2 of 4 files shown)


stdlib/runtime/gc.gr
```diff
@@ -45,8 +45,18 @@ primitive unbox: Box<a> -> a = "@unbox"
 
 exception DecRefError
 
-export let decimalCount32 = box((n: WasmI32) => 0n)
-export let utoa32Buffered = box((a: WasmI32, b: WasmI32, c: WasmI32) => void)
+let decimalCount32Dummy = (n: WasmI32) => 0n
+let utoa32BufferedDummy = (a: WasmI32, b: WasmI32, c: WasmI32) => void
+
+// When these boxes are backpatched, the reference count of each function will
```
peblair, line 51: Should we maybe emit a warning or something when using boxes in `--compilation-mode=runtime`? It seems like this could bite us

stdlib/runtime/gc.gr
```diff
@@ -45,8 +45,18 @@ primitive unbox: Box<a> -> a = "@unbox"
...
+// fall to zero which would cause them to be freed. We're not allowed to free
+// anything in runtime mode, so here we prevent that by setting a higher
+// refcount.
```
peblair, line 54: Can you clarify in this comment that this is the case because of `--compilation-mode=runtime`? It wasn't clear when I initially read it

peblair (approved):
One minor comment, but this looks :fire:

stdlib/runtime/gc.gr
ospencer: Maybe? The plan is for only `malloc.gr` and `gc.gr` to use runtime mode, so it might not be necessary. But it's not really just a problem for boxes; it's for anything that gets allocated in runtime mode that gets mutated by things outside of the runtime.

I actually thought about a couple of solutions for this:
- The fake reference count of runtime objects is `1`; could just set it to `0xefffffff`. While this would work, I'd be worried about some large program that could decRef enough that the problem would occur, and it'd be a hell of a time debugging that.
- Could alter `free` to never free anything in runtime land. Actually fairly easy to implement with lowish perf overhead, though I dislike that the reference counts for runtime values would be essentially random. That's more of an aesthetic problem though since the count doesn't matter at all.
- Disallow mutable boxes/arrays/record fields in runtime mode. That'd completely prevent this (which is handy) though of course we couldn't use those constructs (which isn't a big deal since we don't need them, but it'd prevent doing backrefs for debugging like this, etc.).

But also maybe this comment is fine 🤷 

Up to you if you want to make an issue for it.

stdlib/runtime/gc.gr
ospencer: Done!

marcusroberts (approved):
Nice to see real Grain code back!

stdlib/runtime/numberUtils.gr
```diff
@@ -936,101 +968,108 @@ export let utoa32Buffered = (buf, value, radix) => {
...
 }
 
+@unsafe
 export let utoa32 = (value, radix) => {
   if (WasmI32.ltS(radix, 2n) || WasmI32.gtS(radix, 36n)) {
     throw Exception.InvalidArgument(
       "toString() radix argument must be between 2 and 36"
     )
   }
   let str = if (WasmI32.eqz(value)) {
```
phated, line 978: Can you just remove `let str` here and at the end where you return it?

stdlib/runtime/numberUtils.gr
```diff
@@ -936,101 +968,108 @@ export let utoa32Buffered = (buf, value, radix) => {
...
 }
 
+@unsafe
 export let utoa64 = (value, radix) => {
   if (WasmI32.ltS(radix, 2n) || WasmI32.gtS(radix, 36n)) {
     throw Exception.InvalidArgument(
       "toString() radix argument must be between 2 and 36"
     )
   }
   let str = if (WasmI64.eqz(value)) {
```
phated, line 1042: Can you remove this `str` variable?

stdlib/runtime/gc.gr
phated: You know me, I'm always happy with issues where we can have a longer discussion!

stdlib/runtime/string.gr
```diff
@@ -37,21 +37,13 @@ primitive (&&): (Bool, Bool) -> Bool = "@and"
...
-}
-
-@disableGC
-let slConsDisableGc = (a, b) => {
-  Memory.incRef(WasmI32.fromGrain(SLCons))
-  Memory.incRef(WasmI32.fromGrain(a))
-  // for easier chaining, we don't incRef `b`
-  SLCons(a, b)
+  [],
+  [...](String, StringList),
```
phated, line 41: This is 🔥  Love it

phated: Did we have an issue about this somewhere that should be closed by this PR?

phated (approved):
🎉 Awesome!

stdlib/runtime/gc.gr
ospencer: #1156

stdlib/runtime/string.gr
ospencer: Probably https://github.com/grain-lang/grain/issues/552 but that's not _quite_ ready to be closed.

PR #1136, opened 2022-02-23
fix(stdlib)!: Change Buffer#addStringSlice API

### LINKED ISSUE #1071: String.slice and Buffer.addStringSlice have different API signatures
Edited: Original description had flipped functions

While working on #1065, it was uncovered that the API signatures of these are different, `Buffer.addStringSlice` takes `start` and `length` while `String.slice` takes `start` and `end`. This is a problem because `Buffer.addStringSlice` uses `String.slice` underneath and we needed to apply a workaround to find the ~~length~~ end based on length..

We want to align these APIs, but that is a breaking change that we need to batch for our next breaking release.

peblair:
This PR changes the semantics of `Buffer#addStringSlice` to align with those of `String#slice`.

Closes #1071.

### DIFF
--- a/compiler/test/stdlib/buffer.test.gr
+++ b/compiler/test/stdlib/buffer.test.gr
@@ -197,21 +197,28 @@ Buffer.addString(str, buf)
 assert Buffer.toString(buf) == str
 
 // addStringSlice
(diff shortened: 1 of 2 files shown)


phated (approved):
🔥 Excellent!

ospencer (approved):

marcusroberts (approved):
Nice!

PR #1137, opened 2022-02-23
feat(runtime)!: Cyclic GC

### LINKED ISSUE #989: Cyclic GC
We don't currently support cyclic garbage collection. This is an issue for things like recursive `enums` and functions (ref #773).

Before #988, we had a test (or two?) that supposedly exercised cyclic GC, but weren't actually testing the right behavior because they weren't supposed to pass typechecking. This was noticed when the fix in #988 was applied, so the tests were removed.

We need to implement/support cyclical GC and then re-add tests for it.

peblair:
- Tests are still not passing, so this is a draft
- This is currently breaking, since the GC header is changed. Per our conversation about it, we can revisit this change to see if it's needed (if not, then these changes are not breaking).

Closes #989.

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -1460,7 +1460,7 @@ let heap_allocate = (wasm_mod, env, num_words: int) =>
         ),
         Expression.Const.make(
           wasm_mod,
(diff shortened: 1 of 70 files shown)


peblair:
Closing, per comment on #1152 .

PR #1139, opened 2022-02-27
chore(compiler): Selectively include runtime imports in modules

ospencer:


peblair (approved):

compiler/src/codegen/compcore.re
```diff
@@ -3502,20 +3511,8 @@ let validate_module = (~name=?, wasm_mod: Module.t) =>
...
-      ) => {
-    let rt_idx =
-      if (is_runtime_import) {
-        idx + dynamic_offset;
-      } else {
-        idx;
-      };
-    let rt_idx_name = Printf.sprintf("global_%d", rt_idx);
+      (acc_env, idx, {mimp_mod, mimp_name, mimp_type, mimp_kind}) => {
+    let idx_name = Printf.sprintf("global_%d", idx);
```
phated, line 3515: Didn't you add a helper for `global_name`?

compiler/src/codegen/compcore.re
```diff
@@ -3621,10 +3596,10 @@ let compile_wasm_module = (~env=?, ~name=?, prog) => {
 
   let compile_all = () => {
     ignore @@ compile_functions(wasm_mod, env, prog);
+    ignore @@ compile_tables(wasm_mod, env, prog);
```
phated, line 3599: Why does the order matter here?

phated (approved):
Absolutely love the cleanup here!

compiler/src/codegen/compcore.re
ospencer: Compiling imports needs to happen after anything that imports something is done so it knows what imports need to get included (and tables import an offset).

PR #1150, opened 2022-03-03
fix(compiler)!: Enable garbage collection of simply-recursive functions

### LINKED ISSUE #1148: Avoid incRef on recursive functions
As described in #989, we currently cannot collect cyclic references. A large class of these cyclic references are simple recursive functions. As per the community call today, we think that this can be mitigated by making the self-references stored in closures not increment their reference counts (then, when the closure falls out of scope, the closure should be collectable). We should give this a try, and, if we're confident in it, merge it in.

peblair:
Closes #1148. Marking as a draft because there are failing stdlib tests. It _looks_ like this is just because of a problem in random.gr (the other failing modules import it directly or indirectly), but I have not been able to figure out what it is. To reproduce the problem, one can run this program:
```grain
import Random from "sys/random"
Random.random()
```
(Note: This implementation can probably be optimized a bit, but I want to track down the test failure first)

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -2515,15 +2515,42 @@ let collect_backpatches = (env, f) => {
 
 let do_backpatches = (wasm_mod, env, backpatches) => {
   let do_backpatch = ((lam, {variables})) => {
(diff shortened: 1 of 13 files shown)


peblair:
Closing in favor of #1152.

PR #1151, opened 2022-03-04
feat(compiler): Upgrade binaryen to 0.15.0 to support Mac M1 arch

### LINKED ISSUE #848: What is your plan for supporting Apple Silicon?
Do you have any plan to support Apple Silicon? I tried to build grain from source but it seems to be incapable of compiling on M1 and I don't have any idea what is going wrong with the build script. Do you have any idea or plan for this?

phated:
This upgrades Grain to use binaryen.ml 0.15.0 which includes libbinaryen v105 and my new js_of_ocaml implementation of foreigns. The libbinaryen v105 release was the release that removed their python dependency.

This change means that we are now building **both** cmake and libbinaryen from source, which will take a lot longer on a fresh build, but the way I split up the packages should leverage caching very well—which is what differentiates this from the first binaryen.ml version that @ospencer wrote that required the cmake to be re-run on almost any change.

In addition, I've locked down some of our dependencies because the ocaml ecosystem has broken some stuff with esy. 

Oscar also asked that we add a run of the tests being compiled to JS. I added this but we need to skip any tests that would use `Unix`, so I updated the tests to look at the backend type and skip them. I also added this test run to the CI. Let me know if you think it should be a separate "step".

Closes #848


ospencer (approved):
This looks great! Awesome work, and I'm super happy about M1 support! I have a slight preference for the JS tests to be a separate step (more obvious that it was the JS tests that failed in the output) but not going to block on it. 🥳

phated:
> I have a slight preference for the JS tests to be a separate step (more obvious that it was the JS tests that failed in the output)

I have this commit locally, but going to wait for a green run before I push (so we actually have some caching).

peblair:
> we need to skip any tests that would use Unix, so I updated the tests to look at the backend type and skip them.

Do we have a strategy for this? If not, and we can't run all of the tests with the JS runner, then why do we want it at all?

peblair (approved):
See comment, but this looks good!

ospencer:
The JS tests are the compiler compiled to JS (the main version of the Grain compiler we ship, but not the one we test), which has two backends we have to worry about, both js_of_ocaml (which we can feel reasonably confident about) and binaryen.ml (which we hand wrote bindings for). Having these tests even if it's not the whole suite gets us a lot of confidence that things are at least mostly working as intended. Blaine actually found something like 4 bugs in the JS implementation of binaryen.ml from running these tests, which we probably would have never noticed! Being able to check the snapshots is actually fairly sufficient, since if the snapshots match they won't execute any differently. The only thing we really miss out on at the moment is testing linking, but baby steps.

peblair:
Okay, so if I understand correctly, we don't have any plans of removing the non-JS test suite for the foreseeable future?

ospencer:
Oh, no, never.

PR #1152, opened 2022-03-05
feat(compiler): Reduce closure sizes by utilizing `$self` argument when possible

### LINKED ISSUE #1148: Avoid incRef on recursive functions

peblair:
Closes #1148 (though not in the way that the ticket describes). This PR changes the compilation of functions such that self-references are looked up at runtime via argument #0 (`$self`), rather than a closure variable. In addition to making the closures smaller, this should increase the scope of functions which are garbage-collectible (see #989).

I've implemented this patch in two commits, since the change did not modify any snapshots. The first commit introduces a test case, and the second commit shows the diff on that test case's snapshot to illustrate the change (and show that it actually works).

### DIFF
--- a/compiler/src/codegen/transl_anf.re
+++ b/compiler/src/codegen/transl_anf.re
@@ -558,7 +558,14 @@ let compile_lambda =
     (~name=?, id, env, args, body, attrs, loc): Mashtree.closure_data => {
   register_function(id);
   let (body, return_type) = body;
-  let used_var_set = Anf_utils.anf_free_vars(body);
+  // NOTE: we special-case `id`, since we want to
+  //       have simply-recursive uses of identifiers use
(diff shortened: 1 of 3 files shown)


peblair:
We should also probably close #1137 if/when this is merged, since this should un-block us from punting until the WASM GC proposal lands.

ospencer (approved):
Looks great!

phated (approved):
Let's gooooo! 🎉

PR #1161, opened 2022-03-16
chore: Reset ident stamps when compiling

ospencer:
`Ident.reinit` doesn't actually reset anything—it sort of snapshots where each module begins and ends. This PR adds a `Ident.setup` function which sets the current ident stamp back to 1000 (because builtins use 1-999).

### DIFF
--- a/compiler/src/typed/builtin_types.re
+++ b/compiler/src/typed/builtin_types.re
@@ -193,6 +193,6 @@ let builtin_values =
    be defined in this file (above!) without breaking .cmi
    compatibility. */
 
-let _ = Ident.set_current_time(999);
+let _ = Ident.setup();
 let builtin_idents = List.rev(builtin_idents^);
 let builtin_decls = List.rev(builtin_decls^);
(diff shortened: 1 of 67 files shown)


ospencer:
Should also note that this doesn't at all change programs that are compiled through the CLI, just multiple programs compiled programmatically.

compiler/src/typed/builtin_types.re
```diff
@@ -193,6 +193,6 @@ let builtin_values =
    be defined in this file (above!) without breaking .cmi
    compatibility. */
 
-let _ = Ident.set_current_time(999);
+let _ = Ident.setup();
```
ospencer, line 196: This is what made calling this `reset` feel weird.

compiler/test/__snapshots__/enums.aa34084a.0.snapshot
```diff
@@ -118,7 +118,7 @@ enums › adt_trailing
           )
           (i64.store offset=24
            (local.get $0)
-           (i64.const 85899347075)
+           (i64.const 85899347051)
```
ospencer, line 121: This code populates the type information in memory that's used for printing. The stamp is what is used as the identifier for enums at runtime, so this changes to reflect that.

phated: Thanks 🙏  This was going to be my only question

phated (approved):
Awesome! Thanks for digging into this. With this, we'll be able to run the snapshot tests through a JS version of the compiler 🎉

jozanza (approved):

PR #1165, opened 2022-03-18
fix(stdlib): Fix float printing in dtoa

ospencer:
Problem was that there was an unused argument in `genDigits`, and Grain "any" types are GC'd. We should consider adding a warning to `@unsafe` when arguments are unused/any types.


ospencer:
Made https://github.com/grain-lang/grain/issues/1166 to add a warning.

peblair (approved):
Scary! We should add that warning soon.

phated (approved):
:bow:

PR #1173, opened 2022-03-20
feat(compiler): Support pattern matching "or" patterns

### LINKED ISSUE #264: Pattern Matching: Support "or" patterns

ospencer:
This PR refactors the match compiler to properly support or-patterns. The two major changes are how bindings work and how constant patterns are compiled. Slots for each of the bindings needed in each of the branches are created, and during the branch selection process these slots are filled with the proper values. Constant patterns are now compiled the same way constructors are and should be more efficient now.

This opens the door for a number of optimizations we can write, but I left them out of this PR to keep it relatively simple.

You may notice that a fair number of snapshots are a bit bigger now; we need a few optimizations to remove the cruft, which will come after this PR.

Closes #264

### DIFF
--- a/compiler/src/middle_end/linearize.re
+++ b/compiler/src/middle_end/linearize.re
@@ -5,7 +5,6 @@ open Typedtree;
 open Type_utils;
 open Anftree;
 open Anf_helper;
-open Grain_utils.Either;
 
 module MatchCompiler = Matchcomp.MatchTreeCompiler;
 
@@ -116,8 +115,7 @@ let convert_binds = anf_binds => {
 };
 
 let transl_const =
-    (c: Types.constant)
-    : Grain_utils.Either.t(imm_expression, (string, comp_expression)) => {
+    (c: Types.constant): Either.t(imm_expression, (string, comp_expression)) => {
   switch (c) {
   | Const_number(n) => Right(("number", Comp.number(n)))
   | Const_int32(i) => Right(("int32", Comp.int32(i)))
--- a/compiler/src/middle_end/linearize.re
+++ b/compiler/src/middle_end/linearize.re
@@ -5,6 +5,7 @@ open Typedtree;
 open Type_utils;
 open Anftree;
 open Anf_helper;
+open Grain_utils.Either;
 
 module MatchCompiler = Matchcomp.MatchTreeCompiler;
 
@@ -70,9 +71,17 @@ let lookup_symbol = (~allocation_type, mod_, mod_decl, name, original_name) => {
   };
 };
 
-type either('a, 'b) =
-  | Left('a)
-  | Right('b);
+let convert_bind = (body, bind) =>
+  switch (bind) {
+  | BSeq(exp) => AExp.seq(exp, body)
+  | BLet(name, exp, global) =>
+    AExp.let_(~global, Nonrecursive, [(name, exp)], body)
+  | BLetMut(name, exp, global) =>
+    AExp.let_(~mut_flag=Mutable, ~global, Nonrecursive, [(name, exp)], body)
+  | BLetRec(names, global) => AExp.let_(~global, Recursive, names, body)
+  | BLetRecMut(names, global) =>
+    AExp.let_(~mut_flag=Mutable, ~global, Recursive, names, body)
+  };
 
 let convert_binds = anf_binds => {
   let void_comp =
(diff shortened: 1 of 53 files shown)


compiler/src/middle_end/matchcomp.re
```diff
@@ -719,120 +368,200 @@ let rec rotate_if_needed = mtx =>
...
+
+/* [See paper for details; modified slightly]
 
       Row: ([p_1; p_2; ...], a_j)
 
       Pattern p_1 (for row j)       Row(s) of D(P)
       -----------------------       --------------
            c(q_1,...,q_n)               No row
-                 _                [([p_2; ...], a_j)]
+                 _                [([p_1; p_2; ...], a_j)]
```
peblair, line 483: This confuses me. The default matrix is supposed to drop the first pattern (that's the whole idea, as far as I understand). Can you explain?

compiler/src/middle_end/matchcomp.re
```diff
@@ -719,120 +368,200 @@ let rec rotate_if_needed = mtx =>
...
-        Guard(mb, Leaf(i), result);
-      | None => Leaf(i)
+        Guard(mb, pats, aliases, Leaf(i, pats, aliases), result);
+      | None => Leaf(i, pats, aliases)
       };
     rest_tree;
   | _ when !col_is_wildcard(mtx, 0) =>
     /* If the first column contains a non-wildcard pattern */
+
+    let alias = Ident.create("match_explode_alias");
```
peblair, line 551: To clarify, this identifier is only used when there is an alias pattern?

peblair (commented):
Looks pretty good (`matchcomp.re` is definitely cleaner now), but I have a couple of questions before I hit approve.

compiler/src/middle_end/matchcomp.re
ospencer: I like to think of `default_matrix` as being the same as any of the other specialization matrixes, except this one specializes on "any" patterns, but otherwise it's no different—just needs to select the rows where the pattern in the first column always matches. Because it knows that the first column always matches, when it selects a row it _can_ remove that first pattern, but it doesn't have to. In our case, we want to see that pattern since it might contain a binding that we need to handle. Removing that first pattern can potentially make compiling the matrix faster since it never needs to look at that column again, but it makes no difference for the compiled matrix.

The key bit is that we'll always select a row if all of the patterns in that row always match.

With how we handle our value stack, it's much simpler to leave the pattern there since we won't need to manage removing that value from the stack (and dealing with finding that value later when we want to use it in one of the branches). So now, when we reach a leaf, we've got the full expanded row and all of the bindings to go with it! 🎉

peblair: I can get behind your rationale, but maybe we should have a comment explaining any divergence from the paper (since the paper's pattern syntax is a bit more trimmed down, the formula they provide is different from what's now shown in this comment).

marcusroberts:
Does this need formatter support @ospencer

jozanza (approved):

compiler/src/middle_end/linearize.re
```diff
@@ -1375,46 +1350,22 @@ let rec transl_anf_statement =
...
-            ]
-          | Immutable => [BLet(bind, exp_ans, Global({exported: exported}))]
-          };
-        let (tmp, anf_patts) =
-          bind_patts(~mut_flag, ~global=Global({exported: exported}), patt);
-        binds @ [BLet(tmp, exp_ans, Nonglobal), ...anf_patts];
-      | TPatAny => [BSeq(exp_ans)]
-      | _ =>
-        failwith(
-          "NYI: transl_anf_statement: Non-record or non-tuple destructuring in let",
```
phated, line 1415: Do we want tests for this now that it seems that records and tuples can be destructured?

compiler/src/middle_end/matchcomp.re
```diff
@@ -10,6 +10,7 @@ open Grain_utils;
 open Types;
 open Typedtree;
 open Type_utils;
+open Either;
```
phated, line 13: In ocaml stdlib

compiler/src/middle_end/matchcomp.re
phated: Makes sense to me to explain the above in the comment (as @peblair requested).

compiler/src/middle_end/matchcomp.re
```diff
@@ -719,120 +368,200 @@ let rec rotate_if_needed = mtx =>
...
+  | Const_void
+  | Const_bool(_)
+  | Const_char(_)
+  | Const_wasmi32(_) => PhysicalEquality
+  | Const_wasmi64(_)
+  | Const_wasmf32(_)
+  | Const_wasmf64(_) =>
+    failwith(
+      "Pattern matching not supported on low-level i64/f32/f64 types.",
+    );
```
phated, line 533: With these pattern matching updates, how far are we from supporting these?

compiler/src/middle_end/matchcomp.re
```diff
@@ -900,6 +625,34 @@ let rec compile_matrix = mtx =>
           };
         Switch(ConstructorSwitch, switch_branches, default_tree);
       };
+    | ConstantMatrix =>
+      let constants = matrix_head_constants(mtx);
+      let equality_type = equality_type(List.hd(constants));
+
+      // TODO: Optimize physical equality checks into a switch.
```
phated, line 632: Do we want an issue for this?

compiler/src/middle_end/matchcomp.re
```diff
@@ -914,156 +667,503 @@ let rec compile_matrix = mtx =>
...
+        Comp.if_(
+          ~allocation_type=true_comp.comp_allocation_type,
+          Imm.id(cond_id),
+          fold_tree(true_setup, true_comp),
+          fold_tree(false_setup, false_comp),
+        ),
+        [alias_binding, ...cond_setup],
+      );
+    | Fail =>
+      /* FIXME: We need a "throw error" node in ANF */
```
phated, line 875: Fix when? Do we need it in this PR or an issue?

compiler/src/middle_end/matchcomp.re
```diff
@@ -914,156 +667,503 @@ let rec compile_matrix = mtx =>
...
+        | ConstantMatrix =>
+          failwith("Internal error: Explode of a constant matrix")
+        };
+
+      let new_values =
+        List.map(
+          fun
+          | BLet(id, _, _) => Imm.id(id)
+          | _ =>
+            failwith("matchcomp: compile_tree_help: binding was not BLet"),
```
phated, line 966: Is this not "Impossible"?

phated (changes requested):
Looking really good! We need to switch over to ocaml's Either since we upgraded to 4.12 and I had some other comments.

It looks like the snapshots are outdated so I didn't look at them.

compiler/src/middle_end/linearize.re
ospencer: Records and tuples could be destructured before, but this case was handling destructuing that wasn't one of those things (or patterns, constructors, etc.). I'll add some more tests :)

compiler/src/middle_end/matchcomp.re
ospencer: "this is only really possible with wasm GC" far haha

compiler/src/middle_end/matchcomp.re
ospencer: This was just moved here but it was outdated anyway, `Imm.trap()` is quite literally a throw error node haha

compiler/src/middle_end/matchcomp.re
ospencer: https://github.com/grain-lang/grain/issues/1185

ospencer:
@phated @peblair this should be all set!

compiler/src/middle_end/matchcomp.re
peblair: LGTM

peblair (approved):
Awesome work!

phated (approved):
🔥  Let's GOOOO!

PR #1174, opened 2022-03-20
feat(compiler): Support pattern aliases

ospencer:


marcusroberts:
Does the formatter need updating for this?

jozanza (approved):
🔥

compiler/test/suites/pattern_matching.re
```diff
@@ -222,6 +222,18 @@ describe("pattern matching", ({test}) => {
     "or_match_3",
     "match ([5]) { [a, _] | [_, a, _] | [a] => true, _ => false }",
   );
+  // Aliases
+  assertSnapshot("alias_match_1", "match (true) { _ as p => p }");
+  assertSnapshot("alias_match_2", "match (true) { a as b => a && b }");
+  assertSnapshot("alias_match_3", "match (true) { true | false as p => p }");
```
phated, line 228: Can we also have a runnable test for this?

phated (changes requested):
So awesome! I just want a runnable test to know that constant or-pattern alias is doing what I expect

compiler/test/suites/pattern_matching.re
ospencer: Done!

phated: Looks great 👍

phated (approved):
Huge approval from me!

PR #1175, opened 2022-03-22
feat(compiler): Call known functions across module boundaries

ospencer:
This PR uses regular `call`s where possible between modules. This should be fairly significant for performance, as direct `call`s are much faster than `call_indirect`s. Additionally, Binaryen can now provide better optimizations after linking.

### DIFF
--- a/compiler/src/codegen/comp_utils.re
+++ b/compiler/src/codegen/comp_utils.re
@@ -92,7 +92,7 @@ let grain_env_name = "_grainEnv";
 
 let is_grain_env = str => grain_env_name == str;
 
-let get_exported_names = (~local_functions=?, ~local_globals=?, wasm_mod) => {
+let get_exported_names = (~function_names=?, ~global_names=?, wasm_mod) => {
   let num_exports = Export.get_num_exports(wasm_mod);
   let exported_names: Hashtbl.t(string, string) = Hashtbl.create(10);
   for (i in 0 to num_exports - 1) {
@@ -103,16 +103,15 @@ let get_exported_names = (~local_functions=?, ~local_globals=?, wasm_mod) => {
 
     if (export_kind == Export.external_function) {
       let new_internal_name =
-        switch (local_functions) {
-        | Some(local_functions) =>
-          Hashtbl.find(local_functions, internal_name)
+        switch (function_names) {
+        | Some(function_names) => Hashtbl.find(function_names, internal_name)
         | None => internal_name
         };
       Hashtbl.add(exported_names, exported_name, new_internal_name);
     } else if (export_kind == Export.external_global) {
       let new_internal_name =
-        switch (local_globals) {
-        | Some(local_globals) => Hashtbl.find(local_globals, internal_name)
+        switch (global_names) {
+        | Some(global_names) => Hashtbl.find(global_names, internal_name)
         | None => internal_name
         };
       Hashtbl.add(exported_names, exported_name, new_internal_name);
--- a/compiler/src/codegen/comp_utils.re
+++ b/compiler/src/codegen/comp_utils.re
@@ -116,7 +116,6 @@ let get_exported_names = (~function_names=?, ~global_names=?, wasm_mod) => {
         };
       Hashtbl.add(exported_names, exported_name, new_internal_name);
     };
-    ();
   };
   exported_names;
 };
--- a/compiler/src/codegen/comp_utils.re
+++ b/compiler/src/codegen/comp_utils.re
@@ -180,7 +180,7 @@ let write_universal_exports =
                   arguments,
                   call_result_types,
                 )
-              | _ =>
+              | Indirect =>
                 let call_arg_types =
                   Type.create(
                     Array.of_list(
@@ -204,6 +204,7 @@ let write_universal_exports =
                   call_arg_types,
                   call_result_types,
                 );
+              | Unknown => failwith("Impossible: Unknown function call type")
               };
             let function_body =
               switch (rets) {
--- a/compiler/src/codegen/comp_utils.re
+++ b/compiler/src/codegen/comp_utils.re
@@ -92,23 +92,32 @@ let grain_env_name = "_grainEnv";
 
 let is_grain_env = str => grain_env_name == str;
 
-let get_exported_names = (~local_names=?, wasm_mod) => {
+let get_exported_names = (~local_functions=?, ~local_globals=?, wasm_mod) => {
   let num_exports = Export.get_num_exports(wasm_mod);
   let exported_names: Hashtbl.t(string, string) = Hashtbl.create(10);
   for (i in 0 to num_exports - 1) {
     let export = Export.get_export_by_index(wasm_mod, i);
     let export_kind = Export.export_get_kind(export);
-    if (export_kind == Export.external_function
-        || export_kind == Export.external_global) {
-      let exported_name = Export.get_name(export);
-      let internal_name = Export.get_value(export);
+    let exported_name = Export.get_name(export);
+    let internal_name = Export.get_value(export);
+
+    if (export_kind == Export.external_function) {
       let new_internal_name =
-        switch (local_names) {
-        | Some(local_names) => Hashtbl.find(local_names, internal_name)
+        switch (local_functions) {
+        | Some(local_functions) =>
+          Hashtbl.find(local_functions, internal_name)
+        | None => internal_name
+        };
+      Hashtbl.add(exported_names, exported_name, new_internal_name);
+    } else if (export_kind == Export.external_global) {
+      let new_internal_name =
+        switch (local_globals) {
+        | Some(local_globals) => Hashtbl.find(local_globals, internal_name)
         | None => internal_name
         };
       Hashtbl.add(exported_names, exported_name, new_internal_name);
     };
+    ();
   };
   exported_names;
 };
@@ -133,12 +142,23 @@ let write_universal_exports =
       List.iter(
         item => {
           switch (item) {
-          | TSigValue(id, {val_repr: ReprFunction(args, rets)}) =>
+          | TSigValue(
+              id,
+              {
+                val_repr: ReprFunction(args, rets, direct),
+                val_fullpath: path,
+              },
+            ) =>
             let name = Ident.name(id);
             let exported_name = "GRAIN$EXPORT$" ++ name;
-            let internal_name = Hashtbl.find(exported_names, exported_name);
+            let internal_global_name =
+              Hashtbl.find(exported_names, exported_name);
             let get_closure = () =>
-              Expression.Global_get.make(wasm_mod, internal_name, Type.int32);
+              Expression.Global_get.make(
+                wasm_mod,
+                internal_global_name,
+                Type.int32,
+              );
             let arguments =
               List.mapi(
                 (i, arg) =>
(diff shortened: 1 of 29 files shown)


compiler/src/middle_end/linearize.re
```diff
@@ -52,16 +53,29 @@ let lookup_symbol = (~allocation_type, mod_, mod_decl, name, original_name) => {
     let fresh = gensym(name);
     switch (mod_decl.md_filepath) {
     | Some(filepath) =>
+      let shape =
+        switch (repr) {
+        | ReprFunction(args, rets, Direct(_)) =>
```
peblair, line 58: Curious if/how this branch ever matches? Looking at `middle_end` and `typing`, everything looks like it's `Unknown` at this point in the pipeline?

compiler/src/linking/link.re
```diff
@@ -339,7 +307,9 @@ let link_all = (linked_mod, dependencies, signature) => {
   table_offset := 0;
 
   let link_one = dep => {
-    let local_names: Hashtbl.t(string, string) = Hashtbl.create(128);
+    let local_functions: Hashtbl.t(string, string) = Hashtbl.create(128);
+    let local_globals: Hashtbl.t(string, string) = Hashtbl.create(128);
```
peblair, line 311:
```suggestion
    let module_globals: Hashtbl.t(string, string) = Hashtbl.create(128);
```
or something similar? `local_globals` really throws me through a loop (same thing for `comp_utils.re`)

compiler/src/codegen/comp_utils.re
```diff
@@ -133,47 +142,70 @@ let write_universal_exports =
...
-              );
+              switch (direct) {
+              | Direct(name) =>
+                Expression.Call.make(
+                  wasm_mod,
+                  Hashtbl.find(exported_names, name),
+                  arguments,
+                  call_result_types,
+                )
+              | _ =>
```
peblair, line 184: Should we choke if this is `Unknown`?

peblair (commented):
Still wrapping my head around this, but I've left some initial thoughts

stdlib/sys/file.gr
```diff
@@ -248,7 +248,12 @@ let rec combineRightsHelp = (acc, dirflags) => {
         PathUnlinkFile => _RIGHT_PATH_UNLINK_FILE,
         PollFdReadwrite => _RIGHT_POLL_FD_READWRITE,
         SockShutdown => _RIGHT_SOCK_SHUTDOWN,
-        _ => fail "Unknown file right",
+        _ => {
+          ignore(fail "Unknown file right")
```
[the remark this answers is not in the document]
ospencer: Just to mute the warning about it being an unsound type since it's not used!

compiler/src/linking/link.re
ospencer: It's definitely a funny-sounding name 🙂 

Will update it.

stdlib/sys/file.gr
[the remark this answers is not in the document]
peblair: Sorry, I'm still confused. Why does the `a` return value of `fail "Unknown file right"` not unify to `WasmI64` (in which case it wouldn't need to be ignored)?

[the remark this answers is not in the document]
ospencer: It does! But at the WebAssembly level, Grain "any" types are i32s but this block needs to return an i64 like the others. We got around this before by this being a `call_indirect`—we just lied about what type `fail` returned. At runtime, instead of raising `Failure`, this is a call_indirect typechecking error.

Essentially, we need a compiler primitive for `fail` which can print the error and then generate `unreachable`. (But we can't do that until the enum refactor is done because `Failure` would need to be baked into the compiler.)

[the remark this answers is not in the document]
ospencer: (And now since this is a regular `call` instruction, we can't lie about the return type.)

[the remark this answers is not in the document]
peblair: Ah, okay, that makes sense. Now I understand why it was needed as part of this PR. Thanks 👍

marcusroberts (approved):
Wow, that's a lot of work.   I'm not going to spot problems, but I can say the approach looks good to me.

compiler/src/codegen/transl_anf.re
```diff
@@ -539,24 +539,32 @@ let compile_imm = (env, i: imm_expression) =>
...
 let known_function = f =>
   switch (f.imm_desc) {
   | ImmId(id) =>
-    if (Ident_tbl.mem(known_functions, id)) {
-      Some(Ident.unique_name(id));
-    } else {
-      None;
+    switch (Ident_tbl.find_opt(known_functions, id)) {
+    | Some(Internal(id)) => Some(Ident.unique_name(id))
+    | Some(Imported(id, name)) => Some(name)
```
peblair, line 558: double-checking: `name` is unique here (defined at the import site), right?

peblair (approved):
This seems good IMO. One question before merging, but that's it.

compiler/src/codegen/transl_anf.re
ospencer: Yup, unique by module/function pair.

compiler/src/codegen/comp_utils.re
```diff
@@ -92,23 +92,31 @@ let grain_env_name = "_grainEnv";
...
-        || export_kind == Export.external_global) {
-      let exported_name = Export.get_name(export);
-      let internal_name = Export.get_value(export);
+    let exported_name = Export.get_name(export);
+    let internal_name = Export.get_value(export);
+
+    if (export_kind == Export.external_function) {
+      let new_internal_name =
+        switch (function_names) {
+        | Some(function_names) => Hashtbl.find(function_names, internal_name)
```
phated, line 107: This `Hashtbl.find` will never raise?

compiler/src/codegen/comp_utils.re
```diff
@@ -92,23 +92,31 @@ let grain_env_name = "_grainEnv";
...
   let exported_names: Hashtbl.t(string, string) = Hashtbl.create(10);
   for (i in 0 to num_exports - 1) {
     let export = Export.get_export_by_index(wasm_mod, i);
     let export_kind = Export.export_get_kind(export);
-    if (export_kind == Export.external_function
-        || export_kind == Export.external_global) {
-      let exported_name = Export.get_name(export);
-      let internal_name = Export.get_value(export);
+    let exported_name = Export.get_name(export);
+    let internal_name = Export.get_value(export);
```
phated, line 102: Why is the "value" a name?

compiler/src/codegen/comp_utils.re
```diff
@@ -92,23 +92,31 @@ let grain_env_name = "_grainEnv";
...
+        | Some(function_names) => Hashtbl.find(function_names, internal_name)
+        | None => internal_name
+        };
+      Hashtbl.add(exported_names, exported_name, new_internal_name);
+    } else if (export_kind == Export.external_global) {
       let new_internal_name =
-        switch (local_names) {
-        | Some(local_names) => Hashtbl.find(local_names, internal_name)
+        switch (global_names) {
+        | Some(global_names) => Hashtbl.find(global_names, internal_name)
```
phated, line 114: Same question about Hashtbl

compiler/src/codegen/comp_utils.re
```diff
@@ -92,23 +92,31 @@ let grain_env_name = "_grainEnv";
...
         | None => internal_name
         };
       Hashtbl.add(exported_names, exported_name, new_internal_name);
     };
+    ();
```
phated, line 119: Just being explicit or does this need to be here?

compiler/src/codegen/comp_utils.re
```diff
@@ -133,47 +141,71 @@ let write_universal_exports =
...
-                func_ptr,
-                arguments,
-                call_arg_types,
-                call_result_types,
-              );
+              switch (direct) {
+              | Direct(name) =>
+                Expression.Call.make(
+                  wasm_mod,
+                  Hashtbl.find(exported_names, name),
```
phated, line 179: All these `Hashtbl.find` make me nervous. Are we 100% sure they will always have the values?

compiler/src/codegen/compcore.re
```diff
@@ -3323,14 +3322,15 @@ let compile_imports = (wasm_mod, env, {imports}) => {
     | MImportWasm => Ident.name(name)
     | MImportGrain => "GRAIN$MODULE$" ++ Ident.name(name);
 
-  let compile_import_name = name =>
-    fun
```
phated, line 3327: had to double-take on reading the `fun` and then working backwards on this refactor. `fun` is nice, but not clear for my brain during reviews. Nice cleanup! 😄

compiler/src/codegen/transl_anf.re
```diff
@@ -1048,36 +1058,55 @@ let lift_imports = (env, imports) => {
...
+                mimp_mod,
+                mimp_name,
+                mimp_type: process_shape(true, imp_shape),
+                mimp_kind: MImportGrain,
+                mimp_setup: MCallGetter,
+                mimp_used: true,
+              },
+            ],
+          )
         | FunctionShape(_) =>
```
phated, line 1077: Can you explain FunctionShape to me a bit? Since it seems like we're adding support that we didn't have before

compiler/src/linking/link.re
```diff
@@ -135,7 +135,16 @@ let is_function_imported = func =>
 //       constructing the AST (either through constructing two separate instances or by
 //       using Expression.copy())
 
-let rec globalize_names = (local_names, expr) => {
+let rec globalize_names = (function_names, global_names, label_names, expr) => {
```
phated, line 138: Any reason that `function_names`, `global_names` and `label_names` aren't named parameters?

stdlib/sys/file.gr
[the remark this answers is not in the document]
phated: @ospencer Do we have an issue for your comment above? If not, please create it as a follow-up to baking Enum into the compiler.

phated (changes requested):
Snapshots look great!! I had a few questions/comments.

stdlib/sys/file.gr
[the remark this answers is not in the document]
ospencer: https://github.com/grain-lang/grain/issues/1192

compiler/src/codegen/comp_utils.re
ospencer: That's just what Binaryen calls it. Probably just to be consistent with the other getters.

compiler/src/codegen/comp_utils.re
ospencer: It would only raise if the linker failed to provide the names of all the functions/locals, but if that happened the linker would fail far before it ever got here.

compiler/src/codegen/comp_utils.re
ospencer: Yes, and there's no way to recover if it doesn't (compilation failure/bug). We could wrap the whole thing in a `try` with a catch on `Not_found` to say "hey report this if you ever see it" but I didn't feel we really needed that.

compiler/src/codegen/transl_anf.re
ospencer: Of course! Imports take one of two shapes, `GlobalShape` or `FunctionShape`, corresponding to wasm globals or wasm functions. Imports that happen via `foreign wasm` are either `GlobalShape` or `FunctionShape`, and before this PR, Grain imports were only `GlobalShape`, meaning that imported functions were only ever imported as values—the closure for the function. This meant that the only way the function could be called was via a `call_indirect`. 

This change allows us say that we're importing an actual Grain webassembly function (and its closure), so we can call it via a regular `call` instruction rather than _having_ to pull the pointer from the closure.

phated (approved):
Excellent! ❤️

PR #1176, opened 2022-03-22
feat(compiler): Allow function re-exports to use regular call instruction

ospencer:


compiler/src/codegen/transl_anf.re
```diff
@@ -1228,44 +1235,78 @@ let lift_imports = (env, imports) => {
...
+          vid,
           {
             val_repr: ReprFunction(args, rets, _),
             val_fullpath: Path.PIdent(id),
           } as vd,
         ) => {
           switch (Ident_tbl.find_opt(func_map, id)) {
-          | Some({name: Some(name)}) =>
+          | Some(internal_name) =>
+            let external_name = Ident.name(vid);
```
peblair, line 1283: What's the significance of changing to `vid` instead of `id` here (and line 1293)?

compiler/src/codegen/compcore.re
```diff
@@ -3384,53 +3384,51 @@ let compile_imports = (wasm_mod, env, {imports}) => {
...
   let exports = {
-    let exported = Hashtbl.create(14);
+    module StringSet = Set.Make(String);
+    let exported_globals = ref(StringSet.empty);
+    let exported_functions = ref(StringSet.empty);
     /* Exports are already reversed, so keeping the first of any name is the correct behavior. */
     List.filter(
-      ({ex_name}) =>
-        if (Hashtbl.mem(exported, Ident.name(ex_name))) {
+      fun
```
peblair, line 3411: Is this just removing duplicates (this question probably belongs on the other PR)?

peblair (commented):
Looks good (I think); just a couple of questions.

marcusroberts (approved):
I've read the code through, and LGTM but I don't understand some of the deeper concepts yet, so take with a pinch of salt !

compiler/src/codegen/transl_anf.re
ospencer: This way is "guaranteed" to be correct, since it's the name that's included in the module signature rather than the name that's a part of the full path (which is something internal). 1293 was really just a bug 😶 

I guess it wasn't _really_ a bug since it doesn't affect anything, but `id` could potentially be incorrect.

compiler/src/codegen/compcore.re
ospencer: Yeah, that's it. The work it does is a bit different now since the names are already supplied.

peblair (approved):

compiler/src/codegen/transl_anf.re
```diff
@@ -1228,44 +1235,78 @@ let lift_imports = (env, imports) => {
   (imports, setups, env);
 };
 
-let transl_signature = (functions, signature) => {
+let transl_signature = (functions, imports, signature) => {
```
phated, line 1238: Shall we also change these to `~functions` and `~imports`?

compiler/src/codegen/transl_anf.re
```diff
@@ -1228,44 +1235,78 @@ let lift_imports = (env, imports) => {
...
   );
+  List.iter(
+    imp =>
+      switch (imp.imp_shape) {
+      | FunctionShape(_) =>
+        let internal_name =
+          switch (imp.imp_desc) {
+          | GrainValue(mod_, name) => grain_import_name(mod_, name)
+          | WasmFunction(mod_, name) => Ident.unique_name(imp.imp_use_id)
+          | _ => failwith("Wasm or js value had FunctionShape")
```
phated, line 1264: Is this something users can run into? Should we have a more descriptive error?

phated (changes requested):
This is **soooooo** awesome! I looked at a couple snapshots and they look :fire:

I only had a couple comments.

compiler/src/codegen/transl_anf.re
ospencer: Nope! It's not possible, so I'll add an "Impossible" marker to it. We can maybe make the types a little better here in the future to avoid having a case like this (combining imp_desc and imp_shape).

phated (approved):
Approved! Just need CI to turn green

PR #1177, opened 2022-03-22
chore(cli): Clean exe and jsoo code before we build compiler

phated:
In this PR, I'm cleaning all the compiler-generated code before we run the compiler build.

This should solve the codesign sigkill that @ospencer was experiencing on the latest MacOS on M1.

I believe what is happening is that we were overwriting the exe but for some reason MacOS didn't see it as a new file but instead a modification to the old file (inode?). If we delete these exe files before we generate them again, it seems to work fine.

I'm also deleting the `.bc.js` files, too, because I can.

### DIFF
--- a/compiler/package.json
+++ b/compiler/package.json
@@ -15,6 +15,7 @@
     "link": "yarn link",
     "clean": "esy clean",
     "prepare": "esy install",
+    "prebuild": "yarn workspace cli clean",
     "build": "esy",
     "postbuild": "esy copy:exe",
     "build:js": "esy b dune build @js --no-buffer",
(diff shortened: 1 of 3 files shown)


compiler/package.json
```diff
@@ -15,6 +15,7 @@
     "link": "yarn link",
     "clean": "esy clean",
     "prepare": "esy install",
+    "prebuild": "yarn workspace @grain/cli clean",
```
phated, line 18: Reaching across this workspace boundary is gross, but since we **only** want this to run if you do `yarn compiler build` (or `yarn workspace @grain/compiler build`), we need to do it this way 😭

peblair (approved):

Issue #1113, opened 2022-01-20
Syntax: Change -> to => in type signatures


phated:
It's weird that our function type signatures look like `Number -> Number` but writing a function looks like `x => x`. We should consolidate that to using `=>` everywhere.

PR #1178, opened 2022-03-23
chore(compiler)!: Disallow end-of-line before `=>`

phated:
This removes support for newlines before `=>`. This came up while I was trying to work on #1113 with @ospencer. We don't allow newlines before `->`, so menhir got really mad when I switched them to `=>` and Oscar said it would be really hard to resolve.

I started typing out examples of eol before arrow and they looked really terrible:

```grain
let a =
  () => 1
```

```grain
match (a) {
    x
      => 2
}
```

It seems easiest to just remove support for eol before an arrow. I hope y'all agree.

(Is it a chore? Who knows 🤷 )

### DIFF
--- a/compiler/src/parsing/parser.messages
+++ b/compiler/src/parsing/parser.messages
@@ -2354,7 +2354,7 @@ program: FOR WHILE
 
 Expected `(` to set the for loop parameters.
 
-program: FUN ID EOL THICKARROW WHEN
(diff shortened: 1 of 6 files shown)


phated:
I need @ospencer to look at the messages file because I updated it and it seems that it might have been out-of-date from other changes, because it changed a ton.

jozanza (approved):

PR #1181, opened 2022-03-27
chore(compiler): Fix OCaml 4.12.0 and cmdliner warnings

peblair:
Following #1151, the build emits a number of warnings related to Cmdliner's [deprecation of the `Term` evaluation interface](https://github.com/dbuenzli/cmdliner/blob/master/CHANGES.md#new-cmd-module-and-deprecation-of-the-term-evaluation-interface) and OCaml 4.12's [widening of warning 16](https://github.com/ocaml/ocaml/pull/9783). This PR makes the necessary changes to fix those warnings.

### DIFF
--- a/compiler/src/parsing/ast_helper.re
+++ b/compiler/src/parsing/ast_helper.re
@@ -324,7 +324,7 @@ module Top = {
 };
 
 module Val = {
-  let mk = (~loc=?, ~mod_, ~name, ~alias, ~typ, ~prim) => {
+  let mk = (~loc=?, ~mod_, ~name, ~alias, ~typ, ~prim, ()) => {
     let loc = Option.value(~default=Location.dummy_loc, loc);
     {
       pval_mod: mod_,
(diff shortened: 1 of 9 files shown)


compiler/src/parsing/ast_helper.re
```diff
@@ -324,7 +324,7 @@ module Top = {
 };
 
 module Val = {
-  let mk = (~loc=?, ~mod_, ~name, ~alias, ~typ, ~prim) => {
+  let mk = (~loc=?, ~mod_, ~name, ~alias, ~typ, prim) => {
```
ospencer, line 327: I think we like having the labeled arguments everywhere. Would you mind changing this to
```suggestion
  let mk = (~loc=?, ~mod_, ~name, ~alias, ~typ, ~prim, ()) => {
```
in the places you removed the last labeled argument? (The call syntax is still the same as it was before without having to worry about that last unit.)

ospencer: (Actually not all of them, just this one.)

peblair: Done 👍

ospencer (approved):
Looks great!

marcusroberts (approved):
Cmdliner is my nemesis but LGTM!

PR #1183, opened 2022-04-08
feat(compiler)!: Selectively include functions in the global function table

ospencer:
This PR changes the module contract for first-class functions. Previously, modules put all functions in the global function table to allow any function to be called indirectly, just with a reference to the closure. Now, modules that want to be able to call a function indirectly are responsible for putting the function into the function table. This allows us to finally take advantage of tree shaking—Binaryen can completely drop unused functions 🎉 

There are a couple more changes necessary to see the full effect, but with just this change, "hello world" programs are 10% smaller. Some early testing on those other changes has shown close to a 40% module size reduction.

Those other changes include some refactoring of how exports are handled, and completely avoiding closure allocation when it's not needed.

~Draft because we need the fix from https://github.com/grain-lang/binaryen.ml/pull/149, and after that there's one failing test due to https://github.com/WebAssembly/binaryen/issues/4587.~


compiler/src/middle_end/analyze_function_calls.re
```diff
@@ -0,0 +1,80 @@
...
+};
+
+let has_indirect_call = id => {
+  switch (Ident_tbl.find_opt(direct_function_calls, id)) {
+  | Some(count) => count != 0
+  | None => true
+  };
+};
+
+module CSVArg: Anf_iterator.IterArgument = {
```
phated, line 32: What does CSVArg stand for?

compiler/src/typed/translprim.re
```diff
@@ -1571,22 +1571,19 @@ let transl_prim = (env, desc) => {
...
 
+  let value = Typecore.type_expression(env, value);
   let binds = [
     {
-      pvb_pat: {
-        ppat_desc: PPatVar(desc.tvd_name),
-        ppat_loc: loc,
+      vb_pat: {
+        pat_desc: TPatVar(desc.tvd_id, desc.tvd_name),
+        pat_loc: Location.dummy_loc,
```
phated, line 1579: Why is this a dummy_loc?

compiler/src/typed/env.re
```diff
@@ -1870,7 +1874,11 @@ and store_extension = (~check, id, ext, env) => {
       switch (cstr.cstr_args) {
       | [] => ReprValue(WasmI32)
       | args =>
-        ReprFunction(List.map(_ => WasmI32, args), [WasmI32], Unknown)
+        ReprFunction(
+          List.map(_ => WasmI32, args),
+          [WasmI32],
+          Direct(Ident.unique_name(id)),
```
phated, line 1880: Didn't poke around the code, but do we still use `Unknown` in other parts of the code?

phated (approved):
This is awesome! A few questions but here's a preemptive approval!

compiler/src/codegen/compcore.re
```diff
@@ -973,38 +989,31 @@ let compile_imm =
...
     switch (args) {
     | [] => mk_err()
     | _ =>
       let compiled_args = args;
-      Expression.Call_indirect.make(
+      Expression.Call.make(
         wasm_mod,
-        global_function_table,
-        load(~offset=8, wasm_mod, mk_err()),
+        Option.get(err_func_ident),
```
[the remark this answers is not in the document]
ospencer: No, it won't ever fail. `Call` is used when the error is a function, and there's always a func ident in that case (not dynamic, defined a few lines above here).

compiler/src/codegen/compcore.re
```diff
@@ -3598,7 +3635,8 @@ exception WasmRunnerError(Module.t, option(string), string);
 let validate_module = (~name=?, wasm_mod: Module.t) =>
   try(assert(Module.validate(wasm_mod) == 1)) {
   | Assert_failure(_) =>
-    raise(WasmRunnerError(wasm_mod, name, "WARNING: Invalid module"))
+    Module.print(wasm_mod);
```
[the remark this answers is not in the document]
ospencer: That wasn't supposed to be here 😅

compiler/src/typed/env.re
ospencer: Yeah, it's still used.

marcusroberts (approved):
LGTM!

jozanza (approved):

compiler/src/codegen/compcore.re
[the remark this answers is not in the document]
peblair: Still seems a little fragile IMO; I might have rolled the args into the `Some` as well and then used the option for the pattern match

compiler/src/codegen/transl_anf.re
```diff
@@ -1303,6 +1351,55 @@ let transl_signature = (~functions, ~imports, signature) => {
             )
           };
         }
+      | TSigType(tid, {type_kind: TDataVariant(cds)} as td, rs) => {
```
peblair, line 1354: Just so I understand, can you clarify what this is doing? (In GitHub is fine)

compiler/src/middle_end/analyze_function_calls.re
```diff
@@ -0,0 +1,80 @@
...
+let mark_function_call = id => {
+  switch (Ident_tbl.find_opt(direct_function_calls, id)) {
+  | Some(count) => Ident_tbl.replace(direct_function_calls, id, count - 1)
+  | None => ()
+  };
+};
+
+let mark_ident_seen = id => {
+  switch (Ident_tbl.find_opt(direct_function_calls, id)) {
+  | Some(count) => Ident_tbl.replace(direct_function_calls, id, count + 1)
```
peblair, line 20: I think this +1/-1 logic could use a comment (since it's not immediately obvious what this table is tracking)

compiler/src/middle_end/analyze_function_calls.re
```diff
@@ -0,0 +1,80 @@
...
+  | None => ()
+  };
+};
+
+let has_indirect_call = id => {
+  switch (Ident_tbl.find_opt(direct_function_calls, id)) {
+  | Some(count) => count != 0
+  | None => true
```
peblair, line 28: Might warrant a comment why this defaults to `true`

peblair (changes requested):
Two minor requests for comments, but looks good otherwise!

compiler/src/codegen/transl_anf.re
ospencer: Absolutely! It just exports the functions that were created for the types you defined in the module. It won't be necessary anymore when we're not creating functions to create variants.

peblair (approved):

PR #1184, opened 2022-04-09
feat(compiler): Name globals in wasm output

ospencer:
This is more code cleanup than anything else, largely from old logic we needed back before we used Binaryen. Additionally, it'll make it easier for us to debug what's going in in the wasm output since we'll actually see the names of globals getting assigned rather than just `global_0` etc.


peblair (approved):
❤️

marcusroberts (approved):
I can only say LGTM!

jozanza (approved):
🔥

phated (approved):
Love the code cleanup 🎉

PR #1196, opened 2022-04-21
chore: Simplify slim docker image with multistage build

### LINKED ISSUE #1194: Slim docker images are failing
Seems related to building cmake from source. Do we need to add some other tools to the slim dockerfile?

ospencer:
Closes #1194 

Tested this locally—works great (more similar to the regular Dockerfile) and is the same size as before.


phated (approved):
LGTM even though I don't know a lot about docker.

marcusroberts (approved):
Sweet!

PR #1197, opened 2022-04-22
chore(compiler): Upgrade reason-native dependencies

phated:
This is a WIP draft of upgrading our reason-native dependencies to drop the Windows path separator hacks.

There's more work to do and a ton more testing (haven't tried it on Windows yet).


phated:
I tested this on Windows (and the pkg build) and everything seems to work!

PR #1198, opened 2022-04-26
chore(compiler): Add rely resolution to avoid junit

phated:
Currently, Grain `main` doesn't build on Windows. This is because ptime released 1.0.0 which breaks esy's sandbox requirements. Ref https://github.com/dbuenzli/ptime/issues/29

However, we only have the ptime dependency because an older version of rely depended on junit. If we resolve to the WIP 4.0.0 on their `master` branch, it drops this dependency.


ospencer (approved):

jozanza (approved):

marcusroberts (commented):
Nice to see this working on Windows!

PR #1199, opened 2022-04-26
chore: Add yarn global bin to the PATH in CI

ospencer:


phated (commented):
@ospencer why doesn't this get added by setup-node? Seems like a problem that others would hit, right?

ospencer:
Found out that on Windows, `yarn` uses `<prefix>\bin` but `npm` uses just `<prefix>`. I've updated this to just add `yarn global bin` to the PATH both on windows and unix systems instead of trying to match prefixes for npm and yarn.

.github/workflows/ci.yml
```diff
@@ -14,16 +14,26 @@ jobs:
 
     steps:
       - name: Setup Node.js
-        uses: actions/setup-node@v2.1.2
+        uses: actions/setup-node@v3.1.1
```
phated, line 17:
```suggestion
        uses: actions/setup-node@v2.1.2
```
@ospencer let's keep this pinned until we want to do a full upgrade. I think our other workflows still have the v2

phated (approved):
LGTM! Thanks for going down this rabbit hole.

marcusroberts (approved):
Nice to keep the Windows platform equal!

PR #1200, opened 2022-04-27
fix(compiler): Supply correct error for unbound record labels

### LINKED ISSUE #1000: Very unfriendly compiler error
If a user writes this program:

```grain
let foo = ""

foo.charAt(0)
```

Nothing is reported in the LSP, and then when you go to compile the program, the compiler crashes with:

```
grainc.exe: internal error, uncaught exception:
                   (Failure "Impossible: deprecated error type in typetexp")
```

I was only able to debug this by changing the compiler to output the identifier name. How should we handle this better?

ospencer:
Fixes #1000


phated (approved):
Weird to say "record" when the value isn't, but better than what we have now.

marcusroberts (approved):
LGTM!

Issue #972, opened 2021-10-07
Graindoc: Directory support


phated:
We want to be able to run Graindoc recursively on the stdlib directory. I considered writing a script in the package.json file, but I think it would be better if Graindoc was updated to work on a directory if the path isn't a file.

Issue #216, opened 2020-07-13
Compiler: Use Fp.t for type safe filepaths instead of strings


phated:
Currently, almost all of the filepath handling in the compiler is done with strings and those are passed to some methods `Grain_utils.Files` (which will use [`Fp` from Reason-Native](https://github.com/facebookexperimental/reason-native) after #215) and converted back into strings for the output.

I think it would be more type safe to turn any filepaths into `Fp.t` right when they enter the compiler and then work with them throughout.

PR #1206, opened 2022-05-08
chore(compiler): Reimplement readdir with Fs & Fp libraries

phated:
This converts the readdir utility to using the `Fp` and `Fs` libraries. We currently only use this function in our test harness, but I wanted to use it for #972 and working with strings is a nightmare.

We've also had #216 open for a long time, and this is one small step towards that.


ospencer (approved):
Very nice.

marcusroberts (approved):
Great improvement!

PR #1207, opened 2022-05-08
fix: Preserve all configs when compiling

phated:
While working on #972, I noticed that the compiler crashed while trying to do docgen on multiple files, specifically when it encountered files in the runtime directory.

@ospencer did some digging and found that the flags at the top of some runtime files were being set globally in the compiler and then failing on the next compile target. By adding `Grain_utils.Config.preserve_all_configs` around any compile, the flags will always be reset to the flags specified by the run instead of including flags set by a file. (Oscar can let us know if I messed up that description).


ospencer:
I just realized something with this that's quite annoying. For all intents and purposes this definitely solves our problem, but it breaks the "resumable" functionality of `compile_resume` since some of the config that gets set during compilation would be washed away if you did a partial compile and tried to resume it later. This isn't something that we actually do (or really even care to do)—we really only use that functionality to stop compilation at a certain point.

IMO I think we just add a `compile.rei` so that `compile_resume` isn't exposed from the module, so there's no expectation that compilations are resumable outside of `compile.re`. Eventually when we do a nice config refactor, we could expose `compile_resume` again.

WDYT?

phated:
I'm fine with that. Want me to add a "breaking" commit on this branch?

ospencer:
Yeah, that makes sense.

ospencer (approved):
Very nice!

PR #1208, opened 2022-05-08
fix(graindoc): Avoid singletons when building ordered comments

phated:
While working on #972, I found that the singleton used for ordering the comments of a file doesn't work when you try to handle multiple files. It would inject more comments in the map, which could cause bugs or throw errors on multiple module definitions, etc.

This switches the implementation to using a first-class module `OrderedComments` which is then passed to the functions that operate on comments.


ospencer (approved):

marcusroberts (approved):
Nice work

PR #1210, opened 2022-05-09
chore(compiler): Rework filepath utilities

phated:
This builds upon #1206 and reworks some of our use of filepaths to use the `Fp.t` type more. It also moves some filesystem accessing functions into the correct module.

I wanted to make it very explicit when we were operating on string filepaths, so you now need to reach into the `Filepath.String` module to use those.


ospencer (approved):
Looks good.

PR #1211, opened 2022-05-09
feat(compiler)!: Refactor enum constructors

### LINKED ISSUE #1021: Printing records defined in different modules prints an opaque record
From my chat with @ospencer

* Variants/records are created with the module's id so at runtime we can look up the type info.
* Variants are created from the file in which they're defined (because we export functions)
* Records are not which means if you create a record in a different file than which it's defined, it has the module id of that module instead of the other one

This is going be a much large fix because the fix is to refactor it completely.

_Originally posted by @phated in https://github.com/grain-lang/grain/pull/1019#discussion_r743330243_

### LINKED ISSUE #1502: Compiler: Well Formedness Upper Case Types
The parser should read in uppercase types as PTyConstr instances.

ibdafna:
Signed-off-by: Itay Dafna <i.b.dafna@gmail.com>

Closes #1021 
Closes #1502 

This PR changes how enum variant constructors work in Grain, making individual variants handled separately from functions. This change allows us to build variant types into the language rather than just the standard library, which will allow us to provide better error messages, allow for disambiguation of variants from multiple types with the same name, and open up opportunities for features like optional arguments. The tradeoff is that the names of variant constructors are no longer functions, but this feature was only seldom used (only appearing once in Grain's entire standard library). Wrapper functions can be used to provide the same functionality.

Before:
```
List.map(Some, [1, 2, 3])
```

After:
```
List.map(x => Some(x), [1, 2, 3])
```


compiler/src/codegen/compcore.re
```diff
@@ -70,15 +70,6 @@ let panic_with_exception_ident =
...
   Ident.create_persistent("GRAIN$EXPORT$panicWithException");
-let assertion_error_ident = Ident.create_persistent("AssertionError");
-let assertion_error_closure_ident =
-  Ident.create_persistent("GRAIN$EXPORT$AssertionError");
-let index_out_of_bounds_ident =
-  Ident.create_persistent("GRAIN$EXPORT$IndexOutOfBounds");
-let index_non_integer_ident =
-  Ident.create_persistent("GRAIN$EXPORT$IndexNonInteger");
-let match_failure_ident =
-  Ident.create_persistent("GRAIN$EXPORT$MatchFailure");
```
ospencer, line 81: There are no identifiers or closures for variants anymore.

compiler/src/codegen/compcore.re
```diff
@@ -90,6 +81,7 @@ let console_mod = "console";
 let tracepoint_ident = Ident.create_persistent("tracepoint");
 
 let grain_main = "_gmain";
+let grain_type_metadata = "_gtype_metadata";
```
ospencer, line 84: Type metadata is now explicitly provided by a special type metadata function in the module. This allows modules to choose to define only types but still have type information loaded.

compiler/src/codegen/compcore.re
```diff
@@ -1015,45 +958,233 @@ let compile_imm =
...
-let call_error_handler = (wasm_mod, env, err, args) => {
-  let (err_value_ident, err_func_ident) =
-    switch (err) {
-    | Runtime_errors.MatchFailure => (match_failure_ident, None)
-    | Runtime_errors.IndexOutOfBounds => (index_out_of_bounds_ident, None)
-    | Runtime_errors.IndexNonInteger => (index_non_integer_ident, None)
-    | Runtime_errors.AssertionError => (
-        assertion_error_closure_ident,
-        Some(get_wasm_imported_name(exception_mod, assertion_error_ident)),
+/** Heap allocations. */
```
ospencer, line 961: None of this has changed, it's just moved higher so we can do allocations for exceptions (which was previously just calling a function).

compiler/src/codegen/compcore.re
```diff
@@ -2484,6 +2388,9 @@ let compile_prim1 = (wasm_mod, env, p1, arg, loc): Expression.t => {
         Expression.Unreachable.make(wasm_mod),
       ],
     )
+  | Assert => failwith("Unreachable case; should never get here: Assert")
+  | BuiltinId =>
```
ospencer, line 2392: This `BuiltinId` will be how `print` will be able to know about built-in types.

compiler/src/linking/link.re
```diff
@@ -4,24 +4,9 @@ open Grain_typed;
...
-
-module G = Imperative.Digraph.Concrete(String);
-module Topo = Topological.Make(G);
 
-let dependency_graph = G.create(~size=10, ());
 let modules: Hashtbl.t(string, Module.t) = Hashtbl.create(10);
-
-let grain_main = "_gmain";
-let grain_start = "_start";
+let hard_dependencies: Hashtbl.t(string, unit) = Hashtbl.create(10);
```
ospencer, line 9: Hard dependencies are ones that require code to be loaded from the module, whereas soft dependencies just require type information be loaded.

compiler/src/linking/link.re
```diff
@@ -301,17 +285,25 @@ let rec globalize_names = (~function_names, ~global_names, ~label_names, expr) =
 };
 
 let table_offset = ref(0);
-let module_id = ref(0);
+let module_id = ref(Comp_utils.encoded_int32(0));
```
ospencer, line 288: This was a really sad bug that took away a week of my life.

compiler/src/linking/link.re
```diff
@@ -578,24 +608,26 @@ let link_all = (linked_mod, dependencies, signature) => {
...
-  G.add_vertex(dependency_graph, main_module);
-  build_dependency_graph(
+  load_hard_dependencies(
     ~base_dir=Filepath.String.dirname(main_module),
     main_module,
   );
-  let dependencies =
-    Topo.fold((dep, acc) => [dep, ...acc], dependency_graph, []);
+
+  let dependencies = Module_resolution.get_dependencies();
```
ospencer, line 629: This gives us the full dependency graph, which includes files that might not contain necessary code.

compiler/src/typed/builtin_types.re
```diff
@@ -21,12 +21,11 @@ open Path;
 open Types;
 open Btype;
 
-let builtin_idents = ref([]);
-let builtin_decls = ref([]);
+let builtin_idents = Hashtbl.create(64);
```
ospencer, line 24: The hash lets us look these up quickly versus having to traverse the list.

stdlib/runtime/string.gr
```diff
@@ -46,42 +46,33 @@ enum StringList {
 primitive _RUNTIME_TYPE_METADATA_PTR: WasmI32 = "@heap.type_metadata"
 
 @unsafe
-let findTypeMetadata = (moduleId, typeId) => {
+let findTypeMetadata = typeId => {
```
ospencer, line 49: This function is now quite a bit less efficient since it traverses _all_ type information. It should be converted to something faster like a binary tree or a hash, though it's tricky to implement because the metadata itself lives in runtime land where arbitrary allocations are not available. Regardless, I felt that optimization didn't need to be included in this PR and could come separately.

compiler/src/linking/link.re
phated: Can you explain it a bit?

compiler/src/parsing/parser.mly
```diff
@@ -389,6 +392,10 @@ app_expr:
 rcaret_rcaret_op:
   | lnonempty_list(RCARET) RCARET { (String.init (1 + List.length $1) (fun _ -> '>')) }
 
+construct_expr:
+  | type_id lparen lseparated_list(comma, expr) comma? rparen { Exp.construct ~loc:(to_loc $loc) $1 $3 }
+  | type_id %prec LPAREN { Exp.construct ~loc:(to_loc $loc) $1 [] }
```
phated, line 397: what's the difference between `lparen` and `LPAREN`?

phated (commented):
Thanks for the comments inline, they helped a ton! I had one question about the parser (and I'll probably spend a little more time looking at snapshots).

compiler/src/linking/link.re
ospencer: As defined in the data representations doc, module IDs are always tagged and thus GC-able. Except during linking, the linker assigned non-tagged IDS 🤦‍♂️

This wasn't a problem before because the IDs were treated by the runtime as untagged, but some updates while working on this PR caused them to be treated as regular values and made the bug show up.

compiler/src/parsing/parser.mly
ospencer: `LPAREN` is the literal token `(`, and `lparen` is the non-terminal rule we defined to handle them. It's got some extra rules around whitespace.

ospencer: But in this context we're specifying that the reduction should happen with the same precedence as encountering an LPAREN token, in this case instead shifting the LPAREN to try to parse a constructor with arguments instead of reducing a constant constructor.

ospencer: So tl;dr when the parser reads `Some` and the next token is `(`, keep parsing a constructor instead of immediately saying the entire constructor is just `Some`.

marcusroberts (approved):
I had a good read through this - nice to see list constructors cleaned up as well and looks like some nice changes as well as a deep bug fixed.  Really nice work!   

Hopefully with the new code owner rules I can approve this releasing it as a lot of the changes are beyond my current comprehension but I liked the changes I did understand and the code changes read well.

ospencer:
Note to self: need to update `store_type` and `store_extension` in env.re to stop adding constructors as binding names to the environment

ospencer:
`store_type` and `store_extension` updated.

phated (approved):
Love this!

jozanza (approved):

PR #1213, opened 2022-05-11
chore(graindoc): Implement input & output args with a Filepath module

phated:
This is a refactor that I wanted to implement to share input/output arguments between our different CLI executables. This is important because I'll be implementing directory support that will be shared between graindoc and grainformat.

What do y'all think of this approach?


ospencer (approved):
Looks really good!

marcusroberts (approved):
I like this approach

PR #1215, opened 2022-05-11
chore(compiler): Translate default optimization passes from binaryen

phated:
This is a translation of Binaryen's default optimization passes. By not relying on the defaults, we can adjust this to speed up compilation times. The primary reason for doing a 1-to-1 translation was to ensure that everything was behaving correctly and can be seen by the snapshots not changing at all.


ospencer (approved):

marcusroberts (approved):
I know it’s translated but very readable and comprehensible

PR #1219, opened 2022-05-13
Refactor suggestions

phated:
@marcusroberts These are some refactors that I think we should make to the formatter code. It really cuts down on the different code paths and mutable values. `Grain_utils.Fs_access` isn't the *perfect* place for the EOL utils, but I think it's fine for now.


marcusroberts (approved):
Oh, very nice indeed!

marcusroberts:
I like it, that's a great improvement to the code.

PR #1223, opened 2022-05-14
chore: Don't raise signals in tests

ospencer:
It's not very cross-platform (crashes on Windows) and we probably shouldn't raise random signals.


phated (approved):
Thanks for finding this issue! ❤️

marcusroberts (approved):

PR #1224, opened 2022-05-14
chore(stdlib): Update docs in Set module

spotandjake:
Update set to match the wording used in the Map documentation.

PR #1225, opened 2022-05-14
fix(stdlib)!: Add explicit void return type on Set.forEach

spotandjake:
This explicitly annotates the return type of `Set.forEachBucket` as `Void` to ensure the signature is correct.


stdlib/set.gr
```diff
@@ -243,7 +243,7 @@ let rec forEachBucket = (fn, node) => {
   match (node) {
     None => void,
     Some({ key, next }) => {
-      fn(key)
```
phated, line 246: @spotandjake Please add the `@history` docs entry too 🙏

phated: Docs were added in #1224

phated (approved):

marcusroberts (approved):
Nice!

ospencer (approved):

PR #1226, opened 2022-05-15
chore!: Switch from yarn to npm

phated:
Relies on #1092

While working on Windows CI for Grain, I found out that the yarn global path is not available to our test harness for some reason (no idea why). By switching to npm, the grain binary will correctly be available to powershell.exe that is executed from our test runner (upcoming PR).

I was planning to make this change as soon as we dropped node 14 support, as npm 8 has workspace support and then users don't need to worry about having yarn installed.

The one downside to this change is that all of our `yarn` commands are now `npm run`. (don't 🔫  me)


marcusroberts:
I had the same on Windows.  The Grain node CLI was installed here

C:\Users\marcus\AppData\Local\Yarn\bin

which Yarn bin doesn't report as a location.

marcusroberts (approved):
I have no great love for yarn and needing to install yarn after node is just another step, so I'll throw you an approval!

ospencer (approved):
Nothing a few aliases can't fix 🙂

peblair (approved):

PR #1229, opened 2022-05-15
chore(ci): Enable native tests on Windows

phated:
This gets our test harness "child process" runner to work in a Windows CI. It also enables the native tests on Windows.

Shoutout to @ospencer who did a ton of work investigating different shells and ways to call into them. I stole a lot of code from his `oscar/test-spawn` branch.

We're still working to get the JS tests working on Windows, so that step is still skipped.

### DIFF
--- a/compiler/test/runner.re
+++ b/compiler/test/runner.re
@@ -105,46 +105,33 @@ let extract_anf = ({cstate_desc}) =>
 let compile_string_to_final_anf = (name, s) =>
   extract_anf(compile_string(~hook=stop_after_optimization, ~name, s));
 
-let run = (~num_pages=?, file) => {
-  let cli_flags = "-g";
-  let cli_flags =
-    switch (num_pages) {
-    | Some(x) =>
-      Printf.sprintf(
-        "%s --initial-memory-pages %d --maximum-memory-pages %d",
-        cli_flags,
-        x,
-        x,
-      )
-    | None => cli_flags
-    };
+let open_process = args => {
+  // We need to run the tests in powershell on Windows to have the correct environment
+  let shell = Sys.win32 ? "powershell.exe" : "/usr/bin/env";
 
-  let stdlib = Option.get(Grain_utils.Config.stdlib_dir^);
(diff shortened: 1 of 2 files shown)


compiler/test/runner.re
```diff
@@ -105,46 +105,33 @@ let extract_anf = ({cstate_desc}) =>
...
-        "%s --initial-memory-pages %d --maximum-memory-pages %d",
-        cli_flags,
-        x,
-        x,
-      )
-    | None => cli_flags
-    };
+let open_process = args => {
+  // We need to run the tests in powershell on Windows to have the correct environment
+  let shell = Sys.win32 ? "powershell.exe" : "/usr/bin/env";
```
ospencer, line 110: `shell` feels a little weird since `/usr/bin/env` isn't really a shell. Maybe `program`?

ospencer (commented):
Looks great! Thanks for bringing this over the finish line 🎉 

One little question but otherwise this is awesome.

compiler/test/runner.re
phated, line 110:
```suggestion
  let program = Sys.win32 ? "powershell.exe" : "/usr/bin/env";
```

compiler/test/runner.re
```diff
@@ -105,46 +105,33 @@ let extract_anf = ({cstate_desc}) =>
...
-    file,
-  ];
-  let command = String.concat(" ", args);
+  // Powershell doesn't exit with the script's exit code so we need to do this
+  let exit = Sys.win32 ? [|";", "exit", "$LastExitCode"|] : [||];
 
   let (stdout, stdin, stderr) =
-    Unix.open_process_full(command, Unix.environment());
+    Unix.open_process_args_full(
+      shell,
```
phated, line 120:
```suggestion
      program,
```

compiler/test/runner.re
phated: Done!

ospencer (approved):
Excellent!

marcusroberts (approved):
Good stuff!

PR #1231, opened 2022-05-16
chore(compiler): Implement JS externals for Windows readdir

phated:
This implements some missing JS externals for Windows related to `Unix.readdir`. I didn't realize they had separate externals when I first implemented them. I also needed to get one fix to the Unix external.

This also enables the JS tests on Windows so we catch things like this before they land. One change that needed to happen to make them work was to ensure our snapshots always have `LF` line endings (because JSOO doesn't actually handle `\r\n` translations for `open_in` on Windows.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Autodetect text files
+* text=auto
+
+# Ensure snapshots always use LF
+*.snapshot text eol=lf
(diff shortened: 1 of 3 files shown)


ospencer (approved):
Woo!

marcusroberts (approved):
Excellent!

PR #1232, opened 2022-05-16
chore(compiler): Remove depends from dune-project stanzas

phated:
These don't actually do anything since we exclusively use esy, and (for that reason) we are very bad at updating them. We should just delete them. The only thing dune needs for the dune-project stanzas is the name, but a description is nice too.


ospencer (approved):

peblair (approved):
🎉

marcusroberts (approved):

PR #1233, opened 2022-05-17
chore(compiler): Move type string printing into Printtyp module

phated:
This moves the type-to-string utilities that I wrote for graindoc into Printtyp (where they really should be). I noticed that #1131 copy-pasted my utilities. Instead of duplicating this logic, we should be putting them in the correct location.

I also experimented with `string_of_type_sch` and `string_of_type_scheme` and I **think** `string_of_type_sch` does the same thing as my weird `string_of_type_expr` function. I'll have a more thorough testing of that once directory support for graindoc is finished.


marcusroberts (approved):
Nice cleanup!

ospencer (approved):

marcusroberts (approved):

PR #1234, opened 2022-05-19
chore(compiler): Remove Config.base_path entirely

phated:
While doing some Windows debugging, I noticed that we don't actually use `Grain_utils.Config.base_path` anywhere anymore. I believe there was a change where we made things relative to the file being compiled, and this was probably just leftover.


marcusroberts (approved):
Good spot!

ospencer (approved):

PR #1235, opened 2022-05-20
New LSP refactor

phated:
@marcusroberts I wanted to propose these refactors against your `new_lsp` branch. I wanted to set this new LSP up with good patterns that should enable us to rapidly add features.

A few things are going on here:
1. I moved the library code into a `grain_language_server` package (inside of `src`). We need to stop dumping a bunch of bespoke code into the executable directories, as it is a pain to use by other parts of the compiler.
2. I updated a bunch of the code to better take advantage of `[@deriving]` annotations. We can actually deserialize the JSON by using the code generate by deriving yojson, and we can enumerate a ADT by deriving enum.
3. I renamed a lot of files. I didn't go all-in on the naming scheme that I thought would be best (I'd like to see `text_document_did_open.re`, etc) but I tried to better align the files with the messages they are handling. I even added `shutdown.re` and `exit.re` even though they are currently simple.
4. I moved a **lot** of types around. We should be trying to keep the types located with the logic that operate on the types. If something like the request parameters for completions are only used in that module, then they shouldn't be placed into a different module. It just makes things more convoluted.
5. To go with (4), I added interface files to every module. This helps to better understand your APIs and notice when you are leaking abstractions because you can make your types abstract.

There are still improvements to be made, but we can keep making things better as we go along. One thing that I think we need to finish before shipping this is replacing your "log file" logic with proper LSP messages for logging, as I noticed that it is part of the spec.

I also snuck one JS fix in, though we are still having a different issue with the JS LSP on Windows.


marcusroberts (approved):
Well, that's a fine lesson in taking hacky code and making it production ready and beautiful.

Awesome stuff, a *huge* improvement and will make adding extra features easier for me and others.

PR #1237, opened 2022-05-21
Handle more data types and fix printing

phated:
While working on a fix for type printing, I ran into an issue where the LSP wasn't printing types, records, etc as nicely as it could. This applies some fixes to make it nicer.

One thing that I still didn't solve is that `enum` declarations are being printed as `type`. I'm wondering if they are being marked as abstract somehow (cc @ospencer ?)

### DIFF
--- a/compiler/src/diagnostics/modules.re
+++ b/compiler/src/diagnostics/modules.re
@@ -3,6 +3,9 @@ open Grain_typed;
 type export_kind =
   | Function
   | Record
+  | Variant
+  | AbstractType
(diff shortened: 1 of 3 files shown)


phated:
These are close to what ocaml-lsp uses: https://github.com/ocaml/ocaml-lsp/blob/6a38d66c57930c6ef5ac67d740a51e2d9551f93f/ocaml-lsp-server/src/compl.ml#L17

marcusroberts (approved):

PR #1238, opened 2022-05-21
fix(graindoc): Fix type printing for types and abstract types

### LINKED ISSUE #1236: Abstract types fail to print with Printtyp
While helping with the new LSP, I noticed that the type aliases in Regex crash it. This is due to them not being implemented in Printtyp.

phated:
Fixes #1236

This actually implements the prefix when printing abstract types or really any type definition.

Is the `Otyp_constr` correct?


ospencer (approved):
`type` is correct for `Otyp_constr`.

marcusroberts (approved):

PR #1241, opened 2022-05-21
fix(graindoc): Use value_descriptions and type_declarations defined by the module signature

ospencer:
We previously did a lookup on the types in the module's signature, but that type is the original type rather than what the module is trying to expose. This change uses what's defined in the signature.


phated (approved):
Amazing! Thank you so much for digging into that for me 🙇  I love that we can show types like `Map<k, v>` now!

marcusroberts (approved):
Excellent!

PR #1242, opened 2022-05-22
Rework LSP logger to use tracing

phated:
@marcusroberts I spent a bunch of time messing with the vscode plugin to understand the tracing (and why it wasn't showing up). I will send a PR against your pure_client in addition to this.

The gist of this is that the LSP actually sends the `trace` configuration upon initialize, so we don't need a `--debuglsp` flag. We also need to listen for the `$/setTrace` message to adjust the trace logging level.

I didn't use the `~verbose` parameter at all in the conversion, but I think that would be used if we want to send a bunch of json down with the trace messages. That's what the vscode tracing seems to do.


marcusroberts:
@phated  this looks great.  I'd like to retain the ability to send window/logMessage notifications too though?

marcusroberts (approved):
Looks great!

PR #1243, opened 2022-05-22
fix(compiler): Read custom sections in full

ospencer:
Previously we relied on OCaml's `input` to read the custom section from a module. If all the bytes were not read, we skipped the section entirely. This swaps to `really_input` to ensure we read the entire section.

### DIFF
--- a/compiler/src/utils/wasm_utils.re
+++ b/compiler/src/utils/wasm_utils.re
@@ -460,11 +460,8 @@ module BinarySection =
           /* Now we're at the start of the section. Time to read */
           let realsize = size - (pos_in(inchan) - offset);
           let bytes = Bytes.create(realsize);
-          if (input(inchan, bytes, 0, realsize) == realsize) {
-            Some(Spec.deserialize(bytes));
-          } else {
-            process(tl);
-          };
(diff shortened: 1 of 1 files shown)


marcusroberts (approved):

compiler/src/utils/wasm_utils.re
```diff
@@ -460,11 +460,8 @@ module BinarySection =
           /* Now we're at the start of the section. Time to read */
           let realsize = size - (pos_in(inchan) - offset);
           let bytes = Bytes.create(realsize);
-          if (input(inchan, bytes, 0, realsize) == realsize) {
-            Some(Spec.deserialize(bytes));
-          } else {
-            process(tl);
-          };
+          really_input(inchan, bytes, 0, realsize);
```
phated, line 463: Is it okay that this will raise if we can't read it all?

phated (approved):
Just one question. Preemptive approval 👍

compiler/src/utils/wasm_utils.re
ospencer: Indeed! If all the bytes aren't there (when the module reported that they were) means the module is corrupted.

PR #1244, opened 2022-05-22
feat(compiler): Refactor exports

ospencer:
This PR refactors how exports are compiled in Grain. Namely,

- Exports are now fully signature-driven. When determining what to export from the module, we iterate the module signature (rather than collecting things to export along the way).
- A scoping bug that put exported names in scope is now resolved. For example,
```grain
let foo = 5
export foo as bar
print(bar)
```
will now properly result in a type error, whereas previously `bar` was unexpectedly in scope.
- Re-exports from one module to another now incur no runtime cost whatsoever.

Because of these changes, the hello world module size is reduced again by 25%, coming in under 39k.


compiler/src/codegen/transl_anf.re
```diff
@@ -532,9 +513,6 @@ let run_register_allocation = (instrs: list(Mashtree.instr)) => {
   |> run(Types.StackAllocated(WasmF64));
 };
 
-let grain_import_name = (mod_, name) =>
-  Printf.sprintf("gimport_%s_%s", mod_, name);
-
 let wasm_import_name = (mod_, name) =>
```
phated, line 516: Is this one still used?

compiler/src/codegen/transl_anf.re
```diff
@@ -1088,15 +1063,15 @@ let lift_imports = (env, imports) => {
       ) => {
     switch (imp_desc) {
     | GrainValue(mod_, name) =>
-      let mimp_mod = Ident.create_persistent(mod_);
-      let mimp_name = Ident.create_persistent(name);
-      let import_name = grain_import_name(mod_, name);
+      let mimp_mod = mod_;
+      let mimp_name = name;
```
phated, line 1067: Why not rename the bindings in the pattern match?

compiler/src/codegen/transl_anf.re
```diff
@@ -1156,21 +1143,22 @@ let lift_imports = (env, imports) => {
...
+              MGlobalBind(Ident.unique_name(imp_use_id), alloc),
               env.ce_binds,
             ),
         },
       );
     | WasmValue(mod_, name) =>
-      let mimp_mod = Ident.create_persistent(mod_);
-      let mimp_name = Ident.create_persistent(name);
+      let mimp_mod = mod_;
+      let mimp_name = name;
```
phated, line 1153: Why not rename the bindings in the pattern match?

compiler/src/middle_end/optimize_dead_assignments.re
```diff
@@ -14,21 +17,61 @@ let safe_to_remove_import = i =>
   | _ => false
   };
 
+let is_used = ident =>
+  Option.is_some(Ident.find_same_opt(ident, used_symbols^));
```
phated, line 21: We should add some utilities to the Ident module (like `contains`). Probably not needed in this PR.

compiler/src/typed/types.re
```diff
@@ -141,6 +141,7 @@ type value_description = {
   val_type: type_expr,
   val_repr,
   val_kind: value_kind,
+  val_internalpath: Path.t,
```
phated, line 144: Does this remove the need to use `Env.find_value` in #1241 ??

compiler/test/__snapshots__/basic_functionality.0c400bde.0.snapshot
```diff
@@ -5,21 +5,21 @@ basic functionality › modulo6
...
  (export \"memory\" (memory $0))
  (export \"_gmain\" (func $_gmain))
  (export \"_start\" (func $_start))
  (export \"GRAIN$TABLE_SIZE\" (global $GRAIN$TABLE_SIZE))
  (func $_gmain (; has Stack IR ;) (result i32)
-  (call $gimport_pervasives_%
-   (call $wimport_GRAIN$MODULE$runtime/gc_incRef
-    (global.get $wimport_GRAIN$MODULE$runtime/gc_GRAIN$EXPORT$incRef)
-    (global.get $gimport_pervasives_%)
+  (call $%_1131
```
phated, line 19: Feels really weird that Pervasives is just `%` + a random number now.

phated (commented):
This PR is 🔥  - I had a few questions/thoughts

compiler/src/codegen/compcore.re
```diff
@@ -51,11 +51,11 @@ let swap_slots =
 let module_runtime_id = Ident.create_persistent("moduleRuntimeId");
 let reloc_base = Ident.create_persistent("relocBase");
 let table_size = Ident.create_persistent("GRAIN$TABLE_SIZE");
-let grain_env_mod = Ident.create_persistent(grain_env_name);
-let malloc_mod = Ident.create_persistent("GRAIN$MODULE$runtime/malloc");
-let gc_mod = Ident.create_persistent("GRAIN$MODULE$runtime/gc");
-let exception_mod = Ident.create_persistent("GRAIN$MODULE$runtime/exception");
-let console_mod = Ident.create_persistent("console");
+let grain_env_mod = grain_env_name;
```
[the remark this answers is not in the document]
ospencer: I like it because it's a bit more intentional than using `grain_env_name` directly.

compiler/src/codegen/compcore.re
```diff
@@ -51,11 +51,11 @@ let swap_slots =
...
+let malloc_mod = "GRAIN$MODULE$runtime/malloc";
+let gc_mod = "GRAIN$MODULE$runtime/gc";
+let exception_mod = "GRAIN$MODULE$runtime/exception";
+let console_mod = "console";
```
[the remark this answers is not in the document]
ospencer: It's used in the JS runner... I kind of want to get rid of it (but that's for another time).

compiler/src/codegen/transl_anf.re
ospencer: Yeah, it's still used.

compiler/src/typed/types.re
ospencer: Nah, that's more about locations.

compiler/test/__snapshots__/basic_functionality.0c400bde.0.snapshot
ospencer: Just to be clear, this is the mod operator that's been imported from Pervasives, which definitely looks a little weird but that's just the name of the function.

The important difference is that imports aren't special-cased anymore—they're assigned regular idents just like everything else (which makes them much easier to find in the env to do things like say you want to export one 😄)

compiler/src/codegen/compcore.re
```diff
@@ -70,7 +70,7 @@ let malloc_ident = Ident.create_persistent("malloc");
 let malloc_closure_ident = Ident.create_persistent("GRAIN$EXPORT$malloc");
 let incref_ident = Ident.create_persistent("incRef");
 let incref_closure_ident = Ident.create_persistent("GRAIN$EXPORT$incRef");
-let equal_mod = Ident.create_persistent("GRAIN$MODULE$runtime/equal");
+let equal_mod = "GRAIN$MODULE$runtime/equal";
```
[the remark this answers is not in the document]
ospencer: I'm going to reorganize them so they make sense.

phated (approved):
Approved! 🎉

compiler/src/middle_end/linearize.re
```diff
@@ -1557,6 +1521,33 @@ let rec transl_anf_statement =
...
+  | TTopExport(exports) =>
+    List.iter(
+      ({tex_path}) => {
+        let {val_fullpath, val_type, val_repr} =
+          Env.find_value(tex_path, env);
+        switch (val_fullpath) {
+        | Path.PExternal(Path.PIdent(mod_) as p, ident, _) =>
+          let allocation_type = get_allocation_type(env, val_type);
+          let mod_decl = Env.find_module(p, None, env);
+          // Lookup external exports to import them into the module
```
peblair, line 1533: I'm a little confused what this is doing. Can you clarify a little more?

peblair (commented):
I mostly follow this, and it looks good to me. One very minor request for information, and then I think it's good to go.

compiler/src/middle_end/linearize.re
ospencer: Yeah! So the way imports actually make it into the WebAssembly module is via `lookup_symbol` registering the individual values and functions to include them in the module. Previously doing `export foo` was sugar for `export let foo = foo`, making `foo` get used and thus be included as an import, but now an `export` statement skips the extra binding. The lookup here is effectively us saying that we plan to use that value so it actually gets imported.

PR #1245, opened 2022-05-23
Remove fallible string operations in LSP

phated:
I found that the LSP was crashing a lot because it was trying to do operations on strings that were outside of the string boundaries. @marcusroberts we have the `String_utils.slice` utility so you should never be using `String.sub`. And I added a `char_at` utility that returns an option.


marcusroberts (approved):
Nice find and solution!

PR #1246, opened 2022-05-23
Fix hover highlighting by using the correct language

phated:
In talking with @ospencer about what the purpose of "grain-type" syntax in our vscode extension, he helped me notice that the syntax highlighting on our hover tooltips was wrong. We need to use `grain-type` syntax for any type signatures (which all hover items should be).


marcusroberts (approved):
Nice find!

PR #1249, opened 2022-05-23
Fix shutdown/exit by making request params optional

phated:
I noticed that the restart was crashing because it couldn't parse params on the request.


marcusroberts (approved):
Well spotted!

PR #1251, opened 2022-05-24
fix(compiler): Report Win32 instead of Cygwin in JS compiler

phated:
The JS version of the compiler kept crashing because it was unable to locate the stdlib. This was due to JSOO reporting itself as "Cygwin" instead of "Win32". We've pushed for a lot of upstream changes in `Fp` so we use the correct separators on Windows, but that doesn't happen if it thinks it is running in Cygwin.

This hack (placed in `Grain_utils` so pretty much everything will get it applied) will swap us back to reporting `Win32` and make the all our tools work.

### DIFF
--- a/compiler/src/utils/filepath.re
+++ b/compiler/src/utils/filepath.re
@@ -26,6 +26,8 @@ let derelativize = (~base=?, fname: Fp.firstClass) => {
   };
 };
 
+// All uses of `Filename` from OCaml should be constrained to this file because we need to do
+// normalization on the filepaths those functions produce.
 module String = {
   // This module is converting strings into Fp.t and then back into Strings
   // TODO(#216): We should consider switching to type safe Fp.t where ever filepaths are used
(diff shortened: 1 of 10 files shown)


ospencer (approved):

marcusroberts (approved):

phated:
Looks like this might be causing a bug in the relative import resolution. I'll need to debug before we can land it.

compiler/src/utils/filepath.re
```diff
@@ -85,6 +87,44 @@ module String = {
...
+
+  // TODO(#216): We should consider switching to type safe Fp.t where ever filepaths are used
+  let dirname = path => {
+    normalize_separators(Filename.dirname(path));
+  };
+
+  // TODO(#216): We should consider switching to type safe Fp.t where ever filepaths are used
+  let basename = Filename.basename;
+
+  // TODO: This is poorly named
```
ospencer, line 114: Do we want to just rename now?

ospencer (approved):
New version looks great!

compiler/src/utils/filepath.re
phated: Yeah. Any suggestions? It's like `non_module_relative_path`? 🤣

ospencer: Could go the OCaml approach and name it `is_really_relative` 😛

phated: God no

phated: Let's skip for now. We probably want to change the way modules are specified when we completely change the module system.

PR #1253, opened 2022-05-24
Fix highlighing of Module name when hovering

marcusroberts:
Recalculate the location to highlight just the module name


phated:
Pinged @marcusroberts in DM's about an edge case here.

marcusroberts:
Updated to fix the edge case

PR #1255, opened 2022-05-24
chore(cli): Add NEAR smoketest

### LINKED ISSUE #1160: Add smoketest for runtimes with limited features
We currently use the multivalue feature of Binaryen to manipulate the wasm stack via tuples, but don't utilize any multivalue functions. Because the flag for multivalue is turned on, Binaryen _may_ perform an optimization that would introduce a multivalue function/type. This test would verify that this doesn't occur using a runtime that doesn't support multivalue, like Node 14 or the NEAR runner.

jozanza:
Add test for a hello world NEAR smart contract to ensure multivalue functions are not in wasm output

Closes #1160 

Closes https://github.com/grain-lang/near-simple-counter/issues/4


phated:
> Thanks! It's really too bad that NEAR is MVP only for Wasm. I'll make a proposal soon to expand, but I'm sure it'll be met with resistance.

Honestly, their whole ecosystem is a mess. If you notice the failed CI runs, it shows that the runner doesn't even install on Windows...

phated:
> https://github.com/near/workspaces-js has been updated and improved for stability

This is what we are using

phated:
Rebased this on #1261 so I changed the base until that lands

ospencer (approved):

phated (approved):
Thanks @jozanza !

PR #1258, opened 2022-05-25
Hack to make autocompletion work

marcusroberts:
Here's a hack to add checking for context.

Just for @phated  to look at, not for merging


phated:
Included in #1262 but we might end up removing it when we get a better answer.

PR #1260, opened 2022-05-25
Fix module hovers in the LSP

marcusroberts:


phated:
There seems to still be an issue with this:

When I hover "expect" of `Result.expect`, it only highlights "Result":

Screen Shot 2022-05-26 at 3 35 00 PM

phated (approved):
This solved the module hover. I believe more fixes are in a different PR.

PR #1261, opened 2022-05-25
fix(compiler): Refactor WASI polyfill resolution

### LINKED ISSUE #1257: The `--wasi-polyfill` flag fails to work if you specify a file inside a directory
In #1255, @jozanza used `pushd` and `popd` because the command failed to work. I investigated why (because those aren't available in our linux CI) and it turns out that grain compile just fails to work if the `--wasi-polyfill` flag contains a directory, e.g. `--wasi-polyfill __test__/wasi.gr`.

ospencer:
We previously tried to resolve the location of the wasi polyfill relative to the Grain program being compiled, which is incorrect—it's relative to the directory where `grainc` is invoked.

Closes #1257

### DIFF
--- a/compiler/src/linking/link.re
+++ b/compiler/src/linking/link.re
@@ -47,16 +47,17 @@ let is_grain_module = mod_name => {
 };
 
 let wasi_polyfill_module = () => {
-  "GRAIN$MODULE$./"
-  ++ Filename.remove_extension(Option.get(Config.wasi_polyfill^));
+  Filename.remove_extension(Option.get(Config.wasi_polyfill_path()))
+  ++ ".gr.wasm";
 };
 
 let is_wasi_module = mod_name => {
   mod_name == "wasi_snapshot_preview1";
 };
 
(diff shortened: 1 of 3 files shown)


phated (approved):
Rebased #1255 on this and it works! Thanks

marcusroberts (approved):

peblair:
@ospencer Does this work with `// grainc-flags`?

peblair (changes requested):
Looks good, but see question in my previous comment before we merge

jozanza (approved):

ospencer:
@peblair It has the same limitations as `--stdlib` in that the relative path is relative to the `grain` command being invoked rather than to the file (which is only maybe a limitation depending on how you look at it).

We can't delete that interface since it's still used. We could remove it and refactor the code that uses it, but computing the full path is somewhat expensive vs the times we want to just check if it's defined. I felt like it was fine to keep since we also expose the underlying value for `stdlib_dir` as well.

peblair (approved):

PR #1262, opened 2022-05-26
feat(lsp): Add code completion

phated:
We decided to remove this from the rewrite of the LSP server in #1131, so I wanted to extract the current implementation details into a draft.

This is also an interesting view into how a new feature will be added!


marcusroberts:
I think Oscar's work for hover will help here too wrt module exports.

phated:
Closing in favor of #1900

PR #1263, opened 2022-05-27
feat(graindoc): Allow directory input & output

### LINKED ISSUE #972: Graindoc: Directory support

phated:
Closes #972

This adds a first pass at directory support in graindoc. I regenerated the stdlib to show it in action.

I'm guessing we will have some back-and-forth on the error messages (or if those should be error cases), so I wanted to get this open before the weekend.

This logic should be portable to grainformat once we decide it looks good.

### DIFF
--- a/compiler/graindoc/graindoc.re
+++ b/compiler/graindoc/graindoc.re
@@ -4,7 +4,7 @@ open Compile;
 open Grain_typed;
 open Grain_utils;
 open Grain_diagnostics;
-open Filename;
+open Grain_utils.Filepath.Args;
 
 let () =
   Printexc.register_printer(exc =>
@@ -28,13 +28,17 @@ let () =
   );
 
 [@deriving cmdliner]
-type params = {
+type io_params = {
   /** Grain source file for which to extract documentation */
   [@pos 0] [@docv "FILE"]
-  input: Filepath.Args.ExistingFile.t,
+  input: ExistingFileOrDirectory.t,
   /** Output filename */
   [@name "o"] [@docv "FILE"]
-  output: option(Filepath.Args.MaybeExistingFile.t),
+  output: option(MaybeExistingFileOrDirectory.t),
+};
+
+[@deriving cmdliner]
+type params = {
   /**
     The version to use as current when generating markdown for `@since` and `@history` attributes.
     Any future versions will be replace with `next` in the output.
@@ -43,10 +47,10 @@ type params = {
   current_version: option(string),
 };
 
-let compile_typed = (opts: params) => {
-  let input = Filepath.to_string(opts.input);
-
-  switch (Compile.compile_file(~hook=stop_after_typed, input)) {
+let compile_typed = (input: Fp.t(Fp.absolute)) => {
+  switch (
+    Compile.compile_file(~hook=stop_after_typed, Filepath.to_string(input))
+  ) {
   | exception exn =>
     let bt =
       if (Printexc.backtrace_status()) {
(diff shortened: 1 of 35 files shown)


compiler/graindoc/graindoc.re
```diff
@@ -222,13 +223,87 @@ let generate_docs =
...
+    ])
+  | (File(input_file_path), Some(NotExists(output_file_path))) =>
+    `Ok([
+      {input_path: input_file_path, output_path: Some(output_file_path)},
+    ])
+  | (Directory(_), None) =>
+    `Error((
+      false,
+      "Directory input must be used with `-o` flag to specify output directory",
+    ))
```
ospencer, line 272: Some other tools just print _all_ the contents to stdout, but that feels kinda silly.

compiler/graindoc/graindoc.re
```diff
@@ -222,13 +223,87 @@ let generate_docs =
...
+  | (Directory(input_dir_path), Some(Exists(Directory(output_dir_path)))) =>
+    `Ok(enumerate_directory(input_dir_path, output_dir_path))
+  | (Directory(input_dir_path), Some(NotExists(output_dir_path))) =>
+    `Ok(enumerate_directory(input_dir_path, output_dir_path))
+  | (File(input_file_path), Some(Exists(Directory(output_dir_path)))) =>
+    `Error((
+      false,
+      "Using a file as input cannot be combined with directory output",
+    ))
```
ospencer, line 281: I'd just create the file in that directory in this case.

compiler/graindoc/graindoc.re
phated: Yeah, I thought about it but it seems weird. I suppose that'd be one way to "support" outputting all docs from a directory into a single file `grain doc stdlib > docs.md` 🤔

compiler/graindoc/graindoc.re
phated: I started implementing that before I submitted this and it has some **very** non-intuitive behavior.

Since the directory specified is handled recursively, we calculate the "base path" of the contents, such that:
`grain doc stdlib -o foo` will produce a tree containing:
```
$ tree -d foo
foo
├── runtime
│   ├── unsafe
│   └── utils
└── sys
```
This allows us to run the command with the input and output directory as the same path:
`grain doc stdlib -o stdlib` and it will output each markdown file next to the grain file.

Now, if we implement what you are suggesting, the behavior becomes non-intuitive because you can't calculate a base path:
`grain stdlib/runtime/unsafe/constants.gr -o stdlib` would need to create `stdlib/constants.md` because we don't have a base path. You could say that it should use the `-o` directory as the base path, and that would work if the input and output are in the same directory, but what if you wanted it to go somewhere else?
`grain stdlib/runtime/unsafe/constants.gr -o foo` can't use `foo` as a base path, otherwise it'll create a `stdlib` directory inside of it, which is different behavior than directory + directory.

I don't like any of these non-intuitive behaviors and decided to disallow it with this error.

phated: The only way I could see this being supported **and** intuitive is if we add a `--recursive` flag which can only be used with directory input. But I really don't want that extra flag.

ospencer: Yup, that sounds reasonable.

ospencer (approved):
No real complaints from me!

marcusroberts (approved):
I think your arguments for this approach are good.

PR #1264, opened 2022-05-27
Add formatting support

marcusroberts:
Moved the formatting implementation to a Grain module
Call the formatter directly from the LSP
Advertise formatting as available to the LSP client


marcusroberts:
There's a crazy weird bug, don't merge this!

compiler/src/language_server/formatting.re
```diff
@@ -0,0 +1,135 @@
...
+    Option.iter(
+      s =>
+        if (Grain_utils.Config.debug^) {
+          prerr_string("Backtrace:\n");
+          prerr_string(s);
+          prerr_string("\n");
+        },
+      bt,
+    );
+    `Error((false, "Failed to parse code"));
```
phated, line 75: This code won't need to be in the LSP. It is only for the cmdliner stuff

compiler/src/language_server/formatting.re
```diff
@@ -0,0 +1,135 @@
...
+
+      Trace.log("ready to return a formatted result");
+      let res: ResponseResult.t = [{range, newText: formatted_code}];
+      Protocol.response(~id, ResponseResult.to_yojson(res));
+
+    | `Error(_) =>
+      Trace.log("Error formatting");
+      Protocol.error_response(
+        ~id,
+        ~code=-32700,
```
phated, line 130: Do we need an enumeration of these? Where is the list of error codes?

marcusroberts: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#errorCodes

Yes, an enumeration would be ideal

compiler/src/language_server/formatting.re
```diff
@@ -0,0 +1,94 @@
...
+      ~cached_code: Hashtbl.t(Protocol.uri, Typedtree.typed_program),
+      ~documents: Hashtbl.t(Protocol.uri, string),
+      params: RequestParams.t,
+    ) => {
+  switch (Hashtbl.find_opt(documents, params.text_document.uri)) {
+  | None =>
+    Trace.log("The code requested isn't available on the server");
+    Protocol.error_response(
+      ~id,
+      ~code=0,
```
phated, line 53: Need correct error codes (and an enumeration). I'll work on that

compiler/src/language_server/formatting.re
```diff
@@ -0,0 +1,116 @@
...
+        let formatted_code =
+          Format.format_ast(~original_source=lines, ~eol, parsed_program);
+
+        let range: Protocol.range = {
+          range_start: {
+            line: 0,
+            character: 0,
+          },
+          range_end:
+            // Use Int32.max_int to ensure we fit the entire number in JSON
```
phated, line 73: In the LSP branch, I want to change these range types to only work with Int32 so we make sure we don't wrap on large ints

marcusroberts (commented):
That's all come together nicely!

phated (approved):
Merging into the LSP branch

PR #1266, opened 2022-05-27
fix(compiler) Fix the LSP to use the new exports

marcusroberts:


phated (approved):
Looks like it compiles, let's get it into the LSP branch.

PR #1267, opened 2022-05-27
feat(compiler)!: Make optimization level two the default

ospencer:
This PR makes the default optimization level for Grain programs level two of three, which runs Grain optimization passes but not Binaryen (wasm-opt) optimizations. This change optimizes (pun intended) for developer experience—much faster rebuilds during development, but we can still get all of the extra optimizations when it's time to ship to production.

On my machine with this change, rebuilds take 1s in JS and 120ms natively. (And for those curious, linking accounts for 50% of that time in JS, and 33% natively).

Also in this PR is a new `--release` flag for production—currently the equivalent of passing `-O3` to enable all optimizations, but in the future may be extended.


ospencer:
Going to close this because I'm just going to remove the optimization levels entirely.

PR #1269, opened 2022-05-27
chore(ci): Allow esy cache import to fail without killing run

phated:
The esy cache import seems to fail on Windows more than I'd like. Let's just make this step fallible in our workflows so it doesn't crash the run (it'll just take longer).


ospencer (approved):

marcusroberts (approved):
I've seen so many CI crashes, this is great!

PR #1270, opened 2022-05-27
feat(compiler)!: Replace optimization levels with compilation profiles

### LINKED ISSUE #214: CI: Add test phase with optimizations disabled
Related to #213 - we should run the test suite with optimizations disabled as a separate phase of our CI. This would help us catch bugs in non-optimized codegen.

### LINKED ISSUE #648: grain compile very slow
On Linux Mint, I installed grain using the command line from the docs which got me `grain` 0.3.1. When I run `grain demo.gr` or `grain compile demo.gr` with a file containing only `print("helo world")`, some `*.wasm` files are produced under `target/`, but then the executable is <del>seemingly stuck and does not print anything to the terminal.</del>  <ins>silent for around 10 minutes or so and only then prints out the message; re-running the same command on an unchanged file still takes almost a minute. Is this to be expected?</ins>

### LINKED ISSUE #1114: Investigate slowness in binaryen optimizations
The binaryen optimizations take a huge chunk of our compile times. We need to investigate which ones and see if we can drastically reduce this; otherwise, I think we should make our default optimization level be `-02`

ospencer:
This PR replaces compiler optimization levels with compilation profiles, i.e. higher levels of optimization can be achieved via `grain compile --release` instead of having to worry about what optimization levels are.

`grainc` gains a new `--profile` flag, while the `grain` cli provides shorthands (`--release` is a shorthand for `--profile=release`).

As such, this PR also disables Binaryen (wasm-opt) optimizations by default, but when compiling for production, `--release` can be used to do extensive optimizations. This change optimizes (pun intended) for developer experience—much faster rebuilds during development, but we can still get all of the extra optimizations when it's time to ship to production.

On my machine with this change, `grain compile hello.gr` (a simple hello world program) rebuilds in 1s in JS and in 120ms natively, down roughly **75%** in JS vs when full optimizations are enabled. 

New docs will be provided on the website to make this behavior clear, and it'll be announced in the v0.5 blog post!

Closes #214
Closes #648 
Closes #1114


phated (approved):
Thanks for taking my advice on ditching the `-O` flag 🙇  Looking great and sets us up for a nicer silo integration!

marcusroberts (approved):
Oooh, feels so grown up!

cli/bin/grain.js
```diff
@@ -123,6 +138,10 @@ program
     "--elide-type-info",
     "don't include runtime type information used by toString/print"
   )
+  .profileOption(
+    "--release",
+    "produce WebAssembly modules suitable for production"
```
peblair, line 143: Not sure how I feel about `suitable for production`. We don't want to just say `with the release profile`?

compiler/src/compile.re
```diff
@@ -165,13 +165,7 @@ let next_state = (~is_root_file=false, {cstate_desc, cstate_filename} as cs) =>
...
     | TypedWellFormed(typed_mod) =>
       Linearized(Linearize.transl_anf_module(typed_mod))
-    | Linearized(anfed) =>
-      switch (Grain_utils.Config.optimization_level^) {
-      | Level_one
-      | Level_two
-      | Level_three => Optimized(Optimize.optimize_program(anfed))
-      | Level_zero => Optimized(anfed)
-      }
+    | Linearized(anfed) => Optimized(Optimize.optimize_program(anfed))
```
peblair, line 168: So, to make sure I understand, we have no way of disabling optimizations with this change?

compiler/test/__snapshots__/arrays.0f9e7d37.0.snapshot
```diff
@@ -4,7 +4,9 @@ arrays › array_access
  (type $none_=>_i32 (func (result i32)))
  (type $none_=>_none (func))
  (import \"_grainEnv\" \"mem\" (memory $0 0))
+ (import \"_grainEnv\" \"tbl\" (table $tbl 0 funcref))
```
peblair, line 7: So, these changes are just because the binaryen optimizations are disabled, right?

peblair (approved):
LGTM, but I left a couple of questions for my own understanding

compiler/src/compile.re
ospencer: Not the Grain ones, anyway! We've flirted with the idea of removing that choice anyway because Grain programs don't work without the Grain optimizations 😅 

We may have some sort of debug mode for development (though passing compiler hooks via the command line would be a lil nicer I think).

compiler/test/__snapshots__/arrays.0f9e7d37.0.snapshot
ospencer: Yup yup.

cli/bin/grain.js
ospencer:
```suggestion
    "compile using the release profile (production mode)"
```

PR #1271, opened 2022-05-28
fix(compiler): Report binop locations properly

### LINKED ISSUE #1129: Locations of rational literals are reported incorrectly
image

ospencer:
This problem is supposed to be solved by Menhir's `$sloc`, but that only works in the current rule. We'd essentially need to inline all of the right side of binops to get the location reporting right, but this is much simpler.

Closes #1129


phated (approved):

marcusroberts (approved):

peblair (approved):
Nice approach 👍

PR #1272, opened 2022-05-28
feat(compiler): Convert Sys libraries to @unsafe

### LINKED ISSUE #1146: Import only necessary WASI APIs
Currently, Grain seems to emit a wasm binary that imports almost all WASI APIs, regardless of whether they are actually used or not. This means that on a WASM runtime that does not support all WASI APIs, the Grain app will not run, even if the Grain app is actually calling only the WASI APIs supported by that runtime.

Is it possible to improve Grain so that it imports only the APIs that are actually used, instead of always importing all WASI APIs?

I'd also like to note that WASI is making big changes on its API[^1][^2], so reducing imports will also benefit future compatibility.

For example,

```
import Process from "sys/process"
Process.exit(1)
```

This code only calls the `proc_exit` WASI API through Grain's standard library, but the emitted wasm module imports almost all WASI APIs.

```wat
  (import "wasi_snapshot_preview1" "fd_write" (func (;0;) (type 2)))
  (import "wasi_snapshot_preview1" "args_get" (func (;1;) (type 0)))
  (import "wasi_snapshot_preview1" "args_sizes_get" (func (;2;) (type 0)))
-- snip --
  (import "wasi_snapshot_preview1" "path_rename" (func (;36;) (type 11)))

```

[^1]: FS APIs will have big changes https://github.com/WebAssembly/WASI/issues/469#issuecomment-1045251844
[^2]: Some APIs will be removed, such as `proc_raise`, which Grain always imports. https://github.com/WebAssembly/WASI/issues/26

ospencer:
Closes #1146 as this removes the sys functions' self-references, which allows them to be dropped from the module if unused.

### DIFF
--- a/stdlib/sys/file.gr
+++ b/stdlib/sys/file.gr
@@ -1,4 +1,3 @@
-/* grainc-flags --no-gc */
 /**
  * @module Sys/File: Utilities for accessing the filesystem & working with files.
  *
@@ -56,6 +55,7 @@ export enum LookupFlag {
 }
 
 // TODO(#775): This has specific ordering requirements because of ambiguous type inference
+@unsafe
 let rec combineLookupFlagsHelp = (acc, dirflags) => {
   match (dirflags) {
     [hd, ...tl] => {
@@ -67,6 +67,7 @@ let rec combineLookupFlagsHelp = (acc, dirflags) => {
     [] => acc,
   }
 }
+@unsafe
 let combineLookupFlags = dirflags => {
   combineLookupFlagsHelp(0n, dirflags)
 }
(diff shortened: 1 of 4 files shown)


marcusroberts (approved):
Looks good!

stdlib/sys/file.gr
```diff
@@ -738,144 +728,69 @@ export let rec fdDatasync = (fd: FileDescriptor) => {
...
-let orderedFdflags = wasmSafeCons(
-  Append,
-  wasmSafeCons(
-    Dsync,
-    wasmSafeCons(Nonblock, wasmSafeCons(Rsync, wasmSafeCons(Sync, [])))
-  )
-)
-let orderedRights = wasmSafeCons(
+let orderedFdflags = [Append, Dsync, Nonblock, Rsync, Sync]
+let orderedRights = [
```
phated, line 745: Thank god!

phated (approved):
Nice!

PR #1273, opened 2022-05-28
fix hover for record declarations and definitions

marcusroberts:
Look inside records properly to highlight the parts inside


marcusroberts:
I didn't go down the Kept rabbit hole yet, looking for some advice on that one

marcusroberts:
No longer needed after @ospencer's fine work

PR #1274, opened 2022-05-28
feat(grainfmt): Allow directory input & output

### LINKED ISSUE #973: Grainfmt:  Directory support
Inspired by #972 the formatter should recursively format all Grain files in a directory if directory instead of a file is specified

phated:
Closes #973

This adds directory input & output support for grain format. The implementation is very close to the graindoc implementation and they can probably be consolidated in the future.

The big breaking changes to support this are:
1. We no longer have an `--in-place` flag. Instead, you need to specify the same file as the "output", e.g. `grain format test.gr -o test.gr`
2. We no longer support input on stdin. The stdin support was primarily to support the old way our vscode plugin operated, so this shouldn't be an issue. (Side note: we should theoretically be able to support stdin, but ppx_deriving_cmdliner crashes when we try to make a positional argument optional 🤷 https://github.com/hammerlab/ppx_deriving_cmdliner/issues/51 )

I ran the directory formatting against the stdlib and the only unformatted file was Regex. So I suggest y'all review that file to look for complaints/bugs with the formatter (and we'll probably want to file those separately).

### DIFF
--- a/cli/bin/exec.js
+++ b/cli/bin/exec.js
@@ -83,10 +83,12 @@ function execGrainformat(
   execOpts = { stdio: "pipe" }
 ) {
   const flags = [];
-  const options = program.opts();
-  program.options.forEach((option) => {
+  // Inherit compiler flags passed to the parent
+  const options = program.parent.options.concat(program.options);
+  const opts = { ...program.parent.opts(), ...program.opts() };
+  options.forEach((option) => {
     if (!option.forward) return;
-    const flag = option.toFlag(options);
+    const flag = option.toFlag(opts);
     if (flag) flags.push(flag);
   });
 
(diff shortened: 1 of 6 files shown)


marcusroberts (approved):
Looks great!  The new LSP formatter doesn’t need stdin as you say so happy to see that go for directory support instead.

regex format looks good too

peblair:
> breaking changes 
@phated Should this be `feat(grainfmt)!:`?

peblair (approved):
Looks good (though see earlier comment), but on the squash can you make sure there is some mention of reformatting Regex, since this will touch the `git blame` for most of that file?

phated:
> Should this be feat(grainfmt)!:?

The specific breaking changes are marked as breaking, so they will show up. All will be included in the squash.

> can you make sure there is some mention of reformatting Regex, since this will touch the git blame for most of that file?

Will do!

PR #1275, opened 2022-05-28
chore(cli)!: Remove graceful flag & behavior

phated:
This removes the `--graceful`/`-g` flag and removes the "graceful" exit behavior. I don't actually think this is necessary/useful and it just makes boilerplate more cluttered.

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -34,21 +34,6 @@ function graincVersion() {
   return exec.grainc("--version", program).toString().trim();
 }
 
-function wrapAction(action, logError = false) {
-  return (...args) => {
-    try {
-      return action(...args);
-    } catch (e) {
-      if (logError) console.error(e);
-      if (program.opts().graceful) {
-        process.exit();
-      } else {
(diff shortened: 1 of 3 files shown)


marcusroberts (approved):
Looks like an improvement to me!

ospencer (approved):

peblair (approved):
This looks good, but @phated can you maybe put a blurb here on GH for why this doesn't break the tests? My understanding was we needed `-g` for the test suite.

phated:
> can you maybe put a blurb here on GH for why this doesn't break the tests? My understanding was we needed -g for the test suite.

I actually don't know this for sure. I believe it is a flag from a bygone era that we don't need since @ospencer refactored the process spawning. Now, it not only collects stdout and stderr, but it also checks status codes. I'll double check with him to ensure this is the correct answer.

ospencer:
Yup, that's correct! In fact, `-g` isn't even used by `grain run` anymore somehow (which is what the test runner uses). Should be safe to remove 👍

PR #1276, opened 2022-05-28
chore(cli): Reduce boilerplate

phated:
This is committed on top of #1275 because I noticed areas for cleanup after I removed that code. This just reduces a whole bunch of boilerplate that we don't need to deal with.


marcusroberts (approved):
Looks good.     Will need to update the LSP branch as we call the LSP code differently now

ospencer (approved):

PR #1280, opened 2022-05-29
fix(graindoc): Improve location lookup so re-exports do not crash it

phated:
The exports refactor & bug fixes caused some of my bad code to break in graindoc. Since no new bindings are created in scope for re-exports, I can't use `Env.find_value` to find the locations anymore. @ospencer walked me through the better way to do this, but I need help to finish the value lookup (see TODO in diff).


peblair:
@phated It should be good to go now.

peblair (approved):

ospencer (approved):

PR #1281, opened 2022-05-29
feat(lsp): Precise hover locations

ospencer:


marcusroberts (commented):
I love this!

marcusroberts (approved):
Transformational!    I bet a huge improvement in performance too, and I bet a load of hidden bugs just got swept away too.

PR #1284, opened 2022-05-29
fix(lsp): Report non-instanced type for identifiers

ospencer:
This reports the original type of an identifier rather than the instanced type. For example, before this change, the LSP would report the type of `print` as `_weak -> Void`, and if hovering `print` in `print(5)`, `Number -> Void`, when it should show `a -> Void` as the instanced is a bit misleading.

The instanced type is correct for other expressions.

A good example of this is `Map.make`. Creating one without setting any values should show the instanced type as being weak as you haven't set any values to define some concrete types:

image

But hovering over just `make` should show the original type, since hey, you're probably going to make another map some day and you'd be pretty confused if the tool told you it only makes weak maps.

image

And the same deal for non-weak maps:

image

image

### DIFF
--- a/compiler/src/language_server/hover.re
+++ b/compiler/src/language_server/hover.re
@@ -90,6 +90,10 @@ let module_lens = (~program: Typedtree.typed_program, p: Path.t) => {
   grain_code_block(String.concat("\n", signatures));
 };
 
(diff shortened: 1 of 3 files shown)


phated (approved):
Didn't follow it all, but the behavior in the screenshot looks good.

PR #1285, opened 2022-05-29
fix(cli)!: Show all global options within help for every command

phated:
Closes #1080

This PR reworks our commander stuff in the CLI so we show all global options for the root/default command **and** all subcommands. It also allows subcommands to specify specific options they need, which you'll notice with the different `-o` options throughout the CLI (some that require only files and some that allow directories or files).

This is a pretty hefty PR, so I hope I didn't miss anything. Please give it a thorough review and test combinations of flags with various commands. I tested with `--release` and `--no-link` during development.

This PR is marked as breaking because the changes caused the tests to fail, which made me realize we were generating this command `grain -S /path/to/stdlib -I /path/to/test/includes run test-file.gr`. We no longer allow options before the subcommands because otherwise our CLI thinks you are running the root command and that `run` is supposed to be a file.

Here are what some of the `--help` usages output:
Screen Shot 2022-05-29 at 4 38 39 PM

Screen Shot 2022-05-29 at 4 38 27 PM

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -12,7 +12,7 @@ const v8 = require("v8");
  */
 v8.setFlagsFromString("--experimental-wasm-return-call");
 
-const program = require("commander");
+const commander = require("commander");
 const exec = require("./exec.js");
 const run = require("./run.js");
 
@@ -26,11 +26,7 @@ function num(val) {
   return Number.parseInt(val, 10);
 }
 
-function graincVersion() {
-  return exec.grainc("--version", program, { stdio: "pipe" }).toString().trim();
-}
-
-class ForwardOption extends program.Option {
+class ForwardOption extends commander.Option {
   // A ForwardOption is forwarded to the underlying program
   forward = true;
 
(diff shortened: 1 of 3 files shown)


marcusroberts:
I like this approach!

cli/bin/grain.js
```diff
@@ -71,144 +79,165 @@ const optionApplicator = (Option) =>
...
-  .action(function (wasmFile) {
-    // The run subcommand inherits all options of the
-    // top level grain command
-    run(wasmFile, program.opts());
-  });
+  .addOption(new commander.Option("-p, --print-output").hideHelp())
+  .action(run);
 
 program
   .command("lsp <file>")
```
marcusroberts, line 230: Will need to remember to update this for the new LSP

ospencer (approved):

marcusroberts (approved):

cli/bin/grain.js
phated: Yeah, I've been doing the merge and resolves

peblair (approved):

PR #1289, opened 2022-05-30
feat(lsp): Display type declarations on hover

ospencer:
This PR displays the declaration of any type that has a declaration.

image
image


marcusroberts:
I find it confusing when it shows me the gory inner workings of Pervasives though

image

As a user do I need to know how bool is implemented.  Or is that actually very cool?

marcusroberts (approved):
Lovely work!

PR #1290, opened 2022-05-31
feat(ci): Link all versions during release

phated:
This PR upgrades the `release-please` tool we use to v13 (v3.x of the action), which requires us to switch up the uploader actions to the one I built for libbinaryen and binaryen.ml.

The upgrade was primarily to switch from the `node-workspace` plugin to the `linked-versions` plugin so the version of every package is exactly the same. We still need to update the versions inside our `@grain/cli` dependencies, so I'm using the `extra-files` feature of release-please to update those versions, which means we no longer use a caret on those versions.

I encountered various problems when updating all of this:
1. npm v7+ broke their `filename` property in `npm pack` output: https://github.com/npm/cli/issues/3405
2. release-please doesn't work with some combination of `grouped-pull-request-title`, `linked-versions` and `node-workspace`: https://github.com/googleapis/release-please/issues/1456
3. release-please doesn't work with both `linked-versions` and `node-workspace`, requiring us to use `extra-files`: https://github.com/googleapis/release-please/issues/1457

You can see an example of the "linked-versions" PR at https://github.com/phated/grain/pull/2 (notice the various "Synchronize Grain versions" changelog entries).

And you can see an example of a full release at https://github.com/phated/grain/actions/runs/2411992148 (the docker dispatch failed because I disabled the docker workflow in my branch)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,15 +8,12 @@ jobs:
     name: Create Release
     runs-on: ubuntu-latest
     outputs:
-      release_created: ${{ steps.release.outputs.release_created }}
-      upload_url: ${{ steps.release.outputs.upload_url }}
+      releases_created: ${{ steps.release.outputs.releases_created }}
       tag_name: ${{ steps.release.outputs.tag_name }}
-      stdlib_upload_url: ${{ steps.release.outputs.stdlib--upload_url }}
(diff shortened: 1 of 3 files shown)


ospencer (approved):
Absolutely amazing work!

marcusroberts (approved):
Great deep dive to fix all these issues!

PR #1291, opened 2022-05-31
chore(cli)!: Simplify version output

phated:
This builds upon #1285 and needs #1290 

Since #1290 ensures all versions are the same across the monorepo, we no longer need to output different versions for the CLI & compiler.

Simplifying the output so it only prints the version allows us to run `grain doc stdlib -o stdlib --current-version=$(grain -v)` in CI. See https://github.com/grain-lang/grain-lang.org/pull/289


ospencer (approved):
Perfecto!

PR #1296, opened 2022-06-03
fix(ci): Add a last-release-sha to release-please

phated:
I was writing docs for the List stdlib module, and I found a really old issue number linked into our release notes. I believe the new version of release-please isn't finding the correct release PR, so we should try to force it with this.


phated:
(Actually, let's call this a fix so we know it'll pick it up).

ospencer (approved):

marcusroberts (approved):

PR #1298, opened 2022-06-03
chore(stdlib): Add docs to List module

phated:
While prepping for release, I noticed that the List module didn't have any graindoc added to it. With the automated docs updates on the website, this would wipe out our List documentation.

I took most of this from the Array module and updated it for the differences. I also noticed some inconsistencies in Array that I'll PR after this.

Please thoroughly review the docs for clarity and consistency (especially with our other docs).

PR #1300, opened 2022-06-04
chore(stdlib)!: Remove `sum` function from the List module

phated:
In #1298, @ospencer said we should remove `List.sum`. This removes the function before 0.5 since it is a breaking change.


marcusroberts (approved):

ospencer (approved):

peblair (approved):
➕ 🙅‍♂️

PR #1302, opened 2022-06-04
fix(graindoc): Remove spaces between parens and infix idents

phated:
While working on the Pervasives docs, I noticed that we printed infix operators like `( + )` but all our website docs were written as `(+)`. So this removes the extra spaces around it.


Neo-wise (approved):

marcusroberts (approved):

ospencer (approved):
Good catch!

PR #1314, opened 2022-06-06
fix(compiler): Properly handle corrupted CMIs

ospencer:
I was able to reproduce with a hex editor, and with this change the file is recompiled as expected.


ospencer:
Yes yes.

PR #1320, opened 2022-06-11
feat(compiler): Support record field updating assignment

### LINKED ISSUE #1318: Parser syntax error on += with record
Given the following code, I receive this parser error:
> Syntax error after 'idx' and before '+='.
> Expected an infix operator to continue the expression or a newline character to terminate it.

```grain
record Foo {
    mut idx: Number
}

let foo = { idx: 1}

foo.idx += 1
```

I believe the `+=` operator should allowed to be used on mutable record fields.

ospencer:
Closes #1318

### DIFF
--- a/compiler/src/formatting/format.re
+++ b/compiler/src/formatting/format.re
@@ -2348,15 +2348,13 @@ and print_expression =
         print_ident(txt),
       ])
     | PExpRecordSet(expression, {txt, _}, expression2) =>
-      Doc.concat([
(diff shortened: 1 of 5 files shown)


phated:
Looks like this needs parser message updates?

ospencer:
Done.

phated (approved):
Looking great! Thanks for digging into this ❤️

marcusroberts (approved):

peblair (approved):

PR #1322, opened 2022-06-12
fix(grainfmt): Support PPatOr and PPatAlias

marcusroberts:
The formatter now correctly formats Or and Alias patterns.

@ospencer and @phated  notice the second example from the blog has the OR formatted onto a single line as it fits.  
This follows the formatter rules, but may not be best style.   We could look to force a new line for an Or operator applied to lists?

### DIFF
--- a/compiler/src/formatting/format.re
+++ b/compiler/src/formatting/format.re
@@ -1308,7 +1308,9 @@ and print_pattern =
     | PPatOr(pattern1, pattern2) => (
         Doc.group(
           Doc.concat([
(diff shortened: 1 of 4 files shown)


ospencer:
Could you add some examples with line breaks? Similar to infix operators like `&&`, the `|` has to appear on the same line as the first part of the pattern.

marcusroberts:
I'm glad you asked as I realized independently I hadn't covered that. I'll add a fix to this pr and a test

ospencer (approved):

peblair (approved):
Looks great!

PR #1323, opened 2022-06-12
feat(compiler): Allow newline in or-patterns

ospencer:
This PR allows
```
match (true) {
  true
  | false => print("ok")
}
```
in addition to
```
match (true) {
  true |
  false => print("ok")
}
```


phated (approved):
While I hate the look of that, I actually like that we can type it (or copy it from ocaml) and then the formatter will move it to the end-of-line 😛

marcusroberts (approved):
Looks good!

PR #1324, opened 2022-06-12
fix(compiler): Support printing tuple types in oprint

ospencer:
This was crashing the LSP when I wrote `type Tup = (String, String)`. Also noticed it didn't print singleton tuples correctly.

### DIFF
--- a/compiler/src/typed/oprint.re
+++ b/compiler/src/typed/oprint.re
@@ -380,6 +380,9 @@ and print_out_type_2 = ppf =>
         print_typlist(print_simple_out_type, ","),
         tyl,
       );
+      if (List.length(tyl) == 1) {
+        pp_print_char(ppf, ',');
+      };
       pp_print_char(ppf, ')');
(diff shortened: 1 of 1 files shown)


phated (approved):
Thanks! I'm feeling like making all these NYI was a big mistake on my part 🤦

phated (changes requested):
Actually, can you add a formatter test?

ospencer:
@phated I don't think the formatter uses Oprint (and I believe we have tuple tests). Did you mean Graindoc?

phated:
> Did you mean Graindoc?

No, I didn't. We don't have graindoc tests. I think you are right that the formatter doesn't use oprint (though it should).

I'll switch back to approving this.

phated (approved):
Please open an issue to write a regression test.

marcusroberts (approved):

peblair (approved):

PR #1328, opened 2022-06-12
fix(lsp): Always respond to hover request

ospencer:
When we don't have a successful compilation to check, this will tell vscode "I heard you but I don't have an answer for you" so it doesn't think it's waiting on us and display a "Loading..." message.


phated (approved):
Since we can't handle cancelRequest, this is the next best thing. 👍

marcusroberts (approved):
Nice fix!

PR #1329, opened 2022-06-13
fix(compiler): Properly handle types not imported

ospencer:


marcusroberts (approved):
Looks like a good solution to me based on your commentary in Discord.

peblair (approved):

PR #1330, opened 2022-06-13
feat(compiler): Use data segments for string data

ospencer:


compiler/src/codegen/compcore.re
```diff
@@ -3562,10 +3628,6 @@ let compile_wasm_module = (~env=?, ~name=?, prog) => {
   );
   validate_module(~name?, wasm_mod);
 
-  switch (Config.profile^) {
-  | Some(Release) => Optimize_mod.optimize(wasm_mod)
-  | None => ()
-  };
```
ospencer, line 3568: This change is to address a Binaryen limitation. I'm not sure that we want to merge with it.

ospencer: Essentially, if optimizations are run, data segments can be removed. Since there is no way to get the name of a segment, the linker is unable to match `memory.init` instructions to the segment they refer to. We get around this by just skipping the optimizations here and relying on the names being the same as the indices at link time, and optimizing the fully linked module at the end. It appears that we get slightly better results by including this optimization here.

ospencer: I opened https://github.com/WebAssembly/binaryen/issues/5984 to track the update needed in Binaryen.

spotandjake (approved):
I really like how much cleaner the outputted code is with this change.

spotandjake:
I am pretty sure this was unblocked with #2104 as we no longer lose the debug information when linking.

ospencer:
Rebased this. Looks like it's largely working with one test failure that needs to be looked into. I also don't like how the data segments are handled; they're lumped in with the type metadata. That'll need to be cleaned up.

PR #1336, opened 2022-06-14
feat(compiler): Allow compilation of empty programs

### LINKED ISSUE #1335: Compiling an empty Grain file should work
As it says on the tin. Ideally, it produces the smallest wasm module possible, `(module)`, but not a requirement.

ospencer:
Closes #1335 

When compiled in release mode, the resulting module is 51 bytes 😄


phated (approved):
🤣

PR #1337, opened 2022-06-14
chore(stdlib): Add Exception Graindoc

ospencer:

PR #1338, opened 2022-06-14
chore(ci): Add phase for checking stdlib docs & formatting

phated:
This adds a step to the CI workflow that runs graindoc and grainformat against the stdlib. This is really just another linting step but it needs to happen after we know we have a working grain binary.

As you can see in the following commits, we are very bad at updating things 😛 

I have been planning to add this, but #1337 reminded me we needed it when @ospencer accepted my change but didn't regen the docs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
           grain -v
 
       # If we have a working grain CLI, we can run grainfmt and graindoc on stdlib
-      - name: Run formatting lint
+      - name: Check stdlib docs and formatting
         if: matrix.os != 'windows-latest'
         run: |
           grain doc stdlib -o stdlib --current-version=$(grain -v)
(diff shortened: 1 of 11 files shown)


ospencer (approved):
Looks reasonable to me! Reminds me that for formatting I really like Prettier's flags, `--write` and `--check`, where `--write` is essentially an alias for `-o` being the same as the input.

I'd be pretty excited if the CI step was `grain format stdlib --check` and `grain doc stdlib --check` since I imagine this will be fairly common. Is that a change you'd be cool with me implementing?

phated:
> I'd be pretty excited if the CI step was grain format stdlib --check and grain doc stdlib --check since I imagine this will be fairly common. Is that a change you'd be cool with me implementing?

Yeah, I thought we had an issue for that but I actually don't see it. Can you create one? Then we can bikeshed what the flag should be named.

marcusroberts (approved):
Nice to see the formatter at work!

PR #1342, opened 2022-06-16
feat(compiler): Unicode-aware lexing

### LINKED ISSUE #799: Reported compiler positions are not unicode-aware
See this example:
image

The 💯 emoji is four bytes, but in reality, it's only one character.

This can probably be fixed at the lexing level. This should be fixed for everywhere unicode is allowed, currently strings, characters, and comments.

ospencer:
Closes #799.

This PR swaps from OCamllex to [Sedlex](https://github.com/ocaml-community/sedlex), a Unicode-aware lexer generator. The major benefit is that Grain locations are now reported accurately, with an emoji like `💯` being reported as a single character rather than 4 separate bytes (and thus no longer ruining the rest of the locations for the file).

As an added bonus, Sedlex has built-in support for Unicode categories and properties, meaning that Grain can support a much wider range of characters for use in identifiers. As such, Grain identifiers are now based on Unicode's `xid` property (see [xid_start](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=[:XIDS:]) for characters that can begin an identifier and [xid_continue](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=[:XIDC:]) for characters that can appear in an identifier after the first).

The following is now valid 😎 
```
enum Caipirinha {
  Cachaça,
  Sugar,
  Lime,
}

let pokémon = "pikachu"

type Über = Number
```

This change may be a significant help for users whose native language is not English (and also just nice to support what modern languages are supporting).

I did some unscientific performance testing of Sedlex vs OCamllex and saw no significant difference in compile times (Sedlex maybe took a few milliseconds longer).


compiler/esy.json
```diff
@@ -60,6 +61,9 @@
     "ocaml": "4.12.0",
     "@opam/fp": "reasonml/reason-native:fp.opam#0ed854f986256e52e59aeecfa90e9af60f105b15",
     "@opam/fs": "reasonml/reason-native:fs.opam#0ed854f986256e52e59aeecfa90e9af60f105b15",
+    "@opam/ppx_deriving_yojson": "ocaml-ppx/ppx_deriving_yojson:ppx_deriving_yojson.opam#0809ea014a6cdd27d5863e05cf330ff74d0114ae",
+    "@opam/ppx_deriving_cmdliner": "hammerlab/ppx_deriving_cmdliner:ppx_deriving_cmdliner.opam#1f086651fe7f8dd98e371b09c6fcc4dbc6db1c7c",
+    "@opam/ocamlfind": "1.9.3",
```
ospencer, line 66: These resolutions are for the new, unreleased versions of the derivers for yojson and cmdliner that support the newer ppxlib that sedlex uses. (They may be published before this PR lands and can be removed.)

compiler/src/parsing/well_formedness.re
```diff
@@ -6,7 +6,6 @@ type wferr =
   | MalformedString(Location.t)
   | IllegalCharacterLiteral(string, Location.t)
   | MultipleModuleName(Location.t)
-  | TypeNameShouldBeUppercase(string, Location.t)
```
ospencer, line 9: This was never actually used because the parser disallows you from using the wrong kind of identifier (but gives a nice error message telling you that you should have used an uppercase identifier).

peblair (approved):
👏👏👏

marcusroberts:
Oooh, very nice!    Does this affect location reporting for the LSP or Formatter at all?

ospencer:
@marcusroberts yeah it does. I modified the formatter to understand that the locations are in unicode code points rather than bytes. It's a tad janky right now because OCaml doesn't really have great unicode support, but I'm hoping to get some updates upstreamed into the unicode libraries we use to ease this a bit.

marcusroberts:
Just building to test, but then an approval is on the way, this is excellent!

marcusroberts (approved):
Excellent!

PR #1343, opened 2022-06-18
feat(stdlib): Implement sin and cos

FinnRG:
For #1017, adds Number.sin and Number.cos


ospencer:
@FinnRG that PR got landed! Do you mind rebasing this and using the new `pi` constant?

FinnRG:
@ospencer Done. I've also generated new docs.

stdlib/number.gr
```diff
@@ -346,3 +346,56 @@ export let isInfinite = (x: Number) => {
  * @since v0.4.5
  */
 export parseInt
+
+/**
+ * Computes how many times pi has to be subtracted to achieve the required bounds for sin.
+ */
+let reduceToPiBound = (radians: Number) => {
+  let quot = floor(radians / pi)
+  quot
```
ospencer, line 355:
```suggestion
  floor(radians / pi)
```

stdlib/number.gr
```diff
@@ -346,3 +346,56 @@ export let isInfinite = (x: Number) => {
...
+}
+
+/**
+ * Computes the sine of a number using Chebyshev polynomials. Requires the input to be bounded to (-pi, pi).
+ */
+let chebyshevSine = (radians: Number) => {
```
ospencer, line 361: Could you add a comment to this function with a link to information about the algorithm?

ospencer (commented):
Perfect, looks great! Just a couple of minor tweaks and this should be all set to land 👍

stdlib/number.gr
FinnRG: If the link I added in the newest commit doesn't load fast enough, please try the [archived version](https://web.archive.org/web/20220416080650/mooooo.ooo/chebyshev-sine-approximation/) instead.

ospencer:
Thanks for adding the link! Reading it over, it actually has me thinking—are we losing out on some precision here? The algorithm is for f32s, but Grain's Number type defaults to f64s for the underlying float representation.

FinnRG:
@ospencer Yes, the algorithm is for f32's. I don't think this is a big problem, because the average error converges around this precision. If this is still a problem, I could port the libm algorithms for sin and cos (https://github.com/rust-lang/libm/blob/master/src/math/sin.rs), these algorithms are quite lengthy, and I would prefer to put them in a separate file so as not  to pollute the Number module.

ospencer (commented):
Seems reasonable! Could you regen the docs one last time?

FinnRG:
@ospencer Done.

ospencer (approved):
Beautiful, thank you!

phated (approved):
I don't know math, but the code and docs look great! Thanks @FinnRG 🎉

PR #1346, opened 2022-06-28
fix(compiler): Fix compilation of closure scope mutable destructuring

ospencer:


marcusroberts (approved):

compiler/test/suites/pattern_matching.re
```diff
@@ -357,6 +357,34 @@ describe("pattern matching", ({test, testSkip}) => {
...
+
+        print(b)
+
+        () => b
+      }
+
+      run()
+    |},
+    "2\n",
+  );
```
phated, line 387: These tests have me wondering about this:
```
let run = () => {
        let mut a = 1
        let mut b = 2
        let mut (x, y) = (a, b)
        b = 7
        print(y)
}
run()
```
What happens?

ospencer: You get `2`!

phated: Can you add that a test?

ospencer: Done!

phated (approved):
Excellent!

PR #1347, opened 2022-06-28
fix(stdlib): Make toNumber functions respect Number invariants

### LINKED ISSUE #1339: `fromNumber` functions not following `Number` invariants
Performing an array access on the value `Int32.fromNumber(0l)` results in an index out of bounds error because the `0` is not downcasted to a simple number.

ospencer:
Fixes #1339


marcusroberts (approved):
Nice fix!

PR #1349, opened 2022-07-01
fix(compiler): Allow defining _start when using --use-start-section

ospencer:
This PR allows you to manually define a function called `_start` when using the `--use-start-section` flag.

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -3624,7 +3624,9 @@ let compile_exports = (wasm_mod, env, {imports, exports, globals}) => {
   };
   List.iteri(compile_export, exports);
   ignore @@ Export.add_function_export(wasm_mod, grain_main, grain_main);
-  ignore @@ Export.add_function_export(wasm_mod, grain_start, grain_start);
+  if (! Config.use_start_section^) {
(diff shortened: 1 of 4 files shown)


marcusroberts (approved):
I like this.     It caused a problem for me when working with some wasm runtimes, so an approval from me for this approach!

phated (approved):
'member when I wanted to name the flag `--start-section` and take the function you wanted to make your start section and Oscar said no?

ospencer:
That wouldn't have quite worked here—this allows you to have a start section _and_ define a function called `_start`, so you can have module initialization code run independently of your "main" function (versus having your main function run as initialization code). This is important for runtimes like Lunatic, where you want to start the module many times (so setting up closures, etc.) but only run a main function one of those times. Code in the start section is always run at module startup per the spec, but any of the other exports can be run later.

It's actually pretty annoying that browsers don't have great WASI support—the most technically correct thing to do would be to take what Grain currently puts in `_start` and just have that be in the start section (so --use-start-section as the default) and this would have been a non-issue.

But alas.

And just to clarify @marcusroberts, this isn't a feature, just a bug fix! I think we still should support some optional main function. By convention with WASI, this is always exposed as `_start`, but languages like C and Rust call them `main`.

PR #1354, opened 2022-07-10
fix(stdlib): Correctly promote numbers to bigints when left-shifting

### LINKED ISSUE #1350: Bit shifting operators are not arbitrary precision
I wrote `print(5 << 64)` and expected to get a really big number, but I got `5`.

peblair:
Closes #1350.

### DIFF
--- a/stdlib/runtime/numbers.gr
+++ b/stdlib/runtime/numbers.gr
@@ -1833,7 +1833,13 @@ export let (<<) = (x: Number, y: Number) => {
   } else {
     let xval = coerceNumberToWasmI64(x)
     let yval = coerceNumberToWasmI64(y)
-    WasmI32.toGrain(reducedInteger(WasmI64.shl(xval, yval))): Number
+    if (WasmI64.leU(WasmI64.clz(xval), yval)) {
+      let xbi = coerceNumberToBigInt(x)
+      let yval = coerceNumberToWasmI32(y)
+      WasmI32.toGrain(reducedBigInteger(BI.shl(xbi, yval))): Number
+    } else {
+      WasmI32.toGrain(reducedInteger(WasmI64.shl(xval, yval))): Number
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


stdlib/runtime/numbers.gr
```diff
@@ -1833,7 +1833,13 @@ export let (<<) = (x: Number, y: Number) => {
   } else {
     let xval = coerceNumberToWasmI64(x)
     let yval = coerceNumberToWasmI64(y)
-    WasmI32.toGrain(reducedInteger(WasmI64.shl(xval, yval))): Number
+    if (WasmI64.leU(WasmI64.clz(xval), yval)) {
```
ospencer, line 1836: It looks like this fast path only works for positive numbers—is there any other quick check that'd also work for negative numbers?

peblair: Heh, this actually has the opposite effect that you are imagining: `clz` of a negative number is zero, so we always promote to a big integer. If you think that's a problem, I can try to refine the condition a bit more, but I just tested this program, and it works as intended:

```grain
import BigInt from "bigint"
print(5 << 64)
print(-5 << 64)
print(BigInt.shl(-5t, 64l))
```
output:
```
92233720368547758080
-92233720368547758080
-92233720368547758080
```

ospencer: Sorry if I wasn't clear—I wasn't trying to say that this wasn't correct, just that exactly your point—`clz` of a negative number is always zero, so negative numbers never make it to the fast path. I think it'd be better to have a check that entered the fast path for both positive and negative numbers.

peblair: @ospencer Got it. I think that the version I just pushed should do it.

ospencer (commented):
Please don't be too annoyed at me, but I like the version you originally did better than this one. `&&` and `||` are implemented with branching, so now it's pretty expensive to do this check, almost to the point that it might be better to just skip the check altogether and always promote. I think I'm fine to have just the positive case have a chance to be fast-tracked.

Another option would be doing an abs of the value and then doing the clz—can do an abs in like 3 instructions so it's not too much more than your original version (but up to you).

peblair:
@ospencer I did the `abs` solution. Does it look okay now?

ospencer (approved):
Looks good! Thanks for bearing with me on the iterations 🙇

phated (approved):
Excellent! I also enjoyed reading the discussion between @ospencer and @peblair 🎉

marcusroberts (approved):
Clever code!

PR #1357, opened 2022-07-15
chore(compiler): Remove deprecated use of Streams

ospencer:
Streams in OCaml have been deprecated. This also cleans up a long-time comment about an issue with streams, so that's a plus.


phated (approved):
Champion! Thank you 🙇

PR #1358, opened 2022-07-15
chore(compiler): Always use labels when calling functions

ospencer:
Newer versions of OCaml have a warning for this, so this just applies labels to unlabeled arguments.

### DIFF
--- a/compiler/dune-workspace
+++ b/compiler/dune-workspace
@@ -1,2 +1,8 @@
 (lang dune 2.0)
+
 (profile release)
+
+(env
+ (_
+  (flags
+   (:standard -warn-error -a+6+31))))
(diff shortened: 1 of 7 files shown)


marcusroberts (approved):
Nice cleanup!

phated (approved):
Does OCaml just assume positional for named arguments when you have more than the expected number of positional arguments? If so, that's kind of mad...

Should we set a compiler flag to turn this warning into an error??

ospencer:
Yeppppppp. I'm not sure if they added the warning in 4.13 or 4.14, but should I make an issue to make that an error?

phated:
> should I make an issue to make that an error?

Would be ideal to turn it on in this PR so we can verify the option works!

phated:
I believe this would be done by adding this to our `dune-workspace` file:

```
(env
  (_ (flags (:standard -warn-error -a+6+31)))
```

Note: `-warn-error -a+31` is supposed to be the default but idk if dune does something differently (I just took this from the ocaml docs and added `+6`)

Issue #730, opened 2021-06-13
What should the behavior of `Array.zip` be on arrays of different sizes?


peblair:
Currently (#719), `Array.zip` successfully returns a zipped array if the two given arrays are of the same length, and `fail`s if they are not. Some open questions:
- Should we adopt Python's `zip` semantics (allow differing lengths, with the result only as long as the shorter of the two)?
- If not, should this be a `Result` return value instead?

peblair:
My vote on this is that we permit arrays of different lengths, with the returned array having a length equal to the shortest argument. This is in line with `zip` in Python and Haskell.

ospencer:
We've decided to permit arrays/lists of different lengths.

PR #1363, opened 2022-07-17
feat(stdlib): Implement List.zip, List.unzip, List.zipWith, Array.reduceRight, Array.zipWith

alexsnezhko3:
- Added zip and unzip to list to bring it up to parity with array
- Added reduceRight to array
- Added zipWith to both array and list

### DIFF
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -565,12 +565,14 @@ export let unique = list => {
  * 
  * @since v0.5.3
  */
-export let rec zip = (list1, list2) => {
-  match ((list1, list2)) {
-    ([], []) => [],
-    ([first1, ...rest1], [first2, ...rest2]) => [(first1, first2), ...zip(rest1, rest2)],
-    _ => []
+export let zip = (list1, list2) => {
+  let rec zipInner = (list1, list2, acc) => {
+    match ((list1, list2)) {
+      ([first1, ...rest1], [first2, ...rest2]) => zipInner(rest1, rest2, [(first1, first2), ...acc]),
+      _ => acc
+    }
   }
+  reverse(zipInner(list1, list2, []))
 }
 
 /**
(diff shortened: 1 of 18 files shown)


ospencer (approved):
This is awesome, thank you! Docs look really good too. 🎉 

(for anyone like me doing mental gymnastics about whether `zipWith` should be called `map2`, we've already got `Option.zipWith`.)

ospencer:
I thought I fixed the snapshots to not change on standard library changes like this, but maybe not? @alexsnezhko3 Can you run `npm run compiler test -- -u` and commit the results?

marcusroberts (approved):
Excellent contribution thank you!

stdlib/list.gr
```diff
@@ -548,6 +548,77 @@ export let unique = list => {
...
+ * @example List.zip([1, 2, 3], [4, 5]) // [(1, 4), (2, 5)]
+ * 
+ * @since v0.5.3
+ */
+export let rec zip = (list1, list2) => {
+  match ((list1, list2)) {
+    ([], []) => [],
+    ([first1, ...rest1], [first2, ...rest2]) => [(first1, first2), ...zip(rest1, rest2)],
+    _ => []
+  }
```
phated, line 573: I think this needs to use a helper function to be made tail-recursive. I believe you can look at `List.length` or `List.reverse` to see the pattern for doing that.

stdlib/list.gr
```diff
@@ -548,6 +548,77 @@ export let unique = list => {
...
+ * @example List.zipWith((a, b) => a * b, [1, 2, 3], [4, 5]) // [4, 10]
+ * 
+ * @since v0.5.3
+ */
+export let rec zipWith = (fn, list1, list2) => {
+  match ((list1, list2)) {
+    ([], []) => [],
+    ([first1, ...rest1], [first2, ...rest2]) => [fn(first1, first2), ...zipWith(fn, rest1, rest2)],
+    _ => []
+  }
```
phated, line 601: I think this needs to use a helper function to be made tail-recursive. I believe you can look at `List.length` or `List.reverse` to see the pattern for doing that.

phated (changes requested):
Just a note about `List.zip` and `List.zipWith` not being tail-recursive (also, correct me if I'm mistaken).

stdlib/list.gr
alexsnezhko3: Would something like this work?
```
let zip = (list1, list2) => {
  let rec zipInner = (list1, list2, acc) => {
    match ((list1, list2)) {
      ([first1, ...rest1], [first2, ...rest2]) => zipInner(rest1, rest2, [(first1, first2), ...acc]),
      _ => reverse(acc)
    }
  }
  zipInner(list1, list2, [])
}
```
I add the next elements in the list to the beginning of the accumulator and then reverse it at the end to bring it back to its normal shape

ospencer (approved):
Implementation looks great!

phated (approved):
🔥 Awesome! Thanks!

@ospencer Before merging, can you comment on https://github.com/grain-lang/grain/issues/730 about the decision for the List zips functions to operate on lists of different lengths? We also need a followup issue to make the breaking change for Array.zip to match this behavior. 🙏

ospencer:
The array issue is https://github.com/grain-lang/grain/issues/1395 👍

PR #1368, opened 2022-07-22
feat: Arm64 docker builds

ospencer:
This PR adds support for the linux/arm64 arch in our published Docker images. Since GitHub actions doesn't yet support ARM, this is done by booting up an AWS Graviton instance and using it as a remote builder to handle the arm64 part of the image build. The Graviton instance is set to automatically stop when it detects that builds are complete.

Additionally, `esy` does not yet have linux/arm64 binaries, so we use a workaround in the Dockerfiles to use one that I built. When `esy` ships a release with linux/arm64 binaries, we can remove the workaround.

If you have an arm64 linux machine (or a Mac with an M1 processor) you can try out the image with
```sh
docker run -it ghcr.io/grain-lang/grain:oscar-docker-arm64
```
or
```sh
docker run -it ghcr.io/grain-lang/grain:oscar-docker-arm64-slim
```
and try compiling and running a Grain program. (If you're not on an arm64 machine you can also do this, but you won't be testing out the new stuff 🙂)

You can see that both arches are contained in the image:
image

The successful CI run can be found here: https://github.com/grain-lang/grain/runs/7566595877?check_suite_focus=true

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -11,6 +11,9 @@ permissions:
   id-token: write # Required to auth to AWS
   contents: read # Required for actions/checkout
   packages: write # Required to upload to ghcr
+env:
+  AWS_EC2_INSTANCE_ID: i-0960f9b8de8285aed
+  AWS_REGION: us-east-1
 jobs:
   push_base_image:
     name: Push base Docker image to multiple registries
@@ -30,14 +33,14 @@ jobs:
         with:
           role-to-assume: arn:aws:iam::881972110175:role/github-actions-openid
           role-session-name: cisession
-          aws-region: us-east-1
+          aws-region: ${{ env.AWS_REGION }}
 
       - name: Start AWS EC2 arm64 instance
         id: instance
         run: |
-          aws ec2 start-instances --instance-ids ${{ secrets.AWS_EC2_INSTANCE_ID }}
-          aws ec2 wait instance-running --instance-ids ${{ secrets.AWS_EC2_INSTANCE_ID }}
-          echo ::set-output name=ip::$(aws ec2 describe-instances --filters "Name=instance-id,Values=${{ secrets.AWS_EC2_INSTANCE_ID }}" --query 'Reservations[*].Instances[*].[PublicIpAddress]' --output text)
+          aws ec2 start-instances --instance-ids ${{ env.AWS_EC2_INSTANCE_ID }}
+          aws ec2 wait instance-running --instance-ids ${{ env.AWS_EC2_INSTANCE_ID }}
+          echo ::set-output name=ip::$(aws ec2 describe-instances --filters "Name=instance-id,Values=${{ env.AWS_EC2_INSTANCE_ID }}" --query 'Reservations[*].Instances[*].[PublicIpAddress]' --output text)
 
       - name: Bring in SSH keys
         uses: grain-lang/ssh-agent@v0.5.4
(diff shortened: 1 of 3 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -21,6 +21,34 @@ jobs:
...
+          aws ec2 start-instances --instance-ids ${{ secrets.AWS_EC2_INSTANCE_ID }}
+          aws ec2 wait instance-running --instance-ids ${{ secrets.AWS_EC2_INSTANCE_ID }}
+          echo ::set-output name=ip::$(aws ec2 describe-instances --filters "Name=instance-id,Values=${{ secrets.AWS_EC2_INSTANCE_ID }}" --query 'Reservations[*].Instances[*].[PublicIpAddress]' --output text)
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+
+      - name: Bring in SSH keys
+        uses: webfactory/ssh-agent@v0.5.4
```
phated, line 36: Please fork this action and use our fork. People can swap out tags out from underneath and hijack things (I also trust you audited it?)

phated: Ideally, the AWS instance would use the public key from grainbot instead of us importing their private key (we'd need to add a public key to grainbot), but I have no idea how hard that would be

phated (commented):
Thoughts about the private key stuff

.github/workflows/docker-publish.yml
ospencer: I made this keypair, though yeah it'd be nice if grainbot was able to have its own private key. From what I can tell GitHub doesn't currently support it.

ospencer:
Alright @phated, this should be all set.

.github/workflows/docker-publish.yml
```diff
@@ -21,6 +25,37 @@ jobs:
...
+          role-to-assume: arn:aws:iam::881972110175:role/github-actions-openid
+          role-session-name: cisession
+          aws-region: us-east-1
+
+      - name: Start AWS EC2 arm64 instance
+        id: instance
+        run: |
+          aws ec2 start-instances --instance-ids ${{ secrets.AWS_EC2_INSTANCE_ID }}
+          aws ec2 wait instance-running --instance-ids ${{ secrets.AWS_EC2_INSTANCE_ID }}
+          echo ::set-output name=ip::$(aws ec2 describe-instances --filters "Name=instance-id,Values=${{ secrets.AWS_EC2_INSTANCE_ID }}" --query 'Reservations[*].Instances[*].[PublicIpAddress]' --output text)
```
phated, line 40:
```suggestion
          aws ec2 start-instances --instance-ids ${{ secrets.AWS_EC2_INSTANCE_ID }}
          aws ec2 wait instance-running --instance-ids ${{ secrets.AWS_EC2_INSTANCE_ID }}
          echo ::set-output name=ip::$(aws ec2 describe-instances --filters "Name=instance-id,Values=${{ secrets.AWS_EC2_INSTANCE_ID }}" --query 'Reservations[*].Instances[*].[PublicIpAddress]' --output text)
```
We discussed not making this a secret since it doesn't **need** to be and might help with debugging.

.github/workflows/docker-publish.yml
```diff
@@ -21,6 +25,37 @@ jobs:
...
+
+      - name: Bring in SSH keys
+        uses: grain-lang/ssh-agent@v0.5.4
+        with:
+          ssh-private-key: ${{ secrets.AWS_SSH_PRIVATE_KEY }}
```
phated, line 45: Org level secret

phated (changes requested):
Couple things we talked about in DMs that got lost in the noise.

ospencer:
@phated Done.

phated (approved):
Looking great! Thanks for digging into the OpenID stuff too 🎉

marcusroberts (approved):
Pure magic!

PR #1372, opened 2022-07-26
feat(stdlib): Add String.lastIndexOf

spotandjake:
This pr adds String.lastIndexOf, it works the same as indexOf but instead of returning the first occurance of the search it returns the last.


ospencer (changes requested):
I think what might be a bit better here is reusing the code from `indexOf`. The only difference between the two would be the `break` line in the loop.

You can turn `indexOf` into a helper function that takes an additional argument, maybe called `findFirst`, and then that `if` becomes
```
      if (Memory.compare(ptr, pptr, psize) == 0n) {
        result = idx
        if (findFirst) break
      }
```
And then you can call the helper for both functions:
```
export let indexOf = (search, string) => {
  indexOf(search, string, true)
}
export let lastIndexOf = (search, string) => {
  indexOf(search, string, false)
}

File: .gitignore [hunk 1]
```diff
@@ -45,3 +45,5 @@ cli/bin/grainlsp.exe
 cli/bin/grainlsp.bc.js
 
 .DS_Store
+
+yarn.lock
```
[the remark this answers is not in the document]
spotandjake: > I think what might be a bit better here is reusing the code from `indexOf`. The only difference between the two would be the `break` line in the loop.
> 
> You can turn `indexOf` into a helper function that takes an additional argument, maybe called `findFirst`, and then that `if` becomes
> 
> ```
>       if (Memory.compare(ptr, pptr, psize) == 0n) {
>         result = idx
>         if (findFirst) break
>       }
> ```
> 
> And then you can call the helper for both functions:
> 
> ```
> export let indexOf = (search, string) => {
>   indexOf(search, string, true)
> }
> export let lastIndexOf = (search, string) => {
>   indexOf(search, string, false)
> }
> ```

I thought about that but won't that be extremely slow on long strings compared to what I am doing.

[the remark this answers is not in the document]
ospencer: Oh I hadn't realized you actually made it iterate backwards.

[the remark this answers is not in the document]
spotandjake: Yup why the check for incrementing the index is different, I  pretty sure my logic is sound and the tests pass but you might want to double check the index increment and make sure the check makes sense.

stdlib/string.gr
```diff
@@ -161,6 +161,66 @@ export let indexOf = (search: String, string: String) => {
...
+    None
+  } else {
+    // Scan The String
+    let mut matchIndex = -1n
+    let mut stringIndex = 0n
+    let searchPointer = search + 8n
+    let stringStartPtr = string + 8n
+    let mut stringPtr = stringStartPtr + WasmI32.sub(stringSize, searchSize)
+    while (stringPtr >= stringStartPtr) {
+      // Compare The Current String Section To The Search String
```
ospencer, line 201:
```suggestion
      let byte = WasmI32.load8U(stringPtr, 0n)
      stringPtr -= 1n
      if ((byte & 0xC0n) == 0x80n) {
        // We're not at the start of a codepoint, so move on
        continue
      }
      
```

stdlib/string.gr
```diff
@@ -161,6 +161,66 @@ export let indexOf = (search: String, string: String) => {
...
+      if (Memory.compare(stringPtr, searchPointer, searchSize) == 0n) {
+        // If They Are Equal Then Save The Index
```
ospencer, line 203:
```suggestion
```

stdlib/string.gr
```diff
@@ -161,6 +161,66 @@ export let indexOf = (search: String, string: String) => {
...
+        matchIndex = stringIndex
+        break
+      }
+      // Increment The Index If We Have Matched A Full Code Point
+      let byte = WasmI32.load8U(stringPtr, 0n)
+      if ((byte & 0xC0n) != 0x80n) {
+        stringIndex += 1n
+      }
```
ospencer, line 211:
```suggestion
      stringIndex -= 1n
```

stdlib/string.gr
```diff
@@ -161,6 +161,66 @@ export let indexOf = (search: String, string: String) => {
...
+      // Decrement The Pointer
+      stringPtr = WasmI32.sub(stringPtr, 1n)
```
ospencer, line 213:
```suggestion
```

stdlib/string.gr
```diff
@@ -161,6 +161,66 @@ export let indexOf = (search: String, string: String) => {
...
+  let string = WasmI32.fromGrain(string)
+  let searchSize = WasmI32.load(search, 4n)
+  let stringSize = WasmI32.load(string, 4n)
+  // Make Sure The Search String Is Not Bigger Then The String As It Will Never Match
+  if (searchSize > stringSize) {
+    None
+  } else {
+    // Scan The String
+    let mut matchIndex = -1n
+    let mut stringIndex = 0n
```
ospencer, line 196: You'll need to import `untagSimpleNumber` for this one.
```suggestion
    let mut stringIndex = untagSimpleNumber(lastIndex)
```

stdlib/string.gr
```diff
@@ -161,6 +161,66 @@ export let indexOf = (search: String, string: String) => {
...
+        stringIndex += 1n
+      }
+      // Decrement The Pointer
+      stringPtr = WasmI32.sub(stringPtr, 1n)
+    }
+    // Return The Result
+    if (matchIndex == -1n) {
+      None
+    } else {
+      Some(strLen - tagSimpleNumber(matchIndex))
```
ospencer, line 219:
```suggestion
      Some(tagSimpleNumber(matchIndex))
```

ospencer (changes requested):
I left some suggestions here—I switched up the logic a bit to avoid a `Memory.compare` when we're not at the start of a valid codepoint.

stdlib/string.gr
```diff
@@ -161,6 +162,64 @@ export let indexOf = (search: String, string: String) => {
...
+ * @returns `Some(position)` containing the starting position of the substring if found or `None` otherwise
+ *
+ * @example String.lastIndexOf("world", "Hello world world") == Some(12)
+ *
+ * @since v0.5.3
+ */
+@unsafe
+export let lastIndexOf = (search: String, string: String) => {
+  // The last possible index in the string given the length of the search
+  let lastIndex = length(string) - length(search)
```
ospencer, line 179:
```suggestion
  let lastIndex = length(string) - length(search)
  
```

stdlib/string.gr
```diff
@@ -161,6 +162,64 @@ export let indexOf = (search: String, string: String) => {
...
+  let (>) = WasmI32.gtU
+  let (>=) = WasmI32.geU
+  let (==) = WasmI32.eq
+  let (!=) = WasmI32.ne
+  let (+) = WasmI32.add
+  let (-) = WasmI32.sub
+  let (&) = WasmI32.and
```
ospencer, line 186:
```suggestion
  let (&) = WasmI32.and
  
```

stdlib/string.gr
```diff
@@ -161,6 +162,64 @@ export let indexOf = (search: String, string: String) => {
...
+        // We're not at the start of a codepoint, so move on
+        continue
+      }
+      if (Memory.compare(stringPtr, searchPtr, searchSize) == 0n) {
+        matchIndex = stringIndex
+        break
+      }
+      stringIndex -= 1n
+    }
+    // Return The Result
```
ospencer, line 214:
```suggestion

```

ospencer (approved):
Great work!

ospencer:
Looks like you need to regenerate the Graindocs.

phated (changes requested):
I'm holding off my review until #1389 lands because this PR contains a bunch of formatting that shouldn't be included in the PR

spotandjake:
> Looks like you need to regenerate the Graindocs.

GrainDoc Seems To Be Up To Date

compiler/test/stdlib/string.test.gr
```diff
@@ -184,6 +195,7 @@ let processBytes = b => {
 // Simple cases
 // formatter-ignore
 assert processBytes(String.encode("ab", String.UTF8)) == [> 0x61l, 0x62l]
+
```
phated, line 198:
```suggestion
```

phated (approved):
Really cool! I like that you are diving into the low-level Grain stuff @spotandjake 🎉

spotandjake:
> Really cool! I like that you are diving into the low-level Grain stuff @spotandjake 🎉

Thank you, after working on my linker I have needed to work with a lot more low level grain stuff and realized it is actually quite friendly and easier to work with.

PR #1373, opened 2022-07-26
feat(stdlib): Add `Number.tan`, `Number.factorial`, `Number.clamp`, `Number.lerp`

spotandjake:
This pr adds `Number.tan`, `Number.factorial`, `Number.clamp`, `Number.lerp` to the Number libary for #1017


stdlib/number.gr
```diff
@@ -398,3 +398,77 @@ export let sin = (radians: Number) => {
 export let cos = (radians: Number) => {
   sin(pi / 2 + radians)
 }
+
+/**
+ * Computes the tan of a number (in radians) using Chebyshev polynomials.
```
ospencer, line 403:
```suggestion
 * Computes the tangent of a number (in radians) using Chebyshev polynomials.
```

stdlib/number.gr
```diff
@@ -398,3 +398,77 @@ export let sin = (radians: Number) => {
...
+ *
+ * @param radians: The input in radians
+ * @returns The computed tangent
+ *
+ * @since v0.5.2
```
ospencer, line 408: These would all be `v0.5.3`.

stdlib/number.gr
```diff
@@ -398,3 +398,77 @@ export let sin = (radians: Number) => {
...
+
+/**
+ * Computes the factorial of a number.
+ *
+ * @param input: The number to compute the factorial of
+ * @returns The computed factorial
+ *
+ * @since v0.5.2
+ */
+export let factorial = input => {
```
ospencer, line 422: Missed opportunity for a recursive implementation! That said, this should fail if the import is not an integer, which makes it glaringly obvious that we're missing functions like `Number.isInteger`, `Number.isFloat`, and `Number.isRational`. https://github.com/grain-lang/grain/issues/1392

You could wait until those are implemented or remove `factorial` from this PR and bring it back later.

stdlib/number.gr
```diff
@@ -398,3 +398,77 @@ export let sin = (radians: Number) => {
...
+export let clamp = (min: Number, max: Number, input: Number) => {
+  if (input < min) min else if (input > max) max else input
+}
+
+/**
+ * Linearly interpolates between the starting and ending value by the given amount.
+ *
+ * @param start: The starting value
+ * @param end: The ending value
+ * @param amount: an amount between 0 and 1
```
ospencer, line 450: This isn't enforced, so any amount outside of the range 0 to 1 produces an extrapolation. Would make sense to update this param description and the full description so that's clear.

stdlib/number.gr
```diff
@@ -398,3 +398,77 @@ export let sin = (radians: Number) => {
...
+ * @param inMin: The minimum value in the input range
+ * @param outMin: The minimum value in the output range
+ * @param inMax: The maximum value in the input range
+ * @param outMax: The maximum value in the output range
+ * @param current: The number to map
+ * @returns The interpolated value
+ *
+ * @since v0.5.2
+ */
+export let map = (inMin, outMin, inMax, outMax, current) => {
```
ospencer, line 471: I think this name would confuse a lot of FP folks. Maybe a name like `linearMap` would be better?

ospencer (changes requested):

stdlib/number.gr
spotandjake: could we reexport it as that like an alias because I think `linearMap` is going to confuse a lot of oop people and cause them to use a local implementation

ospencer: Definitely want it to only have one name. Since Grain is an FP language we cater a bit more to the FP use case 🙂 

`map` might be a fine name; we'll see what some others think when they review.

spotandjake:
I made the changes we talked about in discord i will move `Number.map` to its own pr i wasnt able to run the tests because downloading the modules on my internet is taking forever.

compiler/test/stdlib/number.test.gr
```diff
@@ -256,3 +256,29 @@ assert Number.sin(20) > 0.91294 && Number.sin(20) < 0.91295
...
+// Number.lerp
+assert Number.lerp(0, 5, 0) == 0
+assert Number.lerp(0, 5, 0.5) == 2.5
+assert Number.lerp(0, 5, 1) == 5
+// Number.map
+assert Number.map(0, 0, 1, 100, 0) == 0
+assert Number.map(0, 0, 1, 100, 0.5) == 50
+assert Number.map(0, 0, 1, 100, 1) == 100
+assert Number.map(0, 0, 1, 100, 150) == 100
+assert Number.map(-5, 0, 0, 100, -2.5) == 50
```
phated, line 284:
```suggestion
```
I think you also meant to remove the tests from this PR?

stdlib/number.gr
```diff
@@ -439,3 +439,63 @@ export let sin = (radians: Number) => {
...
+ * @since v0.5.3
+ */
+export let tan = (radians: Number) => {
+  sin(radians) / cos(radians)
+}
+
+/**
+ * Computes the factorial of a number.
+ *
+ * @param input: The number to compute the factorial of, Must be a Non negative integer
```
phated, line 458:
```suggestion
 * @param input: The non-negative integer used to compute the factorial
```

compiler/test/stdlib/number.test.gr
spotandjake: 🤦

stdlib/number.gr
```diff
@@ -439,3 +439,63 @@ export let sin = (radians: Number) => {
...
+ * @returns The computed factorial
+ *
+ * @since v0.5.3
+ */
+export let factorial = input => {
+  // Error If It is Negative, or Not an Integer
+  if (sign(input) is -1 || isInteger(input)) {
```
phated, line 465:
```suggestion
  if (input < 0 || isInteger(input)) {
```
this is better, right?

stdlib/number.gr
```diff
@@ -439,3 +439,63 @@ export let sin = (radians: Number) => {
...
+    fail "Factorial Expects A Non Negative Integer Input"
```
phated, line 466: Do we actually want to fail or return an option? @ospencer

stdlib/number.gr
```diff
@@ -439,3 +439,63 @@ export let sin = (radians: Number) => {
...
+  } else {
+    // Factorial Helper
+    let rec fact = n => if (n is 0) 1 else n * fact(n - 1)
```
phated, line 469:
```suggestion
    let rec fact = n => if (n == 0) 1 else n * fact(n - 1)
```
why use `is`?

stdlib/number.gr
```diff
@@ -439,3 +439,63 @@ export let sin = (radians: Number) => {
...
+    // Perform Factorial
+    fact(input)
+  }
+}
+
+/**
+ * Clamps a number to a given range.
+ *
+ * @param min: The minimum value in the range
+ * @param max: The maximum value in the range
```
phated, line 479: Did someone say Range?

stdlib/number.gr
```diff
@@ -439,3 +439,63 @@ export let sin = (radians: Number) => {
...
+ * Linearly interpolates between the starting and ending value by the given amount.
+ *
+ * @param start: The starting value
+ * @param end: The ending value
+ * @param amount: an amount between 0 and 1, anything outside the range will produce an extrapolation i.e `lerp(0, 100, 2)` will produce 200
+ * @returns The interpolated value
+ *
+ * @since v0.5.3
+ */
+export let lerp = (start: Number, end: Number, amount: Number) => {
```
phated, line 499:
```suggestion
export let linearInterpret = (start: Number, end: Number, amount: Number) => {
```
Hear me out. `lerp` sounds like I'm 🤮

stdlib/number.gr
```diff
@@ -439,3 +439,63 @@ export let sin = (radians: Number) => {
...
+export let clamp = (min: Number, max: Number, input: Number) => {
+  if (input < min) min else if (input > max) max else input
+}
+
+/**
+ * Linearly interpolates between the starting and ending value by the given amount.
+ *
+ * @param start: The starting value
+ * @param end: The ending value
+ * @param amount: an amount between 0 and 1, anything outside the range will produce an extrapolation i.e `lerp(0, 100, 2)` will produce 200
```
phated, line 494: this is a "interpolation" but you are saying that it also "exptrapolates". I don't like that.

phated (changes requested):
I'm having some issues with this stuff.

stdlib/number.gr
ospencer: It should fail. If we really don't want it to fail then we can take Jake's suggestion to define this as the gamma function for non-integers.

spotandjake:
I'm converting this to a draft pr until #1404 is completed

spotandjake:
I think i am going to close this as it will require an entire rewrite to use range and the scope of this pr has changed to just be `map`, `lerp` and `clamp`

PR #1376, opened 2022-07-27
feat(stdlib): Add `String.charCodeAt`

spotandjake:
Adds `String.charCodeAt` to the string libary.


.gitignore
```diff
@@ -45,3 +45,5 @@ cli/bin/grainlsp.exe
 cli/bin/grainlsp.bc.js
 
 .DS_Store
+
+yarn.lock
```
ospencer, line 49: Please remove this as we don't use yarn 🙂

ospencer (commented):
@phated Do you think it makes sense to have this versus having to do `Char.code(Char.charAt(...))`? There's not really much of a performance argument to be made, so it's strictly convenience.

phated:
I believe the code would actually be `Char.code(String.charAt(...))` which requires importing 2 different libraries, right? If both were in the same library, I'd say "no" to having this convenience function, but I think avoiding the double import might make it worth it. Thoughts?

ospencer:
Yeah, that makes sense.

stdlib/string.gr
```diff
@@ -221,19 +221,9 @@ let getCodePoint = (ptr: WasmI32) => {
...
- * @returns The character at the provided position
- *
- * @example String.charAt(5, "Hello world") == ' '
- *
- * @since v0.4.0
- */
+
 @unsafe
-export let charAt = (position, string: String) => {
+let _characterAt = (position, string: String) => {
```
ospencer, line 226: Can you rename this to `charAtHelp`?

stdlib/string.gr
```diff
@@ -274,6 +264,38 @@ export let charAt = (position, string: String) => {
   result
 }
 
+/**
+ * Get the character code at the position in the input string.
```
ospencer, line 268:
```suggestion
 * Get the Unicode code point at the position in the input string.
```

ospencer (changes requested):

spotandjake:
Applied Those Changes

ospencer (approved):
Looks great!

ospencer:
Looks like Graindoc needs to be regenerated here as well.

marcusroberts (approved):
LGTM!

phated (approved):
Thanks a ton @spotandjake 🙇  I rebased this and re-applied the doc generation since the `lastIndexOf` merge caused a conflict.

I'll merge when this passes CI.

PR #1385, opened 2022-07-30
fix(grainfmt): Format rational numbers with parens when needed for precedence

### LINKED ISSUE #1374: Grainfmt: Formatting causes math operations to change
I have format on save enabled on the grain extension and when i was trying to run tests for #1373 the formatter was reformatting the test files in a way that caused the tests to fail.

image
The above image shows the tests running fine when the extension is uninstalled and it cannot format.
image
When the formater is run it makes the changes above.
image
which causes the tests to fail.

i think it might be because before it is formated it is treated as 4 / a fraction. and after formatting it is treated as repeated division.

marcusroberts:
Fixes #1374 

When formatting rational numbers, add a parens when the number is on the rhs to ensure correct precedence

The numbers test suite is formatted and passed, but please check for style


marcusroberts:
Converting to draft, because

assert (2/3) - 4.0 == -3.3333333333333335
formats to
assert 2/3 - 4.0 == -3.3333333333333335

(note the 2/3 retains the correct spacing)

but

assert 4 * (2/3) == 8/3
formats to
assert 4 * 2 / 3 == 8/3

which is incorrect as the AST has 2/3 as a rational.    I'll need to fix this first before review

ospencer (approved):

phated (approved):
LGTM!

PR #1386, opened 2022-07-30
fix(grainlsp): Send empty response if file cannot be formatted

### LINKED ISSUE #1377: Formatter Throws Error On Invalid Grain Code
image
As can be seen in the image above the formatter is throwing an error when the grain code is invalid which can be annoying if you have something like format on save enabled and are saving at incomplete times it would make more sense if it just gave the input code back unformated or was a silent error.

marcusroberts:
Fixes #1377 

If the formatter can't parse the source, don't error but just return the source as is.    It removes an error for when the user is saving an incomplete source that won't compile


compiler/src/language_server/formatting.re
```diff
@@ -91,15 +90,22 @@ let process =
...
-      )
+        range_end:
+          // Use Int32.max_int to ensure we fit the entire number in JSON
+          {
+            line: Int32.to_int(Int32.max_int),
+            character: Int32.to_int(Int32.max_int),
+          },
+      };
+
+      let res: ResponseResult.t = Some([{range, newText: compiled_code}]);
```
ospencer, line 107: The spec says you can just send back `null`. Did you try that?

marcusroberts: I didn't, and I did and it works!   Code updated.

ospencer (approved):
Looks great!

compiler/src/language_server/driver.re
```diff
@@ -41,7 +41,7 @@ let process = msg => {
     Code_file.DidChange.process(~uri, ~compiled_code, ~documents, params);
     Reading;
   | Formatting(id, params) when is_initialized^ =>
-    Formatting.process(~id, ~compiled_code, ~documents, params);
+    Formatting.process(~id, ~documents, params);
```
phated, line 44: Why did you remove `~compiled_code`? It is passed to every "behavior" to keep them uniform, even if it is unused.

compiler/src/language_server/protocol.re
```diff
@@ -221,6 +221,26 @@ let response = (~id=?, result) => {
   flush(stdout);
 };
 
+let null_response = id => {
```
phated, line 224:
```suggestion
let empty_response = id => {
```

Should we call this `empty_response`? More importantly, should we just rework `response` to take and Optional value instead of forcing the `Some(result)`? I think that would probably be the best solution overall.

phated (changes requested):
Some API suggestions

compiler/src/language_server/driver.re
marcusroberts: Oooh, that's very opposite to how I work, I only ever pass in what a function needs so its dependencies are clear, but I'll put it back.

compiler/src/language_server/protocol.re
marcusroberts: I'll rename it.  I made it its own function as id is mandatory for the empty response and currently optional for the regular response.

marcusroberts:
@phated  comments addressed I hope

compiler/src/language_server/driver.re
phated: @marcusroberts I'll explain a little more: When writing this code, I envisioned that each "action" the LSP could handle would be a specialization to a module functor that provides an implementation. I didn't implement it that way to see if this pattern would work (and for the sake of time/effort), but for that to work in the future, we'll need to keep the module signatures the same.

compiler/src/language_server/protocol.re
phated: Makes sense! Thanks for the clarification about the ID

phated (approved):
Excellent! Thanks!

PR #1388, opened 2022-07-31
fix(grainfmt): Use only the correct comments for constraints

### LINKED ISSUE #1378: GrainFmt: Grain Formatter Is Duplicating Comments
image
Above is an image before the formatter runs,

image
After the formatter runs it has decided to duplicate the comments and everytime it runs it duplicates them again

marcusroberts:
Use only the comments applicable for the constraint not the whole expression

Fixes #1378

### DIFF
--- a/compiler/test/formatter_outputs/constraints.gr
+++ b/compiler/test/formatter_outputs/constraints.gr
@@ -1,0 +1,8 @@
+import List from "list"
+let test = test => {
+  // Comments
+  List.forEach(i => {
+    // Comment In ForEach
+    print(i)
+  }, test: List<Number>)
+}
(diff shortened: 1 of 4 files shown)


ospencer (approved):

compiler/test/formatter_outputs/constraints.gr
```diff
@@ -0,0 +1,8 @@
+import List from "list"
+let test = test => {
+  // Comments
+  List.forEach(i => {
+    // Comment In ForEach
+    print(i)
```
phated, line 6: So this is just testing they **don't** move? I ask because the input and output look exactly the same to me.

phated (approved):
Just one question, but LGTM!

compiler/test/formatter_outputs/constraints.gr
marcusroberts: That's correct, they were moving before.

PR #1394, opened 2022-08-01
fix(stdlib): Prevent addBytesSlice throwing error on empty buffer

### LINKED ISSUE #1382: buffer: AddBuffer errors when in buffer is empty
image
As can be observed in the image above it would seem that when I try to add an empty buffer it throws an error, if i set change line 129 to `make(1)` the error doesnt occur.

spotandjake:
Fixes #1382, prevents `Buffer.addBuffer` from throwing an exception when adding an empty buffer.


ospencer (approved):

spotandjake:
Added That Test

phated (approved):
Excellent! Thanks for adding that test @spotandjake 🎉

PR #1397, opened 2022-08-02
feat(stdlib): Implement mutable/immutable priority queues

### LINKED ISSUE #1383: Add Priority Queue/Heap Data structure to standard library
[Priority queues](https://en.wikipedia.org/wiki/Priority_queue) are data structures that internally attach a priority value to each element in them. Among other applications, they allow for processing a collection of elements in the order of highest/lowest priority first as needed, and are also useful for sorting elements (see [heap sort](https://en.wikipedia.org/wiki/Heapsort)). I believe that priority queues are a fairly useful data structure, and may be a good idea to include directly in the standard library

Proposal:
- Create a new standard library module named "priorityqueue" (or maybe an alternate casing such as "priority-queue" or "priorityQueue", so far all of the stdlib modules are a single word so I don't know which casing should be used going forward for multi-word module names)
- The priority queue will be a mutable data structure, and will be implemented as a [binary heap](https://en.wikipedia.org/wiki/Binary_heap) with internal storage of elements in an array
- API:
  - `make(comparatorFn: (a, a) -> Bool) -> PriorityQueue<a>`, taking a comparator function determining how priority should be assigned comparatively to elements based on their "objective" value ordering. For example, with a heap of numbers, returning `(firstArg < secondArg)` would create a min-heap, where the smallest elements will be pulled first when polling the priority queue.
  - `makeSized(size: Number, comparatorFn: (a, a) -> Bool) -> PriorityQueue<a>`: like `make` but allows for defining an initial size for the internal array for performance reasons if desired.
  - `size(pq: PriorityQueue<a>) -> Number`: returns the number of elements in the priority queue.
  - `isEmpty(pq: PriorityQueue<a>) -> Bool`: returns whether or not the priority queue has elements in it.
  - `push(val: a, pq: PriorityQueue<a>) -> Void`: add a new element to the priority queue, inserting it in the appropriate priority order.
  - `pop(pq: PriorityQueue<a>) -> Option<a>`: removes and returns the "highest priority" element in the queue in a Some variant, or None if there are no more elements in the queue.
  - `peek(pq: PriorityQueue<a>) -> Option<a>`: returns the "highest priority" element in the queue in a Some variant, or None if there are no more elements in the queue.
  - `drain(pq: PriorityQueue<a>) -> List<a>`: empties the priority queue, returning a list of all of the elements removing.
  - `fromArray(arr: Array<a>) -> Void`: creates a new priority queue, loading all of the elements from the array into the queue.

alexsnezhko3:
As the name suggests, this PR is for implementation of both mutable and immutable priority queues. The mutable priority queue is implemented as a binary heap, and the immutable priority queue is implemented as a skew binomial queue as described in the paper "Optimal Purely Functional Priority Queues" by Chris Okasaki. (O(1) insertions and peeking, O(log(n)) deletions)

This closes #1383 . The API for an immutable priority queue was not really discussed in this issue, but I made one that seemed reasonable to me; let me know if something should be changed:

- `make(comp: (a, a) -> Number) -> ImmutablePriorityQueue<a>`: creates a new PQ with a comparator for determining priority
- `size(pq) -> Number`
- `isEmpty(pq) -> Bool`
- `push(val: a, pq) -> ImmutablePriorityQueue<a>`: creates a new immutable PQ with new element added
- `peek(pq) -> Option<a>`: returns highest priority val wrapped in Some, or None if the PQ is empty
- `pop(pq) -> ImmutablePriorityQueue<a>`: returns an immutable PQ without the highest priority element in the original PQ (if the original PQ is empty it just returns it)
- `drain(pq) -> List<a>`: returns a list of all of the elements in the PQ in priority order; the name of this one I was the least sure about since it's not really "draining" the PQ since it is immutable, but I just went with this name for now for consistency with the mutable PQ
- `fromList(list: List<a>, comp: (a, a) -> Number)`: creates a new PQ containing all of the elements in the list in the priority determined by the comparator

### DIFF
--- a/compiler/test/stdlib/immutablepriorityqueue.test.gr
+++ b/compiler/test/stdlib/immutablepriorityqueue.test.gr
@@ -63,7 +63,7 @@ for (let mut i = 0; i < numVals; i += 1) {
 
   let newPq = ImmutablePriorityQueue.pop(pq)
   let newMaxPq = ImmutablePriorityQueue.pop(maxPq)
-  let expectedSize = (numVals - i)
+  let expectedSize = numVals - i
   assert ImmutablePriorityQueue.size(pq) == expectedSize
   assert ImmutablePriorityQueue.size(maxPq) == expectedSize
   assert ImmutablePriorityQueue.size(newPq) == expectedSize - 1
@@ -82,7 +82,9 @@ assert ImmutablePriorityQueue.drain(pqWithAll) == sortedList
 assert ImmutablePriorityQueue.drain(maxPqWithAll) == List.reverse(sortedList)
 assert ImmutablePriorityQueue.drain(
   ImmutablePriorityQueue.fromList(Array.toList(lotsOfVals), (a, b) => a - b)
-) == sortedList
+) ==
+sortedList
 assert ImmutablePriorityQueue.drain(
   ImmutablePriorityQueue.fromList(Array.toList(lotsOfVals), (a, b) => b - a)
-) == List.reverse(sortedList)
+) ==
+List.reverse(sortedList)
--- a/compiler/test/stdlib/immutablepriorityqueue.test.gr
+++ b/compiler/test/stdlib/immutablepriorityqueue.test.gr
@@ -1,0 +1,87 @@
+import ImmutablePriorityQueue from "immutablepriorityqueue"
+import List from "list"
+import Array from "array"
+
+let lotsOfVals = [>
+  3, 2, 1, 5, 3, 2, 2, 10, 6, 5,
+  6, 4, 9, 10, 10, 5, 9, 6, 1, 3,
+  3, 6, 6, 1, 3, 9, 6, 7, 4, 6,
+  5, 4, 4, 8, 6, 4, 1, 2, 3, 8,
+  8, 8, 10, 9, 6, 8, 4, 1, 1, 2,
+  10, 2, 8, 10, 8, 7, 4, 8, 10, 1,
+  3, 10, 4, 1, 6, 9, 6, 7, 8, 7,
+  10, 4, 5, 10, 3, 2, 4, 5, 9, 5,
+  3, 5, 10, 8, 7, 1, 6, 1, 4, 8,
+  3, 5, 7, 2, 4, 2, 4, 5, 10, 6,
+  7, 7, 7, 1, 10, 4, 9, 9, 10, 10,
+  2, 5, 2, 4, 3, 10, 9, 8, 3, 5,
+  7, 10, 9, 7, 8, 9, 3, 7, 7, 1,
+  7, 7, 6, 3, 3, 5, 1, 8, 9, 10,
+  7, 5, 9, 2, 9, 1, 6, 3, 10, 6,
+  10, 2, 6, 2, 9, 3, 3, 2, 4, 1,
+  9, 4, 9, 2, 9, 4, 8, 5, 6, 7,
+  4, 6, 5, 4, 3, 4, 2, 9, 10, 1,
+  10, 4, 4, 6, 3, 1, 9, 10, 2, 2,
+  5, 5, 3, 4, 5, 8, 3, 8, 4, 10
+]
+let numVals = Array.length(lotsOfVals)
+
+let sortedVals = Array.copy(lotsOfVals)
+Array.sort((a, b) => a - b, sortedVals)
+
+let mut pq = ImmutablePriorityQueue.make((a, b) => a - b)
+let mut maxPq = ImmutablePriorityQueue.make((a, b) => b - a)
+
+assert ImmutablePriorityQueue.size(pq) == 0
+assert ImmutablePriorityQueue.isEmpty(pq)
+assert ImmutablePriorityQueue.peek(pq) == None
+assert ImmutablePriorityQueue.drain(pq) == []
+
+pq = ImmutablePriorityQueue.pop(pq)
+assert ImmutablePriorityQueue.size(pq) == 0
+
+for (let mut i = 0; i < numVals; i += 1) {
+  let newPq = ImmutablePriorityQueue.push(lotsOfVals[i], pq)
+  let newMaxPq = ImmutablePriorityQueue.push(lotsOfVals[i], maxPq)
+  assert ImmutablePriorityQueue.size(pq) == i
+  assert ImmutablePriorityQueue.size(maxPq) == i
+  assert ImmutablePriorityQueue.size(newPq) == i + 1
+  assert ImmutablePriorityQueue.size(newMaxPq) == i + 1
+  pq = newPq
+  maxPq = newMaxPq
+}
+
+let pqWithAll = pq
+let maxPqWithAll = maxPq
+
+for (let mut i = 0; i < numVals; i += 1) {
+  let val = ImmutablePriorityQueue.peek(pq)
+  assert val == Some(sortedVals[i])
+  let maxVal = ImmutablePriorityQueue.peek(maxPq)
+  assert maxVal == Some(sortedVals[numVals - i - 1])
+
+  let newPq = ImmutablePriorityQueue.pop(pq)
+  let newMaxPq = ImmutablePriorityQueue.pop(maxPq)
+  let expectedSize = (numVals - i)
+  assert ImmutablePriorityQueue.size(pq) == expectedSize
+  assert ImmutablePriorityQueue.size(maxPq) == expectedSize
+  assert ImmutablePriorityQueue.size(newPq) == expectedSize - 1
+  assert ImmutablePriorityQueue.size(newMaxPq) == expectedSize - 1
+  pq = newPq
+  maxPq = newMaxPq
+}
+
+assert ImmutablePriorityQueue.size(pq) == 0
+assert ImmutablePriorityQueue.size(pqWithAll) == numVals
+assert !ImmutablePriorityQueue.isEmpty(pqWithAll)
+assert ImmutablePriorityQueue.peek(pq) == None
+
+let sortedList = Array.toList(sortedVals)
+assert ImmutablePriorityQueue.drain(pqWithAll) == sortedList
+assert ImmutablePriorityQueue.drain(maxPqWithAll) == List.reverse(sortedList)
+assert ImmutablePriorityQueue.drain(
+  ImmutablePriorityQueue.fromList(Array.toList(lotsOfVals), (a, b) => a - b)
+) == sortedList
+assert ImmutablePriorityQueue.drain(
+  ImmutablePriorityQueue.fromList(Array.toList(lotsOfVals), (a, b) => b - a)
+) == List.reverse(sortedList)
--- a/compiler/test/stdlib/immutablepriorityqueue.test.gr
+++ b/compiler/test/stdlib/immutablepriorityqueue.test.gr
@@ -2,6 +2,7 @@ import ImmutablePriorityQueue from "immutablepriorityqueue"
 import List from "list"
 import Array from "array"
 
+// formatter-ignore
 let lotsOfVals = [>
   3, 2, 1, 5, 3, 2, 2, 10, 6, 5,
   6, 4, 9, 10, 10, 5, 9, 6, 1, 3,
--- a/compiler/test/stdlib/immutablepriorityqueue.test.gr
+++ b/compiler/test/stdlib/immutablepriorityqueue.test.gr
@@ -28,7 +28,7 @@ let lotsOfVals = [>
 let numVals = Array.length(lotsOfVals)
 
 let sortedVals = Array.copy(lotsOfVals)
-Array.sort((a, b) => a - b, sortedVals)
+Array.sort(compare, sortedVals)
 
 let mut pq = ImmutablePriorityQueue.make((a, b) => a - b)
 let mut maxPq = ImmutablePriorityQueue.make((a, b) => b - a)
--- a/stdlib/immutablepriorityqueue.gr
+++ b/stdlib/immutablepriorityqueue.gr
@@ -62,9 +62,9 @@ export let make = comp => {
 }
 
 /**
- * Returns the number of elements in the priority queue.
+ * Gets the number of elements in a priority queue.
  * 
- * @param pq: The priority queue to get the size of
+ * @param pq: The priority queue to inspect
  * @returns The number of elements in the priority queue
  *
  * @since v0.5.3
@@ -74,10 +74,10 @@ export let size = pq => {
 }
 
 /**
- * Returns whether or not the priority queue is empty.
+ * Determines if the priority queue contains no elements.
  * 
  * @param pq: The priority queue to check
- * @returns `true` if the priority queue has no elements and `false` otherwise
+ * @returns `true` if the priority queue is empty and `false` otherwise
  *
  * @since v0.5.3
  */
@@ -119,11 +119,11 @@ let skewInsert = (comp, val, pq) => {
 }
 
 /**
- * Returns a new priority queue with the given element inserted into the given priority queue.
+ * Produces a new priority queue by inserting the given element into the given priority queue.
  * 
  * @param val: The value to add into the priority queue
  * @param pq: The priority queue
- * @returns A new priority queue with the given element inserted into the given priority queue
+ * @returns A new priority queue with the given element inserted
  */
 export let push = (val, pq) => {
   let { comp, size, root } = pq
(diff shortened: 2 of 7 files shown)


stdlib/immutablepriorityqueue.gr
```diff
@@ -0,0 +1,324 @@
...
+  // split out the values with nodes of rank 0, which will all be singleton nodes
+  let (rankZeroVals, nonRankZeroNodes) = splitRankZero([], [], hpNode.children)
+
+  let mergedPq = merge(comp, nonHpNodes, nonRankZeroNodes)
+  List.reduce((pq, val) => skewInsert(comp, val, pq), mergedPq, rankZeroVals)
+}
+
+let rec findHighestPriority = (comp, pq) => {
+  match (pq) {
+    // empty list case should never happen; this is here just to satisfy compiler
```
ospencer, line 254:
```suggestion
    // empty list case should never happen; this is here just to satisfy the compiler
```

stdlib/immutablepriorityqueue.gr
```diff
@@ -0,0 +1,324 @@
...
+    },
+  }
+}
+
+/**
+ * Returns a new priority queue without the highest-priority element in the given priority queue
+ * in a `Some` variant, or `None` if the priority queue is empty.
+ * 
+ * @param pq: The priority queue
+ * @returns A new priority queue without the highest-priority element in the input priority queue
```
ospencer, line 269:
```suggestion
 * Returns a new priority queue without the highest priority element in the given priority queue
 * in a `Some` variant, or `None` if the priority queue is empty.
 * 
 * @param pq: The priority queue
 * @returns A new priority queue without the highest priority element in the input priority queue
```

stdlib/immutablepriorityqueue.gr
```diff
@@ -0,0 +1,324 @@
...
+      None => acc,
+      Some(root) => drainRec([root.rootVal, ...acc], pop(pq)),
+    }
+  }
+  List.reverse(drainRec([], pq))
+}
+
+/**
+ * Constructs a new priority queue initialized with the elements in the list, and
+ * with a custom comparator function used to assign priority.
```
ospencer, line 314:
```suggestion
 * Constructs a new priority queue initialized with the elements in the list, using
 * a custom comparator function to assign priority.
```

stdlib/priorityqueue.gr
```diff
@@ -0,0 +1,223 @@
...
+ * @returns An empty priority queue
+ *
+ * @since v0.5.3
+ */
+export let makeSized = (size, comp) => {
+  { size: 0, array: Array.make(size, None), comp }
+}
+
+/**
+ * Creates a new priority queue with a comparator to use for determining element priority.
```
ospencer, line 84: Probably worth copying what I wrote about the comparator for the immutable priority queue here as well.

stdlib/priorityqueue.gr
```diff
@@ -0,0 +1,223 @@
...
+  // reorder heap to ensure that binary heap property of parent nodes having
+  // larger values than their children is upheld
+  siftUp(pq.size - 1, pq)
+}
+
+/**
+ * Returns the highest-priority element in the priority queue in a `Some`
+ * variant, or `None` if the priority queue is empty.
+ * 
+ * @param pq: The priority queue to get the highest-priority element of
```
ospencer, line 150:
```suggestion
 * Returns the highest priority element in the priority queue in a `Some`
 * variant, or `None` if the priority queue is empty.
 * 
 * @param pq: The priority queue to get the highest priority element of
```

stdlib/priorityqueue.gr
```diff
@@ -0,0 +1,223 @@
...
+  } else {
+    pq.array[0]
+  }
+}
+
+/**
+ * Removes and returns the highest-priority element in the priority queue in a `Some`
+ * variant, or `None` if the priority queue is empty.
+ * 
+ * @param pq: The priority queue to remove and get the highest-priority element of
```
ospencer, line 167:
```suggestion
 * Removes and returns the highest priority element in the priority queue in a `Some`
 * variant, or `None` if the priority queue is empty.
 * 
 * @param pq: The priority queue to remove and get the highest priority element of
```

stdlib/priorityqueue.gr
```diff
@@ -0,0 +1,223 @@
...
+    pq.size -= 1
+    // reorder heap to ensure that binary heap property of parent nodes having
+    // larger values than their children is upheld
+    siftDown(0, pq)
+    root
+  }
+}
+
+/**
+ * Clears the priority queue, and returns all of the elements in the priority
```
ospencer, line 189:
```suggestion
 * Clears the priority queue and returns all of the elements in the priority
```

stdlib/priorityqueue.gr
```diff
@@ -0,0 +1,223 @@
...
+    match (pop(pq)) {
+      Some(val) => drainRec([val, ...acc]),
+      None => acc,
+    }
+  }
+  List.reverse(drainRec([]))
+}
+
+/**
+ * Constructs a new priority queue initialized with the elements in the list, and
```
ospencer, line 208:
```suggestion
 * Constructs a new priority queue initialized with the elements in the list using
```

ospencer (commented):
Really well done! This looks absolutely excellent. Left a few documentation suggestions but otherwise LGTM.

One other thought I had: should the immutable priority queue implementation of `pop` just return an empty queue if the input is empty? That's what the `Queue` implementation does.

ospencer:
It looks like you'll also need to run `npm run compiler format` to get the CI to pass.

alexsnezhko3:
> should the immutable priority queue implementation of `pop` just return an empty queue if the input is empty?

I believe that's what I have in my implementation; if the root node on the input priority queue is None it just returns the input queue, which should effectively be the same as creating a new empty queue. I realize now that my doc comment for pop was incorrect though, thanks for bringing that to my attention.

ospencer (approved):
This is sweet! Thanks again for implementing this 🎉

marcusroberts (approved):
Wow, excellent stuff!    Thank you so much for the contribution!

stdlib/immutablepriorityqueue.gr
```diff
@@ -0,0 +1,332 @@
...
+
+let uniquify = (comp, pq) => {
+  match (pq) {
+    [] => [],
+    [node, ...rest] => ins(comp, node, rest),
+  }
+}
+
+// step through the trees of two priority queues in increasing rank order, perform-
+// ing a simple link whenever we find two trees of equal rank
```
phated, line 196:
```suggestion
// step through the trees of two priority queues in increasing rank order,
// performing a simple link whenever we find two trees of equal rank
```
nit

stdlib/immutablepriorityqueue.gr
```diff
@@ -0,0 +1,332 @@
...
+    [node] => node.val,
+    [node, ...rest] => {
+      let currMin = findHighestPriority(comp, rest)
+      if (comp(node.val, currMin) <= 0) node.val else currMin
+    },
+  }
+}
+
+/**
+ * Returns a new priority queue without the highest priority element in the
```
phated, line 269:
```suggestion
 * Produces a new priority queue without the highest priority element in the
```

stdlib/immutablepriorityqueue.gr
```diff
@@ -0,0 +1,332 @@
...
+ * given priority queue. If the input priority queue is empty, this function will
+ * return it.
+ * 
+ * @param pq: The priority queue
+ * @returns A new priority queue without the highest priority element in the input priority queue
```
phated, line 274:
```suggestion
 * @returns A new priority queue without the highest priority element
```

stdlib/immutablepriorityqueue.gr
```diff
@@ -0,0 +1,332 @@
...
+          }
+        )
+      }
+      { comp, size: size - 1, root: newRoot }
+    },
+  }
+}
+
+/**
+ * Returns all of the elements in the priority queue in priority order.
```
phated, line 300:
```suggestion
 * Produces a list of all elements in the priority queue in priority order.
```

stdlib/immutablepriorityqueue.gr
```diff
@@ -0,0 +1,332 @@
...
+}
+
+/**
+ * Constructs a new priority queue initialized with the elements in the list
+ * using a custom comparator function, which is used to determine priority of
+ * elements. The comparator function takes two elements and must return 0 if
+ * both share priority, a positive number if the first has greater priority,
+ * and a negative number if the first has less priority.
+ * 
+ * @param list: A list of values to initialize the priority queue with
```
phated, line 324:
```suggestion
 * @param list: A list of values used to initialize the priority queue
```

stdlib/immutablepriorityqueue.gr
```diff
@@ -0,0 +1,332 @@
...
+ * @param comp: A comparator function used to assign priority to elements
+ * @returns A priority queue containing the elements in the list
```
phated, line 326:
```suggestion
 * @returns A priority queue containing the elements from the list
```

stdlib/immutablepriorityqueue.md
```diff
@@ -0,0 +1,248 @@
...
+
+```grain
+import ImmutablePriorityQueue from "immutablepriorityqueue"
+```
+
+## Types
+
+Type declarations included in the ImmutablePriorityQueue module.
+
+### Immutablepriorityqueue.**ImmutablePriorityQueue**
```
phated, line 20: @ospencer we've never had this type of module name so graindoc doesn't camelcase it. Thoughts?

stdlib/priorityqueue.gr
```diff
@@ -0,0 +1,233 @@
...
+ * @section Values: Functions for working with PriorityQueues.
+ */
+
+let swap = (i1, i2, array) => {
+  let t = array[i2]
+  array[i2] = array[i1]
+  array[i1] = t
+}
+
+let get = (array, i) => Option.unwrap(array[i])
```
phated, line 37:
```suggestion
let get = (array, i) => Option.unwrap(array[i])
```
Should we use `Option.expect` here so it would actually fail with a better error message if it was somehow `None`?

stdlib/priorityqueue.gr
```diff
@@ -0,0 +1,233 @@
...
+    siftUp(parentI, pq)
+  }
+}
+
+/**
+ * Creates a new priority queue with a given internal storage size and a
+ * comparator function, which is used to determine priority of elements. The
+ * comparator function takes two elements and must return 0 if both share
+ * priority, a positive number if the first has greater priority, and a
+ * negative number if the first has less priority.
```
phated, line 74:
```suggestion
 * negative number if the first has less priority.
 *
 * Generally, you won't need to care about the storage size of your priority queue and can use `PriorityQueue.make()` instead.
```

stdlib/priorityqueue.gr
```diff
@@ -0,0 +1,233 @@
...
+ * 
+ * @param size: The number of elements to initialize the internal storage array with
```
phated, line 76:
```suggestion
 * @param size: The initial storage size of the map
```

stdlib/priorityqueue.gr
```diff
@@ -0,0 +1,233 @@
...
+ * @since v0.5.3
+ */
+export let makeSized = (size, comp) => {
+  { size: 0, array: Array.make(size, None), comp }
+}
+
+/**
+ * Creates a new priority queue with a comparator function, which is used to
+ * determine priority of elements. The comparator function takes two elements
+ * and must return 0 if both share priority, a positive number if the first
```
phated, line 89: Same question about "which comes first" if priority is equal

stdlib/priorityqueue.gr
```diff
@@ -0,0 +1,233 @@
+/**
```
phated, line 1: Can you apply the same changes from the suggestions in immutable to the docs in this file?

stdlib/priorityqueue.gr
```diff
@@ -0,0 +1,233 @@
...
+ */
+export let isEmpty = pq => {
+  pq.size == 0
+}
+
+/**
+ * Adds a new element to the priority queue.
+ * 
+ * @param val: The value to add into the priority queue
+ * @param pq: The priority queue to add a new element to
```
phated, line 132:
```suggestion
 * @param pq: The priority queue to update
```

phated (changes requested):
This PR is 🔥

I love seeing high-quality PRs like this!

Most of my suggestions are related to docs. I'm the stickler at trying to achieve consistency. Please keep the updates as separate commits so I can quickly skim 🙇

compiler/test/stdlib/immutablepriorityqueue.test.gr
```diff
@@ -0,0 +1,88 @@
+import ImmutablePriorityQueue from "immutablepriorityqueue"
```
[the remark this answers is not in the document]
alexsnezhko3: Eh, after working with enterprise Java I'm not phased by this stuff anymore lol

alexsnezhko3:
@phated are you saying to have a separate commit for each individual comment resolution?

phated:
> are you saying to have a separate commit for each individual comment resolution?

Nope, you can accept all of the suggestions as a group, but I meant the updates to `priorityqueue.gr` could be a separate commit for me to review (since I didn't make many suggestions).

stdlib/immutablepriorityqueue.gr
```diff
@@ -0,0 +1,332 @@
...
+}
+
+/**
+ * @section Values: Functions for working with ImmutablePriorityQueues.
+ */
+
+/**
+ * Creates a new priority queue with a comparator function, which is used to
+ * determine priority of elements. The comparator function takes two elements
+ * and must return 0 if both share priority, a positive number if the first
```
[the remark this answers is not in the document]
alexsnezhko3: I don't believe so; this algorithm doesn't have any special treatment for equal-priority items.

stdlib/immutablepriorityqueue.gr
```diff
@@ -0,0 +1,332 @@
...
+
+/**
+ * Returns a new priority queue with the given element inserted into the given priority queue.
+ * 
+ * @param val: The value to add into the priority queue
+ * @param pq: The priority queue
+ * @returns A new priority queue with the given element inserted into the given priority queue
+ */
+export let push = (val, pq) => {
+  let { comp, size, root } = pq
```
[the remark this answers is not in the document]
alexsnezhko3: I left it as `pq` because I assumed that grain doc uses the parameter name when generating documentation

stdlib/priorityqueue.gr
alexsnezhko3: In theory this shouldn't happen but I guess that would be a good idea in case there is a bug somewhere

stdlib/immutablepriorityqueue.gr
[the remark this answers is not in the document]
phated: 👍  That makes sense. We don't enforce that right now because the identifiers are not in the module interface, but it's something we want to do in the future.

stdlib/priorityqueue.gr
phated: In the compiler, we often do something like `"impossible by SOMETHING_UNIQUE"` so we can track down those bugs. It looks like regex.gr has a few things like `"impossible (mergeAdjacent)"` and `"Impossible: continue_m"`

stdlib/priorityqueue.gr
alexsnezhko3: Like the immutable implementation, currently equal-priority elements may be ordered arbitrarily (from the user's perspective). I think I've heard of an implementation of binary heaps that allow equal-priority elements to be served in the order they were inserted, but I'm not so sure about how to do that efficiently with the immutable implementation so if API consistency is a priority then maybe this fact is alright?

compiler/test/stdlib/immutablepriorityqueue.test.gr
```diff
@@ -0,0 +1,88 @@
...
+  10, 2, 6, 2, 9, 3, 3, 2, 4, 1,
+  9, 4, 9, 2, 9, 4, 8, 5, 6, 7,
+  4, 6, 5, 4, 3, 4, 2, 9, 10, 1,
+  10, 4, 4, 6, 3, 1, 9, 10, 2, 2,
+  5, 5, 3, 4, 5, 8, 3, 8, 4, 10
+]
+let numVals = Array.length(lotsOfVals)
+
+let sortedVals = Array.copy(lotsOfVals)
+Array.sort((a, b) => a - b, sortedVals)
```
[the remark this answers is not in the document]
ospencer: Yes!

stdlib/immutablepriorityqueue.md
phated: This will be fixed by #1406 so we can merge this PR and then fix up the generated docs in that PR

phated (approved):
I rebased this on main and updated the `Array.sort` to use `compare` (and fixed a formatter fail)

Love this PR. Thanks @alexsnezhko3 🎉

alexsnezhko3:
Happy to help!

PR #1399, opened 2022-08-03
feat: Implement Pervasives.compare

### LINKED ISSUE #305: Check Safety and Efficiency of Numeric Comparison Functions
Currently, our numeric comparison functions (sans `==`) are implemented by coercing numbers into `f64`s and doing the check. This works well enough, but we could probably do it more efficiently if we wrote out a proper (large) implementation of each function. Moreover, we should double-check that we aren't losing anything important precision-wise with this conversion (specifically around the relative ordering of very large `i64` values). **We should be able to guarantee to users that the comparison of two numbers of the same representation (e.g. (i64,i64), (i32, i32), etc.) is correct.**

ospencer:
`compare` can be used for sorting or other kinds of ordering operations. Works on all types. This also improves the accuracy of all of our comparison operators.

Closes #305

### DIFF
--- a/compiler/test/stdlib/pervasives.test.gr
+++ b/compiler/test/stdlib/pervasives.test.gr
@@ -1,0 +1,63 @@
+// compare
+
+// Number comparisons are tested thoroughly by other tests
+assert compare(0, 1) < 0
+assert compare(0, 0) == 0
+assert compare(1, 0) > 0
+// Specific NaN and Infinity tests
+assert compare(1 / 0.0, 987654321) > 0
+assert compare(1 / 0.0, 987654321.) > 0
+assert compare(0.0 / 0.0, -1 / 0.0) < 0
+assert compare(0.0 / 0.0, 987654321.) < 0
+assert compare(0.0 / 0.0, 987654321) < 0
+assert compare(0.0 / 0.0, 0) < 0
+// Booleans
+assert compare(false, true) < 0
+assert compare(true, false) > 0
+assert compare(true, true) == 0
+// Void
+assert compare(void, void) == 0
+// Chars
+assert compare('a', 'c') < 0
+assert compare('d', 'c') > 0
+assert compare('d', 'd') == 0
+assert compare('🌾', '🙂') < 0
+assert compare('💯', '🌾') > 0
+// Strings
+assert compare("", "") == 0
+assert compare("", "abc") < 0
+assert compare("abc", "def") < 0
+assert compare("abc", "d") < 0
+assert compare("ab", "abc") < 0
+assert compare("abc", "abc") == 0
+assert compare("abc", "ab") > 0
+// Tuples
+assert compare((1, 2), (3, 4)) < 0
+assert compare((1, 2), (1, 4)) < 0
+assert compare((1, 2), (1, 2)) == 0
+assert compare(('d', 2), ('c', 2)) > 0
+// Arrays
+assert compare([>], [>]) == 0
+assert compare([>], [> 1]) < 0
+assert compare([> 1, 2], [> 3, 4]) < 0
+assert compare([> 1, 2], [> 1, 4]) < 0
+assert compare([> 1, 2], [> 1, 2]) == 0
+// Enums
+enum Comparable {
+  A,
+  B(Bool),
+  C,
+}
+assert compare(A, A) == 0
+assert compare(B(true), B(true)) == 0
+assert compare(A, B(true)) < 0
+assert compare(C, B(true)) > 0
+assert compare(B(false), B(true)) < 0
+// Records
+record Comparable2 {
+  a: Number,
+  b: Bool,
+  c: Void,
+}
+assert compare({ a: 1, b: true, c: void }, { a: 1, b: true, c: void }) == 0
+assert compare({ a: 1, b: true, c: void }, { a: 1, b: false, c: void }) > 0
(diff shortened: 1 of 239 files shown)


marcusroberts (approved):
Looks excellent!

stdlib/pervasives.gr
```diff
@@ -218,6 +219,19 @@ export (<=)
...
 export (>=)
 
+/**
+ * Compares the first argument to the second argument and produces an integer result.
+ * Provides a consistent ordering over all types and is suitable for sorting and other kinds of ordering.
+ * `compare` treats `NaN` differently than the other comparison operators in that it considers `NaN` equal to itself and smaller than any other number.
+ *
+ * @param num1: The first operand
+ * @param num2: The second operand
+ * @returns A negative integer if the first operand is less than the second operand, `0` if they are equal, or a positive integer otherwise
```
phated, line 229: It's not just `-1` and `1`? Is there some sort of scale?

stdlib/runtime/compare.gr
```diff
@@ -0,0 +1,179 @@
...
+import WasmI64 from "runtime/unsafe/wasmi64"
+import Memory from "runtime/unsafe/memory"
+import Tags from "runtime/unsafe/tags"
+import { tagSimpleNumber } from "runtime/dataStructures"
+
+primitive (!): Bool -> Bool = "@not"
+primitive (||): (Bool, Bool) -> Bool = "@or"
+primitive (&&): (Bool, Bool) -> Bool = "@and"
+
+import { isNumber, cmp as numberCompare } from "runtime/numbers"
```
phated, line 27: Is there a reason this import is hanging out alone?

stdlib/runtime/numbers.gr
```diff
@@ -1721,81 +1730,286 @@ let cmpBigInt = (x: WasmI32, y: WasmI32) => {
...
+      },
+      t when WasmI32.eq(t, Tags._GRAIN_INT64_BOXED_NUM_TAG) => {
+        let yi = boxedInt64Number(y)
+        if (WasmI64.ltS(xi, yi)) -1n else if (WasmI64.gtS(xi, yi)) 1n else 0n
+      },
+      t when WasmI32.eq(t, Tags._GRAIN_BIGINT_BOXED_NUM_TAG) => {
+        WasmI32.sub(0n, cmpBigInt(y, x))
+      },
+      t when WasmI32.eq(t, Tags._GRAIN_RATIONAL_BOXED_NUM_TAG) => {
+        // Rationals and ints are never considered equal
```
phated, line 1834: So `16/16 == 1` is never true? I guess it makes sense but also seems a little weird (especially if I'm adding rationals to get to `1`)

compiler/test/stdlib/pervasives.test.gr
```diff
@@ -0,0 +1,54 @@
+// compare
+
+// Number comparisons are tested thoroughly by other tests
+assert compare(0, 1) < 0
+assert compare(0, 0) == 0
+assert compare(1, 0) > 0
```
phated, line 6: I think we should have a test for `NaN` (given the code comments throughout) and maybe even `Infinity` (?)

phated (commented):
I *love* polymorphic compare! A few questions and comments before approval

compiler/test/stdlib/pervasives.test.gr
```diff
@@ -0,0 +1,54 @@
...
+// Booleans
+assert compare(false, true) < 0
+assert compare(true, false) > 0
+assert compare(true, true) == 0
+// Void
+assert compare(void, void) == 0
+// Chars
+assert compare('a', 'c') < 0
+assert compare('d', 'c') > 0
+assert compare('d', 'd') == 0
```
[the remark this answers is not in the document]
ospencer: I can add some 🙂

stdlib/pervasives.gr
ospencer: There's no scale! I believe languages implement it this way just to make it exhaustive and allow simple implementations like `a - b` that don't need any branching logic. I believe this implementation will ever only return one of `-1, 0, 1` though (I'd need to double-check a few things).

stdlib/runtime/compare.gr
ospencer: Copypasta from `equal` that did it the same way 😛

stdlib/runtime/numbers.gr
ospencer: This might need some clarification! An invariant of the number system is that rationals with a denominator of `1` do not exist, so writing the literal `16/16` gets you `1/1` which Grain just simplifies to simply `1`. So `compare(16/16, 1)` will always be zero, but this case doesn't need to worry about it because of the invariants. They're all listed in this file, but I think it might be nice to have one paragraph at the start that explains them. Shall I make an issue to get that added?

compiler/test/stdlib/pervasives.test.gr
ospencer: I could have sworn I added some!

stdlib/runtime/numbers.gr
phated: Sure, an issue would be nice.

stdlib/runtime/compare.gr
phated: Looks like you updated `equal` but not this file 🙃

phated (approved):
Still needs a rebase but LGTM!

PR #1389, opened 2022-07-31
chore(compiler): Format stdlib tests

ospencer:

Issue #1390, opened 2022-07-31
Grainfmt: Indent binops when wrapping


ospencer:
Right now if you format `foo && bar && baz` (imagine those are all long things that cause wrapping) you get
```
foo &&
bar &&
baz
```
but if would be nice if this formatted to
```
foo &&
  bar &&
  baz
```
for reading clarity.

In particular when I see something like `assert foo == bar` get formatted to
```
assert foo ==
bar
```
I find it a little hard to read vs
```
assert foo ==
  bar
```

However, this should not happen in places like `if` conditions:
```
if (
  foo &&
  bar &&
  baz
) ...
```

PR #1400, opened 2022-08-04
fix(grainfmt): Indent exception primitives if wrapped

marcusroberts:
Tiny fix that indents exception primitives if wrapped, should help with the stdlib tests reformat in #1389

### DIFF
--- a/compiler/src/formatting/format.re
+++ b/compiler/src/formatting/format.re
@@ -2062,7 +2062,7 @@ and print_other_application =
     Doc.concat([
       print_expression(~original_source, ~comments, func),
       Doc.space,
-      print_expression(~original_source, ~comments, first_expr),
+      Doc.indent(print_expression(~original_source, ~comments, first_expr)),
     ])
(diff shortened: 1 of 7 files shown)


compiler/src/formatting/format.re
```diff
@@ -1773,15 +1789,31 @@ and print_infix_application =
...
-    Doc.concat([
-      lhs,
-      Doc.space,
-      Doc.text(function_name),
-      trailing_line_comments,
-      after_comments_docs,
-      force_break_if_line_comment(~separator=Doc.nil, after_comments),
-      rhs,
-    ]);
+    if (true) {
```
phated, line 1792: 🤔

marcusroberts: Nooo.    My commit has  picked up my WIP

Let me force remove the commit

marcusroberts:
@ospencer  highlighted the change wasn't quite as simple as I thought, and my work on #1390 will be needed first.

PR #1405, opened 2022-08-05
chore(stdlib): Resolve #304

### LINKED ISSUE #304: Check Overflow Safety of Numeric Operations
With the introduction of the new number system, there are a couple of loose ends which need to be checked for overflow-safety:

- In `numberPlusMinusSimpleHelp`, we coerce the simple int (`x ∈ (-1 << 30, 1 << 30)`) into a `f32` and add it to the already-f32 RHS. This could probably overflow. What is the best way to make it more hardened?
- In `numberPlusMinusRationalHelp` and `numberTimesDivideRationalHelp`, our `(Rational, Rational)` operations work by performing the grade-school algorithm of multiplying together the appropriate numerators and denominators and then simplifying to a reduced fraction. This is guarded against overflows, but I'm sure there is a smarter way to do these operations which will permit a larger domain of fractions to be usable with these functions.
- In `numberTimesDivideInt64Help`, we perform `(Int64, Float32)` multiplication/division by coercing the `i64` into a `f32` and doing the operation. This should be safe, but I want to make someone else think about it too before declaring it as such.

ospencer:
Closes #304 

The operations have been validated as safe and the rational operations are overflow proof.

### DIFF
--- a/stdlib/runtime/numbers.gr
+++ b/stdlib/runtime/numbers.gr
@@ -895,7 +895,6 @@ let numberAddSubSimpleHelp = (x, y, isSub) => {
         let xval = WasmF32.convertI32S(xval)
         let result = if (isSub) WasmF32.sub(xval, yBoxedVal)
           else WasmF32.add(xval, yBoxedVal)
-        // TODO(#304): is this safe?
         newFloat32(result)
       },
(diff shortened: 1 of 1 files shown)


phated (approved):
LGTM

marcusroberts (approved):
LGTM!

PR #1408, opened 2022-08-05
WIP: Generate release docs in CI

phated:
This is a draft because we would need to build the entire compiler just to regenerate the docs. I'm going to do this by hand right before we release (for now).


phated:
My idea to solve the "build the compiler" part is that we could have built compilers uploaded by our normal CI steps and then this workflow could download and use that compiler. I don't want to implement that right now 🙃

phated:
Closing this as this was completed by #1777

Issue #1403, opened 2022-08-04
Immutable versions of mutable data structures


alexsnezhko3:
As alluded to in #1383, it would be a good idea to include immutable variants for the currently existing mutable data structures (namely `Set` and `Map`). Here is my proposal for these:

- Implement both as balanced binary trees
- `ImmutableSet` API (mostly mimics existing set):
  - `make() -> ImmutableSet<a>`
  - `add(val: a, set) -> ImmutableSet<a>`: create a new set with given value inside of it if it doesn't already exist
  - `contains(set) -> Bool`
  - `remove(val: a, set) -> ImmutableSet<a>`: creates new set w/o given value if it exists
  - `size(set) -> Number`,
  - `isEmpty(set) -> Bool`,
  - `forEach(fn: (a -> Void), set) -> Void` (will do an in-order traversal)
  - `reduce(fn: ((a, b) -> a), init: a, set) -> a` (will do an in-order traversal)
  - `filter(fn: (a -> Bool), set) -> ImmutableSet<a>`: creates new set with only elements in original that fulfill predicate
  - `reject(fn: (a -> Bool), set) -> ImmutableSet<a>`: creates new set with only elements in original that don't fulfill predicate
  - `toList, fromList, toArray, fromArray`: same as in `Set` (toList/toArray will do in-order traversal)
  - `union(set1, set2) -> ImmutableSet<a>`
  - `diff(set1, set2) -> ImmutableSet<a>`
  - `intersect(set1, set2) -> ImmutableSet<a>`
- `ImmutableMap` API (mostly mimics existing map)
  - `make() -> ImmutableMap<k, v>`
  - `set(key: k, val: v, map) -> ImmutableMap<k, v>` (maybe another name would be more appropriate, "set" kind of implies mutation in my mind although I can't think of a better name)
  - `get(key: k, map) -> Option<v>`
  - `contains(key: k, map) -> Bool`
  - `remove(key: k, map) -> ImmutableMap<k, v>`
  - `update(key: k, fn: (Option<v> -> Option<v>), map)`
  - `size(map) -> Number`
  - `isEmpty(map) -> Bool`
  - `forEach(fn: ((k, v) -> Void), map) -> Void` (will do an in-order traversal)
[...]

PR #1414, opened 2022-08-09
feat(stdlib): Implement immutable map and immutable set

alex-snezhko:
I figured I'd include both maps and sets in one PR since their implementation is almost identical. They are both implemented as size-balanced BSTs, following "Implementing Sets Efficiently in a Functional Language" by Stephen Adams. API for these is as was discussed in #1403

### DIFF
--- a/compiler/test/stdlib/immutableset.test.gr
+++ b/compiler/test/stdlib/immutableset.test.gr
@@ -16,7 +16,7 @@ record ResourceData {
 
 // ImmutableSet.isEmpty
 
-let mut e = ImmutableSet.make()
+let mut e = ImmutableSet.empty
 
 assert ImmutableSet.isEmpty(e)
 e = ImmutableSet.remove("🌾", e)
@@ -42,7 +42,7 @@ assert !ImmutableSet.contains("🌳", h)
 
 // ImmutableSet.add
 
-let mut vars = ImmutableSet.make()
+let mut vars = ImmutableSet.empty
 
 vars = ImmutableSet.add(Grain, vars)
 vars = ImmutableSet.add(Sheep, vars)
@@ -50,7 +50,7 @@ vars = ImmutableSet.add(Grain, vars)
 
 assert ImmutableSet.size(vars) == 2
 
-let mut recs = ImmutableSet.make()
+let mut recs = ImmutableSet.empty
 
 recs = ImmutableSet.add({ name: "Grain", emoji: "🌾" }, recs)
 recs = ImmutableSet.add({ name: "Sheep", emoji: "🐑" }, recs)
@@ -81,7 +81,7 @@ assert ImmutableSet.isEmpty(r)
 
 let makeTestSet = () => ImmutableSet.fromList([Grain, Sheep, Brick])
 
-let emptySet = ImmutableSet.make()
+let emptySet = ImmutableSet.empty
 
 let mut filterTestSet = makeTestSet()
 
--- a/compiler/test/stdlib/immutableset.test.gr
+++ b/compiler/test/stdlib/immutableset.test.gr
@@ -81,11 +81,9 @@ assert ImmutableSet.isEmpty(r)
 
 let makeTestSet = () => ImmutableSet.fromList([Grain, Sheep, Brick])
 
-let emptySet = ImmutableSet.empty
-
 let mut filterTestSet = makeTestSet()
 
-ImmutableSet.filter(key => fail "Shouldn't be called", emptySet)
+ImmutableSet.filter(key => fail "Shouldn't be called", ImmutableSet.empty)
 filterTestSet = ImmutableSet.filter(key => key == Sheep, filterTestSet)
 
 assert !ImmutableSet.contains(Grain, filterTestSet)
@@ -96,7 +94,7 @@ assert !ImmutableSet.contains(Brick, filterTestSet)
 
 let mut rejectTestSet = makeTestSet()
 
-ImmutableSet.reject(key => fail "Shouldn't be called", emptySet)
+ImmutableSet.reject(key => fail "Shouldn't be called", ImmutableSet.empty)
 rejectTestSet = ImmutableSet.reject(key => key == Sheep, rejectTestSet)
 
 assert ImmutableSet.contains(Grain, rejectTestSet)
@@ -109,7 +107,7 @@ let reduceTestSet = ImmutableSet.fromList([1, 3, 2, 5, 4])
 
 let result = ImmutableSet.reduce((acc, key) => fail "Shouldn't be called",
 0,
-emptySet
+ImmutableSet.empty
 )
 
 assert result == 0
@@ -128,7 +126,7 @@ assert result == [5, 4, 3, 2, 1]
 
 let forEachTestSet = makeTestSet()
 
-ImmutableSet.forEach(key => fail "Shouldn't be called", emptySet)
+ImmutableSet.forEach(key => fail "Shouldn't be called", ImmutableSet.empty)
 
 let mut called = 0
 
--- a/compiler/test/stdlib/immutableset.test.gr
+++ b/compiler/test/stdlib/immutableset.test.gr
@@ -1,0 +1,199 @@
+import ImmutableSet from "immutableset"
+import List from "list"
+import Array from "array"
+
+// Data types used in multiple tests
+enum Resource {
+  Grain,
+  Sheep,
+  Brick,
+  Wood,
+}
+record ResourceData {
+  name: String,
+  emoji: String,
+}
+
+// ImmutableSet.isEmpty
+
+let mut e = ImmutableSet.make()
+
+assert ImmutableSet.isEmpty(e)
+e = ImmutableSet.remove("🌾", e)
+assert ImmutableSet.isEmpty(e)
+let newE = ImmutableSet.add("🌾", e)
+assert !ImmutableSet.isEmpty(newE)
+assert ImmutableSet.isEmpty(e)
+
+// ImmutableSet.size
+
+let s = ImmutableSet.fromList(["🌾", "🐑", "🧱"])
+
+assert ImmutableSet.size(s) == 3
+
+// ImmutableSet.contains
+
+let h = ImmutableSet.fromList(["🌾", "🐑", "🧱"])
+
+assert ImmutableSet.contains("🌾", h)
+assert ImmutableSet.contains("🐑", h)
+assert ImmutableSet.contains("🧱", h)
+assert !ImmutableSet.contains("🌳", h)
+
+// ImmutableSet.add
+
+let mut vars = ImmutableSet.make()
+
+vars = ImmutableSet.add(Grain, vars)
+vars = ImmutableSet.add(Sheep, vars)
+vars = ImmutableSet.add(Grain, vars)
+
+assert ImmutableSet.size(vars) == 2
+
+let mut recs = ImmutableSet.make()
+
+recs = ImmutableSet.add({ name: "Grain", emoji: "🌾" }, recs)
+recs = ImmutableSet.add({ name: "Sheep", emoji: "🐑" }, recs)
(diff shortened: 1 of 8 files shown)


stdlib/immutablemap.gr
```diff
@@ -0,0 +1,484 @@
...
+/**
+ * Determines if the map contains no key-value pairs.
+ *
+ * @param map: The map to inspect
+ * @returns `true` if the given map is empty or `false` otherwise
+ *
+ * @since v0.5.4
+ */
+export let isEmpty = map => {
+  size(map) == 0
```
ospencer, line 79: I think a `match` to check for the `Empty` case might be slightly better!

stdlib/immutablemap.gr
```diff
@@ -0,0 +1,484 @@
...
+}
+
+// helper function for creating a tree node with correct size from
+// two balanced trees
+let makeNode = (key, val, left, right) => {
+  Tree({ key, val, size: 1 + size(left) + size(right), left, right })
+}
+
+// note: see Figure 1 of paper referenced above for visual illustration of
+// the below rotations
```
ospencer, line 89:
```suggestion
// the rotations below
```

stdlib/immutablemap.gr
```diff
@@ -0,0 +1,484 @@
...
+
+// node rotation moving the left subtree of the right node to the left side
+let singleL = (key, val, left, right) => {
+  let Tree({ key: rKey, val: rVal, left: rl, right: rr, _ }) = right
```
ospencer, line 93: I believe this causes compiler warnings since it's not exhaustive. It's a bit safer to do a pattern match for assignments like these.

stdlib/immutablemap.gr
```diff
@@ -0,0 +1,484 @@
...
+ * Determines if the map contains the given key. In such a case, it will always contain a value for the given key.
+ *
+ * @param key: The key to search for
+ * @param map: The map to search
+ * @returns `true` if the map contains the given key or `false` otherwise
+ *
+ * @since v0.5.4
+ */
+export let rec contains = (key, map) => {
+  get(key, map) != None
```
ospencer, line 208: How about `Option.isSome` or a pattern match?

ospencer (changes requested):
Looking really good! A couple of small things but otherwise this is excellent 😄

stdlib/immutablemap.gr
alexsnezhko3: In this case it _should_ be impossible for the tree to be empty; perhaps I'll mimic an `Option.expect` here with a custom "Impossible: ..." message so that it's at least easier to tell what's going wrong if there's a bug?

alexsnezhko3: Also as a follow-up this case actually doesn't cause a compiler warning

ospencer (approved):
Sorry for the review delay here! This looks absolutely fantastic 🤩

stdlib/immutablemap.gr
```diff
@@ -0,0 +1,495 @@
...
+// Functional Language" by Stephen Adams
+
+record Node<k, v> {
+  key: k,
+  val: v,
+  size: Number,
+  left: ImmutableMap<k, v>,
+  right: ImmutableMap<k, v>,
+},
+enum ImmutableMap<k, v> {
```
phated, line 22:
```suggestion

/**
 * @section Types: Type declarations included in the ImmutableMap module.
 */

enum ImmutableMap<k, v> {
```

You'll want to expose this opaque type in its own section

stdlib/immutablemap.gr
```diff
@@ -0,0 +1,495 @@
...
+  Empty,
+  Tree(Node<k, v>),
+}
+
+/**
+ * @section Values: Functions for working with ImmutableMaps.
```
phated, line 28:
```suggestion
 * @section Values: Functions and constants for working with ImmutableMaps.
```
Because we are changing to `empty` being a constant

stdlib/immutablemap.gr
```diff
@@ -0,0 +1,495 @@
...
+ *
+ * @since v0.5.4
+ */
+export let forEach = (fn, map) => {
+  let rec forEachInner = node => {
+    match (node) {
+      Empty => void,
+      Tree({ key, val, left, right, _ }) => {
+        forEachInner(left)
+        fn(key, val)
```
phated, line 305:
```suggestion
        fn(key, val): Void
```

As with other `forEach` in the stdlib, this needs to be annotated with `Void` otherwise users can return any value. I caught this by looking over the generated markdown

stdlib/immutableset.gr
```diff
@@ -0,0 +1,499 @@
...
+// implementation based on the paper "Implementing Sets Efficiently in a
+// Functional Language" by Stephen Adams
+
+record Node<a> {
+  key: a,
+  size: Number,
+  left: ImmutableSet<a>,
+  right: ImmutableSet<a>,
+},
+enum ImmutableSet<a> {
```
phated, line 20: This needs a section comment. We might be waiting on a formatter fix because of the recursive definition?

stdlib/immutableset.gr
```diff
@@ -0,0 +1,499 @@
...
+/**
+ * Creates a new, empty set.
+ *
+ * @returns An empty set
+ * 
+ * @since v0.5.4
+ */
+export let make = () => {
+  Empty
+}
```
phated, line 43: This needs to become `empty` like mentioned in ImmutableMap

stdlib/immutableset.gr
```diff
@@ -0,0 +1,499 @@
...
+
+// creates a new node after either the left or right trees have just had an
+// element inserted or removed from them, maintaining balance in the tree
+let balancedNode = (key, left, right) => {
+  let makeNodeFn = if (size(left) + size(right) < 2) {
+    makeNode
+  } else if (size(right) > weight * size(left)) {
+    // if the right tree is too much larger than the left then move part of
+    // the right tree to the left side
+    let Tree({ left: rl, right: rr, _ }) = right
```
phated, line 134: See ImmutableMap. These can probably use `unwrapTree`, right?

stdlib/immutableset.gr
```diff
@@ -0,0 +1,499 @@
...
+    if (size(rl) < size(rr)) singleL else doubleL
+  } else if (size(left) > weight * size(right)) {
+    // if the left tree is too much larger than the right then move part of
+    // the left tree to the right side
+    let Tree({ left: ll, right: lr, _ }) = left
```
phated, line 139: See ImmutableMap. These can probably use `unwrapTree`, right?

stdlib/immutableset.gr
```diff
@@ -0,0 +1,499 @@
...
+
+  makeNodeFn(key, left, right)
+}
+
+/**
+ * Produces a new set by inserting the given value into the set. If the value
+ * already exists, the new set will have the same elements as the input set.
+ *
+ * @param key: The value to add
+ * @param set: The set to add to
```
phated, line 155: I don't like "add to" as the end of this sentence. It's better than "modify" like in ImmutableMap but we can do better.

stdlib/immutableset.gr
```diff
@@ -0,0 +1,499 @@
...
+    },
+  }
+}
+
+/**
+ * Produces a new set without the given element. If the value doesn't exist in
+ * the set, the set will be returned unmodified.
+ *
+ * @param key: The value to exclude
+ * @param set: The set to exclude the value from
```
phated, line 220:
```suggestion
 * @param set: The set to exclude from
```

stdlib/immutableset.gr
```diff
@@ -0,0 +1,499 @@
...
+ * 
+ * @since v0.5.4
+ */
+export let forEach = (fn, set) => {
+  let rec forEachInner = node => {
+    match (node) {
+      Empty => void,
+      Tree({ key, left, right, _ }) => {
+        forEachInner(left)
+        fn(key)
```
phated, line 252:
```suggestion
        fn(key): Void
```
Guessing this needs Void annotated too

phated (changes requested):
This PR is excellent! I ran into a few things while reviewing and commented inline. Thanks for all your work and I'm excited to get this updated and merged 🎉

stdlib/immutablemap.gr
```diff
@@ -0,0 +1,495 @@
...
+}
+
+/**
+ * @section Values: Functions for working with ImmutableMaps.
+ */
+
+// semi-arbitrary value chosen for algorithm for determining when to balance
+// trees; no tree can have a left subtree containing this number of times
+// more elements than its right subtree or vice versa
+let weight = 4
```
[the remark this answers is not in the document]
alex-snezhko: If something like this were to be done then should the regular `make` function still be removed?

stdlib/immutablemap.gr
```diff
@@ -0,0 +1,495 @@
...
+
+// creates a new node after either the left or right trees have just had an
+// element inserted or removed from them, maintaining balance in the tree
+let balancedNode = (key, val, left, right) => {
+  let makeNodeFn = if (size(left) + size(right) < 2) {
+    makeNode
+  } else if (size(right) > weight * size(left)) {
+    // if the right tree is too much larger than the left then move part of
+    // the right tree to the left side
+    let Tree({ left: rl, right: rr, _ }) = right
```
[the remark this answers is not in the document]
alex-snezhko: It actually doesn't, which I assume to be a bug? I can make an issue for this.

stdlib/immutablemap.gr
```diff
@@ -0,0 +1,495 @@
...
+
+  makeNodeFn(key, val, left, right)
+}
+
+/**
+ * Produces a new map containing a new key-value pair. If the key already exists in the map, the value is replaced.
+ *
+ * @param key: The unique key in the map
+ * @param value: The value to store
+ * @param map: The map to modify
```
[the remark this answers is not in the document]
alex-snezhko: Maybe something like "The map to use as a basis for insertion"?

stdlib/immutablemap.gr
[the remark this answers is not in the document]
phated: Thanks for opening the issue! Let's switch to unwrapTree though so that we don't need to update this later (once the warning starts to appear)

stdlib/immutablemap.gr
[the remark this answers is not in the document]
phated: I think `make` should always be removed because we would want to be explicit with the name, such as `makeWeighted`. If we ever get optional arguments, that choice might change.

stdlib/immutablemap.gr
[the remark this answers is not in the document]
alex-snezhko: I think "the base map" sounds good personally

stdlib/immutableset.gr
alex-snezhko: Yeah, it looks like you're right. Should I just include `Node` in the section as well with a TODO comment for when the formatter issue gets fixed?

phated: Perhaps you just need a rebase because I believe that was solved with https://github.com/grain-lang/grain/commit/be4b38d87b0ee6ff4544264b8bad4e36820808a2

compiler/test/stdlib/immutableset.test.gr
```diff
@@ -0,0 +1,199 @@
...
+r = ImmutableSet.remove("🐑", r)
+r = ImmutableSet.remove("🧱", r)
+
+assert ImmutableSet.isEmpty(r)
+
+// ImmutableSet.filter
+
+let makeTestSet = () => ImmutableSet.fromList([Grain, Sheep, Brick])
+
+let emptySet = ImmutableSet.empty
```
phated, line 84:
```suggestion
```
Can you remove this `emptySet` and replace the usages with `ImmutableSet.empty` directly?

phated (commented):
Just one change and I think we are good to go.

phated (approved):
Thanks for all the awesome work on this! 🎉

PR #1419, opened 2022-08-16
feat(compiler): Allow custom infix operators

### LINKED ISSUE #1319: Parser disallows defining <<< as an operator
When trying to define the `<<<` operator in some code (below), the parser fails with this error:
> Syntax error after '<<' and before '<'.
> Expected `)` to complete the pattern.

```gr
let (<<<) = (p, q, state) => {
    let x = p(state)
    q(state): Void
    x
}
```

ospencer:
Closes #1319 

I'll use what I wrote in the docs as my PR description 🤓

> Grain also allows the definition of custom infix operators. Custom operators in Grain must be prefixed by one of the operators in the table below. Its precedence is defined by its prefix.

> In addition to these prefixes, the parser accepts the characters `$&*/+-=><^|!?%:.` for use in the names of custom operators.

> It should be noted that `/*` and `//` are disallowed as operators as they denote comments.


jozanza (approved):

marcusroberts (approved):
Looks good and thanks for updating the formatter too!

phated (changes requested):
As discussed on the call. I'd love to implement `??` as `Option.unwrapWithDefault`

ospencer:
@phated Updated to support `??`

peblair (approved):
Looks great!

phated (approved):
Heck yeah! Thanks for `??` @ospencer 🎉

PR #1422, opened 2022-08-18
feat(js-runner): Generate ESM browser bundle

phated:
This adds an ESM browser bundle to the files we generate for js-runner.

I need this to be able to bundle the browser playground with vite.

I tried using `merge` with the browser config, but I couldn't figure out how to "unset" `output.library` which cannot be specified when building a module.


phated:
I'm actually not using this anymore so I'm going to close it.

PR #1423, opened 2022-08-23
chore(compiler): Upgrade to Binaryen 108

phated:
I was discussing with @ospencer about Binaryen versions and he reminded me that Binaryen 108 allowed us to disable the building of their "tools". This update should reduce build times a fair bit.


ospencer (approved):
Hooray!

marcusroberts (approved):

PR #1425, opened 2022-08-26
fix(compiler): Allow building on Fedora 36 by updating ocaml

phated:
This updates ocaml to 4.12.1001, which seems to contain a patch that allows ocaml to build on newer glibc (such as the one shipped with Fedora 36).


marcusroberts (approved):

ospencer (approved):

phated:
Looks like this has the same linking issues as all the new ocaml versions.

PR #1428, opened 2022-09-02
fix(grainfmt): Fix comments in records

### LINKED ISSUE #1415: GrainFmt: Grain Formatter Is Duplicating Comments
Similar to #1378 the grain formatter is duplicating my comments again though this time in a record type definition. the first image shows the state before formatting and second image shows the duplicated comments after, every additional format adds another comment. I am running grain v0.5.3 so the fix for #1378 has been implemeneted in the version i am running.
image
image

marcusroberts:
Fixes #1415


compiler/src/formatting/format.re
```diff
@@ -3644,10 +3644,22 @@ let rec print_data =
       ]);
     };
 
+    let pre_brace_comments = []; // We can't determine from AST if comment comens before or after brace
```
ospencer, line 3647:
```suggestion
    let pre_brace_comments = []; // We can't determine from AST if comment comes before or after brace
```

ospencer (approved):

marcusroberts:
I think I've found another related problem whilst testing #1429, working on it now

marcusroberts:
Found an edge case I need to fix.

marcusroberts:
The edge case is different, I'll raise a new issue

ospencer (approved):

compiler/src/formatting/format.re
```diff
@@ -3672,22 +3684,11 @@ let rec print_data =
...
-              ~remove_comments=after_angle_comments,
-              comments,
-            );
-
           let param_items =
             item_iterator(
               ~get_loc,
               ~print_item,
-              ~comments=cleaned_comments,
+              ~comments=[],
```
phated, line 3691: Does this mean that comments inside parameterized types are no longer support? Do we have a test for something like `type Foo<A /* something */, B> = (A, B)`?

compiler/test/formatter_inputs/records.gr
```diff
@@ -55,8 +74,5 @@ let { // a comment 2
   longlonglongnamenamename3
 } = x
 
-let { // a comment 3
```
phated, line 58: Why was this removed?

phated (changes requested):
A few questions I need answered.

compiler/src/formatting/format.re
marcusroberts: This is broken, but not by this change, so I'll fix

marcusroberts:
I need to fix #1444  first

compiler/src/formatting/format.re
marcusroberts: Fixed in #1460

compiler/test/formatter_inputs/records.gr
marcusroberts: Not sure, it's back now anyway

phated (commented):
This LGTM but I want #1460 to land first

phated:
@marcusroberts please resolve the conflicts with `main` now that #1460 is merged 🙇

marcusroberts:
Conflicts removed!

phated (approved):

PR #1429, opened 2022-09-05
fix(grainfmt): Handle comments within lists

### LINKED ISSUE #1417: GrainFmt: Grain Formatter Removing Comments
When grain formats lists containing comments it deletes them as can be seen by the before and after below.
Before formatting ```
let test = [
  1, // test
  2, // test
  3, // test
``` after formatting ```
let test = [ 1, 2, 3]
```

marcusroberts:
Comments within lists were being discarded.

Fixes #1417

### DIFF
--- a/compiler/src/formatting/format.re
+++ b/compiler/src/formatting/format.re
@@ -1072,7 +1072,7 @@ and resugar_list =
         | Regular(e) =>
           last_item_was_spread := false;
 
-          // do we have any comments on this line?
+          // Do we have any comments on this line?
           // If so, we break the whole list
 
           let (_, item_line, item_char, _) =
--- a/compiler/src/formatting/format.re
+++ b/compiler/src/formatting/format.re
@@ -1149,10 +1149,8 @@ and resugar_list =
   let last_line_breaks_for_comments = ref(false);
   let items_length = List.length(items);
   let list_items =
-    Doc.concat(
       List.mapi(
-        (i, pair) => {
-          let (item, item_comments) = pair;
+        (i, (item, item_comments)) => {
 
           let final_item = items_length - 1 == i;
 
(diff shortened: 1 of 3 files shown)


marcusroberts:
Just checking why the stdlib format has changed, it shouldn't have had an impact

marcusroberts:
Switching to draft, as the format has changed for compiler/test/stdlib/hash.test.gr as a list of records now breaks over line.

marcusroberts:
All good now, but discovered #1428 needs more work, but this can go in independently.

compiler/src/formatting/format.re
```diff
@@ -1065,41 +1065,150 @@ and resugar_list =
...
 
+  // We have to compose this list by hand because of the complexity of if a list item
+  // is followed by a comment, the comma must come before the comment.
+  // It also impacts how we force a new line for a line ending comment at the end of a list
+  // without introducing an extra blank line when bringing the indentation back in again
+
+  let last_line_breaks_for_comments = ref(false);
+  let items_length = List.length(items);
+  let list_items =
+    Doc.concat(
```
phated, line 1152: Can this concat be moved below (line 1199) instead? It feels weird here and makes the nesting harder to read

compiler/src/formatting/format.re
```diff
@@ -1065,41 +1065,150 @@ and resugar_list =
...
+      List.mapi(
+        (i, pair) => {
+          let (item, item_comments) = pair;
```
phated, line 1155: Is `pair` used anywhere or should this destructure happen in the function definition?

phated (commented):
A couple comments. This fix was a bit harder to follow that my other approvals (I think it is due to so many `Doc.concat` combinations) but the tests look good.

marcusroberts:
Thanks for the great comments, fixed as suggested.

ospencer (approved):
Looks great!

phated (approved):

PR #1430, opened 2022-09-05
fix(grainfmt): Write comments found between top level data statements

### LINKED ISSUE #1317: Grainformat: Wipes out docblock comments between types declared together
Given the following chunk of code, the formatter wipes out the comment:

```grain
enum EventType {
  Enter,
  Exit,
},
/**
 *   An event is the start or end of a token amongst other events.
 *   Tokens can “contain” other tokens, even though they are stored in a flat
 *   list, through `enter`ing before them, and `exit`ing after them.
 */
type Event = (EventType, Token, TokenizeContext)
```

marcusroberts:
Fixes #1317 

Formatter was ignoring comments between top level data declarations

Fixed and tests added


phated (approved):
Test looks great and fix makes sense! In the future, we *might* want to refactor to avoid the boxed value, but I think it's good for now.

ospencer (approved):

PR #1432, opened 2022-09-06
fix(utils): Use option converter instead of Some

renatoalencar:
When trying to use --help for grainc, this makes the docs printer for enum look for a None value on the enum list, which yields an Invalid_argument exception, crashing the grainc executable.


ospencer (approved):
Great catch, thank you!

phated (approved):
Thanks!

marcusroberts (approved):
Well spotted!  Thanks!

PR #1436, opened 2022-09-08
fix(grainfmt): Stop adding an extra line after block comments

### LINKED ISSUE #1310: Grainformat: Extra newlines between comments
When two comments come after one another, grainformat will add extra newlines between them. This is a problem because it breaks graindoc.

Here's an example:

```grain
/**
 * Some graindoc documentation
 *
 * @param foo: bar
 * @returns foo
 */
// TODO: We need to do something
let baz = foo => foo
```

In the above example, formatting the code will insert an empty line between the graindoc comment and `// TODO` comment. Graindoc only works if it is on 1 (or 2 because of annotations)  lines before the definition.

marcusroberts:
Fixes #1310

### DIFF
--- a/compiler/src/formatting/comment_utils.re
+++ b/compiler/src/formatting/comment_utils.re
@@ -521,7 +521,7 @@ let rec new_comments_inner =
 
       | 1 => [
           switch (prev_cmt) {
-          | Line(_) 
+          | Line(_)
           | Doc(_) => Doc.nil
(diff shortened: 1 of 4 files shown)


phated (approved):
Looks great! Thanks for the fix @marcusroberts

Now we need to go through the stdlib TODO comments and move them below the docblock 😛

ospencer (approved):

PR #1440, opened 2022-10-03
fix(stdlib): Fix anchoring behavior in Regex.replaceAll

### LINKED ISSUE #1431: Regex `replaceAll` bug with ^ character
I believe if you are attempting to do something like
```
let re = Result.unwrap(Regex.make("^."))
print(Regex.replaceAll(re, "asdf", "-"))
```
The result should be `-sdf`, but the actual value printed is `----`, as if it is taking every character to be the beginning of the string

peblair:
Fixes #1431.

### DIFF
--- a/stdlib/regex.gr
+++ b/stdlib/regex.gr
@@ -3954,11 +3954,12 @@ let regexReplaceHelp =
   let mut out = []
   let rec loop = searchPos => {
     let state = Array.make(rx.reNumGroups, None)
+    let inStart = max(0, searchPos - rx.reMaxLookbehind)
     let poss = searchMatch(
       rx,
       buf,
       searchPos,
-      searchPos,
+      inStart,
       Array.length(buf.matchInputExploded),
       state
     )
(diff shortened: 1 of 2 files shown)


stdlib/regex.gr
```diff
@@ -3954,11 +3954,12 @@ let regexReplaceHelp =
   let mut out = []
   let rec loop = searchPos => {
     let state = Array.make(rx.reNumGroups, None)
+    let inStart = max(0, searchPos - rx.reMaxLookbehind)
     let poss = searchMatch(
       rx,
       buf,
       searchPos,
-      searchPos,
```
peblair, line 3961: This was breaking the anchoring behavior, since the "am I at the start of the input?" check always succeeded due to us saying that the input started at `searchPos` (the current position). For the regular expression `^.`, the lookbehind is `1`, so this will be `0` when `searchPos == 0` and `searchPos - 1` for every other position.

ospencer (approved):

phated (approved):
Thanks for digging into this! 🙇‍♂️

PR #1442, opened 2022-10-04
feat(stdlib): Optimize string trimming

spotandjake:
This pr optimizes`String.trim` to reduce the allocations and makes the function use Wasmi32 types reducing some overhead of the grain number type.


stdlib/string.gr
```diff
@@ -1867,30 +1867,49 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
-    i = charsLength - 1
-    offset = -1
+    i = charsLength - 1n
+    offset = -1n
   }
-  for (; i < charsLength && i > -1; i += offset) {
-    let currentChar = chars[i]
+  let mut count = 0n
+  for (; i < charsLength; i += offset) {
+    // Get The Character
```
ospencer, line 1887:
```suggestion
    // Get the byte
```

ospencer: With the explode we got actual characters, but with this approach, we're just loading bytes.

stdlib/string.gr
```diff
@@ -1867,30 +1867,49 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+    let byte = WasmI32.load8U(stringPtr, i)
```
ospencer, line 1888: It's important to note that since the string is UTF-8, this byte may not necessarily be a full code point.

stdlib/string.gr
```diff
@@ -1867,30 +1867,49 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
-      currentChar != '\u{000C}' && // FORM FEED (FF)
-      currentChar != '\u{0020}' && // Space
-      currentChar != '\u{00A0}' && // No Break Space
-      currentChar != '\u{FEFF}' && // ZERO WIDTH NO-BREAK SPACE
+      byte != 0x09n && // Tab
+      byte != 0x0Bn && // LINE TABULATION
+      byte != 0x0Cn && // FORM FEED (FF)
+      byte != 0x20n && // Space
+      byte != 0xA0n && // No Break Space
+      byte != 0xFEFFn && // ZERO WIDTH NO-BREAK SPACE
```
ospencer, line 1896: ...which comes into play here, where this value is the code point value, but not what's loaded from the single byte. The UTF-8 representation is the byte sequence `EF BB BF`: https://unicode-table.com/en/FEFF/

You'll need to handle this case specially... If the value loaded is `0xEF` I'd probably just load a full word and check the bytes: `(WasmI32.load(stringPtr) >>> 8n) != 0xBFBBEF` (note that the bytes are in reverse order because wasm is little-endian).

ospencer: Also we should add a test.

ospencer (changes requested):

spotandjake:
Applied Those Changes

stdlib/string.gr
ospencer: Can you update the comment to reflect this?

stdlib/string.gr
```diff
@@ -1867,30 +1867,58 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
-  for (; i < charsLength && i > -1; i += offset) {
-    let currentChar = chars[i]
+  let mut count = 0n
+  for (; i < charsLength; i += offset) {
+    // Get The Byte
+    let byte = WasmI32.load8U(stringPtr, i)
     // TODO: Use unicode whitespace property and unicode line terminator once github issue #661 is completed
+    // Check Multi-Byte Characters
+    if (byte == 0xEFn && WasmI32.load(stringPtr, i - 1n) >>> 8n == 0xBFBBEFn) {
+      i += offset * 2n
```
ospencer, line 1895: Can you explain this?

stdlib/string.gr
```diff
@@ -1867,30 +1867,58 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+      count += 3n
+      continue
```
ospencer, line 1897: It took me a moment to wrap my head around why this was `continue` and not `break`. Could you add a comment?

stdlib/string.gr
spotandjake: I can add a comment if you want explaining bassically we want to move the index forward so we are not scanning half the Character but we might be moving backwardss or forwards in the loop so we use `offset * 2` to get the sign.

spotandjake: It is not `i +=2` because when we call `trimEnd`, `trimString` goes in reverse

spotandjake:
Made Those Changes

stdlib/string.gr
ospencer: Grain definitely uses UTF-8, sir! I meant that one byte may not be a full codepoint because UTF-8 codepoints span between 1 and 4 bytes. Could you update the comment again?

stdlib/string.gr
```diff
@@ -1867,30 +1867,78 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+    offset = -1n
   }
-  for (; i < charsLength && i > -1; i += offset) {
-    let currentChar = chars[i]
+  let mut count = 0n
+  for (; i < byteLength; i += offset) {
+    // Get The Byte, May not be a full codepoint as grain uses utf-16
+    let byte = WasmI32.load8U(stringPtr, i)
     // TODO: Use unicode whitespace property and unicode line terminator once github issue #661 is completed
+    // Check For The BOM
```
ospencer, line 1893: This comment feels misplaced.

stdlib/string.gr
```diff
@@ -1867,30 +1867,78 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+    if (offset == 1n) {
```
ospencer, line 1894: This has me thinking the code would be cleaner if we just had an explicit boolean that said we were going forwards or backwards.

stdlib/string.gr
```diff
@@ -1867,30 +1867,78 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+      // We need to check for the start character if we are going backwards
+      if (
+        byte == 0xEFn && WasmI32.load(stringPtr, i - 1n) >>> 8n == 0xBFBBEFn
```
ospencer, line 1897: I know I gave you this incantation but I think we need to explain it a little more. Let me know if you need help.

stdlib/string.gr
```diff
@@ -1867,30 +1867,78 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+      ) {
+        // We need to increment i in the direction of the loop
+        i += 2n
+        // The count is incremented by 3 because that is the length of the BOM
+        count += 3n
+        // We want to skip checking the single bytes
```
ospencer, line 1903:
```suggestion

```

stdlib/string.gr
```diff
@@ -1867,30 +1867,78 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+    } else {
+      // We need to check for the end character if we are going backwards
+      if (
+        byte == 0xBFn && WasmI32.load(stringPtr, i - 3n) >>> 8n == 0xBFBBEFn
+      ) {
+        // We need to decrement i in the direction of the loop
+        i -= 2n
+        // The count is incremented by 3 because that is the length of the BOM
+        count += 3n
+        // We want to skip checking the single bytes
```
ospencer, line 1915:
```suggestion

```

stdlib/string.gr
```diff
@@ -1867,30 +1867,78 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+      byte != 0x0Dn && // LF
+      byte != 0x0An && // CR
+      byte != 0x09n && // Tab
+      byte != 0x0Bn && // LINE TABULATION
+      byte != 0x0Cn && // FORM FEED (FF)
+      byte != 0xA0n // No Break Space
+    ) {
+      break
+    }
+    // Add to count
```
ospencer, line 1930:
```suggestion

```

stdlib/string.gr
```diff
@@ -1867,30 +1867,78 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
   }
-  if (end) slice(0, i + 1, str) else slice(i, charsLength, str)
+  let str = allocateString(byteLength - count)
+  // Copy the string
+  if (end) {
+    Memory.copy(str + 8n, stringPtr, byteLength - count)
+  } else {
+    Memory.copy(str + 8n, stringPtr + count, byteLength - count)
+  }
+  // Return string
```
ospencer, line 1940:
```suggestion

```

ospencer (changes requested):

stdlib/string.gr
spotandjake: `Get The Byte, May not be a full codepoint as grain uses utf-8, a codepoint may be 1-4 bytes`, is this wording preferred I didn't mean to write `utf-16` oops

spotandjake:
Applied Those Changes

stdlib/string.gr
```diff
@@ -1867,30 +1867,76 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+  if (fromEnd) {
+    i = byteLength - 1n
+    offset = -1n
   }
-  for (; i < charsLength && i > -1; i += offset) {
-    let currentChar = chars[i]
-    // TODO: Use unicode whitespace property and unicode line terminator once github issue #661 is completed
+  let mut count = 0n
+  for (; i < byteLength; i += offset) {
+    // Get The Byte, May not be a full codepoint as grain uses utf-8, a codepoint may be 1-4 bytes
```
ospencer, line 1890:
```suggestion
    // Get the byte, not necessarily a full UTF-8 codepoint
```

stdlib/string.gr
```diff
@@ -1867,30 +1867,76 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+    let byte = WasmI32.load8U(stringPtr, i)
+    // TODO(#661): Use unicode whitespace property and unicode line terminator
```
ospencer, line 1892: We can leave this here, but just a note that this may not be possible to do anymore because of the byte-level approach.

stdlib/string.gr
```diff
@@ -1867,30 +1867,76 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+    // Handle The BOM Character which is a full codepoint, The Bytes are 0xEF,0xBB,0xBF
```
ospencer, line 1893:
```suggestion
```

stdlib/string.gr
```diff
@@ -1867,30 +1867,76 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+    if (!fromEnd) {
+      if (
+        byte == 0xEFn && // Check for the BOM identifier at the beginning of the codepoint
+        // Check for the full BOM codepoint, This is revered because wasm is little-endian
```
ospencer, line 1897:
```suggestion
        // Check for the full BOM codepoint, 0xEFBBBF. Bytes are reversed because wasm is little-endian
```

stdlib/string.gr
```diff
@@ -1867,30 +1867,76 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+        WasmI32.load(stringPtr, i - 1n) >>> 8n == 0xBFBBEFn
+      ) {
+        // We need to increment i in the direction of the loop
```
ospencer, line 1900:
```suggestion
```

stdlib/string.gr
```diff
@@ -1867,30 +1867,76 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+        i += 2n
+        // The count is incremented by 3 because that is the length of the BOM
```
ospencer, line 1902:
```suggestion
```

stdlib/string.gr
```diff
@@ -1867,30 +1867,76 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+        count += 3n
+        continue
+      }
+    } else {
+      if (
+        byte == 0xBFn && // Check for the BOM identifier at the end of the codepoint
+        // Check for the full BOM codepoint, This is revered because wasm is little-endian
```
ospencer, line 1909:
```suggestion
        // Check for the full BOM codepoint, 0xEFBBBF. Bytes are reversed because wasm is little-endian
```

stdlib/string.gr
```diff
@@ -1867,30 +1867,76 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
-    // TODO: Use unicode whitespace property and unicode line terminator once github issue #661 is completed
+  let mut count = 0n
+  for (; i < byteLength; i += offset) {
+    // Get The Byte, May not be a full codepoint as grain uses utf-8, a codepoint may be 1-4 bytes
+    let byte = WasmI32.load8U(stringPtr, i)
+    // TODO(#661): Use unicode whitespace property and unicode line terminator
+    // Handle The BOM Character which is a full codepoint, The Bytes are 0xEF,0xBB,0xBF
+    if (!fromEnd) {
+      if (
+        byte == 0xEFn && // Check for the BOM identifier at the beginning of the codepoint
```
ospencer, line 1896:
```suggestion
        byte == 0xEFn && // Check for the first BOM byte
```

stdlib/string.gr
```diff
@@ -1867,30 +1867,76 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+      ) {
+        // We need to increment i in the direction of the loop
+        i += 2n
+        // The count is incremented by 3 because that is the length of the BOM
+        count += 3n
+        continue
+      }
+    } else {
+      if (
+        byte == 0xBFn && // Check for the BOM identifier at the end of the codepoint
```
ospencer, line 1908:
```suggestion
        byte == 0xBFn && // Check for the last BOM byte
```

stdlib/string.gr
```diff
@@ -1867,30 +1867,76 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+        // Check for the full BOM codepoint, This is revered because wasm is little-endian
+        WasmI32.load(stringPtr, i - 3n) >>> 8n == 0xBFBBEFn
+      ) {
+        // We need to decrement i in the direction of the loop
```
ospencer, line 1912:
```suggestion
```

stdlib/string.gr
```diff
@@ -1867,30 +1867,76 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+        i -= 2n
+        // The count is incremented by 3 because that is the length of the BOM
```
ospencer, line 1914:
```suggestion
```

spotandjake:
Applied Those Suggestions

ospencer (approved):
Looks good!

stdlib/string.gr
```diff
@@ -1867,30 +1867,71 @@ export let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+      byte != 0xA0n // No Break Space
+    ) {
+      break
+    }
+    count += 1n
   }
-  if (end) slice(0, i + 1, str) else slice(i, charsLength, str)
+  let str = allocateString(byteLength - count)
+  // Copy the string
+  if (end) {
```
ospencer, line 1929:
```suggestion
  if (fromEnd) {
```
looks like there's one last thing to fix here

PR #1446, opened 2022-10-14
chore(tests): Use alternate string syntax (#479)

ribeirotomas1904:
Please, feel free to request any changes

### DIFF
--- a/compiler/test/suites/exceptions.re
+++ b/compiler/test/suites/exceptions.re
@@ -27,13 +27,20 @@ describe("exceptions", ({test, testSkip}) => {
   );
   assertRunError(
     "throw_exception_2",
-    "exception HorribleError(String, Bool, Number); let _ = throw HorribleError(\"oh no\", false, 1/3)",
+    {|
+      exception HorribleError(String, Bool, Number)
+      let _ = throw HorribleError("oh no", false, 1/3)
+    |},
     "HorribleError\\(\"oh no\", false, 1/3\\)",
   );
   assertRunError(
     ~check_exists=false,
     "throw_exception_3",
-    "exception HorribleError(String, Bool, Number); let _ = throw HorribleError(\"oh no\", false, 1/3); print(\"shouldn't be printed\")",
+    {|
+      exception HorribleError(String, Bool, Number)
+      let _ = throw HorribleError("oh no", false, 1/3)
+      print("shouldn't be printed")
+    |},
     "shouldn't be printed",
   );
   assertRunError(
--- a/compiler/test/suites/imports.re
+++ b/compiler/test/suites/imports.re
@@ -14,175 +14,313 @@ describe("imports", ({test, testSkip}) => {
   /* import * tests */
   assertRun(
     "import_all",
-    "import * from \"exportStar\"; {print(x); print(y(4)); print(z)}",
+    {|
+      import * from "exportStar"
+      {
+        print(x)
+        print(y(4))
+        print(z)
+      }
+    |},
     "5\n4\nfoo\n",
   );
   assertRun(
     "import_all_except",
-    "import * except {y} from \"exportStar\"; {print(x); print(z)}",
+    {|
+      import * except {y} from "exportStar"
+      {
+        print(x)
+        print(z)
+      }
+    |},
     "5\nfoo\n",
   );
   assertSnapshot(
     "import_all_except_multiple",
-    "import * except {x, y} from \"exportStar\"; z",
+    {|
+      import * except {x, y} from "exportStar"
+      z
+    |},
   );
   assertSnapshot(
     "import_all_constructor",
-    "import * from \"tlists\"; Cons(2, Empty)",
+    {|
+      import * from "tlists"
+      Cons(2, Empty)
+    |},
   );
   assertSnapshot(
     "import_all_except_constructor",
-    "import * except {Cons} from \"tlists\"; Empty",
+    {|
+      import * except {Cons} from "tlists"; Empty
+    |},
   );
   assertSnapshot(
     "import_all_except_multiple_constructor",
-    "import * except {Cons, append} from \"tlists\"; sum(Empty)",
+    {|
+      import * except {Cons, append} from "tlists"; sum(Empty)
+    |},
   );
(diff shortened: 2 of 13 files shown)


compiler/test/suites/pattern_matching.re
```diff
@@ -205,15 +221,15 @@ describe("pattern matching", ({test, testSkip}) => {
   );
   assertSnapshot(
     "constant_match_2",
-    "match ((\"foo\", 5, false)) { (\"bar\", 5, false) => false, (\"foo\", _, true) => false, (\"foo\", _, false) => true, _ => false }",
+    {|match (("foo", 5, false)) { ("bar", 5, false) => false, ("foo", _, true) => false, ("foo", _, false) => true, _ => false }|},
```
ospencer, line 224: It could be nice to split these ones onto multiple lines 🙂

ospencer (commented):
Overall looks great!

compiler/test/suites/pattern_matching.re
ribeirotomas1904: about the ones that are like that but are not using the alternate string syntax, should they be kept the same or should I break lines with the normal string syntax? as this one
image

ospencer: I think either is fine.

ribeirotomas1904: updated

compiler/test/suites/exceptions.re
```diff
@@ -27,13 +27,20 @@ describe("exceptions", ({test, testSkip}) => {
   );
   assertRunError(
     "throw_exception_2",
-    "exception HorribleError(String, Bool, Number); let _ = throw HorribleError(\"oh no\", false, 1/3)",
+    {|
+      exception HorribleError(String, Bool, Number)
+      let _ = throw HorribleError("oh no", false, 1/3)
+    |},
     "HorribleError\\(\"oh no\", false, 1/3\\)",
```
phated, line 34: Should we be using the alternative string syntax for these "expected" outputs too? Unless I'm mistaken, we don't need to escape `(` and `)` in the alternative string syntax (which is for regexp stuff in standard strings).

compiler/test/suites/imports.re
```diff
@@ -14,175 +14,313 @@ describe("imports", ({test, testSkip}) => {
...
-    "import * from \"exportStar\"; {print(x); print(y(4)); print(z)}",
+    {|
+      import * from "exportStar"
+      {
+        print(x)
+        print(y(4))
+        print(z)
+      }
+    |},
     "5\n4\nfoo\n",
```
phated, line 25: What about these "expected" outputs? It's weird that we need to use `\n` in the string, isn't it?

compiler/test/suites/records.re
```diff
@@ -21,7 +21,10 @@ describe("records", ({test, testSkip}) => {
   );
   assertRun(
     "record_multiple",
-    "export record Rec {foo: Number, bar: String, baz: Bool}; print({foo: 4, bar: \"boo\", baz: true})",
+    {|
+      export record Rec {foo: Number, bar: String, baz: Bool}
+      print({foo: 4, bar: "boo", baz: true})
+    |},
     "{\n  foo: 4,\n  bar: \"boo\",\n  baz: true\n}\n",
```
phated, line 28: This should at least use the alternative string syntax so we don't need to escape the quotes, even if we decide to keep the `\n`

compiler/test/suites/stdlib.re
```diff
@@ -14,72 +14,174 @@ describe("stdlib", ({test, testSkip}) => {
...
-    "import * from \"list\"; [1, 2, 3] == [1, 2, 3]",
+    {|
+      import * from "list"
+      [1, 2, 3] == [1, 2, 3]
+    |},
+  );
+  assertSnapshot(
+    "stdlib_equal_4",
+    {|
+      import * from "list"
```
phated, line 39: @ospencer why does this and every test below it import list?

compiler/test/suites/strings.re
```diff
@@ -107,7 +107,11 @@ describe("strings", ({test, testSkip}) => {
   );
   assertParseWithLocs(
     "string_loc_multi_line",
-    "\"foo\nbar\nbaz\nqux\nquux\"",
+    {|"foo
+bar
+baz
+qux
+quux"|},
```
phated, line 114: I see, so the indentation here is taken into account with the alternative string syntax. 🤔

phated (changes requested):
@ribeirotomas1904 Thanks for taking the time to make these changes! I had a few questions in the PR (mostly posed to the rest of @grain-lang/core before you need to make any changes).

compiler/test/suites/stdlib.re
ospencer: I believe the original intent was to test that imports work, but we have lots of tests for that now.

compiler/test/suites/imports.re
marcusroberts: I agree.

compiler/test/suites/records.re
marcusroberts: Yep, agreed.

compiler/test/suites/strings.re
marcusroberts: Hmm, I guess you have too.  It doesn't look right, even though it is right.

compiler/test/suites/exceptions.re
ospencer: I looks like the `\` is what's escaped, but I', not sure why. But I suppose it would be good to do these too to avoid the escapes.

phated:
@ribeirotomas1904 It seems we have answers from some more of the team... are you still interested in finishing up these changes?

phated:
@ribeirotomas1904 will you have time to work on this so it can land for v0.5.5?

ribeirotomas1904:
yes, I will make the changes asap

phated:
Hey @ribeirotomas1904, any chance you'll be able to bring this up to date so we can finish it up? Otherwise I'd like to close it so someone else feels comfortable picking it up.

PR #1450, opened 2022-10-15
fix(compiler): Panic immediately when out of memory

ospencer:
This is a bit of a fun bug I discovered while working on refactoring constructors. When an `OutOfMemory` is thrown, if any memory is allocated during the chain of exception printer calls, we wind up in an infinite loop: we're out of memory, so trying to allocate something to report that we're out of memory ends up entering the out of memory flow again.

This PR adds `Exception.panic` in the runtime to print a string and immediately hit `unreachable`, and changes the implementation of `morecore` to immediately panic when out of memory instead of following the normal exception handling logic which could potentially try to allocate memory.

This seemed like a reasonable approach to me, since (in the future) users shouldn't be allowed to catch an `OutOfMemory` error.

I didn't consider this a breaking change because the API changes are only in the runtime.

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -66,9 +66,10 @@ let decref_closure_ident = Ident.create_persistent("GRAIN$EXPORT$decRef");
 
 /* Exceptions */
 let exception_mod = "GRAIN$MODULE$runtime/exception";
-let print_exception_ident = Ident.create_persistent("printException");
-let print_exception_closure_ident =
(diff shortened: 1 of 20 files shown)


peblair (approved):
Nice find. FWIW, there [is precedent](https://stackoverflow.com/a/1692243) in other languages to support catching `OutOfMemory`, but I agree that it's not something we should leave in support for (and we can always add it back later).

jozanza (approved):

Issue #971, opened 2021-10-07
Graindoc: Support multi-line examples


phated:
Discussed in discord with @spotandjake

This was originally an issue because we processed each line with the same regexp, but we now have a double-pass system where each attribute parses itself. I'm wondering if we can expand the first (simpler) attribute regexp to parse multiple lines as an attribute.

PR #1452, opened 2022-10-20
feat(stdlib): System path library

### LINKED ISSUE #1384: stdlib: add path libary
Add a libary for working with file paths to help once #211

alex-snezhko:
A path module for the standard library, which should close #1384. For the most part the API and implementation of this module is based on https://github.com/reasonml/reason-native/tree/master/src/fp, with a few differences and some additional features added.

This module doesn't interact with the actual file system, rather just providing functions for manipulating the data representation of paths.

A summary of the API:
- main type: `Path<a>`. Like ReasonNative's Fp, this type is parameterized in order to be able to strongly type relative/absolute paths.
- `Relative` and `Absolute` are opaque types that can be used to parameterize a path
- `absolute(String) -> Option<Path<Absolute>>`: parses an absolute path, returns `None` if invalid
- `relative(String) -> Option<Path<Relative>>`: parses a relative path, returns `None` if invalid
- `append(Path<a>, Path<Relative>) -> Path<a>`: append a relative path to a path
- `appendString(Path<a>, String) -> Option<Path<a>>`: like `append` but parses the string path
- `relativeTo(Path<a>, Path<a>) -> Result<Path<Relative>, RelativizationError>`: constructs a relative path (from 2 of the same path types) that would take one path to the other. Fails if either both are absolute paths but on different roots e.g. `/bin` and `C:/Users` or if there is not enough context to create a relative path from 2 relative paths e.g. getting from `../here` to `./there`
- `parent(Path<a>) -> Path<a>`: returns the path of the parent directory of the path
- `name(Path<a>) -> Option<String>`: returns the "file name" portion of the path if it exists
- `stem(Path<a>) -> Option<String>`: returns the "file name" portion of the path without the extension if it exists
- `extension(Path<a>) -> Option<String>`: returns the extension of the "file name" portion of the path if it exists
- `root(Path<Absolute>) -> AbsoluteRoot`
- `toString(Path<a>, Encoding) -> String`: stringifies the path with either POSIX (`/`) or Windows-style (`\`) path separators

This is somewhat of a proof-of-concept, so let me know if there are any differing opinions on how paths should be encoded or if anything about the API should change.


stdlib/path.gr
```diff
@@ -0,0 +1,411 @@
...
+ * 
+ * @param path: The path to get the name of
+ * @returns A `Some` variant containing the name of the path, or `None` if the path does not have one
+ * 
+ * @example Path.name(Option.unwrap(Path.relative("./dir/file.txt"))) // Some("file.txt")
+ * @example Path.name(Option.unwrap(Path.relative(".."))) // None
+ * 
+ * @since v0.5.4
+ */
+export let name = (path: Path<a>) =>
```
phated, line 347: Should we name this `filename` or `basename`? Trying to think inline with other stdlibs

compiler/test/stdlib/path.test.gr
```diff
@@ -0,0 +1,165 @@
...
+    Some("file"),
+    Some(".txt"),
+  ),
+  (
+    ".\\dir\\file.txt",
+    ".",
+    "./.\\dir\\file.txt",
+    Some(".\\dir\\file.txt"),
+    Some(".\\dir\\file"),
+    Some(".txt"),
```
phated, line 37: Can we get a double extension test? I'm thinking something like `webpack.config.js`

compiler/test/stdlib/path.test.gr
```diff
@@ -0,0 +1,165 @@
+import Path from "path"
+import Result from "result"
+import Option from "option"
+import List from "list"
+
+// test that a variety of properties hold for valid relative/absolute paths
+let relTests = [
+  (
```
phated, line 8: I'm wondering if these should be a record like `TestDsl` or something? I keep mixing up the tuple meanings while reviewing

compiler/test/stdlib/path.test.gr
```diff
@@ -0,0 +1,165 @@
...
+  ("dir/../../", "../..", "..", None, None, None),
+  (".git", ".", "./.git", Some(".git"), Some(".git"), None),
+  (".", "..", ".", None, None, None),
+  ("..", "../..", "..", None, None, None),
+  ("../../file", "../..", "../../file", Some("file"), Some("file"), None),
+  ("./", "..", ".", None, None, None),
+  (
+    "dir//file.txt///",
+    "dir",
+    "./dir/file.txt",
```
phated, line 26: We don't want to keep a single trailing separator to indicate it is a directory path? I often waffle on these

stdlib/path.gr
alex-snezhko: `filename` seems to me like it's specifically tailored to paths that point to files rather than directories (which I guess are also files technically but you don't generally think of them that way). I was initially partial to `basename` but usually that is paired with `dirname` for the `parent` operation, implying that a "name" (which to me implies a string) would be returned, whereas `parent` actually returns a `Path` and not a string. I'm coming around more to the idea of `basename` though, I can see how it's clearer than just `name`... Any thoughts about changing or keeping the name `parent`?

compiler/test/stdlib/path.test.gr
alex-snezhko: I could've sworn I included a test like this on a `.tar.gz` gzip extension, but I guess I removed it accidentally.

compiler/test/stdlib/path.test.gr
alex-snezhko: Yeah good point, initially these tuples started out with only 3 elements but then grew into these monstrosities :smile:

compiler/test/stdlib/path.test.gr
alex-snezhko: Sorry, I'm not exactly sure what you mean

compiler/test/stdlib/path.test.gr
```diff
@@ -0,0 +1,165 @@
...
+    Path.Drive("C"),
+    Some("me.txt"),
+    Some("me"),
+    Some(".txt"),
+  ),
+  (
+    "label:/me",
+    "label:/",
+    "label:/me",
+    Path.Drive("label"),
```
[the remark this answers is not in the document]
alex-snezhko: Ah yeah I guess drive labels do have to be one letter, good point. I can also change the implementation to use a `Char` instead of a `String` for getting the drive label.

compiler/test/stdlib/path.test.gr
phated: Sorry, I was wondering if we should keep a trailing slash on paths that are parsed with a trailing slash. Usually this is done by tools that need to track a path is exactly for a directory (since paths don't really encode that)

alex-snezhko: Hmm, so are you saying to maintain some sort of information about whether a path is a directory or not? Then if it has a trailing slash when parsed it can be determined with certainty that it is a directory, whereas if it doesn't it is ambiguous?

phated: Yeah, that's what I'm wondering (if we should maintain that). It's something that I've needed many times in gulp

alex-snezhko: Gotcha, do you think this idea of keeping the "directory-ness" of a path should be extended further into this library? For instance maybe encoding it in the type in the same way that absolute vs relative is to make the type something like `Path<Absolute, Directory>`. I can see some possible additional upsides of doing this, like only allowing `stem` and `extension` for `Path<a, File>`, `append` for `Path<a, Directory>`, etc.

phated: Oh! That's very interesting. How much more complicated do you think that'd be?

alex-snezhko: I don't think that it would add a lot of complexity, the only changes I can foresee are the need for more parsing functions to capture each possible case e.g. `relativeFile`, `absoluteDirectory`, etc. Maybe also some functions that allow you to convert a file path to a directory path (less sure about the need for this though)?

alex-snezhko: I added changes to accommodate this; let me know what you think

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+export let isDescendantOf = (base: Path<a, Directory>, path: Path<a, c>) => {
+  let (b1, _, s1) = base
+  let (b2, _, s2) = path
+  b1 == b2 && segsDescendantOf(List.reverse(s1), List.reverse(s2))
+}
+
+/**
+ * Returns the path corresponding to the parent directory of the given path.
+ * 
+ * @param path: The path to get the parent of
```
phated, line 445:
```suggestion
 * @param path: The path to search
```
Alternatively: "inspect", "calculate"? "The base path"?

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * @since v0.5.5
+ */
+export let isDescendantOf = (base: Path<a, Directory>, path: Path<a, c>) => {
+  let (b1, _, s1) = base
+  let (b2, _, s2) = path
+  b1 == b2 && segsDescendantOf(List.reverse(s1), List.reverse(s2))
+}
+
+/**
+ * Returns the path corresponding to the parent directory of the given path.
```
phated, line 443:
```suggestion
 * Retrieves the path corresponding to the parent directory of the given path.
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ */
+export let parent = (path: Path<a, b>) =>
+  match (path) {
+    (base, _, [_, ...rest]) => (base, Dir, rest),
+    (Rel(upDirs), _, []) => (Rel(upDirs + 1), Dir, []),
+    (Abs(_) as base, _, []) => (base, Dir, []),
+  }: Path<a, Directory>
+
+/**
+ * Returns the base name (named final segment) of a path.
```
phated, line 458:
```suggestion
 * Retrieves the final segment (basename) of a path.
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * 
+ * @param path: The path to get the name of
```
phated, line 460:
```suggestion
 * @param path: The path to inspect
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * @returns A `Some` variant containing the name of the path, or `None` if the path does not have one
+ * 
+ * @example Path.name(Option.unwrap(Path.relativeFile("./dir/file.txt"))) // Some("file.txt")
+ * @example Path.name(Option.unwrap(Path.relativeDir(".."))) // None
+ * 
+ * @since v0.5.5
+ */
+export let name = (path: Path<a, b>) =>
```
phated, line 468:
```suggestion
export let basename = (path: Path<a, b>) =>
```
Oscar said he liked `basename` the most.

Also, wondering about this API. Should it take a `PathWithType` instead? That way you could do something like `Path.basename(Path.fromString())` instead of needing to unwrap the option.

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+    (base, _, [_, ...rest]) => (base, Dir, rest),
+    (Rel(upDirs), _, []) => (Rel(upDirs + 1), Dir, []),
+    (Abs(_) as base, _, []) => (base, Dir, []),
+  }: Path<a, Directory>
+
+/**
+ * Returns the base name (named final segment) of a path.
+ * 
+ * @param path: The path to get the name of
+ * @returns A `Some` variant containing the name of the path, or `None` if the path does not have one
```
phated, line 461:
```suggestion
 * @returns `Some(basename)` containing the basename of the path or `None` if the path does not have one
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * 
+ * @example Path.name(Option.unwrap(Path.relativeFile("./dir/file.txt"))) // Some("file.txt")
```
phated, line 463:
```suggestion
 * @example Path.name(Option.unwrap(Path.relativeFile("./dir/file.txt"))) == Some("file.txt")
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * @example Path.name(Option.unwrap(Path.relativeDir(".."))) // None
```
phated, line 464:
```suggestion
 * @example Path.name(Option.unwrap(Path.relativeDir(".."))) == None
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * @since v0.5.5
+ */
+export let name = (path: Path<a, b>) =>
+  match (path) {
+    (_, _, [name, ..._]) => Some(name),
+    _ => None,
+  }
+
+/**
+ * Returns the name of the file path without the extension.
```
phated, line 475:
```suggestion
 * Retrieves the basename of a file path without the extension.
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * 
+ * @param path: The path to get the stem of
```
phated, line 477:
```suggestion
 * @param path: The path to inspect
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * @returns A `Some` variant containing the stem of the file path, or `None` if the path does not have one
```
phated, line 478:
```suggestion
 * @returns `Some(stem)` containing the stem of the file path or `None` if the path does not have one
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+          else Some(name)
+        },
+        None => Some(name),
+      }
+    },
+    _ => None,
+  }
+
+/**
+ * Returns the extension on the name of the file path.
```
phated, line 498:
```suggestion
 * Retrieves the extension on the basename of a file path.
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * 
+ * @param path: The path to get the extension of
```
phated, line 500:
```suggestion
 * @param path: The path to inspect
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * @returns A `Some` variant containing the extension of the file path, or `None` if the path does not have one
```
phated, line 501:
```suggestion
 * @returns `Some(extension)` containing the extension of the file path or `None` if the path does not have one
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+          else None
+        },
+        None => None,
+      }
+    },
+    _ => None,
+  }
+
+/**
+ * Returns the root of the absolute path.
```
phated, line 521:
```suggestion
 * Retrieves the root of an absolute path.
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * 
+ * @param path: The absolute path to get the root of
```
phated, line 523:
```suggestion
 * @param path: The path to inspect
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+    (Abs(root), _, _) => root,
+    _ => fail "Impossible: malformed absolute path data",
+  }
+
+/**
+ * Converts the given `Path` into a string, using the specified path encoding.
+ * A trailing slash is added to directory paths.
+ * 
+ * @param path: The `Path` to convert to a string
+ * @param encoding: The `PathEncoding` to use to represent the path as a string
```
phated, line 539:
```suggestion
 * @param separator: The `Separator` to use to represent the path as a string
```
Should we call these Separator throughout? I think that's more common than encoding.

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+  match (path) {
+    (Abs(root), _, _) => root,
+    _ => fail "Impossible: malformed absolute path data",
+  }
+
+/**
+ * Converts the given `Path` into a string, using the specified path encoding.
+ * A trailing slash is added to directory paths.
+ * 
+ * @param path: The `Path` to convert to a string
```
phated, line 538:
```suggestion
 * @param path: The path to convert to a string
```
This might be helpful to be the first thing in the file (actually second, after `fromString` which needs to be added)

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * @since v0.5.5
+ */
+export let root = (path: Path<Absolute, b>) =>
+  match (path) {
+    (Abs(root), _, _) => root,
+    _ => fail "Impossible: malformed absolute path data",
+  }
+
+/**
+ * Converts the given `Path` into a string, using the specified path encoding.
```
phated, line 535:
```suggestion
 * Converts the given `Path` into a string, using the specified path separator.
```

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * A trailing slash is added to directory paths.
+ * 
+ * @param path: The `Path` to convert to a string
+ * @param encoding: The `PathEncoding` to use to represent the path as a string
+ * @returns A string representing the given path
+ * 
+ * @example Path.toString(myPath, Path.Windows)
```
phated, line 542:
```suggestion
```
I don't like examples that can't be copy-pasted and `myPath` doesn't exist.

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * 
+ * @since v0.5.5
+ */
+export let toString = (path: Path<a, b>, encoding) => {
```
phated, line 546: This needs a matching `fromString` because you will often be working with strings that you don't know whether to pass them to one of the 4 parsing functions.

phated (changes requested):
This is looking sweet! My big request is to add a `fromString` function that parses any type of filepath string and returns the appropriate type. I believe this will need to return a `PathWithType` (but correct me if I'm wrong).

The most important aspects of this module are the entry/exit points (`fromString` and `toString`) so they should be the first functions you see in the functions section.

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+        (Rel(upDirs), _, []) => (Rel(upDirs + 1), ft, []),
+        (Abs(_), _, []) => path,
+      }
+    },
+  }
+}
+
+// splits a path on forward slashes
+let lexPath = pathStr => {
+  let pathStr = String.trim(pathStr)
```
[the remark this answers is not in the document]
alex-snezhko: Hmm, in hindsight maybe not actually; I initially thought that the main OSes disallow leading and trailing spaces in file names but it turns out that's not always the case so it might be best to remove the trim

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+/**
+ * Represents an absolute path's anchor point.
+ */
+export enum AbsoluteRoot {
+  Root,
+  Drive(Char),
+}
+
+// Rel Number represents number of directories up from base point
+enum BasePath<a> {
```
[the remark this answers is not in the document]
alex-snezhko: This enum depends on something that should be shown in the docs (`AbsoluteRoot`); I suppose the types could be made mutually recursive and this enum could be placed above the section, but this seems like a somewhat fragile solution. As a more long-term solution, do you think the documentation generator should be made to ignore types without docblock comments (or as an alternative, maybe include a mechanism similar to `// formatter-ignore`)?

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * (relative|absolute, file|directory).
+ */
+type Path<a, b> = (BasePath<a>, FileType<b>, List<String>)
+
+/**
+ * Wraps each possibile path type to a variant; can be used to dynamically
+ * determine the type of the path wrapped in cases where it is needed. It is
+ * recommended to only make use of this type in cases where it is unavoidable.
+ */
+export enum PathWithType {
```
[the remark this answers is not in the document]
alex-snezhko: Ok, and should the existence of a trailing slash be used to identify directories whereas strings will be parsed as file paths otherwise?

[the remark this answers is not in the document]
phated: Yep! Either a trailing slash or `.` or `..` would indicate a directory

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+export let relativeDir = pathStr => {
+  parseRel(pathStr, Dir): Option<Path<Relative, Directory>>
+}
+
+/**
+ * Wraps a path with its corresponding concrete path type, in order to allow
+ * dynamically checking the path type.
+ * 
+ * @param path: The path to wrap
+ * @returns A `PathWithType` wrapping the path with its type
```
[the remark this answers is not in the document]
alex-snezhko: "The path wrapped with its type" maybe?

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * 
+ * @since v0.5.5
+ */
+export let wrapWithType = (path: Path<a, b>) => {
```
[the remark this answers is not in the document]
alex-snezhko: `wrap` seems somewhat unclear to me (what is it wrapping with?) so I just went with this name for some sort of compromise; reason-native uses the name `firstClass` which I like even less personally. Open to other ideas though

stdlib/path.gr
```diff
@@ -0,0 +1,571 @@
...
+ * @param source: The source path to find a path from
+ * @param dest: The destination path to find the path to
+ * @returns An `Ok` variant containing the relative path if successfully parsed or `Err` otherwise
+ * 
+ * @example // relativeTo(/usr, /usr/bin) => ./bin
+ * @example // relativeTo(/home/me, /home/me) => .
+ * @example // relativeTo(/usr/bin, /etc) => ../../etc
+ * @example // relativeTo(.., ../../thing) => ../thing
+ * @example // relativeTo(/usr/bin, C:/Users) => Err(DifferentRoots)
+ * @example // relativeTo(../here, ./there) => Err(ImpossibleRelativization)
```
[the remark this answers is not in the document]
alex-snezhko: This isn't technically legal since the arguments should be `Path` objects and not strings so I'm thinking this might mislead potential users; Doing it the "proper" way results in somewhat long lines but maybe that's not the worst thing

stdlib/path.gr
alex-snezhko: The assumption I was going under was that you'd usually know what type of path you'd be working with and that `PathWithType` would only be used in extenuating circumstances; therefore all of the significant parts of this API were made with the assumption that "bare" `Path`s would be passed in.

In any case, most of the other functions in this library use the bare Paths over `PathWithType` so if this change is made I feel that it would be inconsistent to only have it for this function and not others like `parent`, `toString` etc.

stdlib/path.gr
alex-snezhko: Sure, I was initially thinking `encoding` as in theory the platform could affect more than just path separators but in the current implementation that's all it does so I suppose separator is clearer

stdlib/path.gr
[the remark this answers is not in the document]
phated: Generally we've been using mutually recursive type definitions and then separating them with the Section comment. That's really the solution for now. When we rework the module system and type exports, we'll probably have a cleaner solution (instead of **every** type being exported from the module).

stdlib/path.gr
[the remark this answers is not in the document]
phated:
```suggestion
 * @returns The path wrapped with a unifying enum
```
This doesn't sound quite right, but maybe something along these lines? I'm thinking of it like "the enum wrapper is actually something to unify disparate path types".

stdlib/path.gr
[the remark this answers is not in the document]
phated: If we start talking about this as a "unifying enum" or something like that, we could call this `toEnum` maybe?

stdlib/path.gr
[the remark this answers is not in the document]
phated: Oh, you are right. That makes this really difficult. Maybe we exclude these until we have multi-line examples? (ref #971)

stdlib/path.gr
phated: Yeah, I'm struggling with the duality of having a nice API vs type safety. I really like what you've done with `absoluteFile`/`relativeFile` and `absoluteDir`/`relativeDir` but it trades safety for making them **really** hard to use.

I wonder if we should should make this first version of the Path module to work with the PathWithType, and once we have the in-file `module` syntax, we can add a `module Relative` and `module Absolute` that operate on their specific `Path<a, b>` types. That way you can opt-in to more type safe APIs, something like:
```grain
let myPath = Path.fromString("/foo/bar/baz.csv")

let myPath = match (myPath) {
  AbsoluteFile(p) => p,
  AbsoluteDirectory(p) => Path.Absolute.append(p, "index.js"),
  RelativeFile(p) => Path.Absolute.join(Process.cwd(), p),
  RelativeDirectory(p) => Path.Absolute.append(Path.Absolute.join(Process.cwd(), p), "index.js")
}

let myPath = Path.wrapWithType(myPath)

File.readFile(myPath)
```

A lot of this thought process is driven behind what the API for the filesystem will look like...and I think I'd want it to take `PathWithType` for ease of use while still being typed.

What do you think?

stdlib/path.gr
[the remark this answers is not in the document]
alex-snezhko: Hmm, if what it's doing is unifying the types then what do you think about `unify`?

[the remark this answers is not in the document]
phated: Yeah, I like `unify` but we need to get feedback from @ospencer on whether that is too much jargon.

stdlib/path.gr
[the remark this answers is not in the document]
alex-snezhko: If I remove the module names it's actually not _too_ bad, the lengths become more manageable imo

e.g. `relativeTo(unwrap(absoluteDir("/usr")), unwrap(absoluteDir("/usr/bin"))) // a Path containing "./bin"`

stdlib/path.gr
alex-snezhko: Hmm, that's an interesting idea; so you're saying for now to make the paths one type and have them be checked dynamically by the functions, then change the functions to throw exceptions or return `Result`s when used on invalid path types?

stdlib/path.gr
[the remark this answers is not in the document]
phated: Going back to the idea I provided in another comment. If these all operated on the `PathWithType`, then these examples turn into `relativeTo(fromString("/usr"), fromString("/usr/bin"))` which looks pretty solid

stdlib/path.gr
alex-snezhko: I thought about this some more and thought of another idea that is similar to the one you suggested: for now we stop exposing the `Path<a, b>` type and rename `PathWithType` to just `Path`, changing the types of all functions to use this `Path` and giving `Result`s whenever the path types are invalid for an operation.

Then once the in-file module system drops there could be one submodule called `Typed` that reimplements all of the module's functions but instead using typed paths. This may also be more ergonomic than `Absolute`, `Relative`, etc as all of the "typed" stuff could be in one submodule rather than many.

An example showcasing what the APIs would look like:
```
Path.isDescendantOf: (base: Path, path: Path) -> Result<Bool, IncompatiblePathType>

Path.Typed.isDescendantOf: (base: Typed.Path<a, Directory>, path: Typed.Path<a, c>) -> Bool
```

What do you think?

phated: Yeah, let's do that! Oscar was also asking us to use the name `Path` instead of `PathWithType` but I didn't have a good replacement name for `Path<a, b>` (what are you going to name it for now?)

alex-snezhko: Hmm, maybe `TypedPath<a, b>` or something? After the module system is added then we could just do something like this maybe
```
type TypedPath<a, b> = // ...

module Typed {
  type Path<a, b> = TypedPath<a, b>
}
```

phated: Sure, let's go with that and people can chime in if they have a better idea 🙇

alex-snezhko: One more question from an API point of view: how should situations where an operation is done on a path type it shouldn't be applicable to be handled? The way I initially outlined would be to enclose all such possible scenarios in a `Result`, returning `Err(IncompatiblePathType)` in situations where the path type is incompatible. However, alternatively perhaps this approach shouldn't always be used in places where it isn't super necessary?

For example, `relativeTo` is an example where doing this approach would make sense, as simply returning `Path` wouldn't be a good idea in error cases for obvious reasons. However, looking at `isDescendantOf` which returns a boolean under normal circumstances, `false` could be just used in these situations instead of an `Err` value, which still makes sense logically.

Overall using this simplification will make the API easier to use but may hide some details about why an operation failed.  Which approach do you think is better?

alex-snezhko: Also on a tangential note, currently `extension` and `stem` return `None` if the file path doesn't have an extension or stem. Should this behavior be changed to return empty string instead in these situations? The decision of which approach to use would also affect these functions

phated: I think that returning a Result is absolutely necessary, even if the "okay" case is `Ok(true)`. This should be a relatively fine API because you can pattern match like:
```grain
match (Path.isDescendantOf(...)) {
  Ok(true) => print("hooray"),
  Ok(false) => print("not descendant"),
  Err(_) => print("whoops")
}
```
The reason for this is that (without Result) `false` would be conflated with too many meanings. The specific details will likely be needed by things like the filesystem API.

I'm not sure about the `extension` and `stem`, as I could see it being fine either way. Empty strings might be nicer in that case.

alex-snezhko: Yeah, empty string inside a result; it would give `Err` if the path is a directory and `Ok("")` in the case of a file path with no stem/extension.

Also I just thought of another approach for `isDescendantOf` that may give even more clarity: what do you think about repurposing this function to instead be called something like `ancestry`:
```
enum AncestryStatus {
  Descendant, // if the first argument is the parent of the second
  Ancestor, // if the first argument descends from the second
  Neither, // if the first argument is neither a descendant or ancestor of the second
  DifferentRoots, // if the arguments are both absolute but are on different roots
  IncompatibleBases, // if one argument is relative and other is absolute
}

export let ancestry = (path1, path2) -> AncestryStatus
```
Alternatively, the success cases can be factored out from the error cases:
```
enum AncestryStatus {
  Descendant, // if the first argument is the parent of the second
  Ancestor, // if the first argument descends from the second
  Neither, // if the first argument is neither a descendant or ancestor of the second
}

enum AncestryError {
  DifferentRoots, // if the arguments are both absolute but are on different roots
  IncompatibleBases, // if one argument is relative and other is absolute
}

export let ancestry = (path1, path2) -> Result<AncestryStatus, AncestryError>
```
This would for example allow you to determine the difference between an ancestor and an unrelated path, which both currently result in `false`; let me know what you think.

phated: I like it! I think i prefer the `Result<AncestryStatus, AncestryError>` because then you can use all of our Result utilities with it.

compiler/test/stdlib/path.test.gr
```diff
@@ -0,0 +1,286 @@
...
+    expParent: "../..",
+    expStr: "../../file",
+    expName: Some("file"),
+    expStem: "file",
+    expExt: "",
+  },
+  {
+    pathStr: ".\\dir\\file.txt",
+    expParent: ".",
+    expStr: "./.\\dir\\file.txt",
```
phated, line 75: See discussion in discord about how we should add an escape hatch for people that need to parse Windows-style paths

compiler/test/stdlib/path.test.gr
```diff
@@ -0,0 +1,286 @@
...
+  { pathStr: "/", isDir: true, isAbs: true },
+  { pathStr: "/file", isDir: false, isAbs: true },
+  { pathStr: ".", isDir: true, isAbs: false },
+  { pathStr: "./file", isDir: false, isAbs: false },
+]
+
+List.forEach(({ pathStr, isDir, isAbs }) => {
+  let path = fs(pathStr)
+  assert isDir == Path.isDirectory(path)
+  assert isAbs == Path.isAbsolute(path)
```
phated, line 174: 👍 I like these helpers

compiler/test/stdlib/path.test.gr
```diff
@@ -0,0 +1,286 @@
...
+// assert Path.relativeFile("./dir/.") == None
+// assert Path.relativeFile("..") == None
+// assert Path.relativeDir("/usr/bin") == None
+// assert Path.relativeDir("C:/Users/me") == None
+// assert Path.absoluteDir("C:\\Users\\me") == None
+// assert Path.absoluteFile("/usr/file.txt/") == None
+// assert Path.absoluteFile("/usr/..") == None
+// assert Path.absoluteFile("./file.txt") == None
+// assert Path.absoluteFile("file.txt") == None
+// assert Path.absoluteDir("../..") == None
```
phated, line 193:
```suggestion
```
Drop these until the new module stuff happens?

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+ * 
+ * @since v0.5.5
+ */
+import String from "string"
+import List from "list"
+import Option from "option"
+import Result from "result"
+import Char from "char"
+
+// this module is mostly a port of https://github.com/reasonml/reason-native/blob/a0ddab6ab25237961e32d8732b0a222ec2372d4a/src/fp/Fp.re
```
phated, line 29:
```suggestion
// this module is influenced by https://github.com/reasonml/reason-native/blob/a0ddab6ab25237961e32d8732b0a222ec2372d4a/src/fp/Fp.re
```
We've changed *a lot*

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+/**
+ * Represents an absolute path's anchor point.
+ */
+export enum AbsoluteRoot {
+  Root,
+  Drive(Char),
+}
+
+// Dummy record names put here just to distinguish the two. These could be
+// replaced with opaque types if they get added to the language
```
phated, line 117:
```suggestion
```
Old comment?

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+/**
+ * Represents an error that can occur when appending paths.
+ */
+export enum AppendError {
+  AppendToFile,
+  AppendAbsolute,
+}
+
+/**
+ * Represents the status of an ancestry check between two paths
```
phated, line 153:
```suggestion
 * Represents the status of an ancestry check between two paths.
```

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+  }
+}
+
+let relativeFile = pathStr => {
+  let pathOpt = parseRel(pathStr, File)
+  Option.map(toTyped, pathOpt): Option<TypedPath<Relative, File>>
+}
+
+let relativeDir = pathStr => {
+  // parseRel(pathStr, Dir)//: Option<TypedPath<Relative, Directory>>
```
phated, line 328:
```suggestion
```
Not needed?

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+  let pathOpt = parseRel(pathStr, Dir)
+  Option.map(toTyped, pathOpt): Option<TypedPath<Relative, Directory>>
+}
+
+/**
+ * Parses a path string into a path wrapped with the path's type. If the path
+ * type is known in advance, the specialized parsing functions (`relativeDir`,
+ * `absoluteFile`, etc) should be used instead.
```
phated, line 336:
```suggestion
  * Parses a path string into a `Path` type, which encodes details about the path in the type.
```
The `relativeDir` and `absoluteFile` functions are not exposed yet.

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+ * 
+ * @param pathStr: The path string to parse
```
phated, line 338:
```suggestion
 * @param pathStr: The string to parse as a path
```

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+ * @returns `Some(wrappedPath)` if the path is successfully parsed or `None` otherwise
```
phated, line 339:
```suggestion
 * @returns The path with details encoded within the type
```

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+ * 
+ * @example fromString("/bin/") // an absolute Path referencing the directory /bin/
```
phated, line 341: Idea: What if these examples showed the type returned?
```suggestion
 * @example fromString("/bin/"): AbsoluteDir(TypedPath<Absolute, Directory>)
```
What do you think?

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+    [Slash | DriveTok(_), ..._] => true,
+    _ => false,
+  }
+  let isDir = !isFilePath(tokens)
+
+  match ((isAbs, isDir)) {
+    (true, true) => AbsoluteDir(Option.unwrap(absoluteDir(pathStr))),
+    (true, false) => AbsoluteFile(Option.unwrap(absoluteFile(pathStr))),
+    (false, true) => RelativeDir(Option.unwrap(relativeDir(pathStr))),
+    (false, false) => RelativeFile(Option.unwrap(relativeFile(pathStr))),
```
phated, line 363: Should these `unwrap`s be `expect` so it can throw `Impossible:` messages if encountered?

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+ */
+export let isAbsolute = path => {
+  let (base, _, _) = pathInfo(path)
+  match (base) {
+    Abs(_) => true,
+    _ => false,
+  }
+}
+
+let toString = (path, separator) => {
```
phated, line 400: Why'd you make this a separate function from the `toString` implementation? This might still be handy for our discussion in discord about having `fromString` be posix and `fromPlatformString` to take the separator 🤔

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+      },
+    (Abs(_), _, _) => fail "Impossible: relative path encoded as absolute path",
+  }: PathInfo
+
+/**
+ * Creates a new path by appending a relative path segment to a directory path.
+ * 
+ * @param path: The base path
+ * @param toAppend: The relative path to append
+ * @returns `Ok(path)` combining the base and appended paths, or `Err(err)` if the paths are incompatible for being appended
```
phated, line 464:
```suggestion
 * @returns `Ok(path)` combining the base and appended paths or `Err(err)` if the paths are incompatible
```

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+    _ => Ok(segsAncestry(List.reverse(s1), List.reverse(s2))),
+  }
+}
+
+/**
+ * Determines the relative ancestry betwen two paths.
+ * 
+ * @param base: The first path to consider
+ * @param path: The second path to consider
+ * @returns `Ok(ancestryStatus)` with the relative ancestry between the paths if they are compatible, or `Err(err)` if they are incompatible
```
phated, line 574:
```suggestion
 * @returns `Ok(ancestryStatus)` with the relative ancestry between the paths if they are compatible or `Err(err)` if they are incompatible
```

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+ * 
+ * @param path: The path to inspect
+ * @returns `Some(path)` containing the basename of the path, or `None` if the path does not have one
+ * 
+ * @example basename(fromString("./dir/file.txt")) == Some("file.txt")
+ * @example basename(fromString(".."))) == None
+ * 
+ * @since v0.5.5
+ */
+export let basename = (path: Path) => {
```
phated, line 633: Do we want basename to return `Option`? I think it's the only function in this module that returns `Option` now, is that correct?

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+      }
+    },
+    _ => "",
+  }
+
+/**
+ * Retrieves the basename of a file path without the extension.
+ * 
+ * @param path: The path to inspect
+ * @returns `Ok(path)` containing the stem of the file path, or `Err(err)` if the path is a directory path
```
phated, line 656:
```suggestion
 * @returns `Ok(path)` containing the stem of the file path or `Err(err)` if the path is a directory path
```

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+      }
+    },
+    _ => "",
+  }
+
+/**
+ * Retrieves the extension on the basename of a file path.
+ * 
+ * @param path: The path to inspect
+ * @returns `Ok(path)` containing the extension of the file path, or `Err(err)` if the path is a directory path
```
phated, line 690:
```suggestion
 * @returns `Ok(path)` containing the extension of the file path or `Err(err)` if the path is a directory path
```

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+  match (path) {
+    (Abs(root), _, _) => root,
+    _ => fail "Impossible: malformed absolute path data",
+  }
+
+/**
+ * Retrieves the root of the absolute path
+ * 
+ * @param path: The path to inspect
+ * @returns `Ok(root)` containing the root of the path, or `Err(err)` if the path is a relative path
```
phated, line 716:
```suggestion
 * @returns `Ok(root)` containing the root of the path or `Err(err)` if the path is a relative path
```

stdlib/path.gr
```diff
@@ -0,0 +1,729 @@
...
+
+// should only be used on absolute paths
+let root = (path: PathInfo) =>
+  match (path) {
+    (Abs(root), _, _) => root,
+    _ => fail "Impossible: malformed absolute path data",
+  }
+
+/**
+ * Retrieves the root of the absolute path
```
phated, line 713:
```suggestion
 * Retrieves the root of the absolute path.
```

phated (changes requested):
I **love** these updates. I had some feedback.

compiler/test/stdlib/path.test.gr
```diff
@@ -0,0 +1,286 @@
...
+// let absFile = x =>
+//   Option.expect(
+//     "Expected " ++ x ++ " to be absolute file",
+//     Path.absoluteFile(x)
+//   )
+// let absDir = x =>
+//   Option.expect(
+//     "Expected " ++ x ++ " to be absolute directory",
+//     Path.absoluteDir(x)
+//   )
```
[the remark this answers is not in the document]
alex-snezhko: Oh yeah forgot about the tests, mb

compiler/test/stdlib/path.test.gr
```diff
@@ -0,0 +1,286 @@
...
+    expStem: "file",
+    expExt: "",
+  },
+  {
+    pathStr: "a.tar.gz",
+    expParent: ".",
+    expStr: "./a.tar.gz",
+    expName: Some("a.tar.gz"),
+    expStem: "a.tar",
+    expExt: ".gz",
```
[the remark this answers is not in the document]
alex-snezhko: Ah, good point

stdlib/path.gr
alex-snezhko: Oh these should be referring to `Relative`, `Absolute`, etc; I moved the definitions but forgot to move the comment

stdlib/path.gr
alex-snezhko: Hmm, I was thinking of hiding `TypedPath` and keeping `Path` opaque until typed paths get added when these will become more relevant, although I could re-expose them if you think it'd be beneficial

stdlib/path.gr
alex-snezhko: As a side note I can add `Helper` to the names to make them clearer e.g. `toStringHelper`

stdlib/path.gr
phated: That's hard. I mostly want them exported so you can print a path and get the details. By abstracting them, you can't pattern match (but there isn't really anything you can do with the inner values)...

stdlib/path.gr
phated: Cool cool. I sort of realized that by the end of the file. Maybe we should add a comment with `TODO: Reuse with typed path implementations` or something?

stdlib/path.gr
alex-snezhko: Yeah I believe so; Another option is to make it return empty string instead for paths with no clear name or also possibly `".."` or `"."`. Which approach do you think would be best?

phated: Why do we want option instead of Result?

stdlib/path.gr
alex-snezhko: I don't know what happened to my grammar skills xD

stdlib/path.gr
alex-snezhko: Hmm, I typically think of `Result` as being for operations that could fail as a result of malformed input or other conditions, whereas `Option` for valid operations that just don't have the value you are asking for. This case seems more like the latter to me, although maybe you have a alternate perspective?

stdlib/path.gr
alex-snezhko: Isn't it still possible to print the value even if it's not exported? I just tried it out and it looks like it still works.

But yeah, the lack of being able to do anything useful with them is why I figured it's not necessary to show right now.

phated: Unless I'm mistaken, printing an opaque type from outside the module will not print the representation.

ospencer: That's how it's supposed to work but we still have some bits to work out :P

alex-snezhko: I just tried on my system and it looks like it does print the representation; is this a bug?

stdlib/path.gr
```diff
@@ -0,0 +1,774 @@
...
+ * provide functionality for interacting with the file system.
+ * 
+ * This module explicitly encodes whether a path is absolute or relative, and
+ * whether it refers to a file or a directory, as part of the `Path` type.
+ *
+ * Paths in this module abide by a special POSIX-like representation/grammar
+ * rather than one defined by a specific operating system. The rules are as
+ * follows:
+ * 
+ * - Path separators are denoted by `/` for POSIX-like paths and additionally `\` for Windows paths
```
phated, line 14:
```suggestion
 * - Path separators are denoted by `/` for POSIX-like paths
```
I don't think we need to mention Windows paths here as long as this distinction is mentioned on `fromPlatformString` and `toPlatformString`

stdlib/path.gr
```diff
@@ -0,0 +1,774 @@
...
+let toPath = (path: PathInfo) => {
+  match (path) {
+    (Abs(_), File, _) as p => AbsoluteFile(toTyped(p)),
+    (Abs(_), Dir, _) as p => AbsoluteDir(toTyped(p)),
+    (Rel(_), File, _) as p => RelativeFile(toTyped(p)),
+    (Rel(_), Dir, _) as p => RelativeDir(toTyped(p)),
+  }
+}
+
+// TODO expose these functions once module system added
```
phated, line 292:
```suggestion
// TODO(#1496): Expose these functions once module system added
```

stdlib/path.gr
```diff
@@ -0,0 +1,774 @@
...
+  let pathOpt = parseRel(pathStr, File, platform)
+  Option.map(toTyped, pathOpt): Option<TypedPath<Relative, File>>
+}
+
+let relativeDir = (pathStr, platform) => {
+  let pathOpt = parseRel(pathStr, Dir, platform)
+  Option.map(toTyped, pathOpt): Option<TypedPath<Relative, Directory>>
+}
+
+// TODO reuse this and the other helper functions for the typed implementations
```
phated, line 334:
```suggestion
// TODO(#1496): Reuse this and the other helper functions for the typed implementations
```

stdlib/path.gr
```diff
@@ -0,0 +1,774 @@
...
+ * @since v0.5.5
+ */
+export let fromString = pathStr => {
+  fromStringHelper(pathStr, Posix)
+}
+
+/**
+ * Parses a path string into a `Path` using the path separators appropriate to
+ * the given platform. Paths will be parsed as file paths rather than directory
+ * paths if there is ambiguity.
```
phated, line 384: Should mention the Windows path stuff here

stdlib/path.gr
```diff
@@ -0,0 +1,774 @@
...
+ * 
+ * @param pathStr: The path string to parse
```
phated, line 386:
```suggestion
 * @param pathStr: The string to parse as a path
```

stdlib/path.gr
```diff
@@ -0,0 +1,774 @@
...
+ * @param platform: The platform whose path separators should be used for parsing
+ * @returns `Some(wrappedPath)` if the path is successfully parsed or `None` otherwise
```
phated, line 388:
```suggestion
 * @returns The path wrapped with details encoded within the type
```

stdlib/path.gr
```diff
@@ -0,0 +1,774 @@
...
+ * 
+ * @param path: The path to inspect
+ * @returns `true` if the path is a directory path or `false` otherwise
+ * 
+ * @example isDirectory(fromString("file.txt")) == false
+ * @example isDirectory(fromString("/bin/")) == true
+ * 
+ * @since v0.5.5
+ */
+export let isDirectory = path => {
```
phated, line 410: Should this be moved below `toString` functions?

stdlib/path.gr
```diff
@@ -0,0 +1,774 @@
...
+/**
+ * Determines whether the path is an absolute path.
+ * 
+ * @param path: The path to inspect
+ * @returns `true` if the path is absolute or `false` otherwise
+ * 
+ * @example isAbsolute(fromString("/Users/me")) == true
+ * @example isAbsolute(fromString("./file.txt")) == false
+ */
+export let isAbsolute = path => {
```
phated, line 424: Should this be moved below `toString` functions?

stdlib/path.gr
```diff
@@ -0,0 +1,774 @@
...
+ * @since v0.5.5
+ */
+export let toString = path => {
+  toStringHelper(pathInfo(path), Posix)
+}
+
+/**
+ * Converts the given `Path` into a string, using the canonical path separator
+ * appropriate to the given platform. A trailing slash is added to directory
+ * paths.
```
phated, line 478: Mention Windows path stuff here

stdlib/path.gr
phated: Let's get some feedback from the rest of the team 😄

phated (changes requested):
🔥 - just a couple of nits about the docs

phated:
Looks like there might be a non-exhaustive warning happening somewhere too.

stdlib/path.gr
alex-snezhko: Should I just keep it as is then?

alex-snezhko:
Ah yeah I'm assuming that's from the
```
let TBase(base) = base
```
This is an enum with a single constructor so it shouldn't ever not match but I'll change the implementation to get rid of the warning

stdlib/path.gr
phated: Sure

phated:
@alex-snezhko seems to be something else because the warning still displays on your latest push

stdlib/path.gr
```diff
@@ -0,0 +1,775 @@
...
+    [Dot | Dotdot | Slash, ..._] => false,
+    _ => true,
+  }
+}
+
+let parseAbs = (pathStr, fileType, platform) => {
+  match (lexPath(pathStr, platform)) {
+    [] => None,
+    [first, ...rest] => {
+      let tokens = [first, ...rest]
```
phated, line 245:
```suggestion
    [first, ...rest] as tokens => {
```
I believe we can use `as` syntax here. I'm also wondering why it doesn't warn about not handling `[first]`... I assume `[first, ...rest]` matches any non-empty list and just makes `rest` an empty list if there is exactly 1 item.

phated:
I merged in `main` and reformatted the test file to fix the failure.

stdlib/path.gr
```diff
@@ -0,0 +1,775 @@
...
+}
+
+let relativeDir = (pathStr, platform) => {
+  let pathOpt = parseRel(pathStr, Dir, platform)
+  Option.map(toTyped, pathOpt): Option<TypedPath<Relative, Directory>>
+}
+
+// TODO(#1496): reuse this and the other helper functions for the typed implementations
+let fromStringHelper = (pathStr, platform) => {
+  let tokens = match (lexPath(pathStr, platform)) {
```
phated, line 336: I'm noticing that we actually lex the path twice for this API. I wonder if the `parse` functions should take a token list instead of the `pathStr`. Thoughts?

stdlib/path.gr
```diff
@@ -0,0 +1,775 @@
...
+  }
+  let segs = match (fileType) {
+    File => segs,
+    Dir => List.append(segs, [""]),
+  }
+  List.join(sep, segs)
+}
+
+/**
+ * Converts the given `Path` into a string, using the POSIX path separator.
```
phated, line 428:
```suggestion
 * Converts the given `Path` into a string, using the `/` path separator.
```

stdlib/path.gr
```diff
@@ -0,0 +1,775 @@
...
+  match (path) {
+    (_, _, [name, ..._]) => Some(name),
+    _ => None,
+  }
+
+/**
+ * Retrieves the basename (named final segment) of a path.
+ * 
+ * @param path: The path to inspect
+ * @returns `Some(path)` containing the basename of the path, or `None` if the path does not have one
```
phated, line 676:
```suggestion
 * @returns `Some(path)` containing the basename of the path or `None` if the path does not have one
```
(pending responses to the below)

stdlib/path.gr
```diff
@@ -0,0 +1,775 @@
...
+      Ok(stem)
+    },
+  }
+}
+
+/**
+ * Retrieves the extension on the basename of a file path.
+ * 
+ * @param path: The path to inspect
+ * @returns `Ok(path)` containing the extension of the file path, or `Err(err)` if the path is a directory path
```
phated, line 732:
```suggestion
 * @returns `Ok(path)` containing the extension of the file path or `Err(err)` if the path is a directory path
```

stdlib/path.gr
```diff
@@ -0,0 +1,775 @@
...
+ */
+export enum AbsoluteRoot {
+  Root,
+  Drive(Char),
+}
+
+/**
+ * Represents a system path.
+ */
+enum Path {
```
phated, line 121:
```suggestion
export enum Path {
```
I think I want this exported... I want to be able to do something like:

```grain
match (Path.fromString("./my/relpath")) {
  AbsoluteFile(_) as path => path,
  AbsoluteDir(_) as path => Path.append(path, "index.md"),
  RelativeFile(_) as path => Path.append(Process.cwd(), path),
  RelativeDir(_) as path => Path.append(Path.append(Process.cwd(), path), "index.md")
}
```
and I don't see a (nice?) way to do that normalization if you make this opaque.

stdlib/path.gr
ospencer: I think option makes sense! In the case that there's no basename, it's not really that it failed, just that there isn't one.

stdlib/path.gr
alex-snezhko: Yeah good idea, I was actually planning on doing this at some point but forgot to implement it.

stdlib/path.gr
alex-snezhko: Ok, I'll add `TypedPath`, `Relative`, etc into this section and write docblocks for them as well then

ospencer (approved):
This is really fantastic, @alex-snezhko. I know a tremendous amount of work went into this, and the result is stunning. And thanks @phated for all of the feedback along the way! Really beautiful, high-quality, idiomatic work.

phated (approved):
Heroic effort! 💪  Thanks for all the work here @alex-snezhko

PR #1458, opened 2022-10-24
fix(stdlib)!: Make `Number.neg(-1) == -1`

spotandjake:
As `Number.neg` is the opposite of `Number.abs` and the purpose of `Number.neg` is to always recieve a negative value i relized an issue in the original implementation where we were just flipping the sign. This pr corrects that making `Number.neg` the opposite of `Number.abs`

* This pr is breaking as we are changing the behavour of an established function.


phated:
@spotandjake please provide your research on this. I've thought of this function as the equivalent to `-n` (applying the negative to the argument).

PR #1462, opened 2022-10-26
fix(grainfmt): Group expressions inside array accessors

### LINKED ISSUE #1461: Formatter: incorrect line splitting within square brackets
If I write something like this (contrived example):
```
export let getBefore = (array, index) => if (array == [>]) "nope" else array[index - 1]
```
The formatter formats it like this:
```
export let getBefore = (array, index) => if (array == [>]) "nope"
else array[index -
1]
```
For some reason it splits the `1` onto a separate line. However, the formatting behavior changes if I were to replace the square brackets on `[index - 1]` with parentheses:
```
export let getBefore = (array, index) => if (array == [>]) "nope"
else array(index - 1)
```

marcusroberts:
Fixes #1461

Handle expressions inside array accessors as groups so they break properly

### DIFF
--- a/compiler/src/formatting/format.re
+++ b/compiler/src/formatting/format.re
@@ -2454,7 +2454,7 @@ and print_expression =
       Doc.concat([
         print_expression(~original_source, ~comments, expression1),
         Doc.lbracket,
-        print_expression(~original_source, ~comments, expression2),
+        Doc.group(print_expression(~original_source, ~comments, expression2)),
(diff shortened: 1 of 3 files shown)


ospencer:
Can you add some tests?

marcusroberts:
Of course, done

ospencer (approved):

phated (approved):

PR #1464, opened 2022-10-31
feat(compiler)!: Early return

### LINKED ISSUE #553: Compiler: Early return
This would add the ability to return early from a function. We've made the decision that if a function is to use a `return` statement, then all implicit returns must be made explicit. The user should see an error that looks something like this:

```
325.  someFunctionThatHasReturnType()
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
326. }

An explicit `return` statement is required when a function returns early.
```

ospencer:
Closes #553 

This PR implements the ability to return early from a function in Grain using the `return` keyword. Here's an example of how this might be used:

```
import Array from "array"

let findFour = arr => {
  for (let mut i = 0; i < Array.length(arr); i += 1) {
    if (arr[i] == 4) {
      return Some(4)
    }
  }

  return None
}
```

`return` in Grain works similarly to other languages with `return`, with one exception: while Grain functions return the value of the last expression implicitly, if a `return` statement is used within a function, _all_ values returned from the function must be made explicit.

This may seem like a stylistic choice, but we felt it was important to properly fit this feature into the language. Some error messages may become confusing if all returns aren't made explicit. For example, look at the code below:

image

It's reasonable to see that this should be a type error. However, when one attempts to explain _exactly why_ this is a type error, it's not obvious. (If you'd like to try to figure it out for yourself, do so now, as the next line will explain.)

The nearest cousin to `return` in terms of how it fits into the language is `throw` or `fail`, and just like `throw` or `fail`, `return` has an "any" type. This is to appease the typechecker—you'll want to place one of these statements just about anywhere in your code, and if using one of these statements typechecked to `Void`, wherever you used it would _also_ have to typecheck to `Void`, and this would be frustrating when trying to throw an error in a single case of a `match` statement, or even a branch of an `if` statement.

Because of `return`'s "any" type, the entire `if` typechecks to `Number` because of the `else` branch. However, the `if` is in the final position of the function body, so it (would be) implicitly returned, which means the `if` must typecheck to `String`, because of the `return` statement which declares the return type of the function to `String`, which in turn means that the `else` branch must also be `String`.

We feel that's a lot of mental hoops to jump through! What's far simpler is an error message that tells you that you need to add a return statement in the `else` branch (or on the whole `if`), and then it's obvious why the types don't line up.

This is why we made the decision to require all returns be explicit if a `return` statement is used. 🙂


compiler/src/formatting/format.re
```diff
@@ -3071,6 +3071,18 @@ and print_expression =
...
+        Doc.text("return"),
+        switch (expr) {
+        | Some(expr) =>
+          Doc.concat([
+            Doc.space,
+            print_expression(~original_source, ~comments, expr),
+          ])
+        | None => Doc.nil
+        },
+      ])
```
phated, line 3085: Need some tests for this?

compiler/test/suites/returns.re
```diff
@@ -0,0 +1,106 @@
+open Grain_tests.TestFramework;
+open Grain_tests.Runner;
+
+describe("early return", ({test, testSkip}) => {
```
phated, line 4: Do we not want any snapshot tests for this?

phated (changes requested):
SoOoOoOo excited for this feature. Really easy to follow implementation too ❤️ 

One question and one request (I think we should have some formatter input/output tests).

ospencer:
@phated Added those tests.

phated (approved):
Delightful!

PR #1465, opened 2022-11-03
fix(grainfmt):  Fix formatting of infix operators

### LINKED ISSUE #1390: Grainfmt: Indent binops when wrapping

marcusroberts:
Fixed #1390 

Lays out bin ops the same as Prettier
Fixes asserts over multiple lines


phated (commented):
Before I review the implementation, I had a couple comments/questions on some of the output.

compiler/src/formatting/format.re
```diff
@@ -1744,25 +1759,42 @@ and print_application =
...
+      ~original_source,
+      ~comments,
+      func,
+    )
   };
 }
 
 and print_infix_application =
     (
+      ~in_conditional: option(bool)=?,
```
phated, line 1782: Small nit, but get we make all of these `in_conditional` flags to not be optional? If we are going to have a boolean flag like this hanging around (you know I don't like them in general), then I'd prefer we are required to specify "yes" or "no" so we don't miss it somewhere.

stdlib/runtime/bigint.gr
```diff
@@ -1827,10 +1827,10 @@ let baseCaseDivRem = (a: WasmI32, b: WasmI32, result: WasmI32) => {
...
   let m = getHalfSize(a) - n
   let qsize = (if (WasmI32.eqz(WasmI32.and(m + 1n, 1n))) {
-        m + 1n
-      } else {
-        m + 2n
-      }) >>
+    m + 1n
+  } else {
+    m + 2n
+  }) >>
```
phated, line 1833: I think you mistook my comment below this to indicate that I didn't want this `if` statement indented, but it was actually about putting the `>>` and `1n` on the same line since they don't actually wrap.

I don't know how I feel about you taking away the indent. @ospencer thoughts?

phated (changes requested):
I have some follow ups

stdlib/runtime/bigint.gr
ospencer: To me it seems like it should be indented the way it was before?

compiler/src/formatting/format.re
```diff
@@ -1846,46 +1877,90 @@ and print_infix_application =
...
+      line_comment_docs,
+      if (after_comments_docs == Doc.nil) {
+        Doc.nil;
+      } else {
+        force_break_if_line_comment(~separator=Doc.nil, line_comments);
+      },
       after_comments_docs,
-      force_break_if_line_comment(~separator=Doc.nil, after_comments),
-      rhs,
+      if (expression_parent == ConditionalExpression) {
```
phated, line 1943: Should we use a pattern match here? That will allow us to expand the `expression_parent_type` in the future and be warned that we need to handle a new case

stdlib/runtime/bigint.gr
```diff
@@ -216,12 +216,12 @@ let setFlag = (ptr, flag, value) => {
...
-      0n
-    } else {
-      1n
-    }),
+      flag *
+        (if (WasmI32.eqz(value)) {
+          0n
+        } else {
+          1n
+        }),
```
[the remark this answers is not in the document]
phated: @marcusroberts I think you "undid" the wrong line

stdlib/runtime/bigint.gr
phated: @marcusroberts this was the line that needed to be reverted

marcusroberts:
Apologies, I've made some changes, broken tests and not fixed formatting, and I have to head to the airport, so I guess this won't make it into the release sorry!

marcusroberts:
Ok, I have an 8 hour flight, let me get this fixed

marcusroberts:
OK, I think that fixes everything!

compiler/src/formatting/format.re
```diff
@@ -1846,46 +1878,96 @@ and print_infix_application =
...
+      | ConditionalExpression =>
+        Doc.concat([
+          Doc.line,
+          if (right_is_leaf) {
+            Doc.group(rhs_expr);
+          } else {
+            rhs_expr;
+          },
+        ])
+      | _ =>
```
phated, line 1959: This should not be using a catch-all. That defeats the purpose of turning it into a switch statement

compiler/src/formatting/format.re
```diff
@@ -2750,8 +2962,9 @@ and print_expression =
         switch (true_expr.pexp_desc) {
         | PExpBlock(expressions) =>
           Doc.concat([
-            Doc.space,
+            Doc.space, // FIX me this needs to indent
```
phated, line 2965: ?

compiler/src/formatting/format.re
```diff
@@ -6,6 +6,11 @@ open Grain_diagnostics;
 
 module Doc = Res_doc;
 
+type expression_parent_type =
+  | ConditionalExpression
+  | GenericExpression
+  | InfixOperation;
```
phated, line 12: Unused currently?

phated (approved):
Thanks for all the work on this @marcusroberts! I'm happy with it now, but we need to make sure @ospencer likes the wrapping on `= if` statements

ospencer (approved):
Looks great, excellent work @marcusroberts!

PR #1467, opened 2022-11-04
fix(grainfmt): Handle chained value bindings properly

### LINKED ISSUE #1454: Formatter: Issue regarding comments with chained definitions
```
export let gammaTwo = z => {
  let g = 7, // g represents the precision desired, p is the values of p[i] to plug into Lanczos' formula
      p = [0.99999999999980993, 676.5203681218851];
}
```
the above is the code before formatting and below is after formatting.
```
export let gammaTwo = z => {
  let g = 7, // g represents the precision desired, p is the values of p[i] to plug into Lanczos' formula p = [
    0.99999999999980993,
    676.5203681218851,
  ]
}
```
In formatter moves the `p = [` up into the comment on the next line making the code invalid.

marcusroberts:
Fixed #1454 

This formats chained value bindings properly so that they now break over lines if over the width or end in line comments.

It does change how we display chained functions, I think it's more consistent now but let me know what you think

### DIFF
--- a/compiler/test/formatter_outputs/chained.gr
+++ b/compiler/test/formatter_outputs/chained.gr
@@ -1,6 +1,6 @@
 let g = 7, // g represents the precision desired, p is the values of p[i] to plug into Lanczos' formula
-p = 8, // another comment
-q = 32
+  p = 8, // another comment
+  q = 32
 
 let myFunction2 = x => {
   let inter = x + 1
--- a/compiler/test/formatter_outputs/chained.gr
+++ b/compiler/test/formatter_outputs/chained.gr
@@ -1,0 +1,23 @@
+let g = 7, // g represents the precision desired, p is the values of p[i] to plug into Lanczos' formula
+p = 8, // another comment
+q = 32
+
+let myFunction2 = x => {
+  let inter = x + 1
+  "some string"
+},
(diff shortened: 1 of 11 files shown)


compiler/test/formatter_outputs/chained.gr
```diff
@@ -0,0 +1,23 @@
+let g = 7, // g represents the precision desired, p is the values of p[i] to plug into Lanczos' formula
+p = 8, // another comment
+q = 32
```
phated, line 3: Like @marcusroberts mentioned in discord, it would be ideal if non-function bindings were indented to line-up with `let` but I don't want functions formatted that way, so I'm fine with this.

phated (approved):
LGTM! Thanks @marcusroberts

compiler/test/formatter_outputs/chained.gr
ospencer: Could we special case functions? I agree that it would look bad for functions, but I'm not a fan of how this looks not indented.

marcusroberts: I now indent everything that's not a lambda.

ospencer (approved):

phated (approved):

Issue #1463, opened 2022-10-27
Lexer: Add Reserved Keywords


spotandjake:
Currently grain plans on adding newer features such as `return` it would help in the future with the possiblity of breaking code if keywords such as `async`, `await`, `namespace`, `module`, `open` were marked as reserved keywords by the lexer and were not allowed to be used as identifiers.

PR #1471, opened 2022-11-07
feat(compiler)!: Add additional reserved keywords

spotandjake:
Added `module`, `macro`, `yield` to Reserved Keywords List, with the goal of making future breaking changes less intrusive. 

If there are any additional reserved keywords that would make sense to add let me know.

Closes: #1463


ospencer (approved):

phated (changes requested):
I'm not sold on these keywords so I don't really want to reserve them.

compiler/src/parsing/lexer.re
```diff
@@ -249,6 +249,11 @@ let rec token = lexbuf => {
   | "try" => positioned(TRY)
   | "throw" => positioned(THROW)
   | "catch" => positioned(CATCH)
+  | "async" => positioned(ASYNC)
+  | "await" => positioned(AWAIT)
```
ospencer, line 253: @phated doesn't want to reserve these two, so let's remove them.

ospencer (changes requested):

spotandjake:
Made those changes

ospencer (approved):

compiler/src/parsing/lexer.re
HTG-YT: Is there explicit reason why though? Perhaps it may come in handy in the future-

spotandjake: Don't want to lock us into a system we might not use.

marcusroberts (approved):

compiler/src/parsing/lexer.re
```diff
@@ -250,6 +250,9 @@ let rec token = lexbuf => {
   | "try" => positioned(TRY)
   | "throw" => positioned(THROW)
   | "catch" => positioned(CATCH)
+  | "module" => positioned(MODULE)
```
ospencer, line 253: Just want to preemptively avoid a merge conflict here :P 

```suggestion
```

compiler/src/parsing/parser.mly
```diff
@@ -42,7 +42,7 @@ module Grain_parsing = struct end
 %token EOL EOF
 
 // reserved tokens
-%token TRY CATCH COLONCOLON
+%token TRY CATCH COLONCOLON MODULE MACRO YIELD
```
ospencer, line 45:
```suggestion
%token TRY CATCH COLONCOLON MACRO YIELD
```

phated (approved):

PR #1443, opened 2022-10-04
feat(stdlib): Add additional functions to Number module

spotandjake:
Adds `Number.pow`, `Number.exp`, `Number.nan`, `Number.fact`, `Number.toDegree`, `Number.toRadian` this adds the unblocked functions from #1373, along with `Number.toDegree` and `Number.toRadian`. 

Ref: #1017

PR #1475, opened 2022-11-10
fix(runtime): Handle bigint mul/div within Number correctly

ospencer:
Resolves the bug discovered in #1443


phated (approved):
Thanks for finding and fixing this! ❤️

peblair (approved):

Issue #1470, opened 2022-11-07
stdlib: Array.rotate rotates opposite to List.rotate


spotandjake:
@alex-snezhko noticed that `Array.rotate` and `List.rotate` rotate in opposite directions with. 

`Array.rotate` - Rotating to the right.
and
`List.rotate` - Rotating to the left.

While you are able to rotate the other direction by using negative indexes, The behaviour seems random from a user perspective and could cause confusion.

alex-snezhko:
Since immutable arrays (#1477) make use of left-rotation like `List`, should `Array` be modified to rotate to the left as well?

ospencer:
Yes. Left rotation for all.

Ruby's doc for it makes it feel really intuitive:

> Returns a [new](https://apidock.com/ruby/Array/new/class) array by rotating self so th[at](https://apidock.com/ruby/Array/at) the element at [count](https://apidock.com/ruby/Array/count) is the [first](https://apidock.com/ruby/Array/first) element of the [new](https://apidock.com/ruby/Array/new/class) array.
> If [count](https://apidock.com/ruby/Array/count) is negative then it rotates in the opposite direction, starting from the end of self where -1 is the [last](https://apidock.com/ruby/Array/last) element.

Issue #775, opened 2021-07-10
Compiler warning if two types have any field names in common


cician:
The compiler emits a warning for the following program. Unless there's something fundamental I don't understand about Grain's type system, I think the two record types should coexist peacefully and stay independent of each other if I use explicit type annotations.

``` grain
record A {
  a: Number
}

record B {
  a: Number,
  b: Number
}

let a = {
  a: 1
} : A

let b = {
  a: 1,
  b: 2
} : B

let c = b.a // Warning: a belongs to several types: B A The first one was selected. Please disambiguate if this is wrong.
```

Same happens in this case:
``` grain
let f = (x: B) => {
  let c = b.a
}
```

alexsnezhko3:
To follow up on this, is there any way to get this warning to go away? I ran into this same situation and have tried disambiguating in various ways but the warning still shows

spotandjake:
> To follow up on this, is there any way to get this warning to go away? I ran into this same situation and have tried disambiguating in various ways but the warning still shows

you can do ```
{ recordInfo } : Type
``` which should prevent the warning.

**Comment by SteveSandersonMS** [comment]
This project looks really interesting!

> you can do { recordInfo } : Type which should prevent the warning

Could you clarify further? Considering the following code:

```grain
record Person { id: Number }
record Company { id: Number }

let person = { id: 123 }: Person
let company = { id: 123 }: Company

print(person.id)
print(company.id) // For some reason this gives a warning even though the preceding line is considered fine
```

I'm guessing it's a compiler bug that `person.id` is allowed whereas `company.id` is not. Is that right?

And assuming so it's possible to work around by disambiguating manually, what's the correct syntax for the final line in this sample?

spotandjake:
The name of this issue has been changed because this isnt exclusive to records.

# Record Issue
The initial issue references Records which can be handled just by adding a `: TypeName` after the record definition.

# Enums
Enums have the same issue take the example program here:
```
module Test

enum NodeA {
  Same, // Same
  DiffB,
}
enum NodeB {
  Same, // Same
  DiffA,
}
let a = Same: NodeB
match (a) {
  DiffA => void,
  Same => void,
}
```
as can be seen in the sample program we are able to dissambiguate at the deffinition the same way we do for records ` Same: NodeB` but we are not able to dissambiguate at the match statement leaving a warning on the line `  Same => void`.

A note is lets say you had a program like:
```
enum Status {
  Success,
  Failure,
}
let x = Success
[...]

PR #1477, opened 2022-11-12
feat(stdlib): Immutable array

alex-snezhko:
An efficient immutable array implementation based off of Elm's Array. For the most part the API is the same as Array with a few exceptions to fit in with immutability.

### DIFF
--- a/compiler/test/stdlib/immutablearray.test.gr
+++ b/compiler/test/stdlib/immutablearray.test.gr
@@ -78,6 +78,7 @@ assert ImmutableArray.init(32, identity) == fromList(noTail)
 
 assert ImmutableArray.make(5, 10) == fromList([10, 10, 10, 10, 10])
 assert ImmutableArray.make(0, 10) == fromList([])
+assert ImmutableArray.make(6, "a") == fromList(["a", "a", "a", "a", "a", "a"])
 
 // ImmutableArray.append
 
--- a/compiler/test/stdlib/immutablearray.test.gr
+++ b/compiler/test/stdlib/immutablearray.test.gr
@@ -37,22 +37,6 @@ assert !ImmutableArray.isEmpty(fromList([1, 2, 3]))
 assert ImmutableArray.length(bltArr) == branchingLongTailLen
 assert ImmutableArray.length(noTailArr) == 32
 
-// ImmutableArray.getOpt
-
-assert ImmutableArray.getOpt(1, arr) == Some(2)
-assert ImmutableArray.getOpt(-1, arr) == Some(3)
-assert ImmutableArray.getOpt(3, arr) == None
-assert ImmutableArray.getOpt(-4, arr) == None
-
-assert ImmutableArray.getOpt(0, bltArr) == Some(0)
-assert ImmutableArray.getOpt(branchingLongTailLen - 1, bltArr) ==
-  Some(branchingLongTailLen - 1)
-assert ImmutableArray.getOpt(branchingLongTailLen, bltArr) == None
-
-assert ImmutableArray.getOpt(0, noTailArr) == Some(0)
-assert ImmutableArray.getOpt(31, noTailArr) == Some(31)
-assert ImmutableArray.getOpt(32, noTailArr) == None
-
 // ImmutableArray.get
 
 assert ImmutableArray.get(1, arr) == 2
@@ -294,6 +278,13 @@ let arrA = fromList([1, 2, 3, 4, 5])
 let arrB = fromList([5, 4, 3, 2, 1])
 
 assert ImmutableArray.unzip(ImmutableArray.zip(arrA, arrB)) == (arrA, arrB)
+
+let arrA = fromList([1, 2, 3, 4])
+let arrB = fromList([3, 2, 1])
+let trimmedA = fromList([1, 2, 3])
+assert ImmutableArray.unzip(ImmutableArray.zip(arrA, arrB)) == (trimmedA, arrB)
+assert ImmutableArray.unzip(ImmutableArray.zip(arrB, arrA)) == (arrB, trimmedA)
+
 assert ImmutableArray.unzip(fromList([])) == (fromList([]), fromList([]))
 
 // ImmutableArray.count
--- a/compiler/test/stdlib/immutablearray.test.gr
+++ b/compiler/test/stdlib/immutablearray.test.gr
@@ -1,0 +1,387 @@
+import ImmutableArray from "immutablearray"
+import List from "list"
+import String from "string"
+import Range from "range"
+
+let fromList = ImmutableArray.fromList
+let arr = fromList([1, 2, 3])
+
+// for testing the case where the tail is empty
+let noTail = Range.map(identity, Range.Exclusive(0, 32))
+let noTailArr = fromList(noTail)
+
+// for testing the case where the tree branches and has a tail of max length
+let branchingLongTailLen = 32 * 33 + 31
+let branchingLongTail = Range.map(identity, Range.Exclusive(0, branchingLongTailLen))
+let bltArr = fromList(branchingLongTail)
+
+// for testing the case where the tree is about to branch
+let almostBranchingLen = 32 * 32 + 31
+let almostBranching = Range.map(identity, Range.Exclusive(0, almostBranchingLen))
+let abArr = fromList(almostBranching)
+
+assert ImmutableArray.empty == fromList([])
+
+// ImmutableArray.length & ImmutableArray.isEmpty
+
+assert ImmutableArray.length(ImmutableArray.empty) == 0
+assert ImmutableArray.isEmpty(ImmutableArray.empty)
+assert ImmutableArray.length(fromList([1, 2, 3])) == 3
+assert !ImmutableArray.isEmpty(fromList([1, 2, 3]))
+assert ImmutableArray.length(bltArr) == branchingLongTailLen
+assert ImmutableArray.length(noTailArr) == 32
+
+// ImmutableArray.getOpt
+
+assert ImmutableArray.getOpt(1, arr) == Some(2)
+assert ImmutableArray.getOpt(-1, arr) == Some(3)
+assert ImmutableArray.getOpt(3, arr) == None
+assert ImmutableArray.getOpt(-4, arr) == None
+
+assert ImmutableArray.getOpt(0, bltArr) == Some(0)
+assert ImmutableArray.getOpt(branchingLongTailLen - 1, bltArr) == Some(branchingLongTailLen - 1)
+assert ImmutableArray.getOpt(branchingLongTailLen, bltArr) == None
+
+assert ImmutableArray.getOpt(0, noTailArr) == Some(0)
+assert ImmutableArray.getOpt(31, noTailArr) == Some(31)
+assert ImmutableArray.getOpt(32, noTailArr) == None
+
+// ImmutableArray.get
+
+assert ImmutableArray.get(1, arr) == 2
+assert ImmutableArray.get(-1, arr) == 3
+assert ImmutableArray.get(-2, arr) == 2
+assert ImmutableArray.get(0, arr) == 1
+
+// ImmutableArray.set
+
+let arr = fromList([1, 2])
+
+let arr = ImmutableArray.set(0, 5, arr)
+assert ImmutableArray.get(0, arr) == 5
+let arr = ImmutableArray.set(-1, 5, arr)
+assert ImmutableArray.get(1, arr) == 5
+let arr = ImmutableArray.set(32, -1, bltArr)
+assert ImmutableArray.get(32, arr) == -1
+let arr = ImmutableArray.set(branchingLongTailLen - 1, -1, bltArr)
+assert ImmutableArray.get(branchingLongTailLen - 1, arr) == -1
+
+// ImmutableArray.init
+
+assert ImmutableArray.init(5, identity) == fromList([0, 1, 2, 3, 4])
+assert ImmutableArray.init(10, x => x * 2) == fromList([0, 2, 4, 6, 8, 10, 12, 14, 16, 18])
+assert ImmutableArray.init(0, x => x * 2) == fromList([])
+assert ImmutableArray.init(branchingLongTailLen, identity) == fromList(branchingLongTail)
+assert ImmutableArray.init(32, identity) == fromList(noTail)
+
+// ImmutableArray.make
+
+assert ImmutableArray.make(5, 10) == fromList([10, 10, 10, 10, 10])
+assert ImmutableArray.make(0, 10) == fromList([])
+
+// ImmutableArray.append
+
+let arr1 = fromList([1, 2])
+let arr2 = fromList([3, 4, 5])
+
+assert ImmutableArray.append(arr1, arr2) == fromList([1, 2, 3, 4, 5])
+assert ImmutableArray.append(fromList([]), fromList([])) == fromList([])
+assert ImmutableArray.append(arr1, fromList([])) == arr1
+assert ImmutableArray.append(fromList([]), arr1) == arr1
+assert ImmutableArray.append(abArr, fromList([almostBranchingLen])) == fromList(List.append(almostBranching, [almostBranchingLen]))
+let val = branchingLongTailLen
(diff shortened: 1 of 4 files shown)


ospencer (commented):
I'll get you a proper, full review, but from just a cursory look, this looks stunning 🤩

stdlib/immutablearray.gr
```diff
@@ -0,0 +1,935 @@
...
+      Leaf(vals) => {
+        Leaf(arraySetCopy(index & bitmask, value, vals))
+      },
+    }
+    arraySetCopy(pos, newVal, node)
+  }
+
+  let { length, shift, root, tail } = array
+  if (index < 0 || index >= length) {
+    array
```
ospencer, line 237: Should we fail in this case?

stdlib/immutablearray.gr
```diff
@@ -0,0 +1,935 @@
...
+  } else if (index >= tailIndex(length)) {
+    { length, shift, root, tail: arraySetCopy(index & bitmask, value, tail) }
+  } else {
+    { length, shift, root: setInner(shift, root), tail }
+  }
+}
+
+// inserts a new tail into the array. If the length of the tail equals the
```
ospencer, line 245:
```suggestion
// Inserts a new tail into the array. If the length of the tail equals the
```

ospencer (commented):
This looks fantastic! Just a few nits on comment style and one question about `set`.

stdlib/immutablearray.gr
marcusroberts: I think so, otherwise it's failing silently.

alex-snezhko: Sure, should it return a result or throw an exception?

alex-snezhko: I just went with throwing `IndexOutOfBounds` for now since it matches regular array behavior. On this topic, do you think that `getOpt` should exist at all, and if so should it be added to `Array` as well?

marcusroberts: Personally, I like having getOpt available.

compiler/test/stdlib/immutablearray.test.gr
```diff
@@ -0,0 +1,430 @@
...
+assert ImmutableArray.get(32, arr) == -1
+let arr = ImmutableArray.set(branchingLongTailLen - 1, -1, bltArr)
+assert ImmutableArray.get(branchingLongTailLen - 1, arr) == -1
+
+// ImmutableArray.init
+
+assert ImmutableArray.init(5, identity) == fromList([0, 1, 2, 3, 4])
+assert ImmutableArray.init(10, x => x * 2) ==
+  fromList([0, 2, 4, 6, 8, 10, 12, 14, 16, 18])
+assert ImmutableArray.init(0, x => x * 2) == fromList([])
```
phated, line 81:
```suggestion
assert ImmutableArray.init(0, x => fail "called init on 0 length") == fromList([])
```
Should have something like this to make sure this function is never called.

compiler/test/stdlib/immutablearray.test.gr
```diff
@@ -0,0 +1,430 @@
...
+// ImmutableArray.toList
+
+assert ImmutableArray.toList(fromList([])) == []
+assert ImmutableArray.toList(fromList([1, 2, 3])) == [1, 2, 3]
+assert ImmutableArray.toList(fromList(branchingLongTail)) == branchingLongTail
+assert ImmutableArray.toList(fromList(noTail)) == noTail
+
+// ImmutableArray.map
+
+assert ImmutableArray.map(incr, fromList([])) == fromList([])
```
phated, line 214:
```suggestion
assert ImmutableArray.map(() => fail "Map iterator should not be called", fromList([])) == fromList([])
```
Suggestion, not tested

compiler/test/stdlib/immutablearray.test.gr
```diff
@@ -0,0 +1,430 @@
...
+assert ImmutableArray.product(arrA, arrB) ==
+  fromList([(1, 2), (1, 3), (1, 4), (2, 2), (2, 3), (2, 4)])
+assert ImmutableArray.product(arrC, arrB) == fromList([(1, 2), (1, 3), (1, 4)])
+assert ImmutableArray.product(arrC, arrD) == fromList([(1, 2)])
+assert ImmutableArray.product(fromList([]), fromList([])) == fromList([])
+
+// ImmutableArray.zip
+
+let arrA = fromList([1, 2, 3, 4, 5])
+let arrB = fromList([5, 4, 3, 2, 1])
```
phated, line 259: Are we supporting different length ImmutableArrays? We are changing the other `zip` functions for that in 0.6

compiler/test/stdlib/immutablearray.test.gr
```diff
@@ -0,0 +1,430 @@
...
+assert ImmutableArray.zip(arrA, arrB) ==
+  fromList([(1, 5), (2, 4), (3, 3), (4, 2), (5, 1)])
+
+// ImmutableArray.zipWith
+
+let addFn = (a, b) => a + b
+let arrA = fromList([1, 2, 3])
+let arrB = fromList([4, 5, 6])
+
+assert ImmutableArray.zipWith(addFn, arrA, arrB) == fromList([5, 7, 9])
```
phated, line 270: See question above

compiler/test/stdlib/immutablearray.test.gr
```diff
@@ -0,0 +1,430 @@
...
+
+assert ImmutableArray.unzip(ImmutableArray.zip(arrA, arrB)) == (arrA, arrB)
+assert ImmutableArray.unzip(fromList([])) == (fromList([]), fromList([]))
+
+// ImmutableArray.count
+
+let arr = fromList([1, 2, 3])
+
+assert ImmutableArray.count(x => x > 0, arr) == 3
+assert ImmutableArray.count(x => x > 0, fromList([])) == 0
```
phated, line 290: should have `fail` like others

compiler/test/stdlib/immutablearray.test.gr
```diff
@@ -0,0 +1,430 @@
...
+
+// ImmutableArray.slice
+let testChars = fromList(['a', 'b', 'c'])
+
+assert ImmutableArray.slice(0, 1, testChars) == fromList(['a'])
+assert ImmutableArray.slice(1, ImmutableArray.length(testChars), testChars) ==
+  fromList(['b', 'c'])
+assert ImmutableArray.slice(0, 0, testChars) == fromList([])
+// Purposefully huge number
+assert ImmutableArray.slice(1, 10000, testChars) == fromList(['b', 'c'])
```
phated, line 368: Why is this okay?

phated (changes requested):
Sorry for the series of single comments. I was told that I probably wouldn't have feedback, so I started and didn't stop.

stdlib/immutablearray.gr
alex-snezhko: Sounds good. On second thought, should I actually rename it to `getOption`? The standard library generally doesn't have a lot of contractions in names.

phated: Since `get` returns a value instead of the ImmutableArray, why did you choose to make it fallible and then have `getOpt`?

compiler/test/stdlib/immutablearray.test.gr
alex-snezhko: Yep, the behavior should be the same as `List.zip` where it has the length of the smaller array. I'll add a test for this

stdlib/immutablearray.gr
alex-snezhko: In order to mirror the behavior for `Array`, where `get` would just return the value at that index and throw `IndexOutOfBounds` otherwise. I figured I'd also add `getOpt` to be used where safety is desired. Also why would `get` return the array?

phated: Honestly, `Array.get` doesn't return an option because it is implemented as `arr[idx]` which throws on index out of bounds. It'd be great if writing `arr[idx]` was the "throws" version and `Array.get` were the safe version.

> Also why would get return the array?

Sorry, I meant in opposition to `set` which should throw with IndexOutOfBounds because it needs to return the ImmutableArray (and returning `Result<ImmutableArray, Err>` is a terrible API).

compiler/test/stdlib/immutablearray.test.gr
```diff
@@ -0,0 +1,446 @@
...
+assert ImmutableArray.zipWith(multFn, arrA, arrB) == fromList([4, 10])
+assert ImmutableArray.zipWith(addFn, fromList([]), arrB) == fromList([])
+
+// ImmutableArray.unzip
+
+let arrA = fromList([1, 2, 3, 4, 5])
+let arrB = fromList([5, 4, 3, 2, 1])
+
+assert ImmutableArray.unzip(ImmutableArray.zip(arrA, arrB)) == (arrA, arrB)
+assert ImmutableArray.unzip(fromList([])) == (fromList([]), fromList([]))
```
phated, line 297: @alex-snezhko does unzip also work with different lengths? Can you test it?

stdlib/immutablearray.gr
alex-snezhko: Ok so are you saying that `getOpt` should be removed and `get` should always return `Option` (changing this behavior for `Array` too)? I still think it would be nice to have a "quick and dirty" way to extract a value from an array though rather than always having to jump through the hoop of unwrapping or handle the `None` case.

I suppose an alternative solution would be to somehow allow the bracket syntax to be used for immutable arrays (either somehow make this an allowable way to construct operators or a built-in syntax for immutable arrays)?

phated: I really hate putting a "return type" in your function name, e.g. `getOption`. It's just bad design.

alex-snezhko: What about `safeGet`?

phated: I think we remove the version that returns an option until we decide to handle for mutable arrays.

stdlib/immutablearray.gr
```diff
@@ -0,0 +1,921 @@
...
+ * @param array1: The array containing elements to appear first
+ * @param array2: The array containing elements to appear second
+ * @returns The new array containing elements from `array1` followed by elements from `array2`
+ *
+ * @example append(fromList([1, 2]), fromList([3, 4, 5])) == fromList([1, 2, 3, 4, 5])
+ *
+ * @since v0.6.0
+ */
+export let append = (array1, array2) => {
+  // Magic number of 4 determined best from benchmarks
```
phated, line 373: Benchmarks in grain or is this like a comment from elm?

stdlib/immutablearray.gr
```diff
@@ -0,0 +1,921 @@
...
+ * array to have the length of the smaller array.
+ * 
+ * @param array1: The array to provide values for the first tuple element
+ * @param array2: The array to provide values for the second tuple element
+ * @returns The new array containing indexed pairs of `(a, b)`
+ *
+ * @since v0.6.0
+ */
+export let zip = (array1, array2) => {
+  // Does the trick :}
```
phated, line 794:
```suggestion
```

stdlib/immutablearray.gr
```diff
@@ -0,0 +1,921 @@
...
+ */
+export let zipWith = (fn, array1, array2) => {
+  fromList(List.zipWith(fn, toList(array1), toList(array2)))
+}
+
+/**
+ * Produces two arrays by splitting apart an array of tuples.
+ *
+ * @param array: The array of tuples to split
+ * @returns An array containing all elements from the first tuple element, and an array containing all elements from the second tuple element
```
phated, line 826:
```suggestion
 * @returns An array containing all elements from the first tuple element and an array containing all elements from the second tuple element
```

stdlib/immutablearray.gr
```diff
@@ -0,0 +1,921 @@
...
+ * @returns The sorted array
+ *
+ * @since v0.6.0
+ */
+export let sort = (comp, array) => {
+  fromList(List.sort(comp, toList(array)))
+}
+
+/**
+ * Rotates array elements by the specified amount to the left.
```
phated, line 903: Double checking that this is "correct" (in light of list and array operating in the opposite directions).

stdlib/immutablearray.md
```diff
@@ -0,0 +1,1069 @@
...
+|----|-----------|
+|`Number`|The number of elements in the array|
+
+Examples:
+
+```grain
+length(fromList([1, 2, 3, 4, 5])) == 5
+```
+
+### ImmutableArray.**getOpt**
```
phated, line 99: Needs to be regenerated to remove this

stdlib/immutablearray.md
```diff
@@ -0,0 +1,1069 @@
...
+|----|-----------|
+|`ImmutableArray<a>`|A new array containing the new element at the given index|
+
+Examples:
+
+```grain
+set(1, 9, fromList([1, 2, 3, 4, 5])) == fromList([1, 9, 3, 4, 5])
+```
+
+### ImmutableArray.**Builder**
```
phated, line 207: Why did this show up here? Does it need to be moved to the top of the file so it is hidden?

phated (changes requested):
Noticed a few more things

stdlib/immutablearray.gr
alex-snezhko: It's a comment from elm; I can make this clear in the comment

stdlib/immutablearray.md
alex-snezhko: It should indeed be moved. Upon moving it up to the top, I am now getting an ambiguity warning about `Builder` and `ImmutableArray` sharing a `tail` field; appears to be #775 . I'll just change `tail` from `Builder` to `btail` for now.

stdlib/immutablearray.gr
alex-snezhko: In my opinion this seems like the more appropriate direction to do a rotation in so I just went with a left rotation, although #1470 hasn't been commented on yet so I don't know what the consensus on what the "correct" behavior is

ospencer (approved):
Excellent work @alex-snezhko!

stdlib/immutablearray.gr
```diff
@@ -0,0 +1,929 @@
...
+}
+
+/**
+ * Retrieves the element from the array at the specified index.
+ * A negative index is treated as an offset from the end of the array.
+ *
+ * @param index: The index to access
+ * @param array: The array to access
+ * @returns The element from the array
+ * @throws IndexOutOfBounds: if the index being accessed is outside the array's bounds
```
phated, line 172:
```suggestion
 * @throws IndexOutOfBounds: When the index being accessed is outside the array's bounds
```

stdlib/immutablearray.gr
```diff
@@ -0,0 +1,929 @@
...
+
+/**
+ * Creates a new array in which the element at the specified index is set to a
+ * new value. A negative index is treated as an offset from the end of the array.
+ *
+ * @param index: The index to update
+ * @param value: The value to store
+ * @param array: The array to update
+ * @returns A new array containing the new element at the given index
+ * @throws IndexOutOfBounds: if the index being updated is outside the array's bounds
```
phated, line 208:
```suggestion
 * @throws IndexOutOfBounds: When the index being updated is outside the array's bounds
```

phated (changes requested):
Just doc nits that I'm going to apply

stdlib/immutablearray.md
```diff
@@ -0,0 +1,1038 @@
...
+
+|type|description|
+|----|-----------|
+|`ImmutableArray<a>`|A new array containing the new element at the given index|
+
+Throws:
+
+`IndexOutOfBounds`
+
+* if the index being updated is outside the array's bounds
```
phated, line 174:
```suggestion
* When the index being updated is outside the array's bounds
```

stdlib/immutablearray.md
```diff
@@ -0,0 +1,1038 @@
...
+
+|type|description|
+|----|-----------|
+|`a`|The element from the array|
+
+Throws:
+
+`IndexOutOfBounds`
+
+* if the index being accessed is outside the array's bounds
```
phated, line 130:
```suggestion
* When the index being accessed is outside the array's bounds
```

PR #1479, opened 2022-11-12
feat(stdlib)!: Make queue and stack mutable & provide Immutable submodules

alex-snezhko:
Mutable array-based implementations for queues and stacks. From a few benchmarks these implementations offer much faster performance than the current immutable `Queue` and `Stack` implementations (likely due to gains in cache locality because of underlying array representation), which may make them more viable for performance-sensitive applications.


phated:
@alex-snezhko I think we're going to hold this PR back from 0.5.5 because we think it is confusing to have Queue and Stack be the immutable versions but Map & Set are mutable (and then have the inverse have the special naming). We're going to deprecated the immutable Queue & Stack in 0.5.5 and then have you make this the breaking change that moves stuff around.

How does that sound?

alex-snezhko:
Sounds good!

marcusroberts (approved):
Easy to read and clear code, great documentation and tests.

Looks very good to me!

alex-snezhko:
#1644 and #1621 should probably be resolved before merging this

compiler/test/stdlib/immutablestack.test.gr
```diff
@@ -0,0 +1,33 @@
+import Stack from "immutablestack"
```
phated, line 1: Is this file going away since they are in the same file now?

phated (changes requested):
@alex-snezhko I've rebased this so I could generate the docs with my submodule support in #1684 - however, I think you need to update the `@since` and `@history` for the various things that moved around. e.g. Since functions moved from the root into the `Immutable` submodule, I think those should be `@since v0.6.0` with a history item saying they were originally a root API (similar to how we did those other PRs).

phated:
Rebased again and regenerated the docs with graindoc fixes.

stdlib/queue.gr
```diff
@@ -1,134 +1,296 @@
...
+/**
+ * A mutable FIFO (first-in-first-out) data structure.
+ */
 abstract record Queue<a> {
-  forwards: List<a>,
-  backwards: List<a>,
+  mut size: Number,
+  mut array: Array<Option<a>>,
+  mut headI: Number,
+  mut tailI: Number, // tail points to where the next element should be inserted
```
ospencer, line 25: Can we call these `headIndex` and `tailIndex`?

ospencer (commented):
Overall looks great! Would just love if these record fields were renamed.

ospencer:
Also looks like the set tests need to be formatted.

phated (approved):
I made the change Oscar wanted, so this should be good to queue.

alex-snezhko:
Thanks @phated :)

PR #1482, opened 2022-11-17
fix(grainfmt): Reduce binop format time

ospencer:
Because the call to `print_expression` was outside of the `if`, we got combinatorial blowup when processing binops. Moving the call to inside the `if` prevents this.


marcusroberts (approved):
Well spotted!

phated (approved):
Thanks for digging into this @ospencer ❤️

PR #1485, opened 2022-11-20
fix(compiler): Raise appropriate exception when modules are missing during dependency graph construction

### LINKED ISSUE #1480: Compiler: Give pretty error when file does not exist
image
When working on #1443 i messed up a file name and it was noticed that it gave a nasty error. rather then a human readable one.

peblair:
Fixes #1480. Additionally, this PR modifies the missing module error message to contain the correct source locations. For example, given the following files:
`tst.gr`:
```grain
import Number from "./numberr"
print(Number.min(2, 3))
```
`numberr.gr`:
```grain
import Number from "number"
import Foo from "foo" // <- does not exist
export let min = Number.min
```
The following error is reported:
```
 λ grain git:(philip/file-missing-import) > grain tst.gr
File "././numberr.gr", line 2, characters 16-21:
Error: Missing file for module foo
```

### DIFF
--- a/compiler/src/parsing/driver.re
+++ b/compiler/src/parsing/driver.re
@@ -126,7 +126,8 @@ let parse = (~name=?, lexbuf, source): Parsetree.parsed_program => {
   };
 };
 
-let scan_for_imports = (~defer_errors=true, filename: string): list(string) => {
+let scan_for_imports =
+    (~defer_errors=true, filename: string): list(loc(string)) => {
   let ic = open_in(filename);
   let lexbuf = Sedlexing.Utf8.from_channel(ic);
   try({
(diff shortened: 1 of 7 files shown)


phated:
Should we normalize the path? In the example `"././numberr.gr"` should probably remove one of the `./`

marcusroberts (approved):
Nice work!

phated (approved):
Looks good to me! I rebased and tested it with my LSP fix and things look great in the problems tab.

Screen Shot 2022-11-29 at 2 20 39 PM

I can't replicate the duplicate relativity in the example (`././numberr.gr`) because my file paths are always absolute. If Philip or anyone else can replicate, please open an issue explaining how you did it.

ospencer (approved):

PR #1486, opened 2022-11-20
fix(compiler): Perform partial match checking on let-binds

### LINKED ISSUE #1453: Non-exhaustive pattern match missing warning
It appears that attempting to do something like this:
```
let a = None
let Some(b) = a
```
does not generate a warning, when it probably should since it is not guaranteed to match.

peblair:
Fixes #1453.

New warnings with the fix:
image

### DIFF
--- a/compiler/src/typed/typecore.re
+++ b/compiler/src/typed/typecore.re
@@ -1991,6 +1991,20 @@ and type_let =
       )
       pat_list
       (List.map2 (fun (attrs, _) e -> attrs, e) spatl exp_list);*/
+  List.iter2(
+    (pat, (attrs, exp)) => {
+      ignore(
+        check_partial(
+          env,
+          pat.pat_type,
+          pat.pat_loc,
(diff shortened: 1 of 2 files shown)


compiler/src/typed/typecore.re
```diff
@@ -1991,6 +1991,20 @@ and type_let =
       )
       pat_list
       (List.map2 (fun (attrs, _) e -> attrs, e) spatl exp_list);*/
+  List.iter2(
```
phated, line 1994: This code looks really close to commented lines 1984-1990, so should we delete those?

phated (approved):
Just one question about cleaning up some commented code but otherwise this looks great to me ❤️

ospencer (approved):

compiler/src/typed/typecore.re
ospencer: Should def delete.

Issue #458, opened 2020-12-10
[Stdlib] Reduce allocations during Array.flatMap on large data sets


phated:
Currently, `Array.flatMap` creates an intermediate array for each iteration, but this could be sped up for large data sets with something like `Array.concat(Array.toList(Array.map()))`

We might want to vary this implementation based on the size of the array we are flatMapping.

phated:
Completed by #1487

PR #1487, opened 2022-11-22
feat(stdlib): Improve perf for array flatMap, array & list some & every

alex-snezhko:
Improves `Array.flatMap` performance addressing #458. Also small performance upgrade for `every` and `some` for `Array` and `List`, as I noticed that in their current state they would always traverse the entire collection even after a suitable match/counterexample was found.

### DIFF
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -233,6 +233,7 @@ export let rec flatMap = (fn, list) => {
 export let rec every = (fn, list) => {
   match (list) {
     [] => true,
+    // The short-circuiting of `&&` makes this tail-recursive
     [first, ...rest] => fn(first) && every(fn, rest),
   }
 }
@@ -250,6 +251,7 @@ export let rec every = (fn, list) => {
 export let rec some = (fn, list) => {
   match (list) {
     [] => false,
+    // The short-circuiting of `||` makes this tail-recursive
     [first, ...rest] => fn(first) || some(fn, rest),
   }
 }
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -230,10 +230,11 @@ export let rec flatMap = (fn, list) => {
  *
  * @since v0.1.0
  */
-export let every = (fn, list) => {
-  reduce((acc, value) => {
-    acc && fn(value)
-  }, true, list)
+export let rec every = (fn, list) => {
+  match (list) {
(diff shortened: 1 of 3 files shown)


phated (approved):
Looks great! Thanks for cleaning these up.

ospencer (changes requested):
The array functions look great! The list functions aren't tail-recursive anymore, but they need to be.

alex-snezhko:
@ospencer I was hesitant about this initially as well, but I think these implementations are actually tail recursive because of the short-circuiting of `&&` and `||`. It also looks like the wasm generated by these implementations is the same as implementations that are more evidently tail-recursive.

ospencer (commented):
I was the one who implemented `&&` and `||` and completely forgot that these are essentially freebies when it comes to tail calls. Hah. Of course, it's not obvious that it's tail recursive. I'm going to drop some comments in there so we don't forget, and this is good to go.

stdlib/list.gr
```diff
@@ -230,10 +230,11 @@ export let rec flatMap = (fn, list) => {
...
  * @since v0.1.0
  */
-export let every = (fn, list) => {
-  reduce((acc, value) => {
-    acc && fn(value)
-  }, true, list)
+export let rec every = (fn, list) => {
+  match (list) {
+    [] => true,
+    [first, ...rest] => fn(first) && every(fn, rest),
```
ospencer, line 236:
```suggestion
    // The short-circuiting of `&&` makes this tail-recursive
    [first, ...rest] => fn(first) && every(fn, rest),
```

stdlib/list.gr
```diff
@@ -246,10 +247,11 @@ export let every = (fn, list) => {
...
  * @since v0.1.0
  */
-export let some = (fn, list) => {
-  reduce((acc, value) => {
-    acc || fn(value)
-  }, false, list)
+export let rec some = (fn, list) => {
+  match (list) {
+    [] => false,
+    [first, ...rest] => fn(first) || some(fn, rest),
```
ospencer, line 253:
```suggestion
    // The short-circuiting of `||` makes this tail-recursive
    [first, ...rest] => fn(first) || some(fn, rest),
```

ospencer (approved):
Thanks @alex-snezhko!

PR #1490, opened 2022-11-29
fix(grainlsp): Properly surface errors in other files

phated:
While looking into #1485, I noticed that the reporting of a missing file was being shown on the line & col of the top-level file we were compiling. This lead me down a rabbit-hole where I found that all errors were being surfaced within the file you were editing (which is very wrong).

This PR changes the error surfacing logic of the LSP to check to see if the error matches the file being compiled. If it matches, the error is surfaced normally; however, if the filenames don't match, we set an error on 0,0 that says which file failed to compile and add its error message to the `related_information` list. This will show "problems" that link to the correct location in the source.

Here's an example for a missing module once #1485 lands:
Screen Shot 2022-11-28 at 7 08 58 PM

### DIFF
--- a/compiler/src/language_server/code_file.re
+++ b/compiler/src/language_server/code_file.re
@@ -39,7 +39,12 @@ let warning_to_diagnostic =
     },
   };
 
-  {range, severity: Warning, message: Grain_utils.Warnings.message(warn)};
+  {
+    range,
+    severity: Warning,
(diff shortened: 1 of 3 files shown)


ospencer (approved):
Looks good!

marcusroberts (approved):
Nice improvement!

PR #1491, opened 2022-11-30
refactor print_trailing_comments to use fold_right

phated:
@marcusroberts this is an example of how you can refactor your `List.append` usage to use `fold_right` and the "list spread" syntax. This should be waaayyy faster than doing all those `List.append` calls (unless I'm crazy).

I also refactored a bit to do away with the boolean, especially because the if statement for each is slightly different.


marcusroberts:
Thanks for this, so simple when seen!

marcusroberts (approved):

PR #1492, opened 2022-12-01
feat(graindoc): Add attribute for exceptions that may be thrown

### LINKED ISSUE #767: `@throws` attribute in Graindoc
In order to document errors in Grain functions, it would be nice if there was a `@throws` attribute that we could include in Graindoc comments. Something like:
```grain
/**
 * @throws SomeError: When you make a mistake
 */
```

phated:
Closes #767 

This adds a `@throws` attribute to graindoc. This is a draft because we need to figure out some design...

Currently, the "exception name" is optional; however, the generated table looks really weird without an exception name. Should we make this required?

Screen Shot 2022-11-30 at 8 28 53 PM

Things look good with the "exception name", but do we want to represent the entire pattern (like you'd match on)? Or just the name of the exception?
Screen Shot 2022-11-30 at 8 43 59 PM

I also started adding the attributes to the stdlib and it felt better to document separate exceptional cases individually, even if they are throwing the same exception.

Please take a look at this and give me some feedback so I can wrap up this feature. 🙏

### DIFF
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -246,7 +246,7 @@ let signature_item_in_range =
   };
 };
 
-// Used for joining multiple `@throws` annotations with the exact same name
+// Used for joining multiple `@throws` annotations with the exact same type
 module StringMap = Map.Make(String);
 
 let to_markdown = (~current_version, docblock) => {
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -246,6 +246,9 @@ let signature_item_in_range =
   };
 };
 
+// Used for joining multiple `@throws` annotations with the exact same name
+module StringMap = Map.Make(String);
+
 let to_markdown = (~current_version, docblock) => {
   let buf = Buffer.create(0);
   Buffer.add_string(buf, Markdown.heading(~level=3, docblock.name));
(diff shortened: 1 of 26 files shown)


ospencer:
I vote for required exception, with the parameters! I think that's the most helpful. I think I personally would want just one entry per exception, and you explain the cases in which it happens. Listing each of the cases separately feels a little unwieldy to me. But I really like this!

phated:
> Listing each of the cases separately feels a little unwieldy to me.

This was my assumption when I started writing the descriptions but they became insanely unwieldy quickly. Take a look at one of the cases where there are 4 exception states and try to think how you'd write that concisely.

ospencer:
Yeah I don't disagree with you, but I just don't like them separate 😩

phated:
Are you more concerned with the process of documenting or the output? We could make this implementation a lot more complicated and join any exceptions that have the same name (and build up a bullet list in the description??).

ospencer:
The output, so maybe you're onto something there!

phated:
> I vote for required exception, with the parameters!

@ospencer additionally, should it be `InvalidArgument(String)` or `InvalidArgument(msg)`?

ospencer:
I like seeing types, so I'd go with `InvalidArgument(String)`.

compiler/graindoc/docblock.re
```diff
@@ -246,6 +246,9 @@ let signature_item_in_range =
   };
 };
 
+// Used for joining multiple `@throws` annotations with the exact same name
```
phated, line 249:
```suggestion
// Used for joining multiple `@throws` annotations with the exact same type
```

compiler/src/diagnostics/comments.re
```diff
@@ -142,6 +146,23 @@ module Attribute = {
     };
   };
 
+  let parse_throws = (~attr, content) => {
+    let re = Str.regexp({|^\([^:]+\):[ ]+\(.+\)$|});
+    if (Str.string_match(re, content, 0)) {
+      let attr_type = Str.matched_group(1, content);
+      let attr_desc = Str.matched_group(2, content);
+      // Name and Type are conflated for exceptions
```
phated, line 154:
```suggestion
```

phated:
This should be ready to go. We now require `@throws` to be provided a type and I changed the output to join all `@throws` with the same type into one heading with a bullet list underneath. The bullet list was annoying to put in the table, so I through that away.

Here's an example of the new output:
Screen Shot 2022-12-03 at 5 49 26 PM

compiler/graindoc/docblock.re
```diff
@@ -353,6 +356,47 @@ let to_markdown = (~current_version, docblock) => {
...
+    |> List.fold_left(
+         (map, attr: Comments.Attribute.t) => {
+           switch (attr) {
+           | Throws({attr_type: Some(attr_type), attr_desc}) =>
+             StringMap.update(
+               attr_type,
+               descs => {
+                 switch (descs) {
+                 | None => Some([attr_desc])
+                 | Some(descs) => Some([attr_desc, ...descs])
```
ospencer, line 371: This never gets reversed, so the descriptions are backwards in the output.

ospencer (approved):
Great work!

marcusroberts (approved):
This looks fantastic to me!

PR #1493, opened 2022-12-01
chore(stdlib): Utilize early return where beneficial

jozanza:
This PR refactors the std lib to use early returns where it's ergonomic/applicable.

### DIFF
--- a/stdlib/runtime/string.gr
+++ b/stdlib/runtime/string.gr
@@ -55,21 +55,19 @@ let findTypeMetadata = typeId => {
   let mut metadataPtr = WasmI32.load(_RUNTIME_TYPE_METADATA_PTR, 0n)
   let mut modData = -1n
   let mut modTypesCount = 0n
-  let mut typeData = -1n
-  while (metadataPtr != 0n && typeData == -1n) {
+  while (metadataPtr != 0n) {
     modData = metadataPtr + 12n
     modTypesCount = WasmI32.load(metadataPtr, 8n)
     for (let mut i = 0n; i < modTypesCount; i += 1n) {
       if (WasmI32.load(modData, 4n) == typeId) {
-        typeData = modData
-        break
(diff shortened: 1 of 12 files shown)


stdlib/runtime/string.gr
```diff
@@ -74,27 +74,24 @@ let getVariantName = variant => {
...
-    while (block < end) {
-      if (WasmI32.load(block, 4n) == variantId) {
-        let length = WasmI32.load(block, 8n)
-        let str = allocateString(length)
-        Memory.copy(str + 8n, block + 12n, length)
-        result = str
-        break
-      }
-      block += WasmI32.load(block, 0n)
+  if (block == -1n) return { -1n }
```
ospencer, line 77:
```suggestion
  if (block == -1n) return -1n
```

stdlib/runtime/string.gr
```diff
@@ -74,27 +74,24 @@ let getVariantName = variant => {
...
+      let str = allocateString(length)
+      Memory.copy(str + 8n, block + 12n, length)
+      return str
     }
-
-    result
+    block += WasmI32.load(block, 0n)
   }
+
+  return { -1n }
```
ospencer, line 94:
```suggestion
  return -1n
```

stdlib/runtime/string.gr
```diff
@@ -104,26 +101,23 @@ let getRecordFieldNames = record_ => {
 
   let mut fields = findTypeMetadata(typeId)
 
-  if (fields == -1n) {
-    -1n
-  } else {
-    fields += 8n
-
-    let fieldArray = allocateArray(arity)
+  if (fields == -1n) return { -1n }
```
ospencer, line 104:
```suggestion
  if (fields == -1n) return -1n
```

ospencer (approved):

PR #1494, opened 2022-12-01
chore(stdlib): Replace most `@disableGC` annotations with `@unsafe`

jozanza:
`@unsafe` is bussin'

Issue #793, opened 2021-08-14
Annotated recursive functions do not compile


peblair:
This program does not compile:
```grain
let rec foo: Number -> Number = (a) => { a + 1 }
```
These program do:
```grain
let foo: Number -> Number = (a) => { a + 1 }
```
```grain
let rec foo = (a) => { a + 1 }
```
The exception is as follows:
```bash
 λ grain git:(callee-owned-values) ✗ > grain tst4.gr
/Users/pblair/views/grain/cli/bin/grainc.exe: internal error, uncaught exception:
                                              (Failure "Non-name not allowed on LHS of let rec.")
```
Some further debugging indicates that this is coming from [here](https://github.com/grain-lang/grain/blob/9386f46304ad958c29bb099570f06193911f7131/compiler/src/middle_end/linearize.re#L1345), with the culprit being a `TPatAlias` which somehow appears in the tree.

ospencer:
Fixed in #920

PR #1495, opened 2022-12-01
chore(stdlib): Re-enable type annotations in Array module

phated:
These didn't get cleaned up when #793 was completed. I noticed them in #1494


ospencer (approved):
Nice!

marcusroberts (approved):
LGTM!

PR #1513, opened 2022-12-02
fix(grainfmt): Handle comment in or near if conditionals better

### LINKED ISSUE #1445: Formatter: Issue Code Formatting
This is invalid code so it probably isnt a priority but it is still annoying when it messes up invalid code more.
 ```
if (true)
// Copy the Separator
  true
```
 gets formatted into 
```
if (true // Copy the Separator) true
```

marcusroberts:
Fixes #1445

As we can't tell from the AST if a trailing comment is inside or outside the parens, we always render it inside

### DIFF
--- a/compiler/src/formatting/format.re
+++ b/compiler/src/formatting/format.re
@@ -2927,6 +2927,38 @@ and print_expression =
           comments,
         );
 
+      let last_comment_different_line =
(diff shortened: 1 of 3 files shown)


ospencer (approved):
Works for me!

phated (approved):

PR #1514, opened 2022-12-03
fix(grainfmt): nicer pretty printing of binops

### LINKED ISSUE #1391: Grainfmt: Use Prettier's logic for infix operators
I think Prettier has a really solid DX around infix operators. If the operator is all the same, they don't add any parens, but they do otherwise. It adds a lot of clarity to the code.

You can see their logic here: https://github.com/prettier/prettier/blob/main/src/language-js/print/binaryish.js

### LINKED ISSUE #1473: Formatter: Incorrect removal of parentheses between operators of same precedence
It seems that the removal of redundant parentheses as addressed by #903 does not always work correctly. For example
```
let a = (2 >> 1) << 1
```
gets formatted to
```
let a = 2 >> 1 << 1
```

The compiler interprets the formatted code right-associatively, performing `1 << 1` before `2 >> 1` giving a result of `0` rather than `2` as expected from the placement of parentheses.

marcusroberts:
Fixes #1391 

Add additional parens to binary operators for clarity when precedence is the same but operators are different.
Except for maths operators, and when the right hand operator is a logical operator


compiler/src/formatting/format.re
```diff
@@ -28,6 +28,33 @@ type expression_parent_type =
...
+  switch (fn) {
+  | "<="
+  | ">="
+  | "=="
+  | "!="
+  | "is"
+  | "&&"
+  | "||" => true
+  | _ => false
+  };
```
ospencer, line 56: These two functions need to be extended to the start of the operators in the same way that `op_precedence` does, but otherwise this looks good.

compiler/src/formatting/format.re
```diff
@@ -1901,7 +1929,24 @@ and print_infix_application =
       | _ => false
       };
 
-    let left_needs_parens = left_is_if || left_grouping_required;
+    // put parens around different operators for clarity.
+    // Except math operations as this makes them look ugly
+    // and logic operations
```
ospencer, line 1934:
```suggestion
    // Put parens around different operators for clarity, except
    // math and logic operations where precedence is well-known
```

compiler/test/formatter_outputs/binops.gr
```diff
@@ -11,3 +11,25 @@ if (
...
+
+let p = 1 + 2 - 3
+
+let p2 = 1 - 2 + 3
+
+let y = 1 / 2 + 3
+
+let ya = 1 / (2 + 3)
+
+let y1 = 1 + 2 /3
```
ospencer, line 33: extra space in here?

ospencer: Oh that's because constants are formatted as-is and it's a rational literal. I'll open an issue about that separately.

ospencer: https://github.com/grain-lang/grain/issues/1516

compiler/src/formatting/format.re
marcusroberts: That's done.

ospencer (approved):
Looks awesome, I've been excited for this one! :D

phated (approved):
Very nice!

PR #1515, opened 2022-12-03
fix(compiler): Fix precedence of >>

### LINKED ISSUE #1473: Formatter: Incorrect removal of parentheses between operators of same precedence

ospencer:
Fixes #1473.

The precedence of `>>` was being parsed as having the precedence of `>`. `>>` is lexed as two `>` characters so that they don't interfere with type arguments.


marcusroberts (approved):
Good fix, thanks!

phated (approved):

PR #1526, opened 2022-12-04
fix(grainfmt): Ensure comments on function arguments are maintained

### LINKED ISSUE #983: Grainfmt: Comment in between arguments on own lines gets moved to weird place
Given the code:

```grain
Result.expect(
    "Unexpected error when writing to stdout",
    // Is stdout correct here?
    File.fdWrite(File.stdout, "Status: 500\n\nInternal Server Error")
)
```

Grainfmt currently generates:

```grain
Result.expect(
    "Unexpected error when writing to stdout",
    File.fdWrite(File.stdout, "Status: 500\n\nInternal Server Error"),
) // Is stdout correct here?
```

But I expect the comment to remain in place because it is on a different line than the parameter.

marcusroberts:
Fixes #983 

Uses new iterator to ensure comments in function argument lists are maintained

### DIFF
--- a/compiler/src/formatting/format.re
+++ b/compiler/src/formatting/format.re
@@ -2210,6 +2210,37 @@ and print_arg = (~original_source, ~comments, arg: Parsetree.expression) => {
   };
 }
 
+and print_args_with_comments =
+    (
+      ~comments: list(Parsetree.comment),
+      ~original_source,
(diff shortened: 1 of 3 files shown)


marcusroberts:
Ugh, broken!   Will close and try again!

PR #1527, opened 2022-12-04
fix(grainfmt): Maintain comments in function arguments

### LINKED ISSUE #983: Grainfmt: Comment in between arguments on own lines gets moved to weird place

marcusroberts:
Fixes #983 

Use the new iterator to format function arguments in function application

### DIFF
--- a/compiler/src/formatting/format.re
+++ b/compiler/src/formatting/format.re
@@ -12,6 +12,7 @@ type iterator_item_type =
   | IteratedRecord
   | IteratedTypeConstructor
   | IteratedPatterns
+  | IteratedArgs
   | IteratedMatchItem
   | IteratedDataDeclarations
   | IteratedRecordLabels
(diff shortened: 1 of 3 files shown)


compiler/src/formatting/format.re
```diff
@@ -23,7 +24,7 @@ type iterator_item_type =
   | IteratedRecordData
   | IteratedValueBindings;
 type expression_parent_type =
-  | ConditionalExpression
+  | ExpressionInParens
```
ospencer, line 27: How about `GroupedExpression`?

ospencer (commented):
Looks good, just one comment!

compiler/src/formatting/format.re
marcusroberts: I like it.  Done.

ospencer (approved):
Thanks @marcusroberts!

phated (approved):
LGTM! Thanks

PR #1531, opened 2022-12-05
feat(compiler)!: Add Uint32 and Uint64 types

### LINKED ISSUE #1222: Uint32 and Uint64 types
The addition of these types would allow better type safety and code clarity. The Int32 and Int64 types would become the signed variants. This signedness would also affect printing.

Proposed literal suffixes: `ul` and `uL`

Last thing to consider would be keeping the unsigned instructions in the Int32 and Int64 libraries / adding signed instructions to the new libraries. I personally think we should remove them—I think a lot of code clarity/type safety would be lost by someone, for example, using the Int32 type but only using unsigned instructions because they were there instead of being required to use the `Uint32` type.

alex-snezhko:
Implemented `Uint32` & `Uint64` types, with the intent of removing the unsigned operations from `Int32`/`Int64`. These types are distinct from `Number` and are only heap-allocable.

Closes #1222


compiler/src/codegen/compcore.re
```diff
@@ -2119,6 +2119,102 @@ let allocate_record = (wasm_mod, env, ttag, elts) => {
   );
 };
 
+// TODO mayb
+let allocate_uint_uninitialized =
+    (~additional_words=?, wasm_mod, env, is_32_bit) => {
```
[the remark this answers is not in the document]
alex-snezhko: On second inspection it doesn't actually look like this function will ever be called (no grain code ever calls `allocateUint32` or `allocateUint64`). Should I just remove this function or leave it in in case it's needed sometime in the future?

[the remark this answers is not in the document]
ospencer: I would leave it.

compiler/test/stdlib/uint64.test.gr
```diff
@@ -0,0 +1,70 @@
+import * from "uint64"
+
+assert fromNumber(5) == 5uL
+assert fromNumber(0) == 0uL
+// 0xffffffffffffffff (Number) gets converted to -1, used decimal number instead
```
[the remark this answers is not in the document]
ospencer: Going to decide that it's definitely a bug! I'll get that fixed up.

I think the overflow is fine at first look; I'll have to think about it some more though.

compiler/src/codegen/value_tags.re
```diff
@@ -21,7 +23,9 @@ let tag_val_of_heap_tag_type =
   | BoxedNumberType => 5
   | LambdaType => 6
   | TupleType => 7
-  | BytesType => 8;
+  | BytesType => 8
+  | Uint32Type => 9
+  | Uint64Type => 10;
```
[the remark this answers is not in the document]
alex-snezhko: I don't have an issue with leaving them like this personally, although I could change them if you have a preference

compiler/src/parsing/parser.mly
```diff
@@ -199,6 +199,8 @@ const:
   | DASH? NUMBER_FLOAT { Const.number (PConstNumberFloat (if Option.is_some $1 then "-" ^ $2 else $2)), $sloc }
   | DASH? INT32 { Const.int32 (if Option.is_some $1 then "-" ^ $2 else $2), $sloc }
   | DASH? INT64 { Const.int64 (if Option.is_some $1 then "-" ^ $2 else $2), $sloc }
+  | UINT32 { Const.uint32 $1, $sloc }
+  | UINT64 { Const.uint64 $1, $sloc }
```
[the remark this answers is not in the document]
alex-snezhko: It shouldn't be too hard to give a suggestion on the hex representation, I can try doing that

stdlib/runtime/compare.gr
```diff
@@ -128,6 +128,18 @@ let rec heapCompareHelp = (heapTag, xptr, yptr) => {
...
+      let xval = WasmI32.load(xptr, 4n)
+      let yval = WasmI32.load(yptr, 4n)
+      return if (WasmI32.ltU(xval, yval)) -1 else if (WasmI32.gtU(xval, yval)) 1
+      else 0
+    },
+    t when t == Tags._GRAIN_UINT64_HEAP_TAG => {
+      let xval = WasmI64.load(xptr, 8n)
+      let yval = WasmI64.load(yptr, 8n)
+      return if (WasmI64.ltU(xval, yval)) -1 else if (WasmI64.gtU(xval, yval)) 1
+      else 0
```
alex-snezhko, line 141: Should probably fix this formatting

alex-snezhko:
Since `BigInt` capitalizes the `Int` should the modules be renamed to `UInt32/UInt64`?

ospencer:
I would say no, but largely as a matter of personal preference. I've seen languages do both, but I believe `Uint` is more common.

ospencer:
(I do see the argument though. I don't care much though—will leave it up to consistency vs. industry standard king @phated)

stdlib/runtime/compare.gr
ospencer: Let's make an issue.

cc @marcusroberts

compiler/src/codegen/compcore.re
```diff
@@ -2047,6 +2047,94 @@ let allocate_record = (wasm_mod, env, ttag, elts) => {
...
+    | Uint32(uint32) => (
+        Uint32Type,
+        [store(~offset=4, ~ty=Type.int32, wasm_mod, get_swap(), uint32)],
+        2,
+        "allocate_uint32",
+      )
+    | Uint64(uint64) => (
+        Uint64Type,
+        [store(~offset=8, ~ty=Type.int64, wasm_mod, get_swap(), uint64)],
+        // Really only 3 words needed, allocate 4 for the sake of alignment
```
ospencer, line 2099: Important to call out that we _need_ 4 because of alignment:
```suggestion
        // Allocate 4 words to store with 8-byte alignment
```

compiler/src/typed/checkertypes.re
```diff
@@ -166,7 +168,65 @@ let constant:
...
+            ~loc,
+            "Uint32 literal %sul exceeds the range of representable 32-bit unsigned integers.",
+            n,
+          ),
+        )
+      | (true, Some(num)) =>
+        Error(
+          Location.errorf(
+            ~loc,
+            "Uint32 literal -%sul contains a sign but should be unsigned; consider using 0x%sul instead.",
```
ospencer, line 191: Nice! Glad you were able to get this working 💪

stdlib/int32.gr
```diff
@@ -152,7 +152,9 @@ provide let div = (x: Int32, y: Int32) => {
  * @param x: The first operand
  * @param y: The second operand
  * @returns The quotient of its operands
- *
+ * 
+ * @deprecated This will be removed in the v0.7.0 release of Grain. Use Uint32 instead for unsigned operations
```
ospencer, line 156: I don't think you need to add the deprecated notices anymore—we'll just remove them as a part of 0.6.

ospencer (commented):
My other thought would be we probably want stdlib functions to convert between int32/uint32 and int64/uint64 directly, but that doesn't have to be a part of this change.

ospencer:
This is looking really good! Mostly a few small things left + a rebase.

compiler/src/codegen/compcore.re
alex-snezhko: True, I think I just worded this poorly: I meant to say that since the data needs 2 words and the tag needs 1 it may be confusing why 4 words are allocated if someone didn't know about the alignment. I think your wording makes sense though

alex-snezhko:
> My other thought would be we probably want stdlib functions to convert between int32/uint32 and int64/uint64 directly, but that doesn't have to be a part of this change.

Hmm, is this a super common thing to do? I was thinking that if someone wanted that they could just do `Int32.fromNumber(Uint32.toNumber(x))`

ospencer:
It'll end up being common because people will need to convert just to do a couple of unsigned or signed operations. What you posted is what they'd have to do, but that's not necessarily a safe conversion.

For example, unit32 number `0xffffffff` converted to a Number would be

ospencer:
wild hotkey slip up, why on earth is that a hotkey

alex-snezhko:
> It'll end up being common because people will need to convert just to do a couple of unsigned or signed operations. What you posted is what they'd have to do, but that's not necessarily a safe conversion.
> 
> For example, unit32 number `0xffffffff` converted to a Number would be 4294967295, but converting that to an int32 would either be an overflow error or truncate, but in reality we just want to keep the same bits.

I see; I can add that into this PR.

compiler/test/stdlib/uint64.test.gr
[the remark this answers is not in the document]
ospencer: Could you remove this now?

stdlib/runtime/compare.gr
ospencer: Did we ever get an issue for this?

ospencer:
Why the change to Random?

stdlib/runtime/compare.gr
alex-snezhko: Yep I made #1654

alex-snezhko:
> Why the change to Random?

Since you suggested to remove `divU`/other unsigned operations for `0.6.0` Random had to be changed as well since it relied on the use of these operators. It looked like `Random` was essentially treating the ints as unsigned anyways so I figured it'd be an appropriate change.

ospencer:
Gotcha. Makes sense!

stdlib/uint32.gr
```diff
@@ -0,0 +1,494 @@
...
+ *
+ * @param number: The value to convert
+ * @returns The Number represented as a Uint32
+ * 
+ * @since v0.6.0
+ */
+@unsafe
+provide let fromNumber = (x: Number) => {
+  let x = WasmI32.fromGrain(x)
+  Memory.incRef(x)
```
ospencer, line 37: This incRef shouldn't be necessary—is there something special about `coerceNumberToUnsignedWasmI32`?

alex-snezhko: It is done for Int32s as well https://github.com/grain-lang/grain/blob/main/stdlib/runtime/numbers.gr#L2268 so I figured it'd just be safe to also include it even though I wasn't quite sure why it existed for signed ints

ospencer: Ah okay, that's a little different. In that case, since we're passing a Grain value as a WasmI32, the compiler doesn't know that it needs to incRef the value, so we do it manually. So it's really _just_ because of the `fromGrain`. You can skip the `fromGrain` here and pass `x` directly and the compiler will inject the incRef call for you.

alex-snezhko: Ah ok I see

stdlib/random.gr
```diff
@@ -113,8 +115,9 @@ let checkInitialized = (random: Random) => {
  * @returns The randomly generated number
  *
  * @since v0.5.0
+ * @history v0.6.0: Changed from `nextInt32` to `nextUint32`
```
phated, line 118:
```suggestion
 * @since v0.6.0
 * @history v0.5.0: Originally named `nextInt32`
```
The function is "new" so we need to reverse the docs

stdlib/random.gr
```diff
@@ -128,18 +131,20 @@ provide let nextInt32 = (random: Random) => {
  * @returns The randomly generated number
  *
  * @since v0.5.0
+ * @history v0.6.0: Changed from `nextInt64` to `nextUint64`
```
phated, line 134:
```suggestion
 * @since v0.6.0
 * @history v0.5.0: Originally named `nextInt64`
```

stdlib/random.gr
```diff
@@ -152,22 +157,18 @@ provide let nextInt64 = (random: Random) => {
  * @returns The randomly generated number
  *
  * @since v0.5.0
+ * @history v0.6.0: Changed from `nextInt32InRange` to `nextUint32InRange`
```
phated, line 160:
```suggestion
 * @since v0.6.0
 * @history v0.5.0: Originally named `nextInt32InRange`
```

stdlib/random.gr
```diff
@@ -182,20 +183,16 @@ provide let nextInt32InRange = (random: Random, low: Int32, high: Int32) => {
  * @returns The randomly generated number
  *
  * @since v0.5.0
+ * @history v0.6.0: Changed from `nextInt64InRange` to `nextUint64InRange`
```
phated, line 186:
```suggestion
 * @since v0.6.0
 * @history v0.5.0: Originally named `nextInt64InRange`
```

stdlib/runtime/bigint.gr
```diff
@@ -504,6 +504,32 @@ provide let toInt64 = (num: WasmI32) => {
...
+    // if (limb <= _SMAX32_I64) {
+    //   if (flagIsSet(num, _IS_NEGATIVE)) {
+    //     WasmI64.mul(-1N, limb)
+    //   } else {
+    //     limb
+    //   }
+    // } else {
+    //     // throw Exception.NumberNotIntlike
+    //   throw Exception.Overflow
+    // }
```
phated, line 527: Delete? Make work? Add an issue with a TODO?

stdlib/runtime/dataStructures.gr
```diff
@@ -55,6 +55,22 @@ provide primitive allocateInt32: () -> WasmI32 = "@allocate.int32"
...
+ * Allocates a new Uint32.
+ *
+ * @returns The pointer to the empty Uint32
+ */
+@unsafe
+provide primitive allocateUint32: () -> WasmI32 = "@allocate.uint32"
+
+/**
+ * Allocates a new Uint32 with a prepopulated value
+ * @param value: The value to store
```
phated, line 68:
```suggestion
 * Allocates a new Uint32 with a prepopulated value.
 *
 * @param value: The value to store
```

stdlib/runtime/dataStructures.gr
```diff
@@ -71,6 +87,22 @@ provide primitive allocateInt64: () -> WasmI32 = "@allocate.int64"
...
+ * Allocates a new Uint64.
+ *
+ * @returns The pointer to the empty Uint64
+ */
+@unsafe
+provide primitive allocateUint64: () -> WasmI32 = "@allocate.uint64"
+
+/**
+ * Allocates a new Uint64 with a prepopulated value
+ * @param value: The value to store
```
phated, line 100:
```suggestion
 * Allocates a new Uint64 with a prepopulated value.
 *
 * @param value: The value to store
```

stdlib/sys/random.gr
```diff
@@ -21,12 +21,12 @@ from DataStructures use { tagSimpleNumber, newInt32, newInt64 }
 /**
  * Produce a random 32-bit integer. This function can be slow, so it's best to seed a generator if lots of random data is needed.
  *
- * @returns `Ok(num)` of a random Int32 if successful or `Err(exception)` otherwise
+ * @returns `Ok(num)` of a random Uint32 if successful or `Err(exception)` otherwise
  *
  * @since v0.5.0
```
phated, line 26: Needs an update here and a history entry for the original name.

stdlib/sys/random.gr
```diff
@@ -36,18 +36,18 @@ provide let randomInt32 = () => {
...
+  return Ok(WasmI32.toGrain(newUint32(rand)): Uint32)
 }
 
 /**
  * Produce a random 64-bit integer. This function can be slow, so it's best to seed a generator if lots of random data is needed.
  *
- * @returns `Ok(num)` of a random Int64 if successful or `Err(exception)` otherwise
+ * @returns `Ok(num)` of a random Uint64 if successful or `Err(exception)` otherwise
  *
  * @since v0.5.0
```
phated, line 47: Needs an update here and a history entry for the original name.

stdlib/uint32.gr
```diff
@@ -0,0 +1,490 @@
...
+provide let toNumber = (x: Uint32) => {
+  WasmI32.toGrain(
+    reducedUnsignedInteger(
+      WasmI64.extendI32U(WasmI32.load(WasmI32.fromGrain(x), _VALUE_OFFSET))
+    )
+  ): Number
+}
+
+/**
+ * Converts a Int32 to a Uint32.
```
phated, line 58:
```suggestion
 * Converts an Int32 to a Uint32.
```

stdlib/uint64.gr
```diff
@@ -0,0 +1,489 @@
...
+ */
+@unsafe
+provide let toNumber = (x: Uint64) => {
+  WasmI32.toGrain(
+    reducedUnsignedInteger(WasmI64.load(WasmI32.fromGrain(x), _VALUE_OFFSET))
+  ): Number
+}
+
+/**
+ * Converts a Int64 to a Uint64.
```
phated, line 57:
```suggestion
 * Converts an Int64 to a Uint64.
```

phated (changes requested):
Looking really good! I had a few comments/suggestions

compiler/test/stdlib/uint32.test.gr
```diff
@@ -0,0 +1,88 @@
+module Uint32Test
+
+let (===) = (==)
```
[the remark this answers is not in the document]
alex-snezhko: I thought it would just make it more convenient; in either case the `(==)` function in Uint32 is tested in this file which would require something like `Uint32.(==)(a, b)` which also feels awkward (unless there is a way to use it as an infix operator here that I don't know about)

stdlib/runtime/bigint.gr
alex-snezhko: Forgot to delete, mb

stdlib/runtime/dataStructures.gr
alex-snezhko: There are also a few others in this file I can fix

compiler/test/stdlib/uint32.test.gr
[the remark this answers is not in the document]
phated: We could separate `Pervasive.(==)` and `Uint32.(==)` tests using blocks and use `from Uint32 use *` inside the blocks, right?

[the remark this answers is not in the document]
alex-snezhko: True; I think that would still require `==` to be aliased for the block with `use *` though so maybe I could just import the functions used individually and alias `Uint32.(==)` as `eq` or something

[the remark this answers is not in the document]
alex-snezhko: Since other comparison operators were also involved I just ended up grouping them at the end and adding a `from` statement specifically for comparison operators after the other tests

phated (approved):
Love it! Thanks a ton @alex-snezhko 🎉

PR #1534, opened 2022-12-06
fix(compiler): Prevent stack overflows when compiling long blocks

### LINKED ISSUE #1530: JS compiler stack overflows with deep dependency chains
We added a very long code block when implementing `parseFloat`, which has revealed a problem with the compiler's stack size when compiled to JS. When compiling a long dependency chain, several module compilations can sit on the stack while compiling dependencies. We'll need to adjust how this compilation happens so only one module is actively compiled at once.

This can be observed by compiling `regex.gr` with no build cache.

ospencer:
Fixes #1530

### DIFF
--- a/compiler/src/compile.re
+++ b/compiler/src/compile.re
@@ -14,6 +14,7 @@ type compilation_state_desc =
   | Initial(input_source)
   | Parsed(Parsetree.parsed_program)
   | WellFormed(Parsetree.parsed_program)
+  | DependenciesCompiled(Parsetree.parsed_program)
   | TypeChecked(Typedtree.typed_program)
   | TypedWellFormed(Typedtree.typed_program)
   | Linearized(Anftree.anf_program)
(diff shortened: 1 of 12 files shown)


compiler/src/compile.re
```diff
@@ -153,7 +153,15 @@ let next_state = (~is_root_file=false, {cstate_desc, cstate_filename} as cs) =>
       };
       Well_formedness.check_well_formedness(p);
       WellFormed(p);
-    | WellFormed(p) => TypeChecked(Typemod.type_implementation(p))
+    | WellFormed(p) =>
+      if (is_root_file) {
```
peblair, line 157: I would vote to make this a separate phase (which no-ops for non-root files).

peblair (approved):
One comment, but this otherwise looks great. Thanks for the fix!

jozanza (approved):

Issue #1476, opened 2022-11-11
Stdlib: Move `Math.pow` into runtime for use in pervasives


spotandjake:
Move `Math.pow` into the runtime for use in pervasives maybas `**`

PR #1538, opened 2022-12-08
feat(stdlib)!: Move Power to runtime and add Operator

spotandjake:
This pr moves power to the runtime and adds pervasives as an operator. I didnt see anywhere to test the operator or functions in the runtime but power is well tested in the number test suite.

Closes: #1476


stdlib/pervasives.md
```diff
@@ -532,6 +532,32 @@ Returns:
 |----|-----------|
 |`Number`|The modulus of its operands|
 
+### Pervasives.**(**)**
```
phated, line 535: @spotandjake I think you found a bug with graindoc. Since `*` is bolding, this isn't shown correctly.

spotandjake: f that is a good point, I guess ideally characters should be escaped.

phated:
Blocked by https://github.com/grain-lang/grain/issues/1550

stdlib/pervasives.md
phated: This was fixed by #1685 so this PR is unblocked now.

PR #1540, opened 2022-12-09
chore: Optimize local docker storage

mohitsaxenaknoldus:


Dockerfile
```diff
@@ -20,13 +20,11 @@ RUN npm ci --ignore-scripts
...
 COPY --from=esy /app/_release /grain/node_modules/esy/platform-linux-arm64
 # Manually run esy's postinstall script
-RUN cd node_modules/esy && npm run postinstall
-
-# Necessary because we disabled scripts during the original install
-RUN npm run prepare
-
-# Build the compiler and CLI
-RUN npm run compiler build
+RUN cd node_modules/esy && npm run postinstall \
```
phated, line 23: Why did you combine these? I thought docker was only cacheable at each command (`RUN`, `COPY`, etc). We want the esy postinstall to be cached separately from the prepare, since esy almost never changes.

phated (changes requested):

Dockerfile
mohitsaxenaknoldus: Check now,

Dockerfile
```diff
@@ -21,12 +21,10 @@ RUN npm ci --ignore-scripts
...
 RUN cd node_modules/esy && npm run postinstall
-
 # Necessary because we disabled scripts during the original install
-RUN npm run prepare
-
-# Build the compiler and CLI
-RUN npm run compiler build
+RUN npm run prepare \
+    # Build the compiler and CLI
+    && npm run compiler build
```
ospencer, line 27: We don't want to combine these because we want the results of the prepare script (which never changes) to be cached separately from the build (which changes on every commit).

Dockerfile-slim
```diff
@@ -21,16 +21,13 @@ RUN npm ci --ignore-scripts
...
-
-# Remove build files
-RUN rm -rf compiler/_esy
+RUN cd node_modules/esy && npm run postinstall \
+    # Necessary because we disabled scripts during the original install
+    && npm run prepare \
+    # Build the compiler and CLI
+    && npm run compiler build \
+    # Remove build files
+    && rm -rf compiler/_esy
```
ospencer, line 30: Similar idea here, esy's postinstall script, the prepare script, and the build script should all be their own RUN statements, but combining the build and the removal of the build files into a single step is an improvement :)

ospencer (changes requested):

Dockerfile
```diff
@@ -21,10 +21,8 @@ RUN npm ci --ignore-scripts
 COPY --from=esy /app/_release /grain/node_modules/esy/platform-linux-arm64
 # Manually run esy's postinstall script
 RUN cd node_modules/esy && npm run postinstall
-
 # Necessary because we disabled scripts during the original install
 RUN npm run prepare
-
```
ospencer, line 27: Can you remove changes to this file?

Dockerfile-slim
```diff
@@ -22,15 +22,12 @@ RUN npm ci --ignore-scripts
 COPY --from=esy /app/_release /grain/node_modules/esy/platform-linux-arm64
 # Manually run esy's postinstall script
 RUN cd node_modules/esy && npm run postinstall
-
```
ospencer, line 25: Can you add this space back?

Dockerfile-slim
```diff
@@ -22,15 +22,12 @@ RUN npm ci --ignore-scripts
...
 # Necessary because we disabled scripts during the original install
-RUN npm run prepare
```
ospencer, line 27: Can you leave this as its own step?

Dockerfile-slim
```diff
@@ -25,12 +25,10 @@ RUN cd node_modules/esy && npm run postinstall
 
 # Necessary because we disabled scripts during the original install
 RUN npm run prepare
-
 # Build the compiler and CLI
```
ospencer, line 28:
```suggestion

# Build the compiler and CLI
```

ospencer (approved):

phated (approved):

PR #1541, opened 2022-12-09
feat(docs)!: Update approvals process using CODEOWNERS

### LINKED ISSUE #976: Process: Add codeowners to the project
As the core team has grown, we've started to get a little bogged down by needing so many reviewers (in some cases). We discussed this on a team call and decided that we'll instead use codeowners as the required reviewers on breaking changes, etc.

GitHub supports codeowners using a file—documented at https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

We need to decide who are the codeowners for specific parts of the project. Please use this issue to discuss before we implement it!

phated:
Closes #976

This introduces a breaking change to our approval process (and thus requires approval from the entire core team). We want to switch to using a CODEOWNERS file and only require one approval from a code owner for non-breaking changes. We'll still require **all** code owner approvals for breaking changes.

Please comment inline on the CODEOWNERS file if you think owners of specific things should be changed—@ospencer and I did the best we could.

From my understanding of CODEOWNERS, this might be clumsy for us to use. GitHub doesn't create a full set of all owners for all touched files, but instead assigns the last match in the file, so order matters. If this doesn't end up working for us, we can look into something else, but this is integrated into the platform. 🤷

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,78 @@
+# In the format of:
+# FILE @PRIMARY_OWNER ...@SECONDARY_OWNERS
+
+# Order of precendence is top-to-bottom. The last line to match a file will assign the codeowners
+
+# Deliberately no catch-all so we know when we need to add something to this file
+
+/.gitpod.Dockerfile                         @marcusroberts @ospencer
+/.gitpod.yml                                @marcusroberts @ospencer
+/README.md                                  @ospencer @phated @peblair
+
+/.vscode/                                   @phated @ospencer
+
+/docs/                                      @ospencer @phated @peblair @marcusroberts @jozanza
+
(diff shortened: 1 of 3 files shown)


ospencer (approved):
I helped assign folks, but I approve of the new process!

marcusroberts (approved):
This is a great idea and the assignments look good to me!

peblair (approved):

jozanza (approved):

phated:
We've received all the approvals from the team 🎉  Since this is only a breaking *process* change, we aren't going to merge it into main as a breaking change—it doesn't affect the grain releases at all.

That way we can start testing out/benefitting from this process change but still make it possible to release patch versions of 0.5.x if something breaks during Advent of Code.

PR #1542, opened 2022-12-10
feat(compiler): Lower stack usage

ospencer:
This PR further lowers the compiler's stack usage via some refactoring:
- free variable analysis now happens in its own analysis pass
- purity analysis has been refactored to use the anf_iterator (which is great cleanup as that analysis predated the iterator!)
- anf_mapper now uses a worklist to map the tree instead of the call stack


phated (approved):
This is excellent! I also built the playground off of this PR and it works in my test environment 🎉

jozanza (approved):

PR #1543, opened 2022-12-10
fix(stdlib): Fix NaN comparisons

### LINKED ISSUE #1459: Bug: `<=` and `>=` do not return the correct result when used with `nan`
image
As can be seen in the imagine above `<=` and `>=` are returning true when comparing agaist `nan` when they should be returning `false`, `<` and `>` are unaffected.

ospencer:
Closes #1459

### DIFF
--- a/stdlib/number.gr
+++ b/stdlib/number.gr
@@ -16,6 +16,7 @@ import {
   isInteger,
   isRational,
   isBoxedNumber,
+  isNaN,
   scalbn,
 } from "runtime/numbers"
 import Atoi from "runtime/atoi/parse"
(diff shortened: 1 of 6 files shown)


stdlib/number.gr
```diff
@@ -814,8 +815,7 @@ export let isFinite = (x: Number) => {
 }
 
 /**
- * Checks if a number contains the NaN value (Not A Number).
- * Only boxed floating point numbers can contain NaN.
```
ospencer, line 818: Noticed we were leaking implementation details in the docs, so I removed it.

jozanza (approved):

phated (approved):
This is 🔥. I think removing the `totalOrdering` flag made the code much more clear (as well as more accurate)!

PR #1345, opened 2022-06-20
chore: Upgrade to OCaml 4.14

ospencer:

PR #1547, opened 2022-12-13
chore(compiler): Update OCaml to 4.14 & update dependencies

phated:
This updates the compiler to use OCaml 4.14 - this required forking https://github.com/esy-ocaml/ocaml and updating flexdll to 0.42 which supports the `-l:` linker flag—submitted upstream as https://github.com/esy-ocaml/ocaml/pull/63. We needed this because OCaml 4.12.1+ made changes related to MinGW and [no longer supports passing the `-static` flag](https://github.com/ocaml/ocaml/issues/11531).

I also updated dependencies while I was making changes to the OCaml stuff.

This also fixes the issue with Grain failing to build on newer GCC because that was fixed in the upstream OCaml (we didn't have an issue).

I didn't port over @ospencer's cleanup from #1345 but we can rebase that one and just use it for the cleanup stuff.


ospencer (approved):
Truly excellent work; let's get this over the line!!

peblair (approved):

PR #1548, opened 2022-12-14
feat(compiler): Cache parsetrees during compilation process

### LINKED ISSUE #599: Compiler: Cache parsetrees from dependency resolution
See also https://github.com/grain-lang/grain/pull/590

We currently read and parse a file to determine its dependencies, and have to read and parse it again later when compiling. We should be able to reuse them.

ospencer:
Closes #599 

Provides about a 10% compilation speedup in JS.

### DIFF
--- a/compiler/src/compile.re
+++ b/compiler/src/compile.re
@@ -283,6 +283,7 @@ let compile_wasi_polyfill = () => {
 };
 
 let reset_compiler_state = () => {
+  Driver.reset();
   Ident.setup();
   Ctype.reset_levels();
   Env.clear_imports();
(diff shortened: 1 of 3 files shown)


compiler/src/compile.re
```diff
@@ -294,6 +294,7 @@ let reset_compiler_state = () => {
 let compile_string =
     (~is_root_file=false, ~hook=?, ~name=?, ~outfile=?, ~reset=true, str) => {
   if (reset) {
+    Driver.reset();
```
phated, line 297: Shouldn't this be in `reset_compiler_state` instead of on it's own?

phated (approved):
Looks great!

peblair (approved):

PR #1552, opened 2022-12-16
feat(stdlib)!: Change array rotation direction

### LINKED ISSUE #1470: stdlib: Array.rotate rotates opposite to List.rotate

alex-snezhko:
Closes #1470 , making `Array.rotate` rotate to the left to match `ImmutableArray` and `List` behavior


phated (approved):
LGTM 👍 Thanks @alex-snezhko

PR #1554, opened 2022-12-17
fix(compiler)!: Apply correct rules for parsing Unicode whitespace

ospencer:
While working on the language reference, I realized that the category used for whitespace wasn't quite right, disallowing some common whitespace while allowing some uncommon ones.

This PR states explicitly that Grain follows https://unicode.org/reports/tr31/#Pattern_Syntax for Unicode allowed in the syntax of the language.

Whitespace in Grain now properly adheres to _Pattern_White_Space_, with additional Grain semantics described below.

Whitespace includes:

Spaces, namely
- Horizontal tab, `U+0009`
- Vertical tab, `U+000B`
- Space, `U+0020`
- Left-to-right mark, `U+200E`
- Right-to-left mark, `U+200F`

Line separators, namely
- Line feed, `U+000A`
- Form feed, `U+000C`
- Carriage return, `U+000D`
- Next line, `U+0085`
- Line separator, `U+2028`
- Paragraph separator, `U+2029`

Line separators act as end-of-statement characters in Grain. Note that this is distinct from file line endings—Grain supports only LF and CRLF (relevant for compiler error messages and tooling).

### DIFF
--- a/compiler/src/parsing/lexer.re
+++ b/compiler/src/parsing/lexer.re
@@ -79,6 +79,9 @@ let collect_comment = (comment_type, source, loc, lexbuf) => {
   comments := [comment_type(source, loc), ...comments^];
 };
 
+// Grain follows the Unicode properties for programming languages outlined in
+// https://unicode.org/reports/tr31/#Pattern_Syntax
+
 let dec_digit = [%sedlex.regexp? '0' .. '9'];
 let hex_digit = [%sedlex.regexp? '0' .. '9' | 'A' .. 'F' | 'a' .. 'f'];
 let oct_digit = [%sedlex.regexp? '0' .. '7'];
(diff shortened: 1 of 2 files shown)


compiler/src/parsing/lexer.re
```diff
@@ -113,7 +116,9 @@ let dec_float = [%sedlex.regexp?
 
 let unsigned_float = [%sedlex.regexp? dec_float];
 
-let uident = [%sedlex.regexp? (lu, Star(xid_continue))];
+let uident = [%sedlex.regexp?
+  (Intersect(xid_start, lu), Star(xid_continue))
```
ospencer, line 120: This is equivalent to what it was before, but I felt it made sense to make it more explicit since it's not obvious that `Lu` is a subset of `xid_start`.

phated (approved):
LGTM

PR #1556, opened 2022-12-18
fix(compiler): Provide appropriate error during invalid array access

ospencer:
This PR fixes an issue where a potential non-integer or heap-allocated index for an array get/set may not result in an error. It also provides a better error for non-integer indexes.

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -75,6 +75,8 @@ let assertion_error_closure_ident =
   Ident.create_persistent("GRAIN$EXPORT$AssertionError");
 let index_out_of_bounds_ident =
   Ident.create_persistent("GRAIN$EXPORT$IndexOutOfBounds");
+let index_non_integer_ident =
+  Ident.create_persistent("GRAIN$EXPORT$IndexNonInteger");
 let match_failure_ident =
   Ident.create_persistent("GRAIN$EXPORT$MatchFailure");
 
@@ -131,6 +133,15 @@ let required_global_imports = [
     mimp_id: index_out_of_bounds_ident,
     mimp_mod: exception_mod,
     mimp_name: Ident.name(index_out_of_bounds_ident),
+    mimp_type: MGlobalImport(Types.Unmanaged(WasmI32), true),
+    mimp_kind: MImportWasm,
+    mimp_setup: MSetupNone,
+    mimp_used: false,
+  },
+  {
+    mimp_id: index_non_integer_ident,
+    mimp_mod: exception_mod,
+    mimp_name: Ident.name(index_non_integer_ident),
     mimp_type: MGlobalImport(Types.Unmanaged(WasmI32), true),
     mimp_kind: MImportWasm,
     mimp_setup: MSetupNone,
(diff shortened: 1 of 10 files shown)


compiler/src/codegen/runtime_errors.re
```diff
@@ -4,5 +4,6 @@ open Sexplib.Conv;
 [@deriving sexp]
 type grain_error =
   | IndexOutOfBounds
+  | IndexNonInteger
```
phated, line 7: Does the order of these affect anything? Does it make it breaking?

compiler/src/codegen/compcore.re
```diff
@@ -1188,108 +1280,44 @@ let compile_array_op = (wasm_mod, env, arr_imm, op) => {
...
     let idx = compile_imm(wasm_mod, env, idx_imm);
     let val_ = compile_imm(wasm_mod, env, val_imm);
-    let get_idx = () => get_swap(1);
-    let set_arr = () => set_swap(2, get_arr_value());
-    let get_arr = () => get_swap(2);
-    /* Check that the index is in bounds */
     Expression.Block.make(
       wasm_mod,
-      gensym_label("MArrayGet"),
+      gensym_label("MArraySet"),
```
phated, line 1298: Typo previously?

compiler/src/codegen/compcore.re
```diff
@@ -1124,56 +1136,136 @@ let compile_array_op = (wasm_mod, env, arr_imm, op) => {
...
-              Expression.Const.make(wasm_mod, const_int32(-1)),
+              Op.add_int32,
+              get_swap(1),
+              load(~offset=4, wasm_mod, get_swap(2)),
             ),
-            get_idx(),
           ),
-          IndexOutOfBounds,
-          [],
+          Expression.Null.make(),
```
phated, line 1237: Why do both of these end with a `Null.make()`? Is it to typecheck in wasm or something?

phated (commented):
Couple of questions

compiler/src/codegen/runtime_errors.re
ospencer: Nope!

compiler/src/codegen/compcore.re
ospencer: Yes yes!

compiler/src/codegen/compcore.re
ospencer: Binaryen does one-sided ifs by passing null as the false branch. We could probably change the API to accept an option, though. But I don't really mind this API.

phated (approved):

compiler/src/codegen/compcore.re
```diff
@@ -1124,56 +1136,136 @@ let compile_array_op = (wasm_mod, env, arr_imm, op) => {
...
+            Expression.Binary.make(
+              wasm_mod,
+              Op.and_int32,
+              get_swap(1),
+              Expression.Const.make(wasm_mod, const_int32(1)),
+            ),
+          ),
+          Expression.Block.make(
+            wasm_mod,
+            gensym_label("IndexNotSimpleInteger"),
```
peblair, line 1162: I feel like this is a symptom that `XX.toNumber` should reduce its argument to the smallest containing integer, rather than be a no-op...

peblair (approved):
LGTM; I left one piece of food for thought, but that's beyond the scope of this PR.

compiler/src/codegen/compcore.re
ospencer: I think how it works now is correct (since the flexibility is the point of Number), but we should definitely have a utility that does let you round/truncate. It gets kinda hairy for floats, since we support bigints. My basic math tells me that the worst case conversion of a float64 to a bigint would be a bigint with 16 limbs, so maybe not actually so bad.

That said though, I could see us converting whole numbers to ints. `3.0` becoming `3`, that sort of thing.

But yeah, food for thought.

PR #1557, opened 2022-12-18
feat(compiler): Allow lists and arrays on the left side of access expressions

ospencer:
This essentially allows you to write expressions like `[> 1, 2, 3][1]`.

A happy side effect is a bunch of cleanup of error messages, since the nonterminals are no longer inlined with separate parser states.


marcusroberts (approved):
LGTM!

phated (approved):
Nice! And this isn't breaking so we could release as a 0.5.10 with the array access non-int fix?

PR #1560, opened 2022-12-22
Update docker-publish.yml

### LINKED ISSUE #1529: CI: Implement concurrency for docker workflow
We're having issues with the docker publish workflow when we merge a few PRs at nearly the same time. We can avoid this issue by implementing `concurrency` on the workflow itself. Ref https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

mohitsaxenaknoldus:
Fixes #1529

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -14,6 +14,7 @@ permissions:
 env:
   AWS_EC2_INSTANCE_ID: i-0960f9b8de8285aed
   AWS_REGION: us-east-1
+concurrency: ${{ github.ref }}
 jobs:
   push_base_image:
     name: Push base Docker image to multiple registries


.github/workflows/docker-publish.yml
```diff
@@ -14,6 +14,7 @@ permissions:
 env:
   AWS_EC2_INSTANCE_ID: i-0960f9b8de8285aed
   AWS_REGION: us-east-1
+concurrency: ${{ github.ref }}
```
phated, line 17: I don't think this is correct. This would still run multiple docker builds upon different references. We want to ensure exactly 1 docker build is happening at a given time so the build machine doesn't run out of resources.

phated (changes requested):

.github/workflows/docker-publish.yml
mohitsaxenaknoldus: Ok, let's make it something static then.

mohitsaxenaknoldus:
@phated What should be the new timeout?

PR #1561, opened 2022-12-24
chore(compiler): Update snapshots

ospencer:


phated (approved):
We should really require all branches are up-to-date before they can be merged. I think that's a branch protection we can enable.

PR #1562, opened 2022-12-27
fix(grainfmt): Fix grouping of expressions on LHS of fn application, record access etc

### LINKED ISSUE #1559: Incorrect formatting of grouped expressions
Namely, binops, function calls, array access, and record field access.

This code
```
let printFoo = () => print("foo")
let printBar = () => print("bar")

(if (true) printFoo else printBar)()
```
formats to
```
if (true) printFoo else printBar()
```
which in this case is syntactically correct, but a very different program.

Here are some silly examples, but they demonstrate the problem well: `(a + b).label`, `(a + b)[2]`

marcusroberts:
Look for the special case of expressions on the left hand side of function application, record/array accessors and group with parens as appropriate.

Fixes #1559


marcusroberts:
@ospencer  does this look like it covers all the cases you can think of?

ospencer:
I would probably reverse the logic and group everything that _isn't_ one of the things to be grouped. Then you can just follow this: https://github.com/grain-lang/grain/blob/main/compiler/src/parsing/parser.mly#L545

So if it is not a:
- function call
- constant
- tuple
- identifier
- array access
- record access
- record
- block
- list
- array

Then it needs parens. That'll help us in the future too if we ever add more constructs; it's easier to tell what needs to be updated.

Maybe you might want to make a helper function for the above, so it's easy to modify if we update the parser rules.

ospencer:
That'll also err on the side of being ugly but correct in the worst case.

marcusroberts:
Thanks, that makes sense!

marcusroberts:
Thanks @ospencer that's done I think.

compiler/src/formatting/format.re
```diff
@@ -3672,6 +3674,52 @@ and print_expression =
 
   expression_doc;
 }
+
+and group_expression = (expr: Parsetree.expression) => {
```
ospencer, line 3678: would probably call this `is_grouped_access_expression`

ospencer (commented):
Overall looks great! Just some thoughts around naming.

compiler/src/formatting/format.re
marcusroberts: Agreed.  Renamed accordingly.

compiler/src/formatting/format.re
```diff
@@ -3672,6 +3674,52 @@ and print_expression =
...
+  | PExpArrayGet(_)
+  | PExpArraySet(_)
+  | PExpRecordGet(_)
+  | PExpRecordSet(_)
+  | PExpRecord(_)
+  | PExpBlock(_)
+  | PExpArray(_) => false
+  | PExpApp(func, _) =>
+    let func_name = get_function_name(func);
+    func_name != list_cons;
```
ospencer, line 3692: Regular applications are allowed on the left side, so the `PExpApp` case should always be `false` like the others.

ospencer (commented):
One small bug (related to the marshal format change)

compiler/src/formatting/format.re
```diff
@@ -25,7 +25,8 @@ type iterator_item_type =
   | IteratedValueBindings;
 type expression_parent_type =
   | GroupedExpression
-  | GenericExpression;
+  | GenericExpression
+  | NeedsLeftSideGrouped;
```
ospencer, line 29: This comment somehow went to the ether but was there when I tried to comment on this again 🤨 

This name is now pretty confusing because `NeedsLeftSideGrouped` is what checks if it needs to do something special for grouping and `GroupedExpression` doesn't. I'd maybe just call it `GroupedAccessExpression` and rename `GroupedExpression` to `GroupedInfixExpression`.

compiler/src/formatting/format.re
marcusroberts: Oh, of course,  and that captures lists too.

compiler/src/formatting/format.re
```diff
@@ -24,8 +24,9 @@ type iterator_item_type =
   | IteratedRecordData
   | IteratedValueBindings;
 type expression_parent_type =
-  | GroupedExpression
-  | GenericExpression;
+  | GroupedInfixExpression
+  | GenericExpression
+  | GroupedAccessExpression;
```
ospencer, line 29: After reading it now, since it's the parent, I think it makes more sense to drop the `Grouped` part so it becomes just `InfixExpression` and `AccessExpression` 👍

ospencer (approved):
Looks great, excellent work @marcusroberts!

phated (approved):
Excellent!

PR #1565, opened 2022-12-30
feat(compiler): Record spread syntax

### LINKED ISSUE #778: Record spread syntax
I forget, do we have record spread syntax? `{ ...config, caseSensitive: caseSensitive }` (caseSensitive could also be punned)

_Originally posted by @phated in https://github.com/grain-lang/grain/pull/680#discussion_r667370729_

alex-snezhko:
Syntax for creating a copy of a record while changing some field values. Currently the "naive" approach of copying fields one at a time is used, but bulk copying was discussed as a possible future improvement.

Closes #778

### DIFF
--- a/compiler/src/middle_end/linearize.re
+++ b/compiler/src/middle_end/linearize.re
@@ -757,35 +757,56 @@ let rec transl_imm =
         ),
       ],
     );
-  | TExpRecord(args) =>
+  | TExpRecord(base, args) =>
+    let base_imm = Option.map(transl_imm, base);
     let tmp = gensym("record");
-    let definitions =
-      Array.to_list @@ Array.map(((desc, def)) => def, args);
-    let definitions =
-      List.map(
-        fun
-        | Kept(_) => assert(false)
-        | Overridden(name, def) => (name, def),
-        definitions,
-      );
     let (new_args, new_setup) =
       List.split(
         List.map(
-          (({txt: name, loc}, expr)) => {
-            let (var, setup) = transl_imm(expr);
-            (
-              (Location.mkloc(Identifier.string_of_ident(name), loc), var),
-              setup,
-            );
-          },
-          definitions,
+          fun
+          | (ld, Kept) => {
+              let (base_var, _) = Option.get(base_imm);
+              let fieldtmp = gensym("field");
+              (
+                (None, Imm.id(~loc, ~env, fieldtmp)),
+                [
+                  BLet(
+                    fieldtmp,
+                    Comp.record_get(
+                      ~loc,
+                      ~env,
+                      ~allocation_type,
+                      Int32.of_int(ld.lbl_pos),
+                      base_var,
+                    ),
+                    Nonglobal,
+                  ),
+                ],
+              );
+            }
+          | (_, Overridden({txt: name, loc}, expr)) => {
+              let (var, setup) = transl_imm(expr);
+              (
+                (
+                  Some(
+                    Location.mkloc(Identifier.string_of_ident(name), loc),
+                  ),
+                  var,
+                ),
+                setup,
+              );
+            },
+          Array.to_list(args),
         ),
       );
     let (typath, _, _) = Typepat.extract_concrete_record(env, typ);
     let ty_id = get_type_id(typath);
     (
       Imm.id(~loc, ~env, tmp),
-      List.concat(new_setup)
+      List.concat(
+        Option.to_list(Option.map(((_, setup)) => setup, base_imm))
+        @ new_setup,
+      )
       @ [
         BLet(
           tmp,
--- a/compiler/src/middle_end/linearize.re
+++ b/compiler/src/middle_end/linearize.re
@@ -763,8 +763,9 @@ let rec transl_imm =
     let (new_args, new_setup) =
       List.split(
         List.map(
-          fun
-          | (ld, Kept) => {
+          arg =>
+            switch (arg) {
+            | (ld, Kept) =>
               let (base_var, _) = Option.get(base_imm);
               let fieldtmp = gensym("field");
               (
@@ -783,8 +784,7 @@ let rec transl_imm =
                   ),
                 ],
               );
-            }
-          | (_, Overridden({txt: name, loc}, expr)) => {
+            | (_, Overridden({txt: name, loc}, expr)) =>
               let (var, setup) = transl_imm(expr);
               (
                 (
--- a/compiler/src/parsing/ast_helper.re
+++ b/compiler/src/parsing/ast_helper.re
@@ -24,6 +24,10 @@ type listitem('a) =
   | ListItem('a)
   | ListSpread('a, Location.t);
 
+type recorditem =
+  | RecordItem(loc(Identifier.t), expression)
+  | RecordSpread(expression, Location.t);
+
 type id = loc(Identifier.t);
 type str = loc(string);
 type loc = Location.t;
@@ -192,8 +196,47 @@ module Exp = {
     mk(~loc?, ~attributes?, PExpConstant(a));
   let tuple = (~loc=?, ~attributes=?, a) =>
     mk(~loc?, ~attributes?, PExpTuple(a));
-  let record = (~loc=?, ~attributes=?, a) =>
-    mk(~loc?, ~attributes?, PExpRecord(a));
+  let record = (~loc=?, ~attributes=?, a, b) =>
+    mk(~loc?, ~attributes?, PExpRecord(a, b));
+  let record_fields = (~loc=?, ~attributes=?, a) =>
+    switch (a) {
+    | [] => assert(false)
+    | [base, ...rest] =>
+      let (spread_base, record_items) =
+        switch (base) {
+        | RecordItem(id, expr) => (None, [(id, expr)])
+        | RecordSpread(expr, _) => (Some(expr), [])
+        };
+      let record_items =
+        List.fold_left(
+          (acc, expr) => {
+            switch (expr) {
+            | RecordItem(id, expr) => [(id, expr), ...acc]
+            | RecordSpread(_, loc) =>
+              switch (spread_base) {
+              | None =>
+                raise(
+                  SyntaxError(
+                    loc,
+                    "A record spread can only appear at the beginning of a record expression.",
+                  ),
+                )
+              | Some(_) =>
+                raise(
+                  SyntaxError(
+                    loc,
+                    "A record expression may only contain one record spread.",
+                  ),
+                )
+              }
+            }
+          },
+          record_items,
+          rest,
+        );
+      let record_items = List.rev(record_items);
+      record(~loc?, ~attributes?, spread_base, record_items);
+    };
   let record_get = (~loc=?, ~attributes=?, a, b) =>
     mk(~loc?, ~attributes?, PExpRecordGet(a, b));
   let record_set = (~loc=?, ~attributes=?, a, b, c) =>
--- a/compiler/src/parsing/ast_helper.re
+++ b/compiler/src/parsing/ast_helper.re
@@ -200,7 +200,7 @@ module Exp = {
     mk(~loc?, ~attributes?, PExpRecord(a, b));
   let record_fields = (~loc=?, ~attributes=?, a) =>
     switch (a) {
-    | [] => assert(false)
+    | [] => failwith("Impossible: empty record field list")
     | [base, ...rest] =>
       let (spread_base, record_items) =
         switch (base) {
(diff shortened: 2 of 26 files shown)


compiler/src/parsing/ast_helper.re
```diff
@@ -192,8 +196,47 @@ module Exp = {
     mk(~loc?, ~attributes?, PExpConstant(a));
   let tuple = (~loc=?, ~attributes=?, a) =>
     mk(~loc?, ~attributes?, PExpTuple(a));
-  let record = (~loc=?, ~attributes=?, a) =>
-    mk(~loc?, ~attributes?, PExpRecord(a));
+  let record = (~loc=?, ~attributes=?, a, b) =>
+    mk(~loc?, ~attributes?, PExpRecord(a, b));
+  let record_unprocessed = (~loc=?, ~attributes=?, a) =>
```
ospencer, line 201: The names of everything in this module are pretty semantic, so I'd probably fold this and `record` into one function just called `record`.

compiler/src/parsing/parser.messages
```diff
@@ -3285,14 +3297,76 @@ program: LBRACE UIDENT COMMA UIDENT COMMA EOL WHILE
 ## may provide an INCOMPLETE view of the future (what was expected next).
 ## In state 3, spurious reduction of production nonempty_list(eol) -> EOL
 ## In state 5, spurious reduction of production eols -> nonempty_list(eol)
-## In state 96, spurious reduction of production comma -> COMMA eols
+## In state 94, spurious reduction of production comma -> COMMA eols
+##
+program: LBRACE ELLIPSIS BIGINT COMMA ELLIPSIS BIGINT ARROW
```
ospencer, line 3302: Technically I think we're looking for a comma followed by more fields or a }.

compiler/src/parsing/parser.messages
```diff
@@ -3285,14 +3297,76 @@ program: LBRACE UIDENT COMMA UIDENT COMMA EOL WHILE
...
+## In state 69, spurious reduction of production annotated_expr -> non_binop_expr
+## In state 203, spurious reduction of production non_stmt_expr -> annotated_expr
+## In state 59, spurious reduction of production expr -> non_stmt_expr
+## In state 208, spurious reduction of production spread_record_field -> ELLIPSIS expr
+## In state 224, spurious reduction of production lseparated_nonempty_list_inner(comma,record_field) -> spread_record_field
 ##
 
 Expected more record fields or an immediate `}` to complete the record expression.
 
+program: LBRACE ELLIPSIS BIGINT ARROW
```
ospencer, line 3329: Same thing here, next token should be a comma or a `}`.

compiler/src/parsing/parser.messages
```diff
@@ -6026,7 +6098,7 @@ Expected `(` followed by an expression for the while loop condition.
 
 program: MATCH LPAREN FLOAT32 RPAREN LBRACE NUMBER_INT EOL WHILE
 ##
-## Ends in an error in state: 777.
+## Ends in an error in state: 591.
```
ospencer, line 6101: Looks like you regenerated all of the error states (which we really need to do) but it'd be a bit better if we did that in  a separate PR. I think I counted three new programs here; could you revert this file and add just those three changes and we'll regenerate in another PR? 🙏

compiler/src/typed/typecore.re
```diff
@@ -738,13 +740,23 @@ and type_expect_ =
...
+        begin_def();
+        let ty = newconstr(p', instance_list(env, decl.type_params));
+        end_def();
+        generalize_structure(ty);
+        (ty, opt_exp_opath);
       };
     };
 
-    let closed = true;
+    let closed = Option.is_none(b);
```
ospencer, line 759: Is this right? We don't have open record types at all.

ospencer (commented):
Looking really stellar! Just a couple of questions from me but this looks like it'll be ready to land pretty soon :)

compiler/src/parsing/ast_helper.re
alex-snezhko: I created a separate `record_unprocessed` function here because in `ast_mapper` a call to `record` is made as well, and I felt it might be a bit wasteful to do all of this preprocessing at `ast_mapper` when `PExpRecord` would have already been preprocessed at that point. If you think this isn't that big of a deal then I could remove it though.

compiler/src/parsing/parser.messages
alex-snezhko: Hmm, I didn't actually write these programs. After making changes to `parser.mly` I ran `npm run parser:check-errors` and these programs were autogenerated by Menhir.

compiler/src/typed/typecore.re
alex-snezhko: Not sure, maybe this isn't necessary. To be honest, most of the details of the typing implementation go over my head and I only changed this because this is how it appears in the OCaml compiler, and so I figured that maybe the new inclusion of record extensions could influence what the "correct" value of this variable should be.

compiler/test/formatter_outputs/records.gr
```diff
@@ -110,3 +110,19 @@ let s = { // comment 1
...
+
+let y = { ...x, bar: 2 }
+
+let y = {
+  ...x,
+  longlonglongnamenamename2: 12345,
+  longlonglongnamenamename3: 12345,
+}
+
+let y = { /* comment 1 */ /* comment 2 */ /* comment 3 */ /* comment 4 */
```
phated, line 124: This seems wrong

phated (changes requested):
One thing I saw from my phone

compiler/test/formatter_outputs/records.gr
ospencer: This is just how we format this currently; fixing it is probably out of scope of this PR. Can make an issue to address though.

phated: I thought @marcusroberts literally just fixed these?

ospencer: If you remove the `...` this is the behavior on main.

compiler/src/parsing/parser.messages
alex-snezhko: Upon re-reading this I think I may have misunderstood what you meant: are you saying that the error message should indicate that more fields or } should be given or are you saying that this is what should be in the program on the line highlighted?

ospencer: Yeah, just that the error message here isn't _quite_ right for the input program.

ospencer: If you didn't change anything in this file then it probably was already a bug and we can just make an issue for it :)

(by the way, are you sure you didn't run `npm run parser:update-errors`? `check-errors` shouldn't have made any changes to this file).

alex-snezhko: Oh yeah I did run `parser:update-errors`, I forgot to mention that. But yes you're right about this message; originally I wrote the parser to treat the expression after ellipsis differently from fields but now that this distinction is made in `ast_helper` this should have changed.

compiler/src/parsing/parser.messages
alex-snezhko: Hmm, should an immediate `}` be allowed here?
```
record_exprs:
  // Don't ever parse {x} as a record
  | non_punned_record_field comma? { [$1] }
  | punned_record_field comma { [$1] }
  | record_field comma lseparated_nonempty_list(comma, record_field) comma? { $1::$3 }
```
Since the `ELLIPSIS BIGINT` is the `record_field` here, the grammar would seem to indicate that only a comma must follow here.

ospencer: That's true!

compiler/test/formatter_outputs/records.gr
marcusroberts: I'll take a look

marcusroberts: I've raised an issue for this https://github.com/grain-lang/grain/issues/1575 

It is on main though, so not a problem with this PR

marcusroberts: Fixed in #1578

alex-snezhko: Thanks for the quick fix! I can rebase this branch off of main when your PR gets merged.

compiler/src/parsing/ast_helper.re
alex-snezhko: Another idea that would make the names more semantic: how about still keeping these functions separate but changing the name of `record_unprocessed` to `record_fields` as this function essentially processes the record fields into a processed `record`

compiler/test/formatter_outputs/records.gr
alex-snezhko: Sweet, thanks!

compiler/src/parsing/ast_helper.re
ospencer: I think that makes sense.

ospencer (approved):
Thanks @alex-snezhko! Awesome work here 😄

compiler/src/middle_end/linearize.re
```diff
@@ -757,35 +757,56 @@ let rec transl_imm =
...
-          (({txt: name, loc}, expr)) => {
-            let (var, setup) = transl_imm(expr);
-            (
-              (Location.mkloc(Identifier.string_of_ident(name), loc), var),
-              setup,
-            );
-          },
-          definitions,
+          fun
+          | (ld, Kept) => {
```
phated, line 767: Generally, I try to avoid the `fun` keyword because it isn't friendly for new folks jumping into the compiler.

compiler/src/middle_end/linearize.re
```diff
@@ -757,35 +757,56 @@ let rec transl_imm =
...
+              let (base_var, _) = Option.get(base_imm);
```
phated, line 768: This can never be `None`?

compiler/src/parsing/ast_helper.re
```diff
@@ -192,8 +196,47 @@ module Exp = {
     mk(~loc?, ~attributes?, PExpConstant(a));
   let tuple = (~loc=?, ~attributes=?, a) =>
     mk(~loc?, ~attributes?, PExpTuple(a));
-  let record = (~loc=?, ~attributes=?, a) =>
-    mk(~loc?, ~attributes?, PExpRecord(a));
+  let record = (~loc=?, ~attributes=?, a, b) =>
+    mk(~loc?, ~attributes?, PExpRecord(a, b));
+  let record_fields = (~loc=?, ~attributes=?, a) =>
+    switch (a) {
+    | [] => assert(false)
```
phated, line 203: do we want this to be an `Impossible:`?

compiler/src/parsing/parser.messages
```diff
@@ -3229,6 +3238,30 @@ program: LBRACE UIDENT COMMA EOL UIDENT COLON BREAK WHILE
...
+##
+## lseparated_nonempty_list_inner(comma,record_field) -> lseparated_nonempty_list_inner(comma,record_field) . comma punned_record_field [ RBRACE EOL COMMA ]
+## lseparated_nonempty_list_inner(comma,record_field) -> lseparated_nonempty_list_inner(comma,record_field) . comma non_punned_record_field [ RBRACE EOL COMMA ]
+## lseparated_nonempty_list_inner(comma,record_field) -> lseparated_nonempty_list_inner(comma,record_field) . comma spread_record_field [ RBRACE EOL COMMA ]
+## record_exprs -> spread_record_field comma lseparated_nonempty_list_inner(comma,record_field) . option(comma) [ RBRACE EOL ]
+##
+## The known suffix of the stack is as follows:
+## spread_record_field comma lseparated_nonempty_list_inner(comma,record_field)
+##
+## WARNING: This example involves spurious reductions.
```
phated, line 3253: @ospencer Does this mean we should worry about something?

ospencer:
Looks like you need a `npm run compiler format`.

ospencer:
Jinx :P

phated (changes requested):
This PR is fantastic. I saw a few things (mostly questions) but overall it was a delight to review!

compiler/src/middle_end/linearize.re
alex-snezhko: Yeah a `Kept` field should only be possible if the record has a base, in which case the field would be copied over

compiler/src/parsing/parser.messages
ospencer: Nope! The lines below explain what this means, but the tl;dr is that there's other possible continuations than what's listed above.

compiler/src/middle_end/linearize.re
alex-snezhko: It appears that `fun` is used in a bunch of other places in the compiler; should an issue be made to remove them all?

phated: > should an issue be made to remove them all?

Probably not, we can clean them up as encountered (I'm also guessing a lot of them are from the ocaml source code).

phated (approved):
:shipit:

PR #1573, opened 2023-01-05
fix(stdlib): build and return stdlib array in File.fdReaddir instead …

av8ta:
…of manually managing array memory

I found all sorts of errors when using File.fdReaddir. Sometimes it would hang without outputting data. Sometimes it would output data and then hang. Sometimes it would return cleanly with malformed data. Sometimes but not often it returned the right result!

Most frequent error:
```
RuntimeError: memory access out of bounds
    at wasm://wasm/0011eea6:wasm-function[55]:0x374f
    at wasm://wasm/0011eea6:wasm-function[60]:0x4068
    at wasm://wasm/0011eea6:wasm-function[72]:0x49b6
    at wasm://wasm/0011eea6:wasm-function[484]:0x2c159
    at wasm://wasm/0011eea6:wasm-function[492]:0x2ea9a
    at wasm://wasm/0011eea6:wasm-function[493]:0x2eea4
    at wasm://wasm/0011eea6:wasm-function[495]:0x2f496
    at wasm://wasm/0011eea6:wasm-function[496]:0x2f4d7
    at wasm://wasm/0011eea6:wasm-function[680]:0x47ade
    at wasm://wasm/0011eea6:wasm-function[682]:0x47b4e
```

Examples of malformed data returned:
```grain
// what's up with file x missing and the inode value?
Ok([> <record value>, {
  inode: <enum value>,
  filetype: RegularFile,
  path: "y"
}])

Ok([> <unknown heap tag type: 0x3f9898 | value: 0x3f97d0>, <enum value>, {
  inode: '',
  filetype: ": ",
  path: "z"
}])
```

While debugging I noticed that the when I printed `dirent` here `let dirent = { inode, path, filetype }` it always looked correct so I just changed the arr returned to stdlib array and appended to that. All the problems seem resolved :)

### DIFF
--- a/stdlib/sys/file.gr
+++ b/stdlib/sys/file.gr
@@ -34,6 +34,7 @@ import {
 } from "runtime/dataStructures"
 
 import List from "list"
+import { append as appendToArray } from "array"
 
 /**
  * @section Types: Type declarations included in the File module.
@@ -1114,7 +1115,7 @@ export let fdReaddir = (fd: FileDescriptor) => {
         None => {
           Memory.free(bufUsed)
 
-          let arr = allocateArray(numEntries)
+          let mut arr: Array<DirectoryEntry> = [>]
 
           for (let mut i = numEntries - 1n; i >= 0n; i -= 1n) {
             let ent = WasmI32.load(bufs, 0n)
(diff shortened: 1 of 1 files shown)


ospencer:
I see the bug in the original code—I'm going to adjust it to remove that bug. Hope you don't mind me pushing to your branch!

av8ta:
> I see the bug in the original code—I'm going to adjust it to remove that bug. Hope you don't mind me pushing to your branch!

No problem!

phated (approved):

ospencer:
@av8ta Could you try this out and verify it still works?

ospencer (approved):
Thanks for the bug report and potential fix! Made it obvious what was wrong here 🙌

av8ta:
> Thanks for the bug report and potential fix! Made it obvious what was wrong here raised_hands

I had a feeling you'd fix it more low-level!

PR #1577, opened 2023-01-05
fix(compiler): Fix error location reporting of unbound modules and labels

ospencer:
This reports the errors inline instead of at the top of the file, like so:

image

image


phated (approved):
lgtm

marcusroberts (approved):
Nice!

PR #1584, opened 2023-01-08
feat(compiler)!: Module system

### LINKED ISSUE #190: Compiler: Importing Records generates warnings

### LINKED ISSUE #270: Compiler: Disambiguate type imports
Currently, there's syntactically no way to determine in this import

```grain
import { Foo } from '...'
```

if `Foo` is a type or a data constructor. The compiler understands this, but users and syntax highlighters are unable to tell the difference.

### 1. Separate type import keyword

```grain
import data { Foo } from '...'
```
or
```grain
import-data { Foo } from '...'
```

### 2. A keyword to differentiate a type value
```grain
import { data Foo } from '...'
```
or
```grain
import { type Foo } from '...'
```

### 3. Keywords for all values
```grain
import (
  data Foo
  val Bar
) from '...'
```

### LINKED ISSUE #271: Compiler: Rethink type exports
Currently, if a type is not explicitly exported but a value that uses that type _is_ exported, an abstract version of that type is exported, and can subsequently be imported by other modules. This works fine, but it's a little awkward since this abstract export happens quietly and may come as a surprise to some.

In Discord, we discussed potentially adding a way to explicitly export a type as abstract:

```grain
export abstract Foo
```

and then failing to compile if the module would leak a type that isn't exported (with a helpful error message).

cc @phated

### LINKED ISSUE #324: Compiler: Investigate if the Includemod.compunit check is necessary after typechecking
See https://github.com/grain-lang/grain/pull/281#discussion_r484176183.

### LINKED ISSUE #681: Resolve type checker warnings in `regex.test.gr`

### LINKED ISSUE #997: Type printing when importing * (star)
While writing some tests for type aliases, I noticed that our LSP was showing a weird type.

Screen Shot 2021-10-21 at 12 30 23 PM

Whenever a type is imported using `import *`, it doesn't have a name to bind to so the compiler creates a name for it and adds a `%` to the beginning to it (so it can't be written by the user).

This isn't a problem in itself, but with the LSP showing codelens of the types, it is very confusing. We should change our type printer to change the output of this type for the LSP and errors. I was thinking something like `<imported>.Foo` but you lose what module it came from, so maybe `<imported ModuleName>.Foo`. Thoughts?

### LINKED ISSUE #1340: Lang: Module import syntax
This was discussed in another issue, but I can't find it quickly so I wanted to open a top-level issue. If you find the discussion, please link.

To improve LSP support (and general code readability), we plan to change the module importing syntax from "file last" to "file first". While not the final syntax, this could look something like:
```grain
from "foo" import { bar }
```

This will likely be a part of the larger module system rewrite that includes module definitions inside files, new foreign syntax, and type exports.

### LINKED ISSUE #1380: Typechecker does not resolve the same module with two different names as the same module
Reproduction: https://github.com/spotandjake/BriskLinker @fad4f7235157ff36d1bc4a5084e9927fa481bd3b

Running `grain index.gr` you get:
```
File "/linker/BuildFile.gr", line 84, characters 57-67:
Error: This expression has type %WasmDecoder.WasmImportDescription
       but an expression was expected of type
         WasmDecoder.WasmImportDescription
WasmDecoder::WasmImportDescription is abstract because no corresponding cmi file was found in path.
```

This should not occur.

A workaround is provided via this patch:
```diff
diff --git a/linker/BuildFile.gr b/linker/BuildFile.gr
index 53147c2..4a9d2e6 100644
--- a/linker/BuildFile.gr
+++ b/linker/BuildFile.gr
@@ -5,11 +5,11 @@ import List from "list"
 import Int32 from "int32"
 import Bytes from "bytes"
 import Stream from "./Stream"
-import { getTypeSection } from "./Decoder/DecodeSection"
-import { WasmImport, WasmImportDescription } from "./Decoder/WasmDecoder"
+import DecodeSection, { getTypeSection } from "./Decoder/DecodeSection"
+import WasmDecoder from "./Decoder/WasmDecoder"
 import Encoder from "./encode"
 // Type Imports
-import { FileData } from "./Decoder/DecodeFile"
+import DecodeFile from "./Decoder/DecodeFile"
 // Buffer Manipulation
 let appendBufferByte = (buffer, byte) => {
   Buffer.addInt8(Int32.fromNumber(byte), buffer)
@@ -29,7 +29,7 @@ export let createTypeSection = (wasmBinary, dependencyChain) => {
   // For Each Dependency
   for (let mut i = 0; i < Array.length(dependencyChain); i += 1) {
     // Get The Dependency
-    let dependency = dependencyChain[i]: FileData
+    let dependency = dependencyChain[i]: DecodeFile.FileData
     // Get The Type Section
     let typeStream = getTypeSection(dependency.wasmSections)
     // Skip This Dependency If There Is Not A Type Section
@@ -66,9 +66,9 @@ export let createImportSection = (wasmBinary, dependencyChain) => {
   // For Each Dependency
   for (let mut i = 0; i < Array.length(dependencyChain); i += 1) {
     // Get The Dependency
-    let dependency = dependencyChain[i]: FileData
+    let dependency = dependencyChain[i]: DecodeFile.FileData
     // Map Each Import
-    List.forEach((wasmImport: WasmImport) => {
+    List.forEach((wasmImport: WasmDecoder.WasmImport) => {
       // TODO: Link Together Brisk Imports
       if (wasmImport.briskImport) {
         void
diff --git a/linker/Decoder/DecodeFile.gr b/linker/Decoder/DecodeFile.gr
index d551dd1..a07b19d 100644
--- a/linker/Decoder/DecodeFile.gr
+++ b/linker/Decoder/DecodeFile.gr
@@ -2,17 +2,17 @@
 import List from "list"
 import String from "string"
 import Stream from "../Stream"
-import { WasmSection, decodeLinkingSection, decodeFile, decodeImportSection } from "./DecodeSection"
-import { WasmImport } from "./WasmDecoder"
+import DecodeSection, { decodeLinkingSection, decodeFile, decodeImportSection } from "./DecodeSection"
+import WasmDecoder from "./WasmDecoder"
 // Types
 export record FileData {
   // Module Data
   importIdentifier: String,
   // Raw Module
-  wasmSections: List<WasmSection>,
+  wasmSections: List<DecodeSection.WasmSection>,
   wasmBinary: Bytes,
   // Parsed Section Data
-  wasmImports: List<WasmImport>,
+  wasmImports: List<WasmDecoder.WasmImport>,
   // References
   functionReferences: List<Number>,
   typeReferences: List<Number>,
@@ -39,12 +39,12 @@ export let decodeFile = wasmBinary => {
   // Parse Section Headers To Determine The Number Of Things Like Functions, Globals, Imports, Exports, And other important data
   // Find Relevant Imports
   // TODO: I Should not need a type cast there, but the inferring seems to be broken
-  let imports = List.map((importData: WasmImport) => {
+  let imports = List.map((importData: WasmDecoder.WasmImport) => {
     importData.briskImport = String.startsWith(linkingInfo.importIdentifier, importData.importModule)
     importData
   }, imports)
-  let briskImports = List.filter((importData: WasmImport) => importData.briskImport, imports)
-  let briskImports = List.map((importData: WasmImport) => {
+  let briskImports = List.filter((importData: WasmDecoder.WasmImport) => importData.briskImport, imports)
+  let briskImports = List.map((importData: WasmDecoder.WasmImport) => {
     String.slice(
       String.length(linkingInfo.importIdentifier),
       String.length(importData.importModule),
diff --git a/linker/Decoder/DecodeSection.gr b/linker/Decoder/DecodeSection.gr
index 4a90d1c..25a4221 100644
--- a/linker/Decoder/DecodeSection.gr
+++ b/linker/Decoder/DecodeSection.gr
@@ -1,7 +1,7 @@
 // Imports
 import List from "list"
 import Stream from "../Stream"
-import { getWasmHeaderCount, decodeWasmImport } from "./WasmDecoder"
+import WasmDecoder, { getWasmHeaderCount, decodeWasmImport } from "./WasmDecoder"
 // Types
 export enum WasmSection {
   // SectionType(Length, Body)
diff --git a/linker/encode.gr b/linker/encode.gr
index a431190..b3c05c6 100644
--- a/linker/encode.gr
+++ b/linker/encode.gr
@@ -2,7 +2,7 @@ import List from "list"
 import Array from "array"
 import String from "string"
 import Char from "char"
-import WasmDecoder, { WasmImportDescription } from "./Decoder/WasmDecoder"
+import WasmDecoder from "./Decoder/WasmDecoder"
 // Encoders
 // Encode Primitives
 export let encodeUIntLeb128 = value => {
diff --git a/linker/index.gr b/linker/index.gr
index 975079b..8a117cc 100644
--- a/linker/index.gr
+++ b/linker/index.gr
@@ -3,7 +3,7 @@ import Set from "set"
 import List from "list"
 import Array from "array"
 import Buffer from "buffer"
-import { decodeFile } from "./Decoder/DecodeFile"
+import DecodeFile, { decodeFile } from "./Decoder/DecodeFile"
 import { buildFile } from "./BuildFile"
 // Linker Entry Function
 export let link = (entryFile, fileReader, pathAbsolute) => {
```

### LINKED ISSUE #1509: Compiler: Get Proper Position Of Errors in other modules
image

### LINKED ISSUE #1518: Compiler: Imported record types aren't inferred
Hey all, I ran into a small issue last night. I'm trying to export a record from a module and then use it in another one. So for example I have two files

person.gr
```
export record Person { name: String }
```

main.gr
```
import { Person } from "./person.gr"
let bob = { name: "Robert" } //fails with, Error: Unbound record label name. Perhaps you need to import its type or write a type definition?
let bob: Person = { name: "Robert" } //works fine
```
I'm using 0.5.4.

phated on discord mentioned:
"Importing a record doesn't introduce it into the module the same way as defining a record does, so it makes sense that the type system can't infer it

However, that's something that we should change/fix with the new type import/export system"

Thanks!

### LINKED ISSUE #1569: Well-formedness checks not adapted for unicode
We have some well-formedness checks that validate casing of module and type names, but they haven't been updated to support Unicode. They likely are passing due to a bug.

Additionally, it seems like the parser has not allowed wrong casing for types and module names. It should, and the well-formedness checks should validate the names.

ospencer:
Closes #190
Closes #270
Closes #271
Closes #324
Closes #681 
Closes #997 
Closes #1340
Closes #1380 
Closes #1509 
Closes #1518 
Closes #1569 

# Module System

This PR introduces the long-awaited module system in Grain, making it far easier to organize code for consumption by other modules. With this change comes new syntax for including code from files and bringing specific names from a module into the current lexical scope.

## The module header

Each module starts with a module header that names the module.

```
module Foo
```

This is the name that will be used by consuming modules.

By convention, the module name should match the name of the file but this is not enforced.

## Including a module from a file

The `import Foo from "path"` has been sunset in favor of a new `include` syntax:

```
include "path"
```

Unlike `import`, `include` serves to _only_ load code from a file and include it as a module. The module written in the file at `"path"` will enter the scope. For example,

```
include "list"
```

would add the `List` module to the current scope, assuming that the module defined in `"list"` is called `List`.

Module names can be aliased at import time:

```
include "list" as MyList
```

After a module has been included, values, modules, and types from that module can be brought into scope via a `from`/`use` statement:

```
include "foo"
from Foo use { foo, Bar, type Baz }
```

Lowercase identifiers represent values, uppercase identifiers represent modules, and uppercase identifiers following the `type` keyword represent types, which also brings all constructors or record fields of that type into scope. Since these are now separate statements, `from`/`use` statements are allowed anywhere, not just the top level:

```
module Main

include "list"

let doWork = () => {
  from List use { map }
  map(x => x + 1, [1, 2, 3])
}
```

## Providing values

The `export` keyword has been replaced by a new `provide` keyword:

```
module Main

provide enum Foo {
  Bar(String)
}

provide let foo = "foo"
```

The syntax to provide values, modules, and types mirrors a `use` statement:

```
provide { foo, Bar, type Baz }
```

## Modules in modules

New module syntax allows you to define modules within your modules:

```
module Main

module Foo {
  provide let hello = "hello"
}

print(Foo.hello)
```

Of course, we can `use` values from `Foo`:
```
module Main

module Foo {
  provide let hello = "hello"
}

from Foo use { hello }
print(hello)
```

And we can provide the entire module:

```
module Main

provide module Foo {
  provide let hello = "hello"
}
```

We can also provide included modules:

```
module Main

include "list" as List
provide { List }
```


peblair:
One comment/quirk:

If we enforce users putting `module Foo` at the top of `foo.gr`, I think this makes the `import "foo" as NotFoo` syntax a bit awkward. `import "foo" as NotFoo` implies to me that `"foo"` _is_ the module, when in reality it's just the file containing `Foo` (the thing being renamed). I guess maybe I am still not fully grokking why we want to have `module Foo` at the top of `foo.gr` (when would we want it to be something else?).

kee-oth:
What is the reason for using the term "expose" instead of the JavaScript-familiar term "export"?

If it's just an arbitrary choice, I would encourage switching to something else. "export" is at least familiar to JS devs which I assume is a large target demographic.

And for me personally, and for possibly others, "expose" has a fairly negative connotation. It brings to mind unwanted publicity, unwanted nudity, betrayal, death from "exposure",  etc.

spotandjake:
> What is the reason for using the term "expose" instead of the JavaScript-familiar term "export"?
> 
> If it's just an arbitrary choice, I would encourage switching to something else. "export" is at least familiar to JS devs which I assume is a large target demographic.
> 
> And for me personally, and for possibly others, "expose" has a fairly negative connotation. It brings to mind unwanted publicity, unwanted nudity, betrayal, death from "exposure", etc.

You can have submodules `export` makes it sound like its accessible outside the file, export is closer to`public` on a js class.

av8ta:
> And for me personally, and possibly others, "expose" has a fairly negative connotation. It brings to mind unwanted publicity, unwanted nudity, betrayal, death from "exposure", etc. It's just an unpleasant word imo.

I see what you mean about "expose", it's likely only appealing to narcissists, and I suspect there are fewer in the dev population than the general population 😁

How about "provides" that's got a fuzzy warm feeling. Sorry for the bikeshedding! btw it doesn't really matter to me; I'll happily use whatever words are chosen so long as they work semantically.

spotandjake:
> > And for me personally, and possibly others, "expose" has a fairly negative connotation. It brings to mind unwanted publicity, unwanted nudity, betrayal, death from "exposure", etc. It's just an unpleasant word imo.
> 
> I see what you mean about "expose", it's likely only appealing to narcissists, and I suspect there are fewer in the dev population than the general population 😁
> 
> How about "provides" that's got a fuzzy warm feeling. Sorry for the bikeshedding! btw it doesn't really matter to me; I'll happily use whatever words are chosen so long as they work semantically.

I do not like provides, I think they said expose comes from `exposing` in elm. I wasn't for it at first but it isn't that bad. if you give it a chance.

ospencer:
@kee-oth Certainly not an arbitrary choice 🙂 

`import` didn't make sense anymore for the requirements of our module loading system which rules out `export` almost _entirely_ because of JavaScript. Users who `export` things expect to be able to `import` them, and given that we don't have `import`, that'd be pretty confusing. Those two words also have a natural symmetry that feels pretty off when one of the two is missing.

Like @spotandjake mentioned, `expose` was inspired by Elm's `exposing` keyword. It's maybe the most accurate word to describe what's occurring (and is much better in that sense than `export`). I really liked how accurate it was, but there was something about it that I didn't like, and that's potentially because of one of the alternate meanings that you listed! We certainly appreciate you calling that out—these sorts of things are very difficult to change in the future, and while we have the opportunity to avoid having a keyword with negative connotation, we should.

@av8ta Appreciate your suggestion as well. After a bunch of deliberation, we settled on `provide` (at least for now—anything can change before this gets merged. It's a draft after all!).

alex-snezhko:
What about `public`? I think it conveys the same meaning as `expose` and is also pretty ubiquitous amongst programming languages (Java & friends, Rust, etc)

ospencer:
`public` is... okay. It has definitely been considered heavily. It's not a verb, which makes exporting values by name awkward `public { foo, bar }` versus the more active `export { foo, bar }`/`expose { foo, bar }`/`provide { foo, bar }`. It's also got a bit of an OOP/academia/CS-degree sort of vibe to it, which I don't think suits what we're after with Grain. Rust made it hip by shortening it to `pub`, but that doesn't feel very Grainy either since we actively try to avoid shorthands.

ospencer:
I'll say that it hasn't been completely dismissed 🙂

spotandjake:
> What about `public`? I think it conveys the same meaning as `expose` and is also pretty ubiquitous amongst programming languages (Java & friends, Rust, etc)

I find `public` really object oriented and makes me start thinking of modules more like classes than containers. I think that `provides` isn't bad but I don't think that completely describes what is going on as well `expose` indicates that somthing is visible to the outside the world. I am a fan of `export` always but I don't think it completely works here. possible suggestion `extern` (it kind of sounds right but it could indicate it is external from the file and also this keyword probs fits somewhere else in the lang better) or `external`, I think it feels grain like though. If we do go with `public` I am defiantly more on board with `pub` it feels less oop like.

kee-oth:
@ospencer I really appreciate you and the Grain team taking my comments into consideration! 🙌 

I really like `provide` as a keyword here. It's very much on the positive end of the connotation spectrum and fits perfectly for the use case imo! And I do like it much more than `public` for the reason you mentioned: it's a verb describing the action. 

And as a side note, I _really_ appreciate Grain staying away from shorthand. I'm a huge advocate of not shortening words as that just leads to people guessing what stuff is called and being inconsistent with names representing the same concept throughout a code base.

ospencer:
@spotandjake We did talk about `extern`/`external` as well, and yeah, typically the languages that use those keywords are referring to things that exist outside of the program.

ospencer:
@av8ta The bikeshedding on this PR may be relatively contained but don't you worry we had plenty of it in our internal chat 🤣

av8ta:
@ospencer I almost didn't comment because I didn't want to be the one to annoy y'all hardworking folks who are providing (:grin:) us such a great language to use! The first word that came to mind was `public` but I immediately dismissed it for precisely the reasons you mentioned as well as what @spotandjake pointed out, that it has an icky object oriented vibe. Didn't expect `provide` to be adopted but am super happy you like it :small_airplane: :ear_of_rice:

compiler/test/suites/imports.re
```diff
@@ -15,175 +15,121 @@ describe("imports", ({test, testSkip}) => {
...
-    "Module 'foo' should have an uppercase name",
+    "include \"list\" as foo;",
+    "Expected an uppercase module identifier",
   );
   assertCompileError(
     "import_module_not_external",
-    "import Foo.Foo from \"list\";",
-    "Module name 'Foo.Foo' should contain only one module.",
+    "include \"list\" as Foo.Foo;",
+    "Expected a newline character to terminate the statement",
```
phated, line 117: This is an interesting error for the problem. Anyway we can make it better?

compiler/test/suites/imports.re
```diff
@@ -204,18 +150,18 @@ describe("imports", ({test, testSkip}) => {
...
     "test_bad_import",
-    "{let x = (1, 2); import * from \"tlists\"; x}",
+    "{let x = (1, 2); include \"tlists\" as TLists; x}",
     "error",
   );
   assertFileRun("test_file_same_name", "list", "OK\n");
   assertSnapshot(
     "annotation_across_import",
-    "import TList, { Empty } from \"tlists\"; let foo : TList.TList<String> = Empty; foo",
+    "include \"tlists\" as TLists; from TLists use { type TList }; let foo : TLists.TList<String> = Empty; foo",
```
phated, line 159: Why "use" the type but then still access it from the module itself?

compiler/test/suites/stdlib.re
```diff
@@ -14,45 +14,75 @@ describe("stdlib", ({test, testSkip}) => {
...
+    "include \"list\" as List; from List use *; true == false",
+  );
+  assertSnapshot(
+    "stdlib_equal_8",
+    "include \"list\" as List; from List use *; [>] == [>]",
+  );
+  assertSnapshot(
+    "stdlib_equal_9",
+    "include \"list\" as List; from List use *; [>] == [> 1]",
+  );
```
phated, line 50: Why do all these include list?

compiler/test/test-libs/noExports.gr
```diff
@@ -1,3 +1,7 @@
-let x = 4;
-let y = 5;
-let z = (foo) => { "bar" };
\ No newline at end of file
+module NoExports
```
phated, line 1: Rename from "exports"

compiler/test/test-libs/onlyXExported.gr
```diff
@@ -1,3 +1,7 @@
-export let x = 4;
-let y = 5;
-let z = (foo) => { "bar" };
\ No newline at end of file
+module OnlyXExported
```
phated, line 1: Reword from "exported" and rename file

compiler/test/test-libs/provideStar.gr
```diff
@@ -0,0 +1,7 @@
+module ProvideStar
```
phated, line 1: Why is this named `ProvideStar` and then doesn't provide star (which I think has been removed)?

compiler/test/test-libs/sameExport.gr
```diff
@@ -1,4 +1,6 @@
-export *
+module SameExport
```
phated, line 1: Rename from "export"

compiler/test/test-libs/unsafeWasmGlobalsExports.gr
```diff
@@ -1,8 +1,10 @@
+module UnsafeWasmGlobalsExports
```
phated, line 1: Rename from "exports"

stdlib/hash.gr
```diff
@@ -1,16 +1,19 @@
 /**
  * @module Hash: Utilities for hashing any value.
- * @example import Hash from "hash"
+ * @example include "hash"
  *
  * @since v0.1.0
  */
 
+module Hash
```
phated, line 8: Does the formatter put this space between?

stdlib/pervasives.gr
```diff
@@ -1,19 +1,26 @@
+/* grainc-flags --no-pervasives */
```
phated, line 1: Why did you choose to do this inside the file instead of skip adding it inside the compiler?

stdlib/pervasives.gr
```diff
@@ -1,19 +1,26 @@
...
+module Pervasives
+
+include "runtime/exception"
+include "runtime/unsafe/memory"
+include "runtime/unsafe/wasmi32"
 
-import { equal as (==) } from "runtime/equal"
-import { compare } from "runtime/compare"
+include "runtime/equal"
+from Equal use { equal as (==) }
```
phated, line 18: Do we really want `runtime/equal` to be available as `Equal`? Why not `RuntimeEqual`?

stdlib/pervasives.gr
```diff
@@ -32,8 +39,9 @@ import {
   (<<),
   (>>>),
   (>>),
-} from "runtime/numbers"
-import { toString, print, concat as (++) } from "runtime/string"
+}
+include "runtime/string"
+from String use { toString, print, concat as (++) }
```
phated, line 44: In relation to my comment above, `runtime/string` making `String` available and `string.gr` doing the same is weird and we should have a better pattern.

stdlib/pervasives.gr
```diff
@@ -112,7 +120,7 @@ export primitive (||): (Bool, Bool) -> Bool = "@or"
  *
  * @since v0.1.0
  */
-export (==)
+provide { (==) }
```
phated, line 123: All of these make me pretty sad. We should rework the place graindoc pulls from to be the original definition so we can use one provide to expose all from the file.

Can you open an issue?

stdlib/queue.gr
```diff
@@ -17,15 +20,15 @@ record Queue<a> {
 }
 
 /**
- * @section Values: Functions and constants for working with queues.
+ * @section Values: Functions and constants for working with immutable arrays.
```
phated, line 23: why did this change?

phated: Looks like you missed the usage in this file too

stdlib/runtime/atof/common.gr
```diff
@@ -28,106 +28,110 @@
...
  * DEALINGS IN THE SOFTWARE.
 */
-
-import WasmI32 from "runtime/unsafe/wasmi32"
-import WasmI64 from "runtime/unsafe/wasmi64"
-import WasmF64 from "runtime/unsafe/wasmf64"
-import { newInt32, newInt64, newFloat64 } from "runtime/dataStructures"
-import { get_POWERS10 } from "runtime/numberUtils"
-import { get_F64_POWERS10_FAST_PATH, get_POWERS5 } from "./table"
+module Common
```
phated, line 31: We need to nail down the naming for these. Naming something just "Common" seems really bad

compiler/test/suites/modules.re
```diff
@@ -0,0 +1,116 @@
...
+  let assertCompileError = makeCompileErrorRunner(test);
+  let assertRun = makeRunner(test_or_skip);
+  let assertFileRun = makeFileRunner(test_or_skip);
+
+  assertSnapshot(
+    "smallest_module",
+    {|
+      module Foo {
+        void
+      }
```
phated, line 18: This is smallest submodule, yeah?

compiler/test/suites/modules.re
```diff
@@ -0,0 +1,116 @@
...
+        provide let foo = "bar"
+      }
+      print(Foo.foo)
+    |},
+    "bar\n",
+  );
+  assertRun(
+    "local_module_use",
+    {|
+      module Foo {
```
phated, line 88: How about a test like:
```
let foo = "foo"

module Foo {
  provide let foo = foo
}

print(Foo.foo)
```
to show that we have everything defined above the module within the scope.

phated (changes requested):
This PR is massive AND massively amazing. Love the new module system ❤️ 

I had a bunch of questions, suggestions for cleanup, and things to think about. I still haven't reviewed the parser messages or snapshots, but wanted to get these sent asap.

cli/__test__/index.gr
```diff
@@ -1,10 +1,11 @@
 /* grainc-flags --no-gc --no-bulk-memory */
```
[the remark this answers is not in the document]
ospencer: https://github.com/grain-lang/grain/issues/1625

compiler/src/codegen/transl_anf.re
```diff
@@ -1310,63 +1310,82 @@ let transl_signature = (~functions, ~imports, signature) => {
...
+          | None =>
+            failwith(
+              "Impossible: path to import not found "
+              ++ Path.name(val_internalpath),
+            )
+          }
+        };
+      exports :=
+        [
+          GlobalExport({
```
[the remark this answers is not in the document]
ospencer: We're talking about WebAssembly exports at this point rather than Grain exports, so `Export` makes more sense here.

compiler/src/codegen/transl_anf.re
```diff
@@ -1310,63 +1310,82 @@ let transl_signature = (~functions, ~imports, signature) => {
...
+      | ReprFunction(args, rets, _) =>
+        switch (Ident_tbl.find_opt(func_map, id)) {
+        | Some(internal_name) =>
+          let external_name = Ident.name(vid);
           exports :=
             [
-              GlobalExport({
-                ex_global_name: Ident.name(vid),
-                ex_global_internal_name: Ident.unique_name(id),
+              FunctionExport({
```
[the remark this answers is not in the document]
ospencer: Same as `GlobalExport`.

compiler/src/formatting/format.re
```diff
@@ -4563,198 +4659,29 @@ let import_print =
...
-
   let path = imp.pimp_path.txt;
 
+  let alias =
+    switch (imp.pimp_alias) {
+    | Some({txt: name}) =>
+      Doc.concat([Doc.space, Doc.text("as"), Doc.space, Doc.text(name)])
+    | None => Doc.nil
+    };
+
```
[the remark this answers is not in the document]
ospencer: No, but AFAIK they weren't before either. We can make an issue to improve comments in `include` statements though.

compiler/src/formatting/format.re
```diff
@@ -5006,23 +5038,53 @@ let format_ast =
...
+      if (leading_comments == []) {
+        Doc.nil;
+      } else {
+        Doc.hardLine;
+      },
+      Doc.text("module"),
+      Doc.space,
+      Doc.text(parsed_program.module_name.txt),
+      Doc.hardLine,
+      Doc.hardLine,
```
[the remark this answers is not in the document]
ospencer: Yeah, to force an empty line after.

compiler/src/middle_end/analyze_closure_scoped_vars.re
```diff
@@ -23,8 +23,8 @@ module CSVArg: Anf_iterator.IterArgument = {
     };
 
   let leave_anf_program = ({signature: {cmi_sign}}) => {
-    Types.(
-      List.iter(
+    let rec process_sig =
+      Types.(
```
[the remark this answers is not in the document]
ospencer: I don't think that'd do enough since there's labels and whatnot that are a part of `Types` that wouldn't get picked up by the annotation.

compiler/src/parsing/ast_helper.re
```diff
@@ -296,6 +296,8 @@ module Exp = {
     mk(~loc?, ~attributes?, PExpReturn(a));
   let constraint_ = (~loc=?, ~attributes=?, a, b) =>
     mk(~loc?, ~attributes?, PExpConstraint(a, b));
+  let use = (~loc=?, ~attributes=?, a, b) =>
```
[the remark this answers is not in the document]
ospencer: I personally really do not like having them mismatched and would prefer we just do them all at once, but I'm not going to hold up on it.

compiler/src/parsing/ast_iterator.re
```diff
@@ -277,45 +295,27 @@ module MB = {
...
+  let iter = (sub, {pimp_alias, pimp_path, pimp_loc}) => {
+    Option.iter(iter_loc(sub), pimp_alias);
     iter_loc(sub, pimp_path);
     sub.location(sub, pimp_loc);
   };
 };
 
-module Ex = {
-  let iter = (sub, exports) =>
+module Pr = {
```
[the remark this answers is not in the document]
ospencer: They all use shortened names, so just matching the style. We can make an issue to have them not use shorted names though.

compiler/src/parsing/ast_iterator.re
```diff
@@ -277,45 +295,27 @@ module MB = {
...
-        },
-      excepts,
+      item => {
+        switch (item) {
+        | PProvideType({name, alias, loc})
+        | PProvideModule({name, alias, loc})
+        | PProvideValue({name, alias, loc}) =>
+          iter_ident(sub, name);
+          Option.iter(iter_ident(sub), alias);
+          sub.location(sub, loc);
```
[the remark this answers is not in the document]
ospencer: I don't think so... the 3 types that do this are all distinct, and it'll really make refactoring a pain/easy to miss something if there was a helper.

compiler/src/parsing/parser.mly
```diff
@@ -310,48 +309,61 @@ value_bind:
...
-  | special_id { PImportModule (mkid [$1] (to_loc $loc)) }
-  | STAR import_exception? { PImportAllExcept (Option.value ~default:[] $2) }
-  | lbrace import_ids? rbrace { PImportValues (Option.value ~default:[] $2) }
+include_alias:
+  | AS UIDENT { mkstr $loc($2) $2 }
 
-import_stmt:
-  | IMPORT lseparated_nonempty_list(comma, import_shape) comma? FROM file_path { Imp.mk ~loc:(to_loc $loc) $2 $5 }
+include_stmt:
+  | INCLUDE file_path include_alias? { Imp.mk ~loc:(to_loc $loc) $2 $3 }
```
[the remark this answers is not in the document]
ospencer: We can un-shorthand the ast_helper modules in another PR.

compiler/src/typed/env.re
```diff
@@ -1992,105 +2010,33 @@ let rec add_signature = (sg, env) =>
...
-    add_l(
-      x => `Constructor(x),
-      filtered_components.comp_constrs,
-      env0.constructors,
-    );
+    add_l(x => `Constructor(x), comps.comp_constrs, env0.constructors);
 
-  let values =
-    add(x => `Value(x), filtered_components.comp_values, env0.values);
+  let labels = add_l(x => `Label(x), comps.comp_labels, env0.labels);
```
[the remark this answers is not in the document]
ospencer: Yes this was a bug, and yes the way to prevent it would be to not do the record spread down below.

compiler/src/typed/typemod.re
```diff
@@ -73,74 +73,27 @@ let extract_sig_open = (env, loc, mty) =>
...
+    Identifier.IdentName(Location.mkloc(mod_name, sod.pimp_loc));
   let path =
     Typetexp.lookup_module(
       ~load=true,
       env,
-      mod_.pimp_loc,
+      sod.pimp_loc,
       mod_name,
-      filepath,
+      Some(filename),
```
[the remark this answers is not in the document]
ospencer: It's whatever the user wrote in the path on their `include` statement.

compiler/src/typed/typemod.re
```diff
@@ -73,74 +73,27 @@ let extract_sig_open = (env, loc, mty) =>
...
-    type_open_(~toplevel?, env, sod);
-  /* )*/
-
-  let od = {
-    /*open_override = sod.popen_override;*/
-    timp_path: path,
-    /*open_attributes = sod.popen_attributes;*/
-    timp_loc: sod.pimp_loc,
-  };
+  let od = {timp_path: path, timp_loc: sod.pimp_loc};
```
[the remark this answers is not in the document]
ospencer: Yes, though the style we use just swaps the `p` from the parsetree version for `t` for the typedtree version. We could maybe rename them all (but that would be separate from this).

compiler/src/typed/typemod.re
```diff
@@ -578,43 +506,59 @@ let type_module = (~toplevel=false, funct_body, anchor, env, sstr /*scope*/) =>
...
+              ),
+            ),
+          alias:
+            Some(
+              Location.mknoloc(
+                Identifier.IdentName(Location.mknoloc(name)),
+              ),
+            ),
+          loc: Location.dummy_loc,
+        });
```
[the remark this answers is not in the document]
ospencer: This handles the `@externalName` attribute, and it doesn't have locations because attributes don't save location information. Could make an issue about that.

compiler/src/typed/typemod.re
```diff
@@ -625,87 +569,95 @@ let type_module = (~toplevel=false, funct_body, anchor, env, sstr /*scope*/) =>
...
+    List.fold_right(
+      (item, (sigs, stmts)) =>
+        switch (item) {
+        | PProvideValue({name: {txt: IdentName(name)}, alias, loc}) =>
+          let id =
+            switch (alias) {
+            | Some({txt: IdentName(alias)}) => Ident.create(alias.txt)
+            | Some(_) => failwith("Impossible: invalid alias")
+            | None => Ident.create(name.txt)
+            };
```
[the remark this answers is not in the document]
ospencer: Nah, they're all a bit different.

compiler/src/typed/typemod.re
```diff
@@ -908,75 +868,68 @@ let type_module = (~toplevel=false, funct_body, anchor, env, sstr /*scope*/) =>
...
+    modules,
+  );
+};
 
 let lookup_implicit_module_spec = m =>
   switch (m) {
-  | Grain_utils.Config.Pervasives_mod => ("Pervasives", "pervasives", true)
-  | Grain_utils.Config.Gc_mod => ("GC", "runtime/gc", false)
+  | Grain_utils.Config.Pervasives_mod => Some(("Pervasives", "pervasives"))
+  | Grain_utils.Config.Gc_mod => None
```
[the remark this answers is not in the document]
ospencer: The old boolean that was a part of this before indicated whether this module was added to the scope of the module. This was important before to make sure that GC got compiled, but since we changed how that happens this no longer needs to attempt to load/compile it. So the `None` here just means "nothing needs to be done for this."

compiler/src/typed/typetexp.re
```diff
@@ -168,9 +168,9 @@ let lookup_module = (~load=false, env, loc, lid, filepath) =>
     lid,
   );
 
-let find_module = (env, loc, lid, filepath) => {
-  let path = lookup_module(~load=true, env, loc, lid, filepath);
-  let decl = Env.find_module(path, filepath, env);
+let find_module = (env, loc, lid) => {
+  let path = lookup_module(env, loc, lid, None);
+  let decl = Env.find_module(path, None, env);
```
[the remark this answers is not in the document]
ospencer: `lookup_module` and `Env.find_module` in most cases need to know the filepath, but particularly the use of `Typetexp.find_module` is for finding a module that's already been loaded, i.e. `from Foo use { bar }`.

compiler/test/formatter_inputs/imports.gr
```diff
@@ -1,28 +1,28 @@
-import Memory from "runtime/unsafe/memory"
-import Tags from "runtime/unsafe/tags"
-import { length, map } from "list"
+module Imports
 
-import WasmI32, { 
+include "runtime/unsafe/memory"
```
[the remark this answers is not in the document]
ospencer: added in cf729b60

compiler/test/formatter_inputs/imports.gr
```diff
@@ -1,28 +1,28 @@
...
+include "runtime/unsafe/tags"
+include "list"
+from List use { length, map }
```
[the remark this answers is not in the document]
ospencer: added in cf729b60

compiler/test/formatter_inputs/only_comments.gr
```diff
@@ -1,4 +1,4 @@
-     
+module OnlyComments
```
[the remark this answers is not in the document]
ospencer: Yeah, I think it's still a pretty useful test, especially since the module header isn't a statement.

compiler/test/formatter_inputs/values.gr
```diff
@@ -1,13 +1,15 @@
...
+include "int32"
+include "int64"
+include "float32"
+include "float64"
+include "bytes"
+include "runtime/unsafe/wasmi32"
+include "runtime/unsafe/wasmi64"
+
+include "runtime/unsafe/wasmf32"
+include "runtime/unsafe/wasmf64"
```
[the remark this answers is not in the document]
ospencer: Me either!

compiler/test/formatter_outputs/imports.gr
```diff
@@ -1,32 +1,26 @@
...
-  shl as (<<),
-} from "runtime/unsafe/wasmi32"
+include "runtime/unsafe/memory"
+include "runtime/unsafe/tags"
+include "list"
+from List use { length, map }
 
-export foreign wasm storage_read: (
+include "runtime/unsafe/wasmi32"
+from WasmI32 use { add as (+), mul as (*), xor as (^), shl as (<<) }
```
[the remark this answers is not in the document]
ospencer: Without the path, it fits within 80 characters now so the formatter doesn't wrap it.

compiler/test/input/brokenImports/broken.gr
```diff
@@ -1,4 +1,6 @@
-import Number from "number"
-import Foo from "./foo"
+module Broken
```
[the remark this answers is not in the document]
ospencer: done in 17830558

compiler/test/input/mixedPatternMatching.gr
```diff
@@ -30,14 +33,14 @@ let test = test => {
       [PList([a]), PAssoc({ foo: [b], _ }), PMulti([], c, { foo: [d], _ })]
     ) =>
       a +
-      b +
-      c +
-      d,
+        b +
+        c +
+        d,
```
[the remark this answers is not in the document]
ospencer: I don't think it's the worst... we deliberately made it this way some time ago, largely because of statements:
```
assert reallyLongFoo &&
  reallyLongBar
```

Could probably be better though/maybe only on statements.

compiler/test/suites/gc.re
```diff
@@ -71,33 +62,95 @@ describe("garbage collection", ({test, testSkip}) => {
...
     "gc3",
     1024,
     "let foo = (s: String) => void\nlet printBool = (b: Bool) => foo(if (b) \"true\" else \"false\")\n\nlet b = true\nfor (let mut i=0; i<100000; i += 1) {\n  printBool(true)\n}",
+    "",
   );
-  assertFileRunGC(
+  assertRunGCError(
     "fib_gc_err",
     1024,
-    "fib-gc",
```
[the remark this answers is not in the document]
ospencer: It's now been deleted. It's gone because now with the module header, the program concatenation we were doing before doesn't work now, so that code just exists here.

compiler/test/suites/imports.re
ospencer: We don't test the exact location here in the test, but the error shows up on the `.` in `Foo.Foo` and says that it was unexpected, which I don't think is too bad of an error.

Before this PR we just parsed the whole identifier here, but the bad thing about that is we end up with a bunch of `Impossible` throughout the compiler. We could maybe go back to parsing it the old way (though slightly better now) and in the action raise a SyntaxError if the user did this and extract the single module name.

compiler/test/suites/imports.re
ospencer: This test is testing that even though you used `Empty` from the `use`d type, your annotation from the type on the module still typechecks.

compiler/test/suites/modules.re
ospencer: Yup!

compiler/test/suites/modules.re
ospencer: Added in 0eacd824

compiler/test/suites/stdlib.re
ospencer: I think it was to test that importing a library works? We can almostly certainly refactor these (not that they test anything anyway, since the entire program should be optimized away). But different PR 😛

stdlib/hash.gr
ospencer: Yes it does.

stdlib/pervasives.gr
ospencer: We'd have to detect that we were compiling Pervasives, and that felt a little brittle to me (what happens if you import Pervasives manually? what happens if you name another module `Pervasives`?), but this was a surefire way to make sure the flag was applied specifically to this file.

stdlib/pervasives.gr
ospencer: I'm with the convention that filename should be the same as the module name. That also works out if we expose the runtime as it's own package, since you'd reference it as `Runtime.Equal`.

stdlib/pervasives.gr
ospencer: I don't think it's weird.

stdlib/pervasives.gr
ospencer: https://github.com/grain-lang/grain/issues/1626

stdlib/runtime/atof/common.gr
ospencer: I think it's more of a smell of the file having a bad name.

compiler/src/typed/typemod.re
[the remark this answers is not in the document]
phated: Could we name this variable something like `include_path` then or something?

compiler/src/codegen/transl_anf.re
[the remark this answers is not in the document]
phated: Why do we call these `GlobalExport` instead of `WasmExport` or even `WasmGlobalExport`?

compiler/src/codegen/transl_anf.re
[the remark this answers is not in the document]
phated: Why do we call these `FunctionExport` instead of `WasmExport` or even `WasmFunctionExport`? Is there a difference between a wasm function export and wasm global export?

compiler/src/formatting/format.re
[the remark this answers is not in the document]
phated: Let's make an issue but also add a formatter input test that shows how it's currently working with a //TODO so we know how it currently formats and has a diff when the issue is fixed.

compiler/src/parsing/ast_iterator.re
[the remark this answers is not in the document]
phated: Let's make an issue 🙏

compiler/src/parsing/parser.mly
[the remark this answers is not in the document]
phated: This wasn't about un-shortening. `Imp` is short for `Import` but this is now `Includes` so it should be `Inc` and then we can un-shorten in another PR.

compiler/src/typed/env.re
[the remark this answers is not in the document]
phated: Why don't know want to remove the record spread? Would it take a lot?

compiler/src/typed/typetexp.re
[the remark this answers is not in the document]
phated: Can you explain a bit more why it needs the filepath? Is that something that will go away when silo is built and grainc only operates on a single file (and doesn't traverse the dep graph)?

compiler/src/typed/typemod.re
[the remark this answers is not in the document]
phated: Let's make an issue 🙏

compiler/src/typed/typemod.re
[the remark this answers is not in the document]
phated: Issue please 🙏

compiler/src/formatting/format.re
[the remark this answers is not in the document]
ospencer: https://github.com/grain-lang/grain/issues/1627

compiler/test/formatter_inputs/values.gr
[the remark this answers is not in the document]
phated: Issue to cleanup or do you want to remove in this PR?

compiler/test/input/mixedPatternMatching.gr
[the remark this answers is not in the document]
phated: But why are these even wrapping in the first place?

compiler/src/parsing/ast_iterator.re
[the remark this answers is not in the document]
ospencer: https://github.com/grain-lang/grain/issues/1628

compiler/test/suites/imports.re
phated: Hmm, I'm on the fence about having Impossible cases throughout the compiler but better error messages are always appreciated 🤔

compiler/test/suites/stdlib.re
phated: Please open an issue 🙏

compiler/src/typed/env.re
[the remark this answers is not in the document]
ospencer: Done in c66043d9

stdlib/hash.gr
phated: Can we not do that?

stdlib/pervasives.gr
phated: I think that's fine. If we ever allow (or do we now?) people to swap out pervasives, it seems weird that they'd need to know to add this flag to the top.

stdlib/pervasives.gr
phated: I think we need to rework our directory-structure-as-namespacing design here and actually do the `Runtime.Equal`, etc changes in a `runtime.gr` module that provides things to the rest of the stdlib because this is chaotic.

I also think that the convention could easily be PascalCaseOfTheFilepath: `runtime/equal` is `RuntimeEqual`, `runtime/unsafe/wasmi32` is `RuntimeUnsafeWasmI32` but it'd be better if we just didn't suck at naming and organizing in the first place.

compiler/src/typed/typetexp.re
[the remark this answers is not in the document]
ospencer: The tl;dr would be that it's how `include` statements work, and we would still need them even if silo existed. It's part of a bigger story of the tangled mess that is how imports work in OCaml vs. how they work in Grain (just referencing a module in OCaml causes it to be imported, and much of this is cruft from that). We could make an issue clean all of that up.

compiler/src/typed/typemod.re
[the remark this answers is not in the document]
ospencer: https://github.com/grain-lang/grain/issues/1629

compiler/src/typed/typemod.re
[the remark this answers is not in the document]
ospencer: https://github.com/grain-lang/grain/issues/1630

compiler/test/formatter_inputs/values.gr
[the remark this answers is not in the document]
ospencer: https://github.com/grain-lang/grain/issues/1631

compiler/test/input/mixedPatternMatching.gr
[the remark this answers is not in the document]
ospencer: I believe it's because they are not grouped, and anything not grouped will wrap if anything on the line wraps (I think).

(I didn't change anything to do with this by the way, I just formatted the file.)

compiler/test/suites/stdlib.re
ospencer: https://github.com/grain-lang/grain/issues/1631

stdlib/hash.gr
ospencer: I suppose it's a stylistic thing, but I much prefer
```
module Main

print("Hello world")
```

to 

```
module Main
print("Hello world")
```

Plus separating that module statement from a block of include statements, etc.

stdlib/pervasives.gr
ospencer: That's going to be a fairly involved change, so I made an issue: https://github.com/grain-lang/grain/issues/1632

compiler/src/typed/typetexp.re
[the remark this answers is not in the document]
phated: Yeah, let's make an issue ❤️

compiler/test/input/mixedPatternMatching.gr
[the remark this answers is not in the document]
phated: That's fine that you didn't change anything. I just think that code review is a good time to point out things that aren't great about the state of things. Can we get an issue? 🙏

stdlib/hash.gr
phated: Sorry for not being clear. I was talking about the newline *above* the `module` statement (between `module Hash` and the module docblock comment).

compiler/src/parsing/parser_header.re
```diff
@@ -98,13 +94,26 @@ let mkid_expr = (loc, ns) =>
...
 
-let make_program = statements => {
+let make_module_alias = ident => {
+  switch (ident.txt) {
+  | IdentName(name) => name
+  | IdentExternal(_) =>
+    raise(
+      SyntaxError(
+        ident.loc,
+        "This module alias should be at most one level deep.",
```
phated, line 104: I think this error is fine, but here's a bikeshed:
```suggestion
        "A module alias cannot contain the `.` character.",
```
```suggestion
        "A module alias cannot contain the dot character.",
```

compiler/src/typed/env.re
```diff
@@ -1992,105 +2010,33 @@ let rec add_signature = (sg, env) =>
...
-      x => `Module(x),
-      filtered_components.comp_modules,
-      env0.modules,
-    );
+  let modules = add(x => `Module(x), comps.comp_modules, env0.modules);
 
   {
-    ...env0,
     summary: Env_open(env0.summary, root),
+    local_constraints: env0.local_constraints,
```
phated, line 2037: Just for my own education, what is a `local_constraint` in the ocaml type system?

compiler/test/formatter_inputs/includes.gr
```diff
@@ -0,0 +1,32 @@
+module Includes
+
+include "runtime/unsafe/memory"
+include "runtime/unsafe/tags"
+include "list"
+include "option" as Opt
```
phated, line 6:
```suggestion
include "option"
  as Opt
```
Is this valid at the parser? Ideally these formatter_inputs would have weird formatting to verify that the output is nice instead of "didn't change anything"

compiler/test/formatter_inputs/includes.gr
```diff
@@ -0,0 +1,32 @@
...
+// TODO(#1627): This comment disappears
+include /* special include */ "array"
```
phated, line 8: What about one with an alias?
```suggestion
// TODO(#1627): This comment disappears
include /* special include */ "array"
include "array" /* special include */  as Foo
include "array" as  /* special include */ Foo
```

compiler/test/formatter_inputs/includes.gr
```diff
@@ -0,0 +1,32 @@
...
+from List use { length, map, forEach as each }
+from Opt use { MutableOpt, ImmutableOpt as Imm, type Opt, type Opt as OptAlias }
```
phated, line 10: We should probably toss some comments in these too?

compiler/test/suites/modules.re
phated: Okay, suggestion to rename the test above.

compiler/test/suites/modules.re
```diff
@@ -0,0 +1,129 @@
...
+  let test_or_skip =
+    Sys.backend_type == Other("js_of_ocaml") ? testSkip : test;
+
+  let assertSnapshot = makeSnapshotRunner(test);
+  let assertCompileError = makeCompileErrorRunner(test);
+  let assertRun = makeRunner(test_or_skip);
+  let assertFileRun = makeFileRunner(test_or_skip);
+
+  assertSnapshot(
+    "smallest_module",
```
phated, line 14:
```suggestion
    "smallest_submodule",
```

compiler/test/test-libs/provideAll.gr
```diff
@@ -0,0 +1,7 @@
+module ProvideStar
```
phated, line 1:
```suggestion
module ProvideAll
```
?

phated (changes requested):
Saw a few more things when I was reviewing the changes you made.

compiler/src/typed/typetexp.re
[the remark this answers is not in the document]
ospencer: https://github.com/grain-lang/grain/issues/1633

compiler/test/__snapshots__/includes.beda767e.0.snapshot
```diff
@@ -8,7 +8,7 @@ imports › import_all_except_multiple
  (import \"_genv\" \"relocBase\" (global $relocBase_0 i32))
  (import \"_genv\" \"moduleRuntimeId\" (global $moduleRuntimeId_0 i32))
  (import \"GRAIN$MODULE$runtime/gc\" \"GRAIN$EXPORT$incRef\" (global $GRAIN$EXPORT$incRef_0 (mut i32)))
- (import \"GRAIN$MODULE$exportStar\" \"GRAIN$EXPORT$z\" (global $z_1138 (mut i32)))
+ (import \"GRAIN$MODULE$../test-libs/provideAll\" \"GRAIN$EXPORT$x\" (global $x_1126 (mut i32)))
```
phated, line 11: This is interesting, did we change the way it's being included or did the representation just change?

compiler/test/input/mixedPatternMatching.gr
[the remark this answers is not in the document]
ospencer: https://github.com/grain-lang/grain/issues/1634

stdlib/hash.gr
ospencer: Oh, yes it does, though that wasn't really my intention. The formatter seems to consider regular blocks and docblocks differently, where if a line wasn't forced here, if you started your file with grainc-flags, for example, it would print the flags like this:
```
/* grainc-flags --no-gc */ module Main
```

We'll probably need to completely rework how start-of-module comments are printed.

compiler/src/parsing/parser_header.re
ospencer: Hmm, maybe. We might want to lean towards saying something along the lines of "you tried to use a name that normally references something within a module, but you're not allowed to do that here."

compiler/src/typed/env.re
ospencer: I believe this is for like when you want to support pattern matching on an inline record constructor and then binding on that record and using it as a standalone value, but you can't let that type escape that local scope.

compiler/test/formatter_inputs/includes.gr
ospencer: Nah, no breaks within this statement are supported.

stdlib/hash.gr
phated: Please open an issue because this really needs to be fixed. Graindoc has a hack that checks 2 lines backward (due to attributes) which would be the only way this is working (though I don't think Graindoc is even updated to match the module to the docblock)

compiler/src/parsing/parser_header.re
phated: Yeah, something like that.

compiler/src/typed/env.re
phated: Interesting! Thanks 😄

compiler/test/formatter_inputs/includes.gr
phated: Really? That seems like an arbitrary decision and it means that line wrapping is totally broken/unavailable for include statements. e.g.
```
include "my/really/really/really/really/really/extremely/deep/path" as SomethingThatShouldWrap
```

compiler/src/parsing/parser.messages
```diff
@@ -1,525 +1,539 @@
...
+## Ends in an error in state: 749.
+##
+## foreign_stmt -> FOREIGN WASM id_str COLON typ . option(as_prefix(id_str)) FROM file_path [ SEMI RBRACE EOL EOF ]
+##
+## The known suffix of the stack is as follows:
+## FOREIGN WASM id_str COLON typ
 ##
 
-Expected a list enum declaration using the syntax `enum List<a> { [], [...](a, List<a>) }`.
+Expected the keyword `from` followed by a file path.
```
phated, line 234: Why does this say `file path`?

compiler/src/parsing/parser.messages
```diff
@@ -1,525 +1,539 @@
...
 ## accurate view of the past (what has been recognized so far), they
 ## may provide an INCOMPLETE view of the future (what was expected next).
-## In state 3, spurious reduction of production nonempty_list(eol) -> EOL
-## In state 5, spurious reduction of production eols -> nonempty_list(eol)
-## In state 96, spurious reduction of production comma -> COMMA eols
+## In state 102, spurious reduction of production qualified_uid -> lseparated_nonempty_list_inner(dot,type_id_str)
 ##
 
-Expected a variant definition or `}` to complete the enum declaration.
+Expected the keyword `use` followed by items to use surrounded by `{}` or `*` to use all items.
```
phated, line 252: Do we like `items` here? Or should it be something like `bindings`?

compiler/src/parsing/parser.messages
```diff
@@ -1,525 +1,539 @@
...
-## In state 106, spurious reduction of production typ -> data_typ
-## In state 88, spurious reduction of production lseparated_nonempty_list_inner(comma,typ) -> typ
-## In state 94, spurious reduction of production option(comma) ->
-## In state 97, spurious reduction of production typs -> lseparated_nonempty_list_inner(comma,typ) option(comma)
-## In state 119, spurious reduction of production option(typs) -> typs
+## FROM qualified_uid USE
 ##
 
-Expected type parameters surrounded by `<`/`>`, a comma followed by more types, a dot followed by an identifier, or `)` to complete the variant definition.
+Expected items to use surrounded by `{}` or `*` to use all items.
```
phated, line 264: Same question about `items` vs another word

compiler/test/__snapshots__/includes.beda767e.0.snapshot
ospencer: We just changed how it's included in the test. The test this snapshot was originally generated with is gone, but I suppose this is similar enough that git thought it was the same.

compiler/src/parsing/parser.messages
ospencer: Because we changed the syntax during development so many times I messed it up 😩

compiler/src/parsing/parser.messages
```diff
@@ -1,525 +1,539 @@
...
-## value_bind -> pattern equal . expr [ SEMI RBRACE EOL EOF COMMA ]
+## attribute -> AT id_str loption(attribute_arguments) . [ TYPE RECORD LET INCLUDE PROVIDE ENUM AT ]
+## attribute -> AT id_str loption(attribute_arguments) . eols [ TYPE RECORD LET INCLUDE PROVIDE ENUM AT ]
 ##
 ## The known suffix of the stack is as follows:
-## pattern equal
+## AT id_str loption(attribute_arguments)
+##
+
+Attributes are only allowed on let bindings, data declarations, and foreign statements.
```
phated, line 514: Do we still like `data declarations` instead of `type declarations`?

compiler/src/parsing/parser.messages
```diff
@@ -878,300 +910,330 @@ program: WASMI64 COLON LPAREN UIDENT COMMA WHILE
...
+## Ends in an error in state: 0.
+##
+## program' -> . program [ # ]
+##
+## The known suffix of the stack is as follows:
+##
 ##
 
-Expected `)` to complete the exception declaration.
+Expected an import, export, expression, or type declaration.
```
phated: No longer `import` or `export`

stdlib/hash.gr
ospencer: https://github.com/grain-lang/grain/issues/1635

compiler/src/parsing/parser.messages
```diff
@@ -1181,145 +1243,475 @@ program: EXPORT UIDENT COMMA EOL WHILE
...
+## export_stmt -> attributes PROVIDE . primitive_stmt [ SEMI EOL EOF ]
+## export_stmt -> attributes PROVIDE . exception_stmt [ SEMI EOL EOF ]
+## export_stmt -> attributes PROVIDE . separated_nonempty_list(comma,aliasable(any_id_str)) [ SEMI EOL EOF ]
+## export_stmt -> attributes PROVIDE . STAR option(export_exception) [ SEMI EOL EOF ]
+##
+## The known suffix of the stack is as follows:
+## attributes PROVIDE
+##
+
+Expected a let binding, data declaration, exception, comma-separated list of identifiers, or `*` to export all values.
```
phated: We don't have export star anymore

compiler/test/formatter_inputs/includes.gr
ospencer: It's more because line breaks aren't currently allowed when aliasing something than an intentional decision. https://github.com/grain-lang/grain/issues/1636

compiler/test/formatter_inputs/includes.gr
ospencer: done

compiler/test/formatter_inputs/includes.gr
ospencer: done

compiler/src/parsing/parser.messages
```diff
@@ -2385,744 +2777,347 @@ program: FUN LPAREN WASMI64 COMMA EOL WHILE
...
+## if_expr -> IF lparen expr rparen eols expr . [ THICKARROW STAR SLASH SEMI RPAREN RCARET RBRACK RBRACE PIPE LCARET INFIX_90 INFIX_80 INFIX_70 INFIX_60 INFIX_50 INFIX_40 INFIX_30 INFIX_120 INFIX_110 INFIX_100 EOL EOF ELSE DASH COMMA COLON ]
+## if_expr -> IF lparen expr rparen eols expr . ELSE expr [ THICKARROW STAR SLASH SEMI RPAREN RCARET RBRACK RBRACE PIPE LCARET INFIX_90 INFIX_80 INFIX_70 INFIX_60 INFIX_50 INFIX_40 INFIX_30 INFIX_120 INFIX_110 INFIX_100 EOL EOF ELSE DASH COMMA COLON ]
+## if_expr -> IF lparen expr rparen eols expr . ELSE eols expr [ THICKARROW STAR SLASH SEMI RPAREN RCARET RBRACK RBRACE PIPE LCARET INFIX_90 INFIX_80 INFIX_70 INFIX_60 INFIX_50 INFIX_40 INFIX_30 INFIX_120 INFIX_110 INFIX_100 EOL EOF ELSE DASH COMMA COLON ]
+##
+## The known suffix of the stack is as follows:
+## IF lparen expr rparen eols expr
 ##
 
-Expected an infix or prefix operator, e.g. `+`, `-`, `!`.
+Expected a newline character to end the `if` statement or `else` to provide an “else” branch.
```
phated, line 2944: Just noticed this uses the "weird" quotes. Probably a copy-paste problem from google docs. Should we adjust?

phated (changes requested):
Finally review the parse.rmessages and noticed some things

compiler/src/parsing/parser.messages
```diff
@@ -231,7 +231,7 @@ program: MODULE UIDENT EOL FOREIGN WASM LIDENT COLON LIDENT ARROW
 ## FOREIGN WASM id_str COLON typ
 ##
 
-Expected the keyword `from` followed by a file path.
+Expected the keyword `from` followed by an external module name surrounded by `"`.
```
phated, line 234: This is related to the foreign wasm statement? I don't like that we are using "external module name" maybe "external wasm namespace"?

stdlib/pervasives.gr
phated: #1637

stdlib/pervasives.gr
phated: #1637

stdlib/runtime/atof/common.gr
phated: #1637

phated (approved):
Thanks for bearing with me on all those reviews! I'm so excited for this new module system. We'll have a fair amount to clean up but we need to get it landed to unblock other work.

So many hearts ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ 

:shipit:

Issue #753, opened 2021-06-22
Tests: Add tests for building and running in unlinked link mode


phated:
During a local refactor, I renamed some values in the JS runner and found that all the tests pass. It wasn't until asking @ospencer that I found out I needed to try an example program with `--no-link` to uncover a bug.

We need to add test for unlinked mode.

spotandjake:
I think we could probably close this issue

ospencer:
Grain no longer supports running programs unlinked as of v0.5.

PR #1585, opened 2023-01-09
feat!: Remove js-runner

### LINKED ISSUE #645: JS Runtime: Expose grainToJSVal
This is a utility that already exists, but it should be exposed for use by users. I imagine this as `GrainModule.toJSVal`, so you can call `module.toJSVal(value)`. Alternatively, the `module.exports` functions can just be wrapped to do this automatically, with the original exports exposed under another name.

### LINKED ISSUE #1537: GrainRunner: Use the real module name over the url
image
Use the real module name over the module url.

ospencer:
Closes #645
Closes #1537

This PR removes the old js-runner. The recommended way to run Grain programs is via standard wasi-supporting tools!

Key points: `-p` is no longer supported by the CLI, and `--no-link` can only be used with `grain compile` (the cli no longer supports running unlinked programs).

Tested and confirmed working with pkg.


cli/bin/grain.js
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/env -S node --experimental-wasi-unstable-preview1 --no-warnings
```
ospencer, line 1: These are node flags, not v8 flags (so they can't be set via `v8.setFlagsFromString`). Unfortunately there isn't a way to only suppress experimental feature warnings.

phated: This won't work on windows...

ospencer: As evident by the failing tests!

peblair (approved):
🪦

marcusroberts (approved):
The times they are a-changing!

ospencer:
Thanks to some investigation by @phated, we learned the two reasons the Windows tests are failing are because the process clock has only millisecond resolution (which is too imprecise for this test to pass) and the thread clock doesn't work. I sent https://github.com/nodejs/uvwasi/pull/181 and https://github.com/nodejs/uvwasi/pull/182 to address these. Hopefully they make it into Node 19, and we'll be able to ship this 😎

spotandjake:
Given this will drop support for running in unlinked mode issue: #753 can probably be closed whenever this gets merged.

cli/bin/grain.js
```diff
@@ -189,28 +175,27 @@ program
...
   .forwardOption("-o <filename>", "output filename")
+  .forwardOption("--no-link", "disable static linking")
   .action(exec.grainc);
 
 program
   .command("run <file>")
-  .description("run a wasm file with grain's javascript runner")
-  .addOption(new commander.Option("-p, --print-output").hideHelp())
-  .action(run);
+  .description("run a wasm file via Node.js")
```
phated, line 197:
```suggestion
  .description("run a wasm file via grain's WASI runner")
```
I think we should say this because eventually this will be wamr.

cli/bin/exec.js
```diff
@@ -114,9 +114,39 @@ function execGrainlsp(options, program, execOpts = { stdio: "inherit" }) {
...
+  const preopens = {};
+  options.dir?.forEach((preopen) => {
+    const [guestDir, hostDir = guestDir] = preopen.split("=");
+    preopens[guestDir] = hostDir;
+  });
+  const doubleDash = process.argv.indexOf("--");
+  const args = process.argv.slice(doubleDash + 1);
+
+  const env = {
+    PREOPENS: JSON.stringify(preopens),
```
phated, line 135: @ospencer Thoughts on doing this to pass the preopens to the WASI runner?

cli/bin/exec.js
```diff
@@ -114,9 +114,39 @@ function execGrainlsp(options, program, execOpts = { stdio: "inherit" }) {
...
+const grainrun = getGrainrun();
+
+function execGrainrun(file, options, program, execOpts = { stdio: "inherit" }) {
+  const preopens = {};
+  options.dir?.forEach((preopen) => {
+    const [guestDir, hostDir = guestDir] = preopen.split("=");
+    preopens[guestDir] = hostDir;
+  });
+  const doubleDash = process.argv.indexOf("--");
+  const args = process.argv.slice(doubleDash + 1);
```
phated, line 132: @ospencer Thoughts on doing this to pass the trailing argv to the WASI runner?

cli/bin/exec.js
```diff
@@ -114,9 +114,39 @@ function execGrainlsp(options, program, execOpts = { stdio: "inherit" }) {
...
+
+  const env = {
+    PREOPENS: JSON.stringify(preopens),
+    NODE_OPTIONS: `--experimental-wasi-unstable-preview1 --no-warnings`,
+  };
+
+  try {
+    execSync(`${grainrun} ${file} ${args.join(" ")}`, { ...execOpts, env });
```
phated, line 140: args are appended here.

phated:
One question I've had since last night: How do we want to support env vars via the CLI? I think we can probably do the same thing as with preopens: `--env=FOO=BAR`. WDYT?

ospencer:
That was my thinking exactly.

cli/bin/exec.js
phated: Going to do it this way and we can change it later if we have a better way.

phated (approved):

Issue #1587, opened 2023-01-09
Enable tail calls by default


ospencer:
Probably drop `experimental-wasm-tail-call` for `--no-tail-call`.

I believe there's currently the opportunity for potential memory leaks during tail calls which should be resolved before enabling it by default.

PR #1589, opened 2023-01-09
feat(compiler)!: Enable tail calls by default

spotandjake:
This pr enables tail calls by default it adds the cli argument `--no-wasm-tail-call` and removes the `--experimental-wasm-tail-calls` cli argument. 

This is marked as a draft because in #1587 it is noted that there are some cases where tail calls currently leak which need to be fixed before this is default.

closes: #1587

### DIFF
--- a/cli/__test__/index.gr
+++ b/cli/__test__/index.gr
@@ -1,4 +1,4 @@
-/* grainc-flags --no-gc --no-bulk-memory */
+/* grainc-flags --no-gc --no-bulk-memory --no-wasm-tail-call */
 module NearTest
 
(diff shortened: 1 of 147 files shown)


peblair:
@phated (apologies if this duplicates Discord discussion) What do you recommend we do about the NEAR smoketest here? Aside from the failing CI, this looks good.

spotandjake:
I was thinking we could pass in the compiler flag to not use tail calls for those tests.

spotandjake:
@ospencer The `gc3` test is failing with a `Maximum memory overflow.

ospencer (approved):

PR #1591, opened 2023-01-10
feat(compiler)!: Include Option and Result as language-supplied types

### LINKED ISSUE #1049: Compiler: Bake Option and Result types into the compiler
This would make Option and Result a part of the language rather than a part of the standard library. This adds opportunities for things like optional arguments, and the ability to use Option/Result in Grain's runtime.

ospencer:
Closes #1049 

The snapshot changes are because Option/Result were removed from Pervasives, which everything depends on so IDs are all now two lower.


phated (approved):
Nice example of how to add new enum types to the compiler! One thing I noticed is that we need to remember to change the runtime/string.gr file

PR #1595, opened 2023-01-11
chore: Update snapshots

ospencer:


phated (approved):
Can we enable the GitHub setting where it'll require the branches to be updated? 😛

ospencer:
Fineeeeeeeeeeeeee. But I'm turning it off when it gets on my nerves :P

PR #1596, opened 2023-01-11
feat(stdlib): fs dev-friendly filesystem api

### LINKED ISSUE #211: Stdlib: user-friendly filesystem API

av8ta:
Closes #211 

Draft because it needs discussion.

### DIFF
--- a/compiler/src/parsing/lexer.re
+++ b/compiler/src/parsing/lexer.re
@@ -252,6 +252,8 @@ let rec token = lexbuf => {
   | "try" => positioned(TRY)
   | "throw" => positioned(THROW)
   | "catch" => positioned(CATCH)
+  | "macro" => positioned(MACRO)
+  | "yield" => positioned(YIELD)
   | "..." => positioned(ELLIPSIS)
   | "." => positioned(DOT)
   | "::" => positioned(COLONCOLON)
--- a/compiler/src/parsing/parser.mly
+++ b/compiler/src/parsing/parser.mly
@@ -42,7 +42,7 @@ module Grain_parsing = struct end
 %token EOL EOF
 
 // reserved tokens
-%token TRY CATCH COLONCOLON
+%token TRY CATCH COLONCOLON MACRO YIELD
 
 // Not a real token, this is injected by the lexer
 %token FUN


av8ta:
I wrote this because while working on my streaming library and after creating array and list sources I got excited about stdin and the rest of the filesystem. Finding the sys/file library a bit clunky to use I figured I ought to write something a bit friendlier so that I had something I enjoyed using for one thing but primarily so I could figure out how it all actually works. I hadn't seen #211 when I started.

I decided to stay close to the sys/file api, I didn't want to get too far away from it until I really understood all the potential pitfalls. So I tried to come up with a balance of an abstraction not too far from the underlying while being more friendly to use.

```grain
export enum FilePath {
  Fd(File.FileDescriptor),
  PwdPath(String), /* fd = File.pwdfd */
  FdPath(File.FileDescriptor, String),
}

export record PathOpenOptions {
  lookupFlag: List<File.LookupFlag>,
  openFlags: List<File.OpenFlag>,
  rights: List<File.Rights>,
  inheriting: List<File.Rights>,
  rwFlags: List<File.FdFlag>,
}
```

For one-off reads and multiple writes life is pretty good using `PwdPath("./path")` but once you get to reading files there's a surprise. If you're reading the whole file in one read you won't realise anything is afoot... but once you're reading in chunks it gets strange from the users' perspective. The FileDescriptors aren't global so each `PwdPath("./path")` call actually gives a new FileDescriptor on each read. Consequently the seek position is back at the beginning of the file and therefore reads the same data as the previous read. This doesn't happen in a similar fashion with writes, as the `File.FdFlag` `Append` flag is passed when opening the file.

Users will likely expect FileDescriptors to behave like file handles.
I think it would be good to provide that experience. Lest we break the principle of least surprise.

Three potential solutions spring to mind:

* 1)  Provide a FileDescriptor store the user can use to manually store their Fd
* 2)  Provide a store users are unaware of that magically handles them
* 3)  Provide a store that is magic but also can be used directly by users wanting more control

I am returning the FileDescriptors with the Result so users can do whatever they like with them and that's what I'm doing when streaming with the hessian readFile source. It just stores the fd in the closure after opening the file. So users can pass any FilePath constructor. It's not horrible, but you need to know it must be done. And they must be closed too!

The PathOpenOptions record is quite nice. The linter tells you what you're missing when writing code and want to manually specify how to open a file. sys/file takes positional arguments for this and it's super hard to remember where they go!

I'm keen to hear your thoughts on this api and am open to other ideas :)

alex-snezhko:
A few thoughts:

- The topic of independence from the `sys/file` module was discussed earlier in #211, where I believe the conclusion was reached that the user shouldn't have to reach for `sys/file` when working with this module (e.g. populating `PathOpenOptions`, working directly with `File.FileDescriptor`). For example, instead of using `File.Filestats`, perhaps a new specialized enum could be created that the user could interact with instead?
- In my opinion the `Err` variant of the `Result`s returned shouldn't be strings but rather variants of an enum (then maybe `systemErrorToString` could be adapted to convert the variants to strings if the user would like. This would make pattern-matching against specific errors easier and more type-safe rather than the user having to remember the exact strings to determine which error was hit.
- As you touched on, I think it would be nice to have the concept of file handles to allow writing to/reading from a file chunk by chunk. Maybe a possible solution here would be to have separate functions for working with file handles and for file-wide operations? For example maybe the `FilePath` you created could be removed and the API would be split into functions that operate on `Path`s and functions that operate on file handles? For example `readFile` could take a `Path` as an argument, open a new `fd` internally, and return the whole file's contents; then something like `readFileHandle` (a better name should probably be chosen) could operate on a `FileHandle` type which internally maintains the same `fd`?
- On the topic of the `PathOpenOptions` record, when #388 gets addressed it should be cleaner to work with this. I'm a fan of keeping the record for now though.

Let me know what you think.

spotandjake:
> A few thoughts:
> 
> * The topic of independence from the `sys/file` module was discussed earlier in [Stdlib: user-friendly filesystem API #211](https://github.com/grain-lang/grain/issues/211), where I believe the conclusion was reached that the user shouldn't have to reach for `sys/file` when working with this module (e.g. populating `PathOpenOptions`, working directly with `File.FileDescriptor`). For example, instead of using `File.Filestats`, perhaps a new specialized enum could be created that the user could interact with instead?
> * In my opinion the `Err` variant of the `Result`s returned shouldn't be strings but rather variants of an enum (then maybe `systemErrorToString` could be adapted to convert the variants to strings if the user would like. This would make pattern-matching against specific errors easier and more type-safe rather than the user having to remember the exact strings to determine which error was hit.
> * As you touched on, I think it would be nice to have the concept of file handles to allow writing to/reading from a file chunk by chunk. Maybe a possible solution here would be to have separate functions for working with file handles and for file-wide operations? For example maybe the `FilePath` you created could be removed and the API would be split into functions that operate on `Path`s and functions that operate on file handles? For example `readFile` could take a `Path` as an argument, open a new `fd` internally, and return the whole file's contents; then something like `readFileHandle` (a better name should probably be chosen) could operate on a `FileHandle` type which internally maintains the same `fd`?
> * On the topic of the `PathOpenOptions` record, when [[Lang] Named (Optional) arguments #388](https://github.com/grain-lang/grain/issues/388) gets addressed it should be cleaner to work with this. I'm a fan of keeping the record for now though.
> 
> Let me know what you think.

Shouldn't The `Err` variants, return `Exceptions` rather than strings or enums.

alex-snezhko:
That would work too, I still feel that the exception should wrap an enum though
```
enum SystemExceptionType { ... }
exception SystemException(SystemExceptionType)
```
The `Err` cases in `Path` just return an enum though

av8ta:
Thanks @alex-snezhko that was helpful feedback, such that while pondering it made me realise I had an incorrect mental model of how the FileDesciptors worked with respect to the sandbox 🤦


|param|type|description|
|-----|----|-----------|
|`dirFd`|`FileDescriptor`|The directory in which ***path resolution starts***|
|`path`|`String`|The path ***to*** the file or directory|

I read these two sentences in the docs and got hung up on "The directory in which ***path resolution starts***"

Knowing the existence of the sandbox and it being a capability system, but not knowing how it worked, I thought the file system was opaque to the code running in wasm, therefore the paths needed to be relative from a preopened file descriptor.

That's why I was pairing file descriptors with relative paths in the loathed FilePath enum. Once the penny dropped with a glance at the wasi docs I realised *why* that enum was quite rightly disliked! 

Opening a file with an absolute path confirmed it's vastly simpler than I thought 😂 so now that I'm on the same page, I agree with what you've proposed @alex-snezhko 

I've sketched out some pseudo code for what reading might look like to gather feedback.

```js
export let readFile = (path: Path.Path) => {
  // open FileDescriptor
  // read file contents until end
  // close FileDescriptor

  contents
}

export let readFileChunk = (handle: Fs.FileHandle, maxBytesToRead: Number) => {
  // store handle if not seen previously
  // read one chunk
  // successive calls to readFileChunk advance

  (chunk, numBytesRead)
}

export let readFileChunks = (handle: Fs.FileHandle, maxBytesToRead: Number, callback) => {
  // store handle if not seen previously
  // read each chunk & call back on each 
  // until end of file

  callback((chunk, numBytesRead))
}

export let readFileSlice = (handle: Fs.FileHandle, range:Range.Range) => {
  // store handle if not seen previously
  // read chunk in range

  (slice, numBytesRead)
}

export let readFileSlices = (handle: Fs.FileHandle, range:Range.Range, maxBytesToRead: Number, callback) => {
  // store handle if not seen previously
  // read chunk in range up to maxBytesToRead
  // callback on each chunk

  callback((chunk, numBytesRead))
}
```

Should `callback` go at the start of the arguments to be aligned with the rest of the stdlib? 

```js
export let readFileSlices = (callback, range:Range.Range, maxBytesToRead: Number, handle: Fs.FileHandle) => {}
```

Therefore the file handle would be at the end since it's the data? Looks awkward to me though. Besides, it's a handle to data, not a collection...

Your thoughts?

PR #1598, opened 2023-01-14
chore(compiler): Move constants from compcore into comp_utils

phated:
This is a refactor that needs to happen prior to binaryen 110 because we want to make a constant for the memory and we'd end up with a dependency cycle with compcore and comp_utils.

### DIFF
--- a/compiler/src/codegen/comp_utils.re
+++ b/compiler/src/codegen/comp_utils.re
@@ -3,6 +3,12 @@ open Binaryen;
 open Grain_typed;
 open Grain_utils;
 
+let grain_main = "_gmain";
+let grain_type_metadata = "_gtype_metadata";
+let grain_start = "_start";
+let grain_env_name = "_grainEnv";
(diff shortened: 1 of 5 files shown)


compiler/src/codegen/comp_utils.re
```diff
@@ -3,6 +3,12 @@ open Binaryen;
 open Grain_typed;
 open Grain_utils;
 
+let grain_main = "_gmain";
+let grain_type_metadata = "_gtype_metadata";
+let grain_start = "_start";
+let grain_env_name = "_grainEnv";
```
ospencer, line 9: Leaving this because of snapshots?

ospencer (approved):

compiler/src/codegen/comp_utils.re
phated: Yep

phated: Same with `"tbl"`

PR #1599, opened 2023-01-14
chore(compiler): Update binaryen.ml to v0.20.1

phated:
This updates to binaryen 110 using our binaryen.ml bindings v0.20.1 - the big change is that binaryen requires all APIs to use an internal memory name when interacting with memories. This is to support multi-memory, I believe. However, we found a bunch of problems and/or bugs with their implementation that we need to raise upstream.

This currently has workarounds for the issues but relies on us dropping the js-runner in #1585. It also is built upon my cleanup in #1598


compiler/src/codegen/compcore.re
```diff
@@ -3278,7 +3283,6 @@ let compile_imports = (wasm_mod, env, {imports}) => {
     );
 
   List.iter(compile_import, imports);
-  Import.add_memory_import(wasm_mod, "mem", grain_env_mod, "mem", false);
```
phated, line 3281: @ospencer Removing this is causing your linking test for the import_memory flag. I have no idea why.

ospencer (approved):

PR #1603, opened 2023-01-16
feat(compiler): Rational number type

### LINKED ISSUE #1524: Rational type & standard library
The Grain `Number` type supports rational numbers, but we should have a standalone Rational type and library.

Proposed syntax: `1/2r`

alex-snezhko:
Implements rational number type and stdlib. Literals are suffixed with `r` (like `1/4r`) and stdlib contains basic arithmetic and comparison operations.

Closes #1524

### DIFF
--- a/compiler/src/parsing/lexer.re
+++ b/compiler/src/parsing/lexer.re
@@ -253,6 +253,10 @@ let rec token = lexbuf => {
   | (unsigned_float, 'w') => positioned(WASMF32(sub_lexeme(lexbuf, 0, -1)))
   | (unsigned_float, 'W') => positioned(WASMF64(sub_lexeme(lexbuf, 0, -1)))
   | (unsigned_int, 't') => positioned(BIGINT(sub_lexeme(lexbuf, 0, -1)))
+  | (unsigned_int, '/', unsigned_int, 'r') =>
+    positioned(RATIONAL(sub_lexeme(lexbuf, 0, -1)))
+  | (unsigned_int, '/', '-', unsigned_int, 'r') =>
+    positioned(RATIONAL(sub_lexeme(lexbuf, 0, -1)))
   | unsigned_int => positioned(NUMBER_INT(Sedlexing.Utf8.lexeme(lexbuf)))
   | "primitive" => positioned(PRIMITIVE)
   | "foreign" => positioned(FOREIGN)
--- a/compiler/src/parsing/lexer.re
+++ b/compiler/src/parsing/lexer.re
@@ -253,9 +253,7 @@ let rec token = lexbuf => {
   | (unsigned_float, 'w') => positioned(WASMF32(sub_lexeme(lexbuf, 0, -1)))
   | (unsigned_float, 'W') => positioned(WASMF64(sub_lexeme(lexbuf, 0, -1)))
   | (unsigned_int, 't') => positioned(BIGINT(sub_lexeme(lexbuf, 0, -1)))
-  | (unsigned_int, '/', unsigned_int, 'r') =>
-    positioned(RATIONAL(sub_lexeme(lexbuf, 0, -1)))
-  | (unsigned_int, '/', '-', unsigned_int, 'r') =>
+  | (unsigned_int, '/', Opt('-'), unsigned_int, 'r') =>
     positioned(RATIONAL(sub_lexeme(lexbuf, 0, -1)))
   | unsigned_int => positioned(NUMBER_INT(Sedlexing.Utf8.lexeme(lexbuf)))
   | "primitive" => positioned(PRIMITIVE)
--- a/compiler/src/typed/checkertypes.re
+++ b/compiler/src/typed/checkertypes.re
@@ -71,6 +71,7 @@ let type_constant =
   | Const_wasmf32(_) => instance_def(Builtin_types.type_wasmf32)
   | Const_wasmf64(_) => instance_def(Builtin_types.type_wasmf64)
   | Const_bigint(_) => instance_def(Builtin_types.type_bigint)
+  | Const_rational(_) => instance_def(Builtin_types.type_rational)
   | Const_bool(_) => instance_def(Builtin_types.type_bool)
   | Const_void => instance_def(Builtin_types.type_void)
   | Const_bytes(_) => instance_def(Builtin_types.type_bytes)
@@ -267,6 +268,30 @@ let constant:
             ~loc,
             "Unable to parse big-integer literal %st.",
             n,
+          ),
+        )
+      }
+    | PConstRational(n, d) =>
+      // TODO(#1168): allow arbitrary-length arguments in rational constants
+      switch (Literals.conv_number_rational(n, d)) {
+      | Some((n, d)) =>
+        // (until above TODO is done, we keep existing behavior and limit to 32-bits (see #1168))
+        Ok(
+          Const_rational({
+            rational_negative: Int32.compare(n, 0l) < 0, // true if rational is less than 0
+            rational_num_limbs: [|Int64.abs(Int64.of_int32(n))|],
+            rational_den_limbs: [|Int64.abs(Int64.of_int32(d))|],
+            rational_num_rep: Int32.to_string(n),
+            rational_den_rep: Int32.to_string(Int32.abs(d)),
+          }),
+        )
+      | None =>
+        Error(
+          Location.errorf(
+            ~loc,
+            "Rational literal %s/%sr is outside of the rational number range of the Number type.",
+            n,
+            d,
           ),
         )
       }
--- a/compiler/test/stdlib/rational.test.gr
+++ b/compiler/test/stdlib/rational.test.gr
@@ -1,33 +1,51 @@
 module RationalTest
 
-let (===) = (==)
-let (!==) = (!=)
-
 include "rational"
-from Rational use *
+from Rational use {
+  fromNumber,
+  toNumber,
+  numerator,
+  denominator,
+  (+),
+  (-),
+  (*),
+  (/),
+}
 
 assert fromNumber(1/2) == 1/2r
 assert fromNumber(0) == 0/1r
 assert fromNumber(1/1) == 1/1r
 assert fromNumber(-1) == -1/1r
 
-assert toNumber(1/2r) === 1/2
-assert toNumber(0/1r) === 0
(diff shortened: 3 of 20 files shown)


compiler/src/parsing/lexer.re
```diff
@@ -214,6 +214,8 @@ let rec token = lexbuf => {
   | (unsigned_float, 'w') => positioned(WASMF32(sub_lexeme(lexbuf, 0, -1)))
   | (unsigned_float, 'W') => positioned(WASMF64(sub_lexeme(lexbuf, 0, -1)))
   | (unsigned_int, 't') => positioned(BIGINT(sub_lexeme(lexbuf, 0, -1)))
+  | (unsigned_int, '/', unsigned_int, 'r') =>
```
peblair, line 217: I almost wish this was `'q'` instead of `'r'`, but I think that'd make people cry...

peblair (commented):
Thanks for submitting this! Having a `Rational` type definitely makes sense. Maybe I missed the discussion, but I am a bit confused about one thing, though -- why do we want to change the `Number` semantics to allow for unreduced rationals (e.g. `4/1r`)? In my opinion, we should have the following:
- Literals such as `4/1r` get compiled down to integers
- The `Rational` type consists of non-fixed-width integers (so simple ints and bigints) and true rationals.

I worry that having the value `1/1` being distinct from `1` is courting bugs in the future with the number library.

ospencer:
@peblair This doesn't change the semantics of Number; `4/1` still gets reduced to `4`, and within the Number type there's no denominator of one. Your gut feeling about the bugs is definitely real, but it's actually the same as how all the other number types work, e.g. `1L` is not represented/reduced to the representation `1`, and all the various number types go through a conversion process when converting to a Number (it's not a no-op).

compiler/src/parsing/lexer.re
ospencer, line 217: This unfortunately doesn't support having a negative denominator. I know it sounds wild, but I think what we want to do here is specifically parse something like `5r` specifically as a rational denominator, and parse this the same way we handle `Number` rationals, but if you used a rational denominator you get a `Rational` literal instead of `Number` literal.

compiler/src/typed/checkertypes.re
```diff
@@ -69,6 +69,7 @@ let type_constant =
   | Const_wasmf32(_) => instance_def(Builtin_types.type_wasmf32)
   | Const_wasmf64(_) => instance_def(Builtin_types.type_wasmf64)
   | Const_bigint(_) => instance_def(Builtin_types.type_bigint)
+  | Const_rational(_) => instance_def(Builtin_types.type_rational)
```
ospencer, line 72: As I was reading this I thought "he didn't add the `Rational` type. How does any of this work?" but it turns out we knew what we were doing all along 😂

stdlib/runtime/numbers.gr
```diff
@@ -1156,6 +1156,74 @@ let numberAddSubBigIntHelp = (x, y, isSub) => {
...
+    if (isDivide) BI.mul(xNumerator, yDenominator)
+    else BI.mul(xNumerator, yNumerator)
+  let denominator =
+    if (isDivide) BI.mul(xDenominator, yNumerator)
+    else BI.mul(xDenominator, yDenominator)
+  reducedFractionBigInt(numerator, denominator, keepRational)
+}
+
+@unsafe
+export let cmpRationals = (x, y) => {
```
ospencer, line 1205: Would call this `cmpRational` to match the style of the others.

ospencer (changes requested):
This looks pretty close to ready to go! Some small notes, though see the one bigger note on parsing.

compiler/src/typed/checkertypes.re
alex-snezhko: Haha yeah not much ended up having to be changed in the compiler

stdlib/runtime/numbers.gr
alex-snezhko: I named in `cmpRationals` because `cmpRational` already exists (it compares a rational number to an unknown number type). I thought adding the `s` would make it clearer that it is known that both are rational

compiler/src/parsing/lexer.re
alex-snezhko: Couldn't you just make the numerator negative though if you wanted a negative number?

alex-snezhko: Also with what you're suggesting how should an error be thrown in the case of `let a = 5r`? Unless I am missing something, it would seem that every single token would have to scanned individually to make sure that a rational denominator has a corresponding numerator.

ospencer: > Couldn't you just make the numerator negative though if you wanted a negative number?

Of course, but we allow negative denominators for Number rational literals, so we should allow it here too.

ospencer: > Also with what you're suggesting how should an error be thrown in the case of `let a = 5r`? Unless I am missing something, it would seem that every single token would have to scanned individually to make sure that a rational denominator has a corresponding numerator.

In this case, we can provide a nice (well-formedness) error and say that this should be a fraction. You wouldn't really need to do anything extra in the parser.

It does though have the negative side effect of forcing us to add that to the parsetree.

An alternate proposal: OCaml just lexes numbers as numbers and accepts a range of suffixes, and carries the suffix with the token, to be interpreted by the typechecker. Maybe we should do that too—then what I'm suggesting makes far more sense. Ast_helper's `Expression.rational` then will accept an un-suffixed number or an `r`-suffixed number in the denominator, and the well-formedness check can see any number with an `r` suffix and can report that it should be a fraction.

Does that make sense?

alex-snezhko: If I'm understanding what you're saying correctly then this would be a pretty big change to the parser, right? Would it have any real advantages over just having the well-formedness error other than not having to have a `PExpRationalDenom` variant? Also what about just adding another lexer rule like
```
| (unsigned_int, '/', '-', unsigned_int, 'r')
```
to allow for negative denominators?

ospencer: Not particularly. I also thought about what you mentioned here... what worries me is operator precedence. 

For `Number`, writing `4 * 2 / 3` parses as `(4 * 2) / 3` because multiplication and division share precedence, and rational literals share that same division precedence. But if you were to write `4 * 2 / 3r` suddenly this parses as `4 * (2 / 3r)` because there's a token that was grabbed by the lexer. Of course, that doesn't work out with the typing, and if it did you'd get the same result. But that's only case if `/` is defined as division, and things can get wonky. So we really want them to work the same. Should we just _also_ lex Number rationals as a single token? Maybe. I'd really need to think about it though. In the operator_precedence doc we do mention that exponentiation has higher precedence than division, though not implemented. You'd likely expect `2 / 3 ** 3` to be `2 / 27`, but if we lex 2/3 as its own token, it'd be `(2/3) ** 3` which is not `2/27`.

alex-snezhko: Sorry, still a bit confused on some of the details you're proposing.
> lexes numbers as numbers and accepts a range of suffixes, and carries the suffix with the token, to be interpreted by the typechecker

I may be misunderstanding, but by this do you mean to say that the various `PConst...` for number types should be removed and the string representing the number should just contain the suffix in the token string like [in OCaml](https://github.com/ocaml/ocaml/blob/trunk/parsing/parsetree.mli#L26)? I'm not sure how this would help necessarily.

And then also to create a new well-formedness check that would reject instances of an `r`-suffixed token (since any valid uses of rational literals would have already been turned into rational number expressions)?

ospencer: That's what I was thinking, but now that I've thought about it more I don't think it's a great solution, or at least, it doesn't belong in this PR. I don't really like the solution of lexing a denominator token either.

I think it's fine to just ignore operator precedence for the sake of the Rational type, and you can lex it as a single token (just allowing negative numerators and/or denominators). If you wrote `2 * 3/4r` you definitely always mean `2 * (3/4r)` and never `(2 * 3)/4r`  syntax error.

ospencer (commented):
One last small comment. The only things I feel are missing from the initial library are `numerator` and `denominator`.

Things that can come later (and we can likely convert this comment to an issue) would be:
- `toIntegerRatio`, to get a tuple of numerator and denominator
- `fromIntegerRatio`, to construct a Rational from two integers
- `toProperFraction`, to get a tuple of an integer and a rational < 1
- `toFloat`
- `inverse`, to invert the rational
- `rationalize`, to get a simpler rational (maybe with an optional epsilon argument)—Python has `limit_denominator` which I maybe like a bit more, you essentially give it a target denominator and it gives you back a rational with a denominator of at most what you requested

ospencer:
@alex-snezhko could you rebase this?

alex-snezhko:
@ospencer rebased

ospencer (approved):
Thanks for putting this together 🤗

compiler/test/stdlib/rational.test.gr
```diff
@@ -0,0 +1,63 @@
+module RationalTest
+
+let (===) = (==)
+let (!==) = (!=)
```
phated, line 4: Looks like we need the same changes here as your other PR so you aren't shadowing `==` and `!=` in the entire file.

compiler/test/stdlib/rational.test.gr
```diff
@@ -0,0 +1,63 @@
...
+
+include "rational"
+from Rational use *
+
+assert fromNumber(1/2) == 1/2r
```
phated, line 9: Why no float tests?

phated (changes requested):
Just a couple changes needed for tests.

compiler/test/stdlib/rational.test.gr
alex-snezhko: It doesn't look like this functionality is currently supported, with only ints being coercible to rationals https://github.com/grain-lang/grain/blob/3907668ce1452cf83aca964cbc1b496472a38e46/stdlib/runtime/numbers.gr#L2437

I'm also not exactly sure how you would implement this; the majority of floats would result in some very nasty rational numbers

phated (approved):
Thanks for updating the tests. This lgtm. I've updated with main so we can get a new CI run since windows was failing.

PR #1604, opened 2023-01-16
fix(stdlib): Allow destructuring of TypedPath tuple

av8ta:
@alex-snezhko I was having some trouble destructuring TypedPath tuples due to type errors. 

I found by exporting: `export type TypedPath<a, b> = (TBase<a>, TFileType<b>, List<String>),` I'm able to destructure it fine now. Is it okay to expose this? Or were you wanting consumers to always interact with Path using provided functions?

```js
match (Path.fromString("./dir1/dir2/dir3")) {
  Path.RelativeFile(x) => {
    let (base, fileType, pathArray) = x           // type error here
    assert pathArray == ["dir3", "dir2", "dir1"]
  },
  _ => fail "not a relative file",
}
```

```
This expression has type Path.TypedPath<Path.Relative, Path.File>
but an expression was expected of type (a, b, c)
```

### DIFF
--- a/compiler/test/stdlib/path.test.gr
+++ b/compiler/test/stdlib/path.test.gr
@@ -269,3 +269,12 @@ assert Path.ancestry(fs("../dir1"), fs("./dir2")) == Ok(Path.NoLineage)
 assert Path.ancestry(fs("./dir1"), fs("../../dir2")) == Ok(Path.NoLineage)
 assert Path.ancestry(fs("./dir1"), fs("/dir2")) == Err(Path.DifferentBases)
 assert Path.ancestry(fs("C:/dir1"), fs("/dir2")) == Err(Path.DifferentRoots)
+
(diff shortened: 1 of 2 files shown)


phated:
This is not supposed to be exported.

alex-snezhko:
I was originally going to leave this hidden but I believe we arrived at exposing it as per [this discussion](https://github.com/grain-lang/grain/pull/1452#discussion_r1037574023)

phated:
The internal representation of the TypedPath should never be exposed. You can still pattern match on the Path type and alias the value to use with other functions in Path (for specialized behavior, etc).

This will be expanded once we have in-file modules so you can use specific APIs for specific types of paths.

av8ta:
That makes sense. 

> You can still pattern match on the Path type and ***alias the value*** to use

Thanks! That was the clue I needed :)

PR #1608, opened 2023-01-18
feat(compiler)!: Remove support of single-argument tuples

### LINKED ISSUE #1605: Remove & disallow single argument tuples
Single argument tuples in Grain don't really serve any purpose and lead to confusing syntax. I also can't find a valuable use case.

For example, they always need a trailing comma:
```grain
let a = (123): Number
let b = (123,): (Number,)
```

This is further confused by enum variants, which don't need a comma if they are single argument:
```grain
enum Foo {
  Bar(Number)
}
```

We should remove single argument tuples (when not used in an enum) from the language. This would also help on some oprint cleanup I'm trying to do.

phated:
Closes #1605

This PR removes single-argument (unary) tuples from Grain syntax. This is only removed at the parser because we still store Boxes as unary tuples. It also cleans up the formatter and oprint which had special logic to support differentiating between `(1)` and `(1,)` but leaves support in the `toString` stdlib function to support printing Boxes.


compiler/src/parsing/parser.messages
```diff
@@ -4156,7 +4156,7 @@ program: LET LPAREN WASMI64 COMMA WHILE
 ## pattern COMMA
 ##
 
-Expected a pattern or an immediate `)` for a single-argument tuple pattern.
+Expected at least another pattern because single-argument tuple patterns are not supported.
```
ospencer, line 4159: People aren't super familiar with single-argument tuples and most languages don't support them, so I don't think we need to mention them at all. This can actually just join the generic `Expected a pattern` message.

compiler/src/parsing/parser.messages
```diff
@@ -4430,7 +4430,7 @@ program: LPAREN WASMI64 COMMA WHEN
 ## expr COMMA
 ##
 
-Expected an expression or `)` for a single-argument tuple.
+Expected at least another pattern because single-argument tuple patterns are not supported.
```
ospencer, line 4433: And this can join the generic `Expected an expression` message.

compiler/src/parsing/parser.messages
```diff
@@ -5933,7 +5927,7 @@ program: WASMI64 COLON LPAREN UIDENT COMMA WHILE
 ## typ COMMA
 ##
 
-Expected a type or an immediate `)` to complete the tuple type.
```
phated, line 5936: I had missed this. How should I make sure I didn't miss others that maybe said "or an immediate `)`"?

compiler/test/suites/tuples.re
```diff
@@ -21,9 +21,14 @@ describe("tuples", ({test, testSkip}) => {
...
   );
-  assertSnapshot("singleton_tup", "(1,)");
-  assertSnapshot("singleton_tup_annotation", "(1,): (Number,)");
+  // Single-argument tuples are not supported at the parser, but `(1)` is parsed as a grouped value
   assertSnapshot("no_non_trailing_comma_singleton_tup", "(1)");
+  assertCompileError("singleton_tup_err", "(1,)", "Expected a pattern");
+  assertCompileError(
+    "singleton_tup_annotation_err",
+    "(1, 1): (Number,)",
+    "Expected another type in the tuple type",
```
phated, line 30: Do we usually test the error messages? That's how I caught my problem commented above

compiler/src/parsing/parser.messages
ospencer: There isn't really a great way; you can compile a list of all of the error messages and skim them.

compiler/test/suites/tuples.re
ospencer: For special cases like this, yes.

stdlib/runtime/string.gr
```diff
@@ -531,7 +531,8 @@ let rec heapValueToString = (ptr, extraIndents, toplevel) => {
         let rparen = ")"
         let mut strings = [rparen]
         if (tupleLength <= 1n) {
-          // Special case: unary tuple
+          // Special case: unary tuple, which is not valid Grain syntax; however, boxed values
+          // are stored as a unary tuple, so we keep this in case one gets printed somehow
```
ospencer, line 535: You can print a box and it prints as a unary tuple. We should make an issue to have real printing for boxes.

compiler/src/parsing/parser.messages
```diff
@@ -5933,7 +5927,7 @@ program: WASMI64 COLON LPAREN UIDENT COMMA WHILE
...
+Expected another type in the tuple type.
```
ospencer, line 5930: Would just move this to the generic `Expected a type` now since the special case to create a tuple type is gone, since if you type `(Number,` it's not yet clear whether you're writing a tuple type or going to go on and write a function type. Unfortunately it's not really obvious from the parser because the lexer injects special `FUN` tokens for functions, but it won't have done that unless you wrote the arrow already.

I guess we can write comments in the messages file, but I fear they'll get lost in the massive sea that's already there 😛 

That plus running `npm run compiler parser:update-errors` would also obliterate them.

stdlib/runtime/string.gr
phated: Is this message fine or do you want me to remove "somehow"?

compiler/src/parsing/parser.messages
phated: I searched for a few phrases that seemed comma, like "comma separated" and "tuple type" and they all still make sense.

marcusroberts (approved):
That cleans up the code nicely and removes a confusing language element

stdlib/runtime/string.gr
ospencer: Yeah, we can just drop the somehow.
```suggestion
          // Special case: unary tuple, which is not valid Grain syntax; however, boxed values
          // are stored as a unary tuple, so we keep this in case one gets printed
```

ospencer (approved):

PR #1611, opened 2023-01-19
fix(compiler): Properly handle parsing return of negative number

### LINKED ISSUE #1609: Return statement does not work on negative values
The following code:
```
let f = () => {
  return -1
}
```
results in a compiler error
```
Error: Syntax error after 'return' and before '-'.
Expected a newline character followed by more statements or `}` to complete the block.
```

ospencer:
Fixes #1609 

Since subtraction is left-associative and return statements are not required to have an expression, parsing `return -1` was causing the `return` itself to be treated as an expression, in the same way that typing `foo(5) - 4` would cause the parser to reduce `foo(5)` to a function call before attempting to parse subtraction.

Thanks @alex-snezhko for the report 🙂

### DIFF
--- a/compiler/src/parsing/parser.mly
+++ b/compiler/src/parsing/parser.mly
@@ -63,7 +63,7 @@ module Grain_parsing = struct end
 %left INFIX_110 DASH
 %left INFIX_120 STAR SLASH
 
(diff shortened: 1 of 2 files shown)


phated (approved):
Thanks for fixing this quickly!

PR #1614, opened 2023-01-22
chore(stdlib): Utilize early return in more of the stdlib

phated:
While rebasing #1493 and poking around in the stdlib, I've found there are more places where we should be using early return.

There are even more but I wanted to chunk my current work to keep the diff easy enough.


marcusroberts (approved):
Certainly improves the comprehensibility of the code!

spotandjake (approved):
Looks good

ospencer (approved):

PR #1615, opened 2023-01-22
chore(stdlib): Light cleanup on constant pattern matches

phated:
This cleans up a few `when` expressions we had on pattern matches in the stdlib that were just comparing to constant values. I assume these are from before we supported constant values in pattern matches.

### DIFF
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -526,7 +526,7 @@ export let part = (count, list) => {
  * @example List.rotate(-7, [1, 2, 3, 4, 5]) // [4, 5, 1, 2, 3]
  *
  * @since v0.1.0
- * 
+ *
  * @history v0.6.0: No longer throws if `count` outside list length bounds
  */
 export let rotate = (count, list) => {
@@ -671,7 +671,7 @@ export let rec drop = (count, list) => {
   } else {
     match ((count, list)) {
       (_, []) => [],
-      (n, _) when n == 0 => list,
+      (0, _) => list,
       (n, [first, ...rest]) => drop(n - 1, rest),
     }
   }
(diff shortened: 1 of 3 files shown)


mortax (approved):
i don't have any standing here but these all look like good,  appropriate clean-up.

spotandjake (approved):
These Changes all look good to me.

marcusroberts (approved):
Nice cleanup

compiler/test/stdlib/number.test.gr
```diff
@@ -630,10 +630,8 @@ assert Number.sign(-10000) == -1
 assert Number.sign(22222) == 1
 assert Number.sign(BI.toNumber(-1t)) == -1
 assert Number.sign(BI.toNumber(1t)) == 1
-// TODO(#693): Replace with Infinity literal when it exists
-assert 1 / Number.sign(0.0) == Float64.toNumber(Float64.infinity)
-// TODO(#693): Replace with -Infinity literal when it exists
-assert 1 / Number.sign(-0.0) == Number.neg(Float64.toNumber(Float64.infinity))
+assert 1 / Number.sign(0.0) == Infinity
+assert 1 / Number.sign(-0.0) == -Infinity
```
ospencer, line 634: There's a bunch more in this file that would be great to use the literal (obviously not the tests _for_ `Number.infinity`). It's essentially all of the `Number.infinity` and `Number.neg(Number.infinity)`.

spotandjake: I was going to mention that. It probably also makes sense to have `Number.infinity` and `Number.nan` use the literals internally rather than `0.0 / 0.0` and `1.0 / 0.0`.

phated: It's almost as if these should have been done with the feature 😉  Anyway, I'm working on updating the file now.

phated: @ospencer this has been removed from this PR because I'm going to remove `infinity` and `NaN` constants in the number module.

ospencer (approved):

Issue #1566, opened 2023-01-01
Make it easier to `String.slice` to the end of a string


spotandjake:
currently if i wanted to do something similar to `"test".slice(1)` in js to get `"est"`, In Grain the only way to currently do this would be `String.slice(1, String.length("test"), "test") == "est"` Negative indexing feels like a fix to this but `String.slice(1, -1, "test") == "es"`, My solutiin to this was `String.slice(0, -0, "test") == "est"` but after asking @ospencer about this in the discord it was noted that we should never ever use `-0`, One possible solution to this would be to make `String.slice` use ranges then we could have inclusive or exclusive ranges but this is somewhat blocked by #1404 and would be a breaking change, another possible nonbreaking solution would be adding a `String.sliceIndex(<splitIndex>, "string")` but that would be less flexible and might get confusing.

TLDR: currently slicing from an index to the end of a string is extremly verbose, it would be nice if it was made simplier.

alex-snezhko:
If ranges were to be used instead then wouldn't you still have the same problem of not being able to use a shortcut? For example you'd still have to call `String.length` for the upper bound of the range

spotandjake:
Grain has `Inclusive` And `Exclusive` ranges. so you could use an inclusive range for that behaviour.

alex-snezhko:
Yes but you'd still have to call `String.length` wouldn't you? It'd either be `Range.Inclusive(0, String.length(s) - 1)` or `Range.Exclusive(0, String.length(s))`

ospencer:
Yeah, we haven't really talked about negative indexing on ranges.

alex-snezhko:
I suppose a hacky alternative would be to allow `Infinity` as the end index although that doesn't seem like a great compromise

phated:
This isn't a problem once we have optional arguments, so I'd vote that we don't "solve" it until then. Once optional arguments exist, the "end" argument will be optional

PR #1616, opened 2023-01-22
feat(compiler)!: Remove Exclusive/Inclusive Ranges, provide as record via compiler

### LINKED ISSUE #1404: Range enum should be provided by compiler
We've found the need to use the `Range` enum in the number library. It seems really heavy-handed to include the Range module for the data type.

We should move the `Range` enum to Pervasives and re-export it from the Range module (for backwards compatibility).

Eventually, `Range` could be a data type implemented inside the compiler and maybe even receive syntax.

phated:
Closes #1404 

This adds the Range enum as a language-supplied type, which is a prerequisite to having language syntax for ranges.

The new Range is a `record Range<a> { rangeStart: a, rangeEnd: a }`. It is now up to APIs to decide if the range is interpreted as Inclusive or Exclusive, and is no longer limited to only Numbers, which also allow APIs to decide how to interpret the types.


alex-snezhko:
I think if `Exclusive` ranges were to be kept then special syntax should be added for them, otherwise I think it'd be better to remove them altogether as it's awkward to have syntax for one way but not the other (also doing `end - 1` for exclusive ranges isn't too big of a deal at the end of the day).

Also, thoughts on adding unbounded ranges like `3..`? I find them pretty useful in other languages that have them for slicing and whatnot (this would also allow the various `slice` functions to take a `Range` rather than a second optional argument as discussed in [#1566](https://github.com/grain-lang/grain/issues/1566#issuecomment-1369090465) which seems nicer imo). `Range.map` could then be replaced with a generator function instead of returning a list to accommodate for not being able to create an infinite list (or an exception could just be thrown in this case :shrug: )

phated:
I've updated this PR to turn `Range` into `record Range<a> { range_start: a, range_end: a }` as we discussed on the community call. It is now up to APIs to decide if the range is interpreted as Inclusive or Exclusive, and is no longer limited to only Numbers, which also allow APIs to decide how to interpret the types.

alex-snezhko:
Should the casing of the record fields be changed to camelcase to match the rest of the language?

phated:
Probably, but I hate it. 🤣

alex-snezhko:
What about just `start` and `end`? Since `Range` will become a builtin type then maybe the `range` prefix can just be omitted by the same logic that `Some` is not called `OptionSome`, `Ok` is not called `ResultOk`, etc?

phated:
We discussed `start` and `end` but it's not a good idea to use commonly used words in records provided by the compiler because they will result in ambiguous record warnings when people write custom records with those common words.

ospencer:
Thoughts on `Inclusive` and `Exclusive` submodules (or just an inclusive submodule)? I feel like it'd be pretty nice to be able to do `Range.Inclusive.inRange()` in those times I want an inclusive range.

phated:
>  just an inclusive submodule

I like the idea. I suppose I need to go back and get that implementation 😛

compiler/src/typed/builtin_types.re
```diff
@@ -95,6 +99,8 @@ and type_option = var =>
 and type_result = (ok, err) =>
   newgenty(TTyConstr(path_result, [ok, err], ref(TMemNil)))
 and type_list = var => newgenty(TTyConstr(path_list, [var], ref(TMemNil)))
+and type_range = var =>
+  newgenty(TTyRecord([("rangeStart", var), ("rangeEnd", var)]))
```
ospencer, line 103: It's insane for me to suggest this right?
```suggestion
  newgenty(TTyRecord([(Ident.name(ident_range_start), var), (Ident.name(ident_range_end), var)]))
```

stdlib/range.gr
```diff
@@ -126,3 +111,106 @@ provide let map = (fn, range) => {
...
+   * @param range: The range to check within
+   * @returns Whether or not the value is within range
+   *
+   * @example Range.Inclusive.inRange(1, { rangeStart: 0, rangeEnd: 1 }) == true
+   * @example Range.Inclusive.inRange(10, { rangeStart: 0, rangeEnd: 2 }) == false
+   *
+   * @since v0.6.0
+   * @history v0.3.0: Root APIs originally handled Inclusive & Exclusive variants
+   */
+
```
ospencer, line 129: Can the docs touch the implementations or is this a formatter bug?

ospencer (commented):
This looks awesome!

stdlib/range.gr
phated: Formatter bug. We're not blocking on it though. Hopefully it'll be fixed soon.

compiler/src/typed/builtin_types.re
phated: I don't know the difference. I thought this type took a string?

phated: Oh, I read that wrong, it's pulling the string name from the ident

ospencer (approved):
Excellent!

PR #1618, opened 2023-01-24
chore(stdlib)!: Replace Number.nan & Number.infinity constants with keywords

phated:
This removes the `Number.nan` and `Number.infinity` constants and uses `Infinity` and `NaN` keywords. Ideally, this sort of changes will happen with the feature so we aren't adding 2 different breaking changes for it in the changelog.

### DIFF
--- a/compiler/test/stdlib/number.test.gr
+++ b/compiler/test/stdlib/number.test.gr
@@ -2,14 +2,9 @@ import Number from "number"
 import Result from "result"
 import Int32 from "int32"
 import Int64 from "int64"
-import Float32 from "float32"
-import Float64 from "float64"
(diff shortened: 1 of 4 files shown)


ospencer (commented):
I didn't think we needed to get rid of the constants in the library, but I guess they don't serve much of a purpose if we have the literals 🤔

phated:
@ospencer I was rewriting the tests and found that there was a ridiculous matrix of testing variations on using the keywords and constants. It doesn't make sense to have more than one way to access `Infinity` and `NaN` so I removed them and updated everything instead.

ospencer (approved):
Looks good, just needs a Graindoc update.

PR #1623, opened 2023-01-29
feat(compiler)!: Labeled and default arguments

### LINKED ISSUE #388: [Lang] Named (Optional) arguments

ospencer:
Closes #388 

Draft because this needs tests, code cleanup, parser messages updates, and as always, because things may still change 🙂 

# Optional and Labeled Arguments

This PR introduces the ability to provide default values for function arguments and optionally omit those arguments when calling the function. Additionally, this PR allows regular arguments to be supplied by name if desired. If you are familiar with how "keyword arguments" work in Python, you'll find this to be fairly similar.

## Declaring an argument as optional

For an argument to be omitted by the caller, a default value must be provided.

```
let rec join = (separator=" ", strings) => {
  match (strings) {
    [] => "",
    [string] => string,
    [hd, ...tl] => hd ++ separator ++ join(separator=separator, tl)
  }
}
```

The `join` function here can be called without the `separator` argument, and the default value will be used instead:

```
join(["foo", "bar", "baz"]) // "foo bar baz"
```

We can optionally provide our own separator:

```
join(separator=", ", ["foo", "bar", "baz"]) // "foo, bar, baz"
```

And we can provide the `separator` argument in any position:

```
join(["foo", "bar", "baz"], separator=", ") // "foo, bar, baz"
```

## Type signatures

The type signature of `join` is `(?separator: String, strings: List<String>) -> String`. The `?` before `separator` indicates that this parameter is optional. The name of the parameter `strings` is also included in the type signature.

Because the parameter name `strings` is also included in the type signature, we can also provide that parameter with its label:

```
join(separator=", ", strings=["foo", "bar", "baz"]) // "foo, bar, baz"
```

And of course, we can provide these in whatever order we like:

```
join(strings=["foo", "bar", "baz"], separator=", ") // "foo, bar, baz"
```

## Typechecking

Two function types are considered equal if, with labels removed, the types are equal. This is to allow function authors to name their function parameters whatever they like and still be able to pass them as arguments to other functions. For example, even if `List.map` had type `(fn: (value: a) -> b, list: List<a>) -> List<b>`, your mapper function would not need to call its argument `value`.

A function that accepts an optional argument does not typecheck with a function that does not accept that optional argument, even if all other types are equal.

If a function type is inferred by arguments that are supplied positionally, they are unlabeled.

### DIFF
--- a/compiler/src/typed/typecore.re
+++ b/compiler/src/typed/typecore.re
@@ -2794,7 +2794,7 @@ let report_error = (env, ppf) =>
       | labels =>
         fprintf(
           ppf,
-          "@ @[Did you mean to supply an optional argument with label %a?@]@]",
+          "@ @[Did you mean to supply an argument with label %a?@]@]",
           oxford,
           unused_optional_arguments,
         )
@@ -3008,17 +3008,20 @@ let report_error = (env, ppf) =>
       if (is_optional(got)) {
         fprintf(
           ppf,
-          "This function contains the optional, default argument %s@ ",
+          "This function contains the argument %s@ ",
           qualified_label_name(got),
         );
       } else {
         fprintf(
           ppf,
-          "The expected function type contains the optional, default argument %s@ ",
+          "The expected function type contains the argument %s@ ",
           qualified_label_name(expected),
         );
       };
-      fprintf(ppf, "but the matching argument is not optional.@ ");
+      fprintf(
+        ppf,
+        "which has a default value, but the matching argument does not.@ ",
+      );
       fprintf(
         ppf,
         "The expected type is@ %a%t",
--- a/compiler/src/typed/typecore.re
+++ b/compiler/src/typed/typecore.re
@@ -1016,7 +1016,7 @@ and type_expect_ =
         (arg, (args, labels, prelude)) => {
           switch (arg.pla_default) {
           | Some(default) =>
-            let default_value_name = mknoloc("<default_value>");
+            let default_value_name = mknoloc("$default_value");
             let default_loc = default.pexp_loc;
             let scases = [
               MatchBranch.mk(
--- a/compiler/src/typed/typecore.re
+++ b/compiler/src/typed/typecore.re
@@ -2676,7 +2676,7 @@ let report_error = (env, ppf) =>
   | Constructor_arity_mismatch(lid, expected, provided) =>
     fprintf(
       ppf,
-      "@[The constructor %a@ expects %i argument(s),@ but is applied here to %i argument(s)@]",
+      "@[The constructor %a@ expects %i argument(s),@ but is called with %i argument(s)@]",
       identifier,
       lid,
       expected,
@@ -2756,13 +2756,13 @@ let report_error = (env, ppf) =>
         "@[<v>@[<2>This expression has type@ %a@]@ %s@]",
         type_expr,
         typ,
-        "This is not a function; it cannot be applied.",
+        "This is not a function; it cannot be called.",
       );
     }
   | Apply_too_many_arguments(typ, unused_tyargs) => {
       reset_and_mark_loops(typ);
       fprintf(ppf, "@[<v>@[<2>This function has type@ %a@]", type_expr, typ);
-      fprintf(ppf, "@ @[It is applied to too many arguments.@]@]");
+      fprintf(ppf, "@ @[It is called with too many arguments.@]@]");
       let unused_optional_arguments =
         List.filter_map(
           ((l, _)) =>
@@ -2838,7 +2838,7 @@ let report_error = (env, ppf) =>
       };
     }
   | Apply_unknown_label(label, valid_labels) => {
-      fprintf(ppf, "This argument cannot be applied with label %s.", label);
+      fprintf(ppf, "This argument cannot be supplied with label %s.", label);
       spellcheck(ppf, label, valid_labels);
     }
   | Label_multiply_defined(s) =>
@@ -3062,7 +3062,7 @@ let report_error = (env, ppf) =>
   | Not_a_variant_type(lid) =>
     fprintf(ppf, "The type %a@ is not a variant type", identifier, lid)
   | Incoherent_label_order => {
-      fprintf(ppf, "This function is applied to arguments@ ");
+      fprintf(ppf, "This function is called with arguments@ ");
       fprintf(ppf, "in an order different from other calls.@ ");
       fprintf(ppf, "This is only allowed when the real type is known.");
     }
--- a/compiler/src/typed/typecore.re
+++ b/compiler/src/typed/typecore.re
@@ -11,7 +11,7 @@ open Typepat;
 open Disambiguation;
 
 type error =
-  | Arity_mismatch(type_expr, int)
+  | Arity_mismatch(type_expr, option(type_forcing_context))
   | Polymorphic_label(Identifier.t)
   | Constructor_arity_mismatch(Identifier.t, int, int)
   | Label_mismatch(Identifier.t, list((type_expr, type_expr)))
@@ -24,6 +24,9 @@ type error =
       option(type_forcing_context),
     )
   | Apply_non_function(type_expr)
+  | Apply_too_many_arguments(type_expr, list((argument_label, type_expr)))
+  | Apply_too_few_arguments(list((argument_label, type_expr)))
+  | Apply_unknown_label(string, list(string))
   | Label_multiply_defined(string)
   | Label_missing(list(Ident.t))
   | Label_not_mutable(Identifier.t)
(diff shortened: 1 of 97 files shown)


alex-snezhko:
This is awesome! Apologies in advance for the syntax bikeshedding :sweat_smile:, but personally I find the
```
let fn = (label reference) => ...
```
syntax where the label comes first a bit unclear; if I were reading a function defined like this I feel I would very likely have to double-take to remember which one is the label and which is the reference (I even had to double-take to make sure I didn't mix them up after writing the example above :laughing:). In the community meeting I believe you brought up doing the `label as reference` syntax, which in my opinion is clearer since `as` in the context of pattern matching assigns the pattern on the left side of the `as` to the binding on the right side. Is there a reason you decided against `as`?

ospencer:
Yeah, that's totally fair, this syntax comes from Swift.

The thing that sucks about using `as` here is that it works completely differently than every other place in the language. When importing a module and you want to rename it, you use `as`. When you want to change the name of a value on import, you use `as`. When you want to create a binding while writing a pattern, you use `as`. When you want to change the name of a value when you export, you use `as`.

But in this one case, you'd be using `as` to provide a destructuring pattern, which is just very very different than all of the cases I listed above. It gets worse when you see them next to each other:

```
let foo = (pair as (fst, snd) as myTuple) => ...
```

That first `as` says that I'm providing a pattern for this label, whereas the second says I'm providing an alias for my pattern.

I don't think the Swift syntax is perfect, but I do think it's better than overloading `as`.

Certainly still open to suggestions for syntax 😄

ospencer:
Something else that I thought of that I probably shouldn't bring up due to the sheer level of bikeshedding this could cause:

Some time ago we decided that we were going to use double colon `::` for type annotations instead of single colon. If we did, we could still use single colon for argument labels (like we still would for records).

The syntax could then be:
```
let foo = (label: pattern) => ...
```

And of course still just
```
let foo = (label) => ...
```
for short.

ospencer:
Which would be separate from
```
let foo = (label :: Type) => ...
```
and distinguish
```
let foo = (label: pattern :: Type) => ...
```

alex-snezhko:
I see, assuming that explicit typing is changed to `::` then that would be a nicer syntax imo. If that is done then maybe applying arguments should also be done with a colon?
```
let fn = (label: pattern) => ...
fn(label: 1)
```

ospencer:
@alex-snezhko Decided to drop the requirement that labels be required, so it avoids the whole issue entirely. Argument labels are just inferred, and this covers 95% of cases just fine. The only time it's less than ideal is when an argument label is unable to be inferred (such as doing a destructure) in a module signature, but you can do `(fst, snd) as pair` and `pair` becomes the label (it's just not _required_ that the user do this). Some tooling (like Graindoc and the lsp) will likely warn that it's good practice to have all provided values have explicit labels.

Doing it this way keeps the language complexity down tremendously, and I believe now this is a non-breaking change.

alex-snezhko:
Awesome!

compiler/src/typed/typecore.re
```diff
@@ -2690,29 +2756,96 @@ let report_error = (env, ppf) =>
...
+        "@[<v>@[<2>This expression has type@ %a@]@ %s@]",
+        type_expr,
+        typ,
+        "This is not a function; it cannot be applied.",
+      );
+    }
+  | Apply_too_many_arguments(typ, unused_tyargs) => {
+      reset_and_mark_loops(typ);
+      fprintf(ppf, "@[<v>@[<2>This function has type@ %a@]", type_expr, typ);
+      fprintf(ppf, "@ @[It is applied to too many arguments.@]@]");
```
phated, line 2770:
```suggestion
      fprintf(ppf, "@ @[It is called with too many arguments.@]@]");
```

compiler/src/typed/typecore.re
```diff
@@ -2690,29 +2756,96 @@ let report_error = (env, ppf) =>
...
   | Apply_non_function(typ) => {
       reset_and_mark_loops(typ);
-      switch (repr(typ).desc) {
-      | TTyArrow(_) =>
+      fprintf(
+        ppf,
+        "@[<v>@[<2>This expression has type@ %a@]@ %s@]",
+        type_expr,
+        typ,
+        "This is not a function; it cannot be applied.",
```
phated, line 2764:
```suggestion
        "This is not a function; it cannot be called.",
```

compiler/src/typed/typecore.re
```diff
@@ -2690,29 +2756,96 @@ let report_error = (env, ppf) =>
...
+      switch (unused_optional_arguments) {
+      | [] => ()
+      | labels =>
         fprintf(
           ppf,
-          "@[<v>@[<2>This function has type@ %a@]",
-          type_expr,
-          typ,
-        );
+          "@ @[Did you mean to supply an optional argument with label %a?@]@]",
```
phated, line 2802:
```suggestion
          "@ @[Did you mean to supply an argument with label %a?@]@]",
```
I don't think this is correct but I'm trying to not use "optional" terminology, just "default" and "labeled", if we can.

compiler/src/typed/typecore.re
```diff
@@ -2690,29 +2756,96 @@ let report_error = (env, ppf) =>
...
-          typ,
-          "This is not a function; it cannot be applied.",
+          "@[<hov>This function call is missing arguments of type %a@]",
+          print_args,
+          args,
         )
       };
     }
+  | Apply_unknown_label(label, valid_labels) => {
+      fprintf(ppf, "This argument cannot be applied with label %s.", label);
```
phated, line 2846:
```suggestion
      fprintf(ppf, "This argument cannot be called with label %s.", label);
```

compiler/src/typed/typecore.re
```diff
@@ -2866,27 +2997,40 @@ let report_error = (env, ppf) =>
...
+  | Function_label_mismatch({got, expected, expected_type, explanation}) => {
+      reset_and_mark_loops(expected_type);
+      if (is_optional(got)) {
         fprintf(
           ppf,
-          "it should have type@ %a%t",
-          type_expr,
-          ty,
-          report_type_expected_explanation_opt(explanation),
+          "This function contains the optional, default argument %s@ ",
```
phated, line 3016:
```suggestion
          "This function contains the labeled, defaulted argument %s@ ",
```

compiler/src/typed/typecore.re
```diff
@@ -2866,27 +2997,40 @@ let report_error = (env, ppf) =>
...
         );
       } else {
-        fprintf(ppf, "This expression should not be a function,@ ");
         fprintf(
           ppf,
-          "the expected type is@ %a%t",
-          type_expr,
-          ty,
-          report_type_expected_explanation_opt(explanation),
+          "The expected function type contains the optional, default argument %s@ ",
```
phated, line 3022:
```suggestion
          "The expected function type contains the labeled, defaulted argument %s@ ",
```

compiler/src/typed/typecore.re
```diff
@@ -2866,27 +2997,40 @@ let report_error = (env, ppf) =>
...
+          qualified_label_name(expected),
         );
       };
+      fprintf(ppf, "but the matching argument is not optional.@ ");
```
phated, line 3026:
```suggestion
      fprintf(ppf, "but the matching argument is not defaulted.@ ");
```

compiler/test/suites/functions.re
```diff
@@ -206,4 +214,136 @@ truc()|},
...
+    "12\n",
+  );
+  assertRun(
+    "default_args5",
+    {|
+      let concat = (a="1", b) => a ++ b
+      print(concat(b="2"))
+    |},
+    "12\n",
+  );
```
phated, line 298:
```suggestion
  );
  assertRun(
    "default_args6",
    {|
      let concat = (a=b ++ b, b) => a ++ b
      print(concat(b="9"))
    |},
    "999\n",
  );
```

phated (changes requested):
A couple questions and some cleanup requests. I also made some suggestions on the error messages because I'd prefer to not use `optional` in the error messages.

compiler/src/typed/typecore.re
spotandjake: I think `but the matching argument does not have a default value` would be much better then using the word `defaulted` I dont think `optional` worked terribly though but that would require us making it clear that the only way to get an optional argument is by having a default value.

compiler/src/typed/typecore.re
```diff
@@ -1147,6 +1003,81 @@ and type_expect_ =
...
+    let gen_opt = () => {
+      incr(opt_counter);
+      "$option_" ++ string_of_int(opt_counter^);
+    };
+    let (args, labels, prelude) =
+      List.fold_right(
+        (arg, (args, labels, prelude)) => {
+          switch (arg.pla_default) {
+          | Some(default) =>
+            let default_value_name = mknoloc("<default_value>");
```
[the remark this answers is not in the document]
ospencer: It's a binding name for the default value the user provided, but it won't show up to a user. We're in the process of typechecking, so it'll only show up if we showed it to the user from this block.

compiler/src/typed/typecore.re
ospencer: The function is the thing being called, not the argument. Will reword:
```suggestion
      fprintf(ppf, "This argument cannot be supplied with label %s.", label);
```

compiler/src/typed/typecore.re
ospencer: Not a huge fan of `defaulted`. `labeled` here also doesn't feel great because everything is implicitly labeled. I think we should just say "has a default value."

phated (approved):
Let's GO!

PR #1638, opened 2023-01-30
feat(compiler): Suppress warnings about using fromNumber for Int32/64 and Float32/64 tests

ospencer:
With the addition of the module system, the typechecker is now smarter and understands that the `fromNumber` that comes from the `Int32` library is the same as doing `Int32.fromNumber`, so warnings are now produced by those tests. This change rebinds `fromNumber` to suppress the warning while we test the functionality.


phated (approved):
Smart enough to know they are the same in the stdlib but not if we reassign to a binding? 🤣

ospencer:
Yeah 😛

PR #1639, opened 2023-01-30
chore(compiler): Don't arbitrarily import List in stdlib tests

### LINKED ISSUE #1631: Don't import list library in stdlib tests
It's unclear why we do this; I think it's just really old.

https://github.com/grain-lang/grain/pull/1584#discussion_r1090054965

ospencer:
Closes #1631

### DIFF
--- a/compiler/test/__snapshots__/stdlib.179d20b9.0.snapshot
+++ b/compiler/test/__snapshots__/stdlib.179d20b9.0.snapshot
@@ -45,5 +45,5 @@ stdlib › stdlib_equal_4
    (call $_gmain)
   )
  )
- ;; custom section \"cmi\", size 280
+ ;; custom section \"cmi\", size 236
 )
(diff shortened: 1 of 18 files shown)


phated (changes requested):
You mentioned that we should be testing this in a different way. Did you mean that it's already being tested in a different way or should we be adding some specific tests to the suite while removing these?

ospencer:
It's the same general problem as many of the snapshots for basic tests that don't produce any output; the snapshots show that the input program is optimized away. This is something that we need to think about and validate that we're testing what we mean to be testing, but it's unrelated to this change.

ospencer:
I reviewed these snapshots, and the call to `==` isn't being optimized away like I thought it would be, so these snapshots are largely showing the right thing. There are a few tests that are testing `is` though that are completely optimized away.

ospencer:
Turns out `==` isn't actually a pure function because of how it checks for cycles. Rough.

phated:
> It's the same general problem as many of the snapshots for basic tests that don't produce any output; the snapshots show that the input program is optimized away. This is something that we need to think about and validate that we're testing what we mean to be testing, but it's unrelated to this change.

Do we need an issue opened?

ospencer:
https://github.com/grain-lang/grain/issues/1646

phated (approved):
Thanks!

PR #1641, opened 2023-01-30
feat(compiler): Allow a newline character after 'as' when aliasing

### LINKED ISSUE #1636: Parser: allow line breaks when aliasing
A line break should be allowed after `as` in all cases.

ospencer:
Closes #1636


phated (approved):
Thanks!

PR #1645, opened 2023-01-31
hacking on comment line breaks

phated:
This is a WIP and massive change to the way we do comments. I spent 5 hours on this last night and there are still some bugs, so this needs to be a draft.

Basically, what I've found is that whether any comment should have a hardline or a space or anything seems to be context dependent and we shouldn't be arbitrarily be jamming those onto every `comment_to_doc` we do...and why oh why was there a `nobreak_comment_to_doc` 😭

There seems to be opportunity for a lot more refactor in the comment code but I'm not sure when I'll have a big chunk of time so I wanted to open this and get help from @ospencer and @marcusroberts ❤️

### DIFF
--- a/compiler/src/formatting/comment_utils.re
+++ b/compiler/src/formatting/comment_utils.re
@@ -265,22 +265,10 @@ let get_comments_to_end_of_line =
 
 let comment_to_doc = (comment: Parsetree.comment) => {
   let comment_string = Comments.get_comment_source(comment);
-  let newline =
-    switch (comment) {
-    | Line(_)
-    | Shebang(_) => Doc.hardLine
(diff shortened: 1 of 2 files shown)


marcusroberts:
Glad to see this as comments are the most problematic area of the formatter.  If only they were in the AST!

phated:
@marcusroberts I wouldn't say they are the most problematic area, just that they are impossible to fix when they are broken with the current design. Being in the AST still doesn't solve all the problems and probably need to have a separate data structure that can be queried instead of storing in a list. but idk.

phated:
Closing this since we're preferring the new formatter in #1810

Issue #1621, opened 2023-01-26
Graindoc: Support new module system


phated:
The new module system is being added in #1584 but it doesn't consider how graindoc should support it. Graindoc needs to be adapted to:
1. Drop the `@module` annotation and just use docblocks attached to  `module` definitions. This also means we shouldn't allow the name to be overridden anymore.
2. Figure out what to generate for inline modules. This also means supporting multiple module docblocks in a file.
3. TBD

PR #1647, opened 2023-02-01
feat(graindoc)!: Replace module attribute with docblock on module header

phated:
This removes our custom `@module` attribute in a docblock and instead relies on a docblock attached to the module header (hooray module anchor points!).

Also, I updated the stdlib to avoid using `@module`, which caught one problem with the `Bigint` module header which generally should be `BigInt` and the docs were different.

One benefit you'll noticed from the regenerated docs is that the docs produced for runtime are more accurate because we have accurate module naming in the files.

Ref #1621 (still plenty to do before it is done).

### DIFF
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -98,12 +98,6 @@ let location_for_ident = (~exports, ident) => {
   snd(Ident.find_name(Ident.name(ident), exports));
 };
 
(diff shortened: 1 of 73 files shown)


ospencer (approved):
I wouldn't say that bigint was wrong before, but that we hadn't officially decided whether it was called `Bigint` or `BigInt`, though I guess we gave it a name in the docs. We should consider updating the casing of the filename.

Excellent work here!

Issue #1644, opened 2023-01-31
Grainformat: remove extraneous newline for function comments in submodules


alex-snezhko:
Formatting the following code
```
module Test

module Inner {
  /**
   * Do stuff
   */
  provide let fn = () => 1
}
```
Produces an extra newline after the function comment when it probably shouldn't
```
module Test

module Inner {
  /**
   * Do stuff
   */

  provide let fn = () => 1
}
```

PR #1652, opened 2023-02-02
chore(stdlib)!: Group mutable/immutable data structures

alex-snezhko:
Groups immutable and mutable versions of the same abstract data types into the same module, with the immutable versions accessible via the `Immutable` submodule.

Draft because this is still blocked on #1621 and #1644


stdlib/array.gr
```diff
@@ -985,3 +988,959 @@ provide let rotate = (n, arr) => {
...
+    length: Number,
+    shift: Number,
+    root: Tree<a>,
+    tail: Array<a>,
+  }
+
+  /**
+   * An empty array.
+   * 
+   * @since v0.6.0
```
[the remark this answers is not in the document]
alex-snezhko: I ended up putting the `@history` tag on the comment for the entire submodule to act as a "catch-all" for all of the functions in the submodule. If you think that is adequate then I think the functions inside would be fine without history.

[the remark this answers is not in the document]
phated: When I'm looking for stuff in the Node.js docs, I jump straight to the thing I care about using the sidebar/index and then expand the history of that API. So I think we should opt to be more complete at the cost of being more verbose in our docblocks.

compiler/test/stdlib/array.test.gr
```diff
@@ -1,7 +1,10 @@
 module ArrayTest
 
+include "list"
 include "array"
+from Array use { Immutable as ImmArray }
```
ospencer, line 5: I kinda like what @phated did to test some other submodules. He made a submodule in the test file to test the submodule, and did all of the assertions in there. What do you think about following that pattern here?

alex-snezhko: Sure, that sounds good

stdlib/array.gr
```diff
@@ -1,5 +1,5 @@
 /**
- * Utilities for working with arrays.
+ * Utilities for working with arrays. An immutable array implementation is available in the `Immutable` submodule
```
phated, line 2:
```suggestion
 * Utilities for working with arrays.
 * 
 * An immutable array implementation is available in the `Immutable` submodule.
```

stdlib/map.gr
```diff
@@ -1,6 +1,5 @@
 /**
- * A Map holds key-value pairs. Any value may be used as a key or value. Operations on a Map mutate the internal state, so it never needs to be re-assigned.
- *
+ * A Map holds key-value pairs. Any value may be used as a key or value. Operations on a Map mutate the internal state, so it never needs to be re-assigned. An immutable map implementation is available in the `Immutable` submodule
```
phated, line 2:
```suggestion
 * A Map holds key-value pairs. Any value may be used as a key or value. Operations on a Map mutate the internal state, so it never needs to be re-assigned.
 *
 * An immutable map implementation is available in the `Immutable` submodule.
```

stdlib/map.gr
```diff
@@ -505,3 +505,526 @@ provide let reject = (predicate, map) => {
...
+  }
+
+  /**
+   * Provides the count of key-value pairs stored within the map.
+   *
+   * @param map: The map to inspect
+   * @returns The count of key-value pairs in the map
+   *
+   * @since v0.6.0
+   * @history v0.5.4: Originally in `"immutableset"` module
```
phated, line 559:
```suggestion
   * @history v0.5.4: Originally in `"immutablemap"` module
```

stdlib/map.gr
```diff
@@ -505,3 +505,526 @@ provide let reject = (predicate, map) => {
...
+    }
+  }
+
+  /**
+   * Determines if the map contains no key-value pairs.
+   *
+   * @param map: The map to inspect
+   * @returns `true` if the given map is empty or `false` otherwise
+   *
+   * @since v0.6.0
```
phated, line 575:
```suggestion
   * @since v0.6.0
   * @history v0.5.4: Originally in `"immutablemap"` module
```

stdlib/map.gr
```diff
@@ -505,3 +505,526 @@ provide let reject = (predicate, map) => {
...
+
+  // semi-arbitrary value chosen for algorithm for determining when to balance
+  // trees; no tree can have a left subtree containing this number of times
+  // more elements than its right subtree or vice versa
+  let weight = 4
+
+  /**
+   * An empty map
+   *
+   * @since v0.6.0
```
phated, line 538:
```suggestion
   * @since v0.6.0
   * @history v0.5.4: Originally in `"immutablemap"` module
```

stdlib/priorityqueue.gr
```diff
@@ -1,5 +1,5 @@
 /**
- * A mutable priority queue implementation. A priority queue is a data structure that maintains elements in a priority order. Elements with higher priority are served before elements with lower priority when extracting from the priority queue.
+ * A priority queue is a data structure that maintains elements in a priority order. Elements with higher priority are served before elements with lower priority when extracting from the priority queue. An immutable priority queue implementation is available in the `Immutable` submodule.
```
phated, line 2:
```suggestion
 * A priority queue is a data structure that maintains elements in a priority order. Elements with higher priority are served before elements with lower priority when extracting from the priority queue.
 *
 * An immutable priority queue implementation is available in the `Immutable` submodule.
```

stdlib/set.gr
```diff
@@ -1,6 +1,5 @@
 /**
- * A Set is an unordered collection of unique values. Operations on a Set mutate the internal state, so it never needs to be re-assigned.
- *
+ * A Set is an unordered collection of unique values. Operations on a Set mutate the internal state, so it never needs to be re-assigned. An immutable set implementation is available in the `Immutable` submodule.
```
phated, line 2:
```suggestion
 * A Set is an unordered collection of unique values. Operations on a Set mutate the internal state, so it never needs to be re-assigned.
 *
 * An immutable set implementation is available in the `Immutable` submodule.
```

phated (commented):
A few changes that I'm going to accept then regen the docs and do any formatting so we can land this.

phated (approved):
Thanks!

PR #1656, opened 2023-02-05
feat!: Rework preopened directories

### LINKED ISSUE #1205: Refactor WASI preopened files
I realized that in C/Rust you can casually use the `open` system call and things just... work. That seemed surprising to me, since all WASI gives us is `path_open`, similar to `openat`, not `open`. I thought `open` was impossible with WASI. That said, I did some digging into wasi-libc's implementation of `open`. It's a couple layers deep, but you can find the important bit here: https://github.com/WebAssembly/wasi-libc/blob/659ff414560721b1660a19685110e484a081c3d4/libc-bottom-half/sources/preopens.c#L212

I don't know if it's documented anywhere, but I've been under the assumption (and it seems further documented in code here) that preopened directories are numbered starting at descriptor 3 and going up from there, but descriptors obtained after the preopened ones are random. wasi-libc just scans descriptors until it hits a bad one to find all of the preopens, and then when `open` is called it tries to open files relative to all of the preopened dirs.

We don't yet have an implementation of `open` (that would probably exist in higher-level file API) but `sys/file` provides `cwdfd` which points at where the first preopened dir would be. The Grain CLI always preopens the current directory, but for other runtimes it's a bit misleading as you would have to also preopen that directory to make it work.

Here some options that I think make sense:
1. Just have `sys/file` provide an `open` utility and get rid of `cwdfd`. This is sort of nice since we'd just have a useful system call in `sys/file`, especially since it deals with some fairly low-level behavior, and often what people are expecting. I dislike it because I see all of the `sys` libraries as pure WASI and this would stray from that.
2. Get rid of `cwdfd`, but add a `preopened` utility:
```
let preopened = idx => {
  // skip stdin, stdout, stderr
  let fd = idx + 3
  // validate that the preopen is good
  // ... call wasi fd_prestat etc
  Ok(FileDescriptor(fd))
}
```
This lets you ask for a preopened directory, positionally.
3. Get rid of `cwdfd`, but add a `getPreopenedDirectories` utility. Similar to the `preopened` utility, but provides you with a list.
4. Get rid of `cwdfd`, ignore any preopen utility nonsense. Just implement the high-level API, and if anyone is trying to do something low level they can just name the file descriptors themselves.

I think the answer is 2 + high-level API.

ospencer:
Closes #1205 

This PR reworks how preopened directories work in Grain. Namely, the Grain CLI no longer automatically preopens the current directory, and instead directories must be explicitly preopened via a new `--dir` flag.

`File.pwdfd` is no more; instead, `File` exposes a new function, `open`, which can open files in preopened directories. Some missing WASI functionality was also added to `File`.

Draft because some light cleanup is needed, plus some licensing needs to be figured out since `open` is based on the wasi-libc implementation. It's not a direct port but is heavily inspired by it, so a license may or may not be needed.

The changes to the js-runner are pretty gross, but I don't really care since it's going away.


stdlib/sys/file.gr
```diff
@@ -577,6 +585,44 @@ provide let fdPread = (fd: FileDescriptor, offset: Int64, size: Number) => {
   return Ok((WasmI32.toGrain(strPtr): Bytes, tagSimpleNumber(nread)))
 }
 
+/**
+ * Get information about a preopened directory.
+ *
+ * @param fd: The file descriptor to check
+ * @returns `Ok(Dir{name})` if successful or `Err(exception)` otherwise
```
phated, line 592:
```suggestion
 * @returns `Ok(Dir{prefix, fd})` if successful or `Err(exception)` otherwise
 *
 * @since v0.6.0
```
Need since docs and it looks like the Dir variant was doc'd wrong

stdlib/sys/file.gr
```diff
@@ -577,6 +585,44 @@ provide let fdPread = (fd: FileDescriptor, offset: Int64, size: Number) => {
...
+
+  let dir = Memory.malloc(8n)
+  let err = Wasi.fd_prestat_get(fd, dir)
+  if (err != Wasi._ESUCCESS) {
+    Memory.free(dir)
+    return Err(Wasi.SystemError(tagSimpleNumber(err)))
+  }
+
+  if (WasmI32.load8U(dir, 0n) != 0n) {
+    fail "Unknown prestat type"
```
phated, line 607: throw an exception and add a `@throws` doc

stdlib/sys/file.gr
```diff
@@ -1664,3 +1710,250 @@ provide let pathRename =
...
+
+/**
+ * Similar to `pathOpen`, but resolves the path relative to preopened directories.
+ *
+ * @param path: The path to the file or directory
+ * @param openFlags: Flags that decide how the path will be opened
+ * @param rights: The rights that dictate what may be done with the returned file descriptor
+ * @param rightsInheriting: The rights that dictate what may be done with file descriptors derived from this file descriptor
+ * @param flags: Flags which affect read/write operations on this file descriptor
+ * @returns `Ok(fd)` of the opened file or directory if successful or `Err(exception)` otherwise
```
phated, line 1934:
```suggestion
 * @returns `Ok(fd)` of the opened file or directory if successful or `Err(exception)` otherwise
 *
 * @since v0.6.0
```
This also needs throws docs because it uses populatePreopens which can throw.

phated (changes requested):
Needs some docs changes, and a throw instead of fail 🙏

ospencer:
@phated That fail was just for debugging, it won't actually throw. I removed it.

phated (approved):
I updated the docs, so this should be good to go.

Issue #802, opened 2021-08-22
Allow function reference to `Pervasives.([...])`


peblair:
There is currently no way to reference the `[...]` function defined in Pervasives (it's a syntax error). We should fix the parser to allow this.

phated:
I believe this is no longer a possibility since the List data type was moved inside the compiler in #1648 but @ospencer can correct me if I'm wrong.

ospencer:
We should remove `Pervasives.cons` and `Pervasives.empty` now.

PR #1657, opened 2023-02-05
chore(stdlib)!: Remove `cons` and `empty` from Pervasives

phated:
Ref https://github.com/grain-lang/grain/issues/802#issuecomment-1418263773


stdlib/pervasives.gr
```diff
@@ -526,28 +526,6 @@ provide primitive unbox: Box<a> -> a = "@unbox"
  * @section List operations: Functions for working with List values.
  */
```
ospencer, line 527: We should remove this section now.

ospencer (changes requested):

ospencer (approved):
Thank you!

PR #1658, opened 2023-02-05
feat(graindoc): Refactored Graindoc parser

### LINKED ISSUE #971: Graindoc: Support multi-line examples

### LINKED ISSUE #1653: Graindoc: Use proper lexer?
Should graindoc use a proper lexer, instead of 2 passes of regexps?

ospencer:
This PR implements a new parser for Graindoc, providing more accurate errors and expanding its capabilities. Namely, attribute descriptions now all support multiple lines, including `@example`!

Either syntax works fine:
```
/**
 * @example let foo = "foo"
 * print(foo)
 */
```

```
/**
 * @example
 * let foo = "foo"
 * print(foo)
 */
```

closes #971 
closes #1653


compiler/src/diagnostics/graindoc_lexer.re
```diff
@@ -0,0 +1,213 @@
...
+  } else {
+    semver;
+  };
+
+// HORIZONTAL TABULATION
+// VERTICAL TABULATION
+// SPACE
+// LEFT-TO-RIGHT MARK
+// RIGHT-TO-LEFT MARK
+let blank = [%sedlex.regexp? 0x09 | 0x0B | 0x20 | 0x200E | 0x200F];
```
phated, line 42: Are some of these defined in `Grain_parsing`? Do we need to redefine them or can we share them?

compiler/src/diagnostics/graindoc_lexer.re
```diff
@@ -0,0 +1,213 @@
...
+}
+
+and default = (state, lexbuf) => {
+  switch%sedlex (lexbuf) {
+  | asterisk => default(state, lexbuf)
+  | "@param" =>
+    state.lexer_mode = Param;
+    PARAM;
+  | "@returns" =>
+    state.lexer_mode = GenericAttribute;
```
phated, line 100: Should this be called like `UnlabeledAttribute` or `FreeTextAttribute`?

compiler/src/diagnostics/comment_attributes.re
```diff
@@ -0,0 +1,37 @@
+exception InvalidAttribute(string);
```
phated, line 1: Why did you split this into a different file?

compiler/src/diagnostics/graindoc_parser.mly
```diff
@@ -0,0 +1,48 @@
+%token PARAM SECTION SINCE HISTORY THROWS RETURNS EXAMPLE DEPRECATED COLON EOL EOF
+%token <string> TEXT IDENT SEMVER CONSTRUCTOR
+
+%right EOL
+
+%start <(Comment_attributes.parsed_graindoc)> graindoc
```
phated, line 6: Can this just be `(option(string), list(t))`
```suggestion
%start <((option(string), Comments.Attribute.attributes))> graindoc
```

phated (changes requested):
This is so awesome! Thanks for looking into it 🙏 I had a few questions

compiler/src/diagnostics/comment_attributes.re
ospencer: There was a circular dependency, parser -> comments -> parser 😢

compiler/src/diagnostics/graindoc_lexer.re
ospencer: I tried to share them, but it didn't seem possible.

compiler/src/diagnostics/graindoc_parser.mly
ospencer: I tried that originally, but it's a syntax error. Seems like you have to provide a named type here.

phated (approved):
LGTM! Thanks 💪

PR #1659, opened 2023-02-08
chore(grainfmt): Consolidate comment logic by removing assumed breaks

phated:
I'm trying to break my cleanups into understandable chunks. This is part 1.


compiler/src/formatting/comment_utils.re
```diff
@@ -535,6 +598,12 @@ let rec new_comments_inner =
           },
           Doc.hardLine,
           comment_to_doc(cmt),
+          switch (cmt) {
+          | Line(_)
+          | Shebang(_) => Doc.hardLine
+          | Doc(_) => Doc.hardLine
+          | _ => Doc.nil
+          },
```
ospencer, line 606: Seems like we should have a helper for this?

ospencer (approved):

Issue #1029, opened 2021-11-13
Support WASM-4 fantasy console


tjpalmer:
It would be great to use Grain for the WASM-4 fantasy console, but there are some requirements first. These include:

- Fitting well within the 64k wasm file "cartridge" size limit
- Ability to import memory rather than export
- Ability to control memory limits
- Ability to prevent using WASM-4's own lower region of memory

[A better reference is here](https://github.com/aduros/wasm4/issues/54#issuecomment-913032066), but I summarized briefly for some level of self-containment.

spotandjake:
I think we might have completed this.

phated:
@spotandjake no, it requires data segments

PR #1660, opened 2023-02-10
feat(compiler)!: Allocate closures only when necessary

### LINKED ISSUE #1410: Refactor closures
Grain currently allocates a closure for every function. However, many functions don't close over anything and are called by name, so the closure isn't necessary. We can save a lot of memory and module size by not allocating closures when unnecessary.

ospencer:
Closes #1410 
Works towards #1029 

This PR only allocates closures when absolutely necessary—when a function actually closes over values or the function may be used in a first-class way. 

This results in a 40% module size reduction for hello world programs, and similar 20-40% module size reductions for most Grain programs. It also improves program performance and module startup time.

This is breaking at the CMI level because consuming modules may need to allocate a closure for an imported function if it plans to use it in a first-class way.

### DIFF
--- a/compiler/test/__snapshots__/basic_functionality.52ca8e0e.0.snapshot
+++ b/compiler/test/__snapshots__/basic_functionality.52ca8e0e.0.snapshot
@@ -140,37 +140,14 @@ basic functionality › func_shadow
   (local $6 i32)
   (local $7 i32)
   (return
-   (block $cleanup_locals.18 (result i32)
+   (block $cleanup_locals.16 (result i32)
     (local.set $0
-     (block $compile_block.17 (result i32)
-      (block $compile_store.9
+     (block $compile_block.15 (result i32)
+      (block $compile_store.8
        (global.set $foo_1113
         (tuple.extract 0
          (tuple.make
-          (block $allocate_closure.7 (result i32)
-           (i32.store
-            (local.tee $0
-             (call $malloc_0
-              (global.get $GRAIN$EXPORT$malloc_0)
-              (i32.const 16)
-             )
-            )
-            (i32.const 6)
-           )
-           (i32.store offset=4
-            (local.get $0)
-            (i32.const 1)
-           )
-           (i32.store offset=8
-            (local.get $0)
-            (i32.const -1)
-           )
-           (i32.store offset=12
-            (local.get $0)
-            (i32.const 0)
-           )
-           (local.get $0)
-          )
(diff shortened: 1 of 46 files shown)


compiler/test/stdlib/immutablepriorityqueue.test.gr
```diff
@@ -4,8 +4,6 @@ include "immutablepriorityqueue"
 include "list"
 include "array"
 
-assert ImmutablePriorityQueue.make(compare) == ImmutablePriorityQueue.empty
-
```
ospencer, line 8: `compare` in one module is no longer necessarily equal to `compare` in another module because they can be different closures. There's no real equivalent of this test now. `==` currently just compares pointers for closures. 

A test that checks something similar to this would be this: `assert ImmutablePriorityQueue.isEmpty(ImmutablePriorityQueue.make(compare))`, but I do not believe that was the essence of this test.

compiler/test/suites/gc.re
```diff
@@ -81,7 +81,7 @@ describe("garbage collection", ({test, testSkip}) => {
   );
   assertRunGCError(
     "fib_gc_err",
-    1024,
+    512,
```
ospencer, line 84: This no longer ran out of memory at 1024 bytes because no closure is allocated when `fib` is called anymore. I adjusted it down.

compiler/test/__snapshots__/basic_functionality.52ca8e0e.0.snapshot
```diff
@@ -140,51 +140,25 @@ basic functionality › func_shadow
...
-            (local.get $0)
-            (i32.const -1)
-           )
-           (i32.store offset=12
-            (local.get $0)
-            (i32.const 0)
-           )
-           (local.get $0)
-          )
+          (i32.const 0)
```
phated, line 150: Is this now one of those tests where we are optimizing the whole thing away?

phated (approved):
One question about the snapshots (are they becoming tests that are completely optimized away?) otherwise looks good to me. We gave @peblair and the rest of the team a lot of time to review.

compiler/test/__snapshots__/basic_functionality.52ca8e0e.0.snapshot
ospencer: The whole closure gets optimized away here (which shows that the feature is working!), but not the rest of the test.

PR #1661, opened 2023-02-10
feat(compiler): Allow importing a memory

ospencer:
Works towards #1029 

For compilers that support this functionality, they just support importing from `env.memory`. We could make this a bit more complicated and allow passing the module name and memory name, but for now I don't think this is necessary.

We don't currently have the ability to add a test for this because the memory import happens after linking.

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -118,6 +118,7 @@ class GrainCommand extends commander.Command {
       "maximum number of WebAssembly memory pages",
       num
     );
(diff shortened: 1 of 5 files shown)


peblair (approved):

phated (commented):
Is these any tests we could write for this?

ospencer:
Test added.

phated (approved):
LGTM! Thanks for figuring out a test

PR #1664, opened 2023-02-11
feat(compiler): Remove Pervasives' dependency on toString when types are elided

ospencer:
This PR removes the forced dependency on `toString` when types are elided at runtime. If there's no type information, all of the code that comes with `toString` is of no help. This reduces module sizes for programs that don't otherwise depend on toString.

Works towards #1029


peblair:
Is it expected that there are no major snapshot changes? I can believe it, but I want to double-check

ospencer:
Nah, no snapshot changes. This would be a change on the snapshot of Pervasives itself, if we had one.

phated (approved):
LGTM

peblair (approved):
Ditto 👍

PR #1665, opened 2023-02-11
feat(compiler): Arbitrary-position spreads

alex-snezhko:
Support for putting the spread syntax at any position in a list (also spread support for arrays).

Example:
```
let listA = [1, 2, 3]
let listB = [...a, 4]
let arrA = [> 1, 2, 3]
let arrB = [> ...a, 4]
```


compiler/src/formatting/format.re
```diff
@@ -1132,6 +1132,52 @@ and resugar_pattern_list_inner = (patterns: list(Parsetree.pattern)) => {
   };
 }
 
+and process_list_comments = items => {
+  // We have to compose this list by hand because of the complexity of if a list item
+  // is followed by a comment, the comma must come before the comment.
+  // It also impacts how we force a new line for a line ending comment at the end of a list
+  // without introducing an extra blank line when bringing the indentation back in again
```
ospencer, line 1139: On my list of things to do is create a new parsetree node for lists, now that it's a part of the compiler. Seems like I should hurry up and do that so we can rebase this on top to prevent this from getting even more complex?

compiler/src/parsing/ast_helper.re
```diff
@@ -384,25 +390,111 @@ module Expression = {
...
+            List.map(
+              expr =>
+                switch (expr) {
+                | ListSpread(expr, loc) => (
+                    PExpSpreadExpr,
+                    {...expr, pexp_loc: loc},
+                  )
+                | ListItem(expr) => (
+                    PExpNonSpreadExpr,
+                    // Still convert to a single-element list to make later compilation steps easier
```
ospencer, line 414: This seems a little unnecessary. To me this looks like you're converting `[a, b, ...c, d, e, ...f]` into `concat([> [a], [b], c, [d], [e], f])`, instead of `concat([> [a, b, ...c], [d, e, ...f]])`.

compiler/src/parsing/ast_helper.re
```diff
@@ -384,25 +390,111 @@ module Expression = {
...
+      collection_concat(
+        ~loc,
+        ~attributes?,
+        PExpArrayConcat,
+        List.map(
+          x => {
+            switch (x) {
+            | ArrayItem(expr) => (
+                PExpNonSpreadExpr,
+                // Still convert to a single-element array to make later compilation steps easier
```
ospencer, line 485: Similar thing here.

compiler/src/typed/typecore.re
```diff
@@ -914,6 +914,27 @@ and type_expect_ =
...
+      };
+    let coll_type = mk_builtin_type(newvar());
+    with_explanation(() => unify_exp_types(loc, env, coll_type, ty_expected));
+    let typed_colls =
+      List.map(
+        ((_, expr)) => type_expect(env, expr, mk_expected(coll_type)),
+        colls,
+      );
+    rue({
+      exp_desc: TExpCollectionConcat(texp_type, typed_colls),
```
ospencer, line 931:
```suggestion
  | PExpCollectionConcat(t, collections) =>
    let (mk_builtin_type, texp_type) =
      switch (t) {
      | PExpListConcat => (Builtin_types.type_list, TExpListConcat)
      | PExpArrayConcat => (Builtin_types.type_array, TExpArrayConcat)
      };
    let collection_type = mk_builtin_type(newvar());
    with_explanation(() => unify_exp_types(loc, env, collection_type, ty_expected));
    let typed_collections =
      List.map(
        ((_, expr)) => type_expect(env, expr, mk_expected(collection_type)),
        collections,
      );
    rue({
      exp_desc: TExpCollectionConcat(texp_type, typed_collections),
```

compiler/src/typed/typedtreeIter.re
```diff
@@ -249,6 +249,7 @@ module MakeIterator =
       iter_expression(a1);
       iter_expression(a2);
       iter_expression(a3);
+    | TExpCollectionConcat(_, colls) => List.iter(iter_expression, colls)
```
ospencer, line 252:
```suggestion
    | TExpCollectionConcat(_, collections) => List.iter(iter_expression, collections)
```

compiler/src/typed/typedtreeMap.re
```diff
@@ -245,6 +245,8 @@ module MakeMap =
           ld,
           map_expression(arg),
         )
+      | TExpCollectionConcat(t, colls) =>
+        TExpCollectionConcat(t, List.map(map_expression, colls))
```
ospencer, line 249:
```suggestion
      | TExpCollectionConcat(t, collections) =>
        TExpCollectionConcat(t, List.map(map_expression, collections))
```

stdlib/runtime/concat.gr
```diff
@@ -0,0 +1,64 @@
...
+@unsafe
+provide let listConcat = listsArray => {
+  let ptr = WasmI32.fromGrain(listsArray)
+  let length = WasmI32.load(ptr, 4n)
+  let mut result = []
+
+  for (let mut i = 0n; WasmI32.ltU(i, length); i = WasmI32.add(i, 1n)) {
+    let list = listsArray[tagSimpleNumber(i)]
+    result = append(result, list)
+  }
```
ospencer, line 25: You'll want to iterate in reverse, since appending is O(length of first list) and the first list is growing. That way total time will be O(k) where k is the sum of the length of all lists.

stdlib/runtime/concat.gr
```diff
@@ -0,0 +1,64 @@
...
+
+  let mut totLength = 0n
+  for (let mut i = 0n; WasmI32.ltU(i, numArrs); i = WasmI32.add(i, 1n)) {
+    let array = arraysArray[tagSimpleNumber(i)]
+    totLength = WasmI32.add(totLength, findLength(array))
+  }
+
+  let result = allocateArray(totLength)
+  let mut byteOffset = result
+  for (let mut i = 0n; WasmI32.ltU(i, numArrs); i = WasmI32.add(i, 1n)) {
```
ospencer, line 47: Looks like it might be good to put some operators in scope.

stdlib/runtime/concat.gr
```diff
@@ -0,0 +1,64 @@
...
+    let array = arraysArray[tagSimpleNumber(i)]
```
ospencer, line 48: I don't usually say this, but for the sake of runtime performance, it's likely worth it to directly load the values instead of using the syntax because of the unnecessary bounds checking.

ospencer (changes requested):
Overall, the approach looks good here. Just a few things to clean up :)

compiler/src/formatting/format.re
alex-snezhko: Sure, if you'd like; I don't think it should be too much work to rebase this off of such a feature

compiler/src/parsing/ast_helper.re
alex-snezhko: True, but I think the overall runtime complexity would be the same since appending a single-element list on the left and appending an element to an existing list are both constant-time

compiler/src/parsing/ast_helper.re
alex-snezhko: Wouldn't it not matter much for arrays since the efficiency of spreads at the end is only relevant for linked lists?

stdlib/runtime/concat.gr
alex-snezhko: True

compiler/src/parsing/ast_helper.re
ospencer: Complexity-wise yes, but you get two allocations for a single-item list (cons cell + empty) versus one (just a cons cell).

compiler/src/parsing/ast_helper.re
ospencer: Just reducing the overall number of allocations. They're pretty expensive, especially once you factor in the GC overhead.

alex-snezhko: Just to clarify, you mean to say that specifically single elements should be grouped with the next spread in the same way that you suggested for lists (like treating `[> a, b, ...c, d, e, ...f]` as `[> [> a, b, ...c], [> d, e, ...f]]`)?

ospencer: In this case, it would be `[> [> a, b], c, [> d, e] f]`.

alex-snezhko: I see, that makes more sense

alex-snezhko:
I addressed the majority of the feedback, will wait until list ADT nodes gets merged to un-bork the formatter :P

alex-snezhko:
What should the fate of this PR be? Should it be tabled until further notice or is it still planned for 0.6? There was some discussion in a community meeting a little while ago to completely change list syntax (using arrows to connect elements or something of the sort), but that would cause the current syntax to be breaking regardless

PR #1667, opened 2023-02-12
feat(compiler)!: Supply primitive types via the compiler

### LINKED ISSUE #1041: Primitive types should be language-supplied
You can override the actual type of a primitive via its type annotation. This program results in a type error:

```
primitive ignore : Number -> Number = "@ignore"
ignore("foo")
```

The annotation should be unified with the actual type (causing a type error at that location) or we should remove the annotations altogether.

ospencer:
Closes #1041

### DIFF
--- a/compiler/src/typed/translprim.re
+++ b/compiler/src/typed/translprim.re
@@ -136,9 +136,9 @@ let prim_map =
         Primitive2(WasmLoadI64({sz: 4, signed: false})),
       ),
       ("@wasm.store_int64", PrimitiveN(WasmStoreI64({sz: 8}))),
-      ("@wasm.store_8_int64", PrimitiveN(WasmStoreI32({sz: 1}))),
-      ("@wasm.store_16_int64", PrimitiveN(WasmStoreI32({sz: 2}))),
-      ("@wasm.store_32_int64", PrimitiveN(WasmStoreI32({sz: 4}))),
+      ("@wasm.store_8_int64", PrimitiveN(WasmStoreI64({sz: 1}))),
+      ("@wasm.store_16_int64", PrimitiveN(WasmStoreI64({sz: 2}))),
+      ("@wasm.store_32_int64", PrimitiveN(WasmStoreI64({sz: 4}))),
       ("@wasm.load_float32", Primitive2(WasmLoadF32)),
       ("@wasm.store_float32", PrimitiveN(WasmStoreF32)),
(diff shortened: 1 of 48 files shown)


compiler/src/typed/translprim.re
```diff
@@ -1486,142 +1486,176 @@ let transl_prim = (env, desc) => {
...
-          WasmMemoryGrow |
-          WasmFromGrain |
-          WasmToGrain
-        ) as p,
-      ) => (
+      (
+        Expression.lambda(~loc, ~attributes, [], Expression.prim0(~loc, p)),
+        Typecore.prim0_type(p),
+      );
+    | Primitive1(BuiltinId as p) =>
```
phated, line 1534: `BuiltinId` always means `Void` or am I missing something?

compiler/test/input/memoryBase/asserts.gr
```diff
@@ -3,9 +3,12 @@ module Asserts
...
 
-primitive heapBase: WasmI32 = "@heap.base"
-primitive heapStart: WasmI32 = "@heap.start"
-primitive typeMetadata: WasmI32 = "@heap.type_metadata"
+@unsafe
+primitive heapBase = "@heap.base"
+@unsafe
+primitive heapStart = "@heap.start"
+@unsafe
+primitive typeMetadata = "@heap.type_metadata"
```
phated, line 11: I thought the changes in the compiler was adding the `disable_gc` attribute to these, so why do we still need `@unsafe`? I'm probably not following it correctly.

phated (approved):
I really like this change! And I noticed it caught some bugs in how the types were defined. I had a couple questions but I don't think they block an approval 🎉

compiler/test/input/memoryBase/asserts.gr
ospencer: I actually removed disable_gc from them because it made it pretty complicated. I liked this a lot better because they're unsafe values, and like anything else, they shouldn't be allowed outside of unsafe contexts.

compiler/src/typed/translprim.re
ospencer: It just means we don't generate a function for it, so you can't `List.map(builtinId, ...)`, for example. This transform is just setting up closures for primitives in the off chance you want to use them in a first-class way, but that's not an option for `BuiltinId` because it's just injecting type information from the compiler.

PR #1668, opened 2023-02-13
chore(stdlib): Clean up extra case in string escape logic

ospencer:
Test for this in `strings.re` continues to pass.


mortax (approved):
lgtm

phated (approved):

PR #1669, opened 2023-02-13
feat(compiler)!: Short integer values

### LINKED ISSUE #672: Feature request: int8, int16, uint8, uint16, uint32, uint64
Hi! Thanks for this amazing work!

In my opinion, Grain lacks some usual integer types: `int8`, `int16`, `uint8`, `uint16`, `uint32`, and `uint64`.

Some advantages to support them:
- better interop with C / Rust, and other system programming languages
- better compatibility with binary serialization formats (FlatBuffers, BARE, ...)
- possibility to express new constraints

alex-snezhko:
Closes #672 with support for signed & unsigned 8-bit and 16-bit integers

```
123s  // 8-bit signed
123S  // 16-bit signed
123us // 8-bit unsigned
123uS // 16-bit unsigned 
```

Details about the internal representation are outlined in my edits to the `data_representation.md` document, but in short:

Signed & unsigned 8/16 bit integers were added into the tag space previously reserved for `Char`. In the internal representation, the leading bits of the 24-bit short-value data space not used for storing the value of numbers are treated as insignificant and exist as an efficiency optimization.


compiler/src/middle_end/linearize.re
```diff
@@ -182,10 +182,20 @@ let transl_const =
...
   | Const_int32(i) =>
     Right(with_bind("int32", tmp => [BLet(tmp, Comp.int32(i), Nonglobal)]))
   | Const_int64(i) =>
     Right(with_bind("int64", tmp => [BLet(tmp, Comp.int64(i), Nonglobal)]))
+  | Const_uint8(i) =>
+    Right(with_bind("uint8", tmp => [BLet(tmp, Comp.uint8(i), Nonglobal)]))
+  | Const_uint16(i) =>
+    Right(
+      with_bind("uint16", tmp => [BLet(tmp, Comp.uint16(i), Nonglobal)]),
+    )
```
ospencer, line 198: These shouldn't need bindings, with the `Left` case being fine. It's the same for chars, but it looks like that was never updated. No need the special Comp helpers for the small ints!

stdlib/bytes.gr
```diff
@@ -385,6 +385,7 @@ provide let getInt8U = (index: Number, bytes: Bytes) => {
   let offset = coerceNumberToWasmI32(index)
   checkIndexIsInBounds(offset, _INT8_BYTE_SIZE, size)
   let n = WasmI32.load8U(ptr + offset, _VALUE_OFFSET)
+  // TODO change
```
ospencer, line 388: Should change.

stdlib/int16.gr
```diff
@@ -0,0 +1,485 @@
...
+  let result = WasmI32.xor(x, 0b110000n)
+  WasmI32.toGrain(result): Int16
+}
+
+/**
+ * @section Operations: Mathematical operations for Int16 values.
+ */
+
+// Note: most of these operations do not attempt to correct trailing bits after
+// an operation for the sake of efficiency
```
ospencer, line 89: I think the 1 extra mask instruction is worth it to avoid a good amount of headache in the future. It'll also simplify the implementations of things like `equal` which can just do a simple check that the values are exactly the same, stuff like `compare` which can focus on simple checks, and simplify upcasts. Really removes a lot of the gotchas when we want to implement new stuff in the future.

ospencer (changes requested):
Overall looks pretty good!

stdlib/bytes.gr
alex-snezhko: Completely forgot about this, my bad 😅 

Also unrelated to this PR: since unsigned 32/64 bit ints exist now should `getInt32, getInt64` be removed in favor of `getInt32U, getInt32S, ...`?

stdlib/int16.gr
alex-snezhko: Sounds good

stdlib/bytes.gr
ospencer: Conveniently the type is called `Int32` so those can stay the same 🙂 
There could be some new ones called `getUint32`, etc.

stdlib/int16.gr
alex-snezhko: Should signed ints have all leading zeros or should negative signed ints have leading 1s? The latter would make comparisons just require a normal i32 signed comparison but would make other operations slightly more expensive as they would require sign-extensions rather than simple masks

stdlib/bytes.gr
alex-snezhko: Should I tack `getUint32`, etc into this PR or make a separate one?

stdlib/int16.gr
ospencer: Yeah, I think they should have leading ones. Kind of annoying, but probably worth it.

stdlib/bytes.gr
ospencer: A separate PR would be better.

alex-snezhko: Ok should I update the 8/16-bit portions of Buffer/Bytes in this PR or make all of the changes in a separate PR after this one is merged for the sake of consistency?

ospencer: Yeah, might as well do it in a new PR.

compiler/src/middle_end/linearize.re
alex-snezhko: Removing the binding for chars causes compilation to fail [here](https://github.com/grain-lang/grain/blob/64e80b69b36a058e4883f4e22b3b9dc0735911ce/compiler/src/codegen/transl_anf.re#L528). Should I add `MConstChar` to the mashtree?

ospencer: Yup, we should have done that when we moved Char from the heap to the stack, but we missed it.

alex-snezhko: One of the downstream effects of removing bindings for small ints (combined with the fact that the functions in the stdlib modules are normal arithmetic operators e.g. `+`, `-`, etc.) is that they now go through constant folding in `optimize_simple_expressions.re`. This essentially means that these standard library function would never actually get called in the stdlib tests, making them basically useless.

Furthermore, I believe the fact that this happens is technically a bug. For example at the moment if I write this code:
```
let (+) = (a: Number, b: Number) => toString(a) ++ " plus " ++ toString(b)
let x = 1 + 2
print(x)
```
The output would be `3` because of the constant folding even though it should be `"1 plus 2"`

compiler/test/stdlib/int16.test.gr
```diff
@@ -0,0 +1,138 @@
...
+  (|),
+  (^),
+  (<<),
+  (>>),
+}
+
+// Suppress warnings about using `fromNumber` on constants, since that's what we want to test.
+let fromNumber = fromNumber
+
+assert fromNumber(5) == 5S
```
alex-snezhko, line 27: Possible consideration: `S` looks pretty similar to `5`, which may make distinguishing `Int16` literals from normal numbers ending in `5` a bit hard. Should the literal notation be changed or is it fine keeping it like this?

ospencer: Nah, I think it's fine. You can always add an underscore for readability; `5_S`.

ospencer: We can also highlight it better in the LSP grammar. https://github.com/grain-lang/grain-language-server/issues/148

stdlib/int16.gr
```diff
@@ -0,0 +1,468 @@
...
+ * @since v0.6.0
+ */
+@unsafe
+provide let toNumber = (value: Int16) => {
+  let num = WasmI32.shrS(WasmI32.fromGrain(value), _DATA_OFFSET)
+  WasmI32.toGrain(tagSimple(num)): Number
+}
+
+@unsafe
+let signExtend = x => WasmI32.shrS(WasmI32.shl(x, _DATA_OFFSET), _DATA_OFFSET)
```
ospencer, line 65: Guess what? There's an instruction for this! https://github.com/grain-lang/grain/blob/main/stdlib/runtime/unsafe/wasmi32.gr#L72

ospencer: You'd need to do the operation untagged, but there is indeed an instruction.

alex-snezhko: Yeah I know about that instruction - I even use `extendS16` for int8 as it is valid for tagged int8s. The reason why I didn't use it here is for the reason you mentioned: it only works for untagged values. In this situation it would be more costly to get working on a tagged 16-bit value (right shift 8 + extendS16 + left shift 8 + mask for tag) vs how I do it here (left shift 8 + signed right shift 8)

ospencer: I suppose that's fair. It's only really true for addition and subtraction though, since most of the other operators are fully untagging and retagging, so it's more expensive for things like multiplication, though I imagine that's less common.

Because of that, it might make sense to use the instruction in most places, but for things like addition where we're using tricks we just have a comment about the special tagging.

ospencer: It might also be a good idea to add new primitives for tagging/untagging the small ints, like we have for chars and simple numbers.

alex-snezhko: > Because of that, it might make sense to use the instruction in most places,

Sure thing; although from an inspection I think the extend instructions would only end up being used in multiplication and left shift.

alex-snezhko: Actually, left shift may be better off with how it currently is (meaning only multiplication would change); aside from the "necessary" steps required regardless of implementation (untagging `amount`, `shl` on the values) the current implementation requires an `and`, an `or`, and an `extendS16`, while switching it to untagging/tagging would require an untag (`shr`), an `extendS8`, and a retag (`shl` + `or`); one instruction is saved using the current implementation.

ospencer (approved):
Beautiful! Needs to be rebased again, but this looks splendid 😊

stdlib/int16.gr
```diff
@@ -0,0 +1,445 @@
...
+provide let fromNumber = (number: Number) => {
+  let val = coerceNumberToShortInt(
+    number,
+    _SMIN16_I32,
+    _SMAX16_I32,
+    _SMIN16_I64,
+    _SMAX16_I64,
+    false
+  )
+  tagInt16(val)
```
phated, line 44: I'm wondering if we should expose this (and co) from the runtime/number file as `coerceNumberToInt16`, etc. Mostly just wondering because it'd keep things more consistent and we're going to eventually want all `@unsafe` stuff in the runtime.

stdlib/int16.gr
```diff
@@ -0,0 +1,445 @@
...
+provide let toNumber = (value: Int16) => {
+  let num = untagInt16(value)
+  WasmI32.toGrain(tagSimple(num)): Number
+}
+
+@unsafe
+let signExtend = x => WasmI32.shrS(WasmI32.shl(x, _DATA_OFFSET), _DATA_OFFSET)
+
+/**
+ * Converts an Uint16 to a Int16.
```
phated, line 65:
```suggestion
 * Converts a Uint16 to an Int16.
```
I tried sounding this out and I think it is flipped.

stdlib/int16.gr
```diff
@@ -0,0 +1,445 @@
...
+ *
+ * @param number: The value to convert
+ * @returns The Uint16 represented as a Int16
```
phated, line 68:
```suggestion
 * @returns The Uint16 represented as an Int16
```

stdlib/int8.gr
```diff
@@ -0,0 +1,443 @@
...
+provide let toNumber = (value: Int8) => {
+  let num = untagInt8(value)
+  WasmI32.toGrain(tagSimple(num)): Number
+}
+
+// Trick: can simply use WasmI32.extendS16 to sign-extend 8-bit ints rather
+// than having to shift right, extendS8, and retag
+
+/**
+ * Converts an Uint8 to a Int8.
```
phated, line 63:
```suggestion
 * Converts a Uint8 to an Int8.
```

stdlib/int8.gr
```diff
@@ -0,0 +1,443 @@
...
+ *
+ * @param number: The value to convert
+ * @returns The Uint8 represented as a Int8
```
phated, line 66:
```suggestion
 * @returns The Uint8 represented as an Int8
```

stdlib/runtime/dataStructures.gr
```diff
@@ -225,3 +225,75 @@ provide primitive tagChar: WasmI32 -> Char = "@tag.char"
...
 provide primitive untagChar: Char -> WasmI32 = "@untag.char"
+
+/**
+ * Tag an int8.
+ *
+ * @param num: The int8 to tag
+ * @returns The tagged int8
+ */
+@unsafe
+provide primitive tagInt8: WasmI32 -> Int8 = "@tag.int8"
```
phated, line 236: We probably want to land this before #1667 so it can be updated for these new primitives. Otherwise you'll need to make the same changes as that PR.

stdlib/runtime/equal.gr
```diff
@@ -13,10 +13,12 @@ from WasmI32 use {
   ltS as (<),
   remS as (%),
   shl as (<<),
+  shrU as (>>),
 }
 include "runtime/unsafe/wasmi64" as WasmI64
 include "runtime/unsafe/tags" as Tags
 include "runtime/numbers" as Numbers
+include "runtime/exception"
```
phated, line 21: Why did these additions need to happen?

stdlib/runtime/numbers.gr
```diff
@@ -2508,6 +2508,88 @@ provide let (>>) = (x: Number, y: Number) => {
 // [NOTE]: Coercion is a *conservative* process! For example, even if a float is 1.0,
 //         we will fail if attempting to coerce to an int!
 
+/**
+ * Converts a Number to an Int32.
+ *
+ * @param number: The value to convert
+ * @returns The Number represented as an Int32
+ *
+ * @since v0.2.0
```
phated, line 2517: Wrong since?

stdlib/runtime/numbers.gr
```diff
@@ -2508,6 +2508,88 @@ provide let (>>) = (x: Number, y: Number) => {
...
+    }
+  }
+  if (WasmI32.gtS(int32, max32) || WasmI32.ltS(int32, 0n)) {
+    throw Exception.Overflow
+  }
+  if (is8bit) WasmI32.and(int32, 0xffffn) else WasmI32.and(int32, 0xffffn)
+}
+
+@unsafe
+provide let coerceNumberToShortInt =
```
phated, line 2553: No docblock?

stdlib/runtime/string.gr
```diff
@@ -562,12 +562,28 @@ toStringHelp = (grainValue, extraIndents, toplevel) => {
...
+        if (toplevel) {
+          string
+        } else {
+          escapeChar(string)
+        }
+      } else if (
+        shortValTag == Tags._GRAIN_INT8_SHORTVAL_TAG ||
+        shortValTag == Tags._GRAIN_INT16_SHORTVAL_TAG
+      ) {
+        NumberUtils.itoa32(shortVal, 10n)
```
phated, line 579: Do these print with their suffix? I am actually realizing that I don't know if other types like this are printed with their suffixes

stdlib/uint16.gr
```diff
@@ -0,0 +1,393 @@
...
+let _TAG_BYTE = 0b100010n
+@unsafe
+let _REMOVE_EXTRA_MASK = 0x00ffffffn
+@unsafe
+let _DATA_MASK = 0xffff00n
+@unsafe
+let _UNTAGGED_DATA_MASK = 0xffffn
+
+/**
+ * Converts a Number to an Uint16.
```
phated, line 28:
```suggestion
 * Converts a Number to a Uint16.
```

stdlib/uint16.gr
```diff
@@ -0,0 +1,393 @@
...
+ *
+ * @param number: The value to convert
+ * @returns The Number represented as an Uint16
```
phated, line 31:
```suggestion
 * @returns The Number represented as a Uint16
```

stdlib/uint16.gr
```diff
@@ -0,0 +1,393 @@
...
+ * @since v0.6.0
+ */
+@unsafe
+provide let fromNumber = (number: Number) => {
+  let val = coerceNumberToShortUint(number, _UMAX16_I32, _UMAX16_I64, false)
+  tagUint16(val)
+}
+
+/**
+ * Converts an Uint16 to a Number.
```
phated, line 42:
```suggestion
 * Converts a Uint16 to a Number.
```

stdlib/uint8.gr
```diff
@@ -0,0 +1,393 @@
...
+let _TAG_BYTE = 0b11010n
+@unsafe
+let _REMOVE_EXTRA_MASK = 0x0000ffffn
+@unsafe
+let _DATA_MASK = 0xff00n
+@unsafe
+let _UNTAGGED_DATA_MASK = 0xffn
+
+/**
+ * Converts a Number to an Uint8.
```
phated, line 28:
```suggestion
 * Converts a Number to a Uint8.
```

stdlib/uint8.gr
```diff
@@ -0,0 +1,393 @@
...
+ *
+ * @param number: The value to convert
+ * @returns The Number represented as an Uint8
```
phated, line 31:
```suggestion
 * @returns The Number represented as a Uint8
```

stdlib/uint8.gr
```diff
@@ -0,0 +1,393 @@
...
+ * @since v0.6.0
+ */
+@unsafe
+provide let fromNumber = (number: Number) => {
+  let val = coerceNumberToShortUint(number, _UMAX8_I32, _UMAX8_I64, true)
+  tagUint8(val)
+}
+
+/**
+ * Converts an Uint8 to a Number.
```
phated, line 42:
```suggestion
 * Converts a Uint8 to a Number.
```

phated (changes requested):
This is awesome. I love seeing a feature implemented top-to-bottom like this.

I had some questions, some docs changes, and we need to open an issue to update Bytes and Buffer to return proper short int values for various functions (e.g. `Buffer.getInt8`)

stdlib/runtime/dataStructures.gr
ospencer: I queued that change before I saw this, but once it gets in I can update this PR.

stdlib/runtime/equal.gr
alex-snezhko: I think I was trying to raise an artificial exception for debugging purposes earlier but forgot to remove these, mb

stdlib/runtime/numbers.gr
alex-snezhko: Yeah looks like I messed up the comment too :sweat_smile:

stdlib/runtime/numbers.gr
alex-snezhko: I actually don't think one is required here since it never gets propagated up to the stdlib; same goes for the one above now that I think of it

stdlib/runtime/string.gr
alex-snezhko: No; the other types don't either (I think it's better without them imo)

compiler/src/typed/parmatch.re
```diff
@@ -1831,8 +1891,13 @@ let untype_constant =
     )
   | Const_number(Const_number_bigint({bigint_rep})) =>
     Parsetree.PConstNumber(Parsetree.PConstNumberInt(bigint_rep))
+  | Const_int8(i) => Parsetree.PConstInt8(Int32.to_string(i))
+  | Const_int16(i) => Parsetree.PConstInt16(Int32.to_string(i))
   | Const_int32(i) => Parsetree.PConstInt32(Int32.to_string(i))
   | Const_int64(i) => Parsetree.PConstInt64(Int64.to_string(i))
+  | Const_uint8(i) => Parsetree.PConstUint8(false, Printf.sprintf("%lu", i))
+  | Const_uint16(i) =>
+    Parsetree.PConstUint16(false, Printf.sprintf("%lu", i))
```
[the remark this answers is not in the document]
alex-snezhko: `Lu` is the ocaml format specifier for 64-bit unsigned numbers while lowercase `lu` is for 32-bit; since ocaml doesn't have 8 or 16 bit numbers natively I just store short numbers in a regular `int32`

stdlib/int16.gr
alex-snezhko: Sure I can do this; pretty much every function in the short int standard libraries are `@unsafe` though so that'll be a lot of code moved :laughing:

stdlib/runtime/dataStructures.gr
alex-snezhko: @ospencer I can go ahead and update them since I have to rebase anyways

phated (approved):
LGTM! Thanks for all the work here!

PR #1670, opened 2023-02-14
chore(compiler): Add Parsetree_iter module with enter/leave hooks

phated:
This replaces `Ast_iterator` with a new module `Parsetree_iter` which operates more like `typetreeIter`. I thought about making it a module functor; however, in keeping with patterns that could translate to Whole Grain, I used records, a `make` function that takes the hooks and passes them into each `iter_*` function.

This also opens up the door for various other hooks that we might need when iterating the parsetree for things like formatting.

In making this change it became apparent that there was code in `typed` that used the old `Ast_iterator` and wasn't used anywhere so I deleted it.


compiler/src/parsing/driver.re
```diff
@@ -167,10 +167,14 @@ let read_imports = ({Parsetree.comments, Parsetree.statements}) => {
...
-  let iterator = {...Ast_iterator.default_iterator, include_: iter_mod};
-  List.iter(iterator.toplevel(iterator), statements);
+
+  let enter_include = inc => {
+    found_includes := [inc.Parsetree.pinc_path, ...found_includes^];
+  };
+
+  let include_iter = Parsetree_iter.(make({...default_hooks, enter_include}));
+
+  List.iter(include_iter.iter_toplevel_stmt, statements);
```
ospencer, line 177: This sort of always irked me; could we change this to `iter_parsed_program`?

phated: Yep! I thought about doing that too but wasn't sure

compiler/src/parsing/parsetree_iter.re
```diff
@@ -0,0 +1,486 @@
...
+and iter_toplevel_stmt =
+    (hooks, {ptop_desc: desc, ptop_attributes: attrs, ptop_loc: loc} as top) => {
+  hooks.enter_toplevel_stmt(top);
+  iter_location(hooks, loc);
+  iter_attributes(hooks, attrs);
+  switch (desc) {
+  | PTopInclude(id) =>
+    hooks.enter_include(id);
+    iter_include(hooks, id);
+    hooks.leave_include(id);
```
ospencer, line 104: We've sort of got two different styles of who's calling `enter` and `leave`, sometimes it's the `iter` that's actually iterating the thing, and sometimes it's the thing that's calling that `iter`.

I think it'd be best to use just one style, and I'd be in favor of the iterator for the thing is the one that calls `enter` and `leave`. For example, `iter_toplevel_stmt` is the one responsible for calling `enter_toplevel_stmt` and `leave_toplevel_stmt`, so here, `iter_include` should internally call `enter_include` and `leave_include`, and the same for the rest of these.

compiler/src/parsing/driver.re
ospencer: You missed this one!

ospencer (changes requested):
Overall looks great! Just one comment around where `enter`/`leave` get called.

compiler/src/parsing/parsetree_iter.re
```diff
@@ -0,0 +1,488 @@
...
+        pdata_params: args,
+        pdata_manifest: manifest,
+        pdata_kind: kind,
+        pdata_loc: loc,
+      } as d,
+    ) => {
+  iter_location(hooks, loc);
+  iter_loc(hooks, name);
+  // TODO: Should this be called before the `iter_location` and `iter_loc` above?
+  hooks.enter_data_declaration(d);
```
ospencer, line 178: Definitely. 
```suggestion
  hooks.enter_data_declaration(d);
  iter_location(hooks, loc);
  iter_loc(hooks, name);
```

ospencer (approved):
Looks great! Just need to resolve that one TODO.

PR #1676, opened 2023-02-17
fix(compiler): Remove number constant folding optimization

### LINKED ISSUE #1675: Inappropriate constant folding for custom-defined infix operators
Currently arithmetic operations on constant numbers get optimized ahead-of-time by the compiler. However, this has the potential to produce incorrect behavior if these arithmetic operations are redefined. For example, consider the following code:
```
let (+) = (a, b) => toString(a) ++ " plus " ++ toString(b)
let x = 1 + 2
```
The value of `x` in the code is the number `3` even though it should be the string `1 plus 2`

ospencer:
Closes #1675 

We will reintroduce this optimization again properly.

### DIFF
--- a/compiler/src/middle_end/optimize_simple_expressions.re
+++ b/compiler/src/middle_end/optimize_simple_expressions.re
@@ -2,69 +2,16 @@ open Anftree;
 open Grain_typed;
 open Types;
 
-let is_int =
-  fun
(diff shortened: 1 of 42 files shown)


phated (approved):
Approving, but please add an `assertRun` regression test for #1675

ospencer:
Test added.

PR #1678, opened 2023-02-18
chore: Update the set documentation

spotandjake:
Space Ghost in the discord found this example to be bad so here is a small pr to fix it.

### DIFF
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -145,13 +145,13 @@ provide let get = (index, array) => {
 /**
  * An alias for normal syntactic array set, i.e. `array[n] = value`.
  *
- * Sets the element at the specified index in the array to the new value.
+ * Sets the element in place at the specified index in the array to the new value.
  * A negative index is treated as an offset from the end of the array.
  *
  * @param index: The index to update
  * @param value: The value to store
  * @param array: The array to update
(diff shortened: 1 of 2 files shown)


stdlib/array.gr
```diff
@@ -145,13 +145,13 @@ provide let get = (index, array) => {
...
  *
- * Sets the element at the specified index in the array to the new value.
+ * Sets the element in place at the specified index in the array to the new value.
  * A negative index is treated as an offset from the end of the array.
  *
  * @param index: The index to update
  * @param value: The value to store
  * @param array: The array to update
- * @example Array.set(1, 9, [> 1, 2, 3, 4, 5]) == [> 1, 9, 3, 4, 5]
+ * @example Array.set(1, 9, [> 1, 2, 3, 4, 5]) // == [> 1, 9, 3, 4, 5]
```
phated, line 154: We need to think of a better way to represent this.

phated (changes requested):
I'm not happy with the way the result is indicated. We should figure out a better way.

stdlib/array.gr
ospencer: The one I suggested was just without the equals; I think it's clear enough. We can do something different like `// array contents [> ...]` but that won't be consistent across the stdlib.

phated: I really don't like comments meant to signify an equals or outcome. Maybe this should just be updated to use multiline examples (since that has landed).

PR #1680, opened 2023-02-18
feat(compiler)!: Explicit abstract types

### LINKED ISSUE #1240: Compiler: Don't expose unnecessary types in module signatures
In Grain, when types are not exported they appear abstract to the consuming module. However, we expose _all_ types defined in that module as abstract ones, when it should only be the ones that are visible via other types or values exposed from the module, i.e. if there's a `type Foo` that's only used internally, it should not appear in the module signature at all.

ospencer:
Previously, any type defined in a module that was not provided was still provided by the compiler as an abstract type. Now, types must explicitly be marked with a new `abstract` keyword to indicate that the type should be visible but the implementation hidden.

This is a breaking change because types not marked with `provide` or `abstract` no longer appear in module signatures. If a value is provided by a module but uses a type that is not also provided or abstract, the compiler shows an error, as this type is private and not accessible by consuming modules.

Closes #1240

### DIFF
--- a/compiler/src/typed/ctype.re
+++ b/compiler/src/typed/ctype.re
@@ -2189,7 +2189,7 @@ let nondep_instance = (env, level, id, ty) => {
    list (nl2, tl2). raise Not_found if impossible */
 let complete_type_list = (~allow_absent=false, env, nl1, lv2, mty2, nl2, tl2) => {
   let id2 = Ident.create("Pkg");
-  let env' = Env.add_module(id2, mty2, None, env);
+  let env' = Env.add_module(id2, mty2, None, Location.dummy_loc, env);
   let rec complete = (nl1, ntl2) =>
     switch (nl1, ntl2) {
     | ([], _) => ntl2
--- a/compiler/test/__snapshots__/exceptions.a68ae348.0.snapshot
+++ b/compiler/test/__snapshots__/exceptions.a68ae348.0.snapshot
@@ -150,5 +150,5 @@ exceptions › exception_4
    (call $_gmain)
   )
  )
- ;; custom section \"cmi\", size 1238
+ ;; custom section \"cmi\", size 1284
 )
--- a/compiler/test/__snapshots__/functions.d9466880.0.snapshot
+++ b/compiler/test/__snapshots__/functions.d9466880.0.snapshot
@@ -15,13 +15,13 @@ functions › func_record_associativity2
  (import \"GRAIN$MODULE$runtime/gc\" \"incRef\" (func $incRef_0 (param i32 i32) (result i32)))
  (import \"GRAIN$MODULE$runtime/gc\" \"decRef\" (func $decRef_0 (param i32 i32) (result i32)))
  (global $GRAIN$TABLE_SIZE i32 (i32.const 1))
- (elem $elem (global.get $relocBase_0) $lam_lambda_1120)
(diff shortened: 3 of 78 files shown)


ospencer:
The errors for non-provided types appearing in provided modules could be expanded to display the specific offending item, but this is a bit complicated and IMO is not necessary for the initial feature.

compiler/src/typed/ctype.re
```diff
@@ -2189,7 +2189,7 @@ let nondep_instance = (env, level, id, ty) => {
    list (nl2, tl2). raise Not_found if impossible */
 let complete_type_list = (~allow_absent=false, env, nl1, lv2, mty2, nl2, tl2) => {
   let id2 = Ident.create("Pkg");
-  let env' = Env.add_module(id2, mty2, None, env);
+  let env' = Env.add_module(id2, mty2, None, Location.dummy_loc, env);
```
phated, line 2192: Why a dummy_loc here?

compiler/test/__snapshots__/exceptions.a68ae348.0.snapshot
```diff
@@ -150,5 +150,5 @@ exceptions › exception_4
    (call $_gmain)
   )
  )
- ;; custom section \"cmi\", size 1238
+ ;; custom section \"cmi\", size 1284
```
phated, line 153: Why would these increase in size instead of decrease?

compiler/test/__snapshots__/functions.d9466880.0.snapshot
```diff
@@ -384,5 +384,5 @@ functions › func_record_associativity2
    (call $_gmain)
   )
  )
- ;; custom section \"cmi\", size 729
+ ;; custom section \"cmi\", size 236
```
phated, line 387: Huge decrease! 🎉

compiler/src/typed/ctype.re
ospencer, line 2192: @phated The dummy_locs are all for places where location information isn't available.

phated (commented):
Only a couple questions then I think this is good to go!

compiler/src/typed/ctype.re
ospencer: I missed getting my comment in soon enough! Just because there's no location information available (this is mostly for module types, a bit of a holdover from OCaml).

compiler/test/__snapshots__/exceptions.a68ae348.0.snapshot
ospencer: We include location information in CMIs now (we previously stripped all location information), so sizes went up. This doesn't affect final binary sizes and will also help with LSP features, like go to definition.

phated (approved):
LGTM! Not sure if you want to give anyone else a chance to bikeshed on `abstract`. I know we asked for alternatives on the community call a few weeks ago and I haven't seen anything.

compiler/test/__snapshots__/exceptions.a68ae348.0.snapshot
ospencer: (I did this because the check just literates the signature, and I essentially am doing a "go to definition" to report the error wherever the error is.)

ospencer:
Not really... I think we gave this a fair amount of bikeshedding, but if we truly find something we like better we can always send it before 0.6 lands.

PR #1681, opened 2023-02-19
chore(graindoc)!: Remove section attribute

phated:
This removes Docblocks that use the `@section` attribute. We plan to replace these "Sections" with submodules (and docblocks on them).

I wanted to remove sections before adding submodule support because the diff gives us a great view into the structure of our stdlib. We need to bikeshed inline on which sections should become submodules and which should remain in the top-level grouping.

I also did some light cleanup and renaming.

Ref #1621

### DIFF
--- a/stdlib/buffer.md
+++ b/stdlib/buffer.md
@@ -15,9 +15,19 @@ No other changes yet.
 include "buffer"
 ```
 
+## Types
+
+Type declarations included in the Buffer module.
+
+### Buffer.**Buffer**
+
+```grain
+type Buffer
+```
+
 ## Values
 
-Functions for working with the Buffer data type.
+Functions and constants included in the Buffer module.
 
 ### Buffer.**make**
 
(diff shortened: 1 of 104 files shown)


stdlib/buffer.md
```diff
@@ -15,9 +15,19 @@ No other changes yet.
...
 ```
 
+## Types
+
+Type declarations included in the Buffer module.
+
+### Buffer.**Buffer**
+
+```grain
+type Buffer
```
ospencer, line 25: I wonder if we want to show `abstract type Buffer` so it's clear that it's an abstract type and not just missing information.

ospencer (commented):
I do not think any of the old sections should be converted to submodules. We didn't use them much, and when we did they were for groupings that made the docs slightly nicer. The closest contenders would be Bytes and Buffer for the operations on ints, but grouping those functions into a submodule feels a bit arbitrary, and the DX would be slightly worse.

What I would want in these cases is something like `@section`—a grouping in the docs where I can give some extra context, but the API should definitely be `Bytes.getInt32`.

If we want to remove sections to make Graindoc less complex, I think that's reasonable, but after looking this PR over, I don't think submodules serve the same purpose.

stdlib/buffer.md
phated: Unrelated to this PR. Please open an issue since this wasn't changed in the abstract PR.

ospencer: This was just a thought.

ospencer (approved):

PR #1683, opened 2023-02-19
feat(compiler)!: Remove 32-bit numbers from `Number` type

### LINKED ISSUE #1555: Refactor Int32 and Float32 out of Number
We can simplify Number quite a bit by separating out Int32 and Float32. 

Grain's simple numbers are 31 bits, so promoting to an Int32 gains only one bit of information. This doubles the number space, which is significant, however no memory is saved versus promoting to an Int64, as Grain's memory allocator is 8-byte aligned.

This will also simplify 64-bit mode, as a 63-bit integer cannot be promoted to an Int32.

All of Grain's number operations (except those on Float32s) return Float64s where applicable, so removing Float32 from Number will remove a number of bugs that can be encountered by casting a Float32 to a Number.

alex-snezhko:
Closes #1555, making `Int32` and `Float32` their own distinct types separate from `Number`. The most fragile part of these changes is making sure that all of the memory load offsets for these types were corrected, but I am fairly confident that I got them all after sweeping through every instance of `8n` in the stdlib/runtime several times and making corrections where necessary :crossed_fingers: .

### DIFF
--- a/stdlib/float32.gr
+++ b/stdlib/float32.gr
@@ -19,6 +19,9 @@ from Numbers use {
   coerceFloat32ToNumber as toNumber,
 }
 
+@unsafe
+let _VALUE_OFFSET = 4n
+
 /**
  * Infinity represented as a Float32 value.
  *
@@ -79,8 +82,8 @@ provide { fromNumber, toNumber }
  */
 @unsafe
 provide let add = (x: Float32, y: Float32) => {
-  let xv = WasmF32.load(WasmI32.fromGrain(x), 8n)
-  let yv = WasmF32.load(WasmI32.fromGrain(y), 8n)
+  let xv = WasmF32.load(WasmI32.fromGrain(x), _VALUE_OFFSET)
+  let yv = WasmF32.load(WasmI32.fromGrain(y), _VALUE_OFFSET)
   let ptr = newFloat32(WasmF32.add(xv, yv))
   WasmI32.toGrain(ptr): Float32
 }
@@ -96,8 +99,8 @@ provide let add = (x: Float32, y: Float32) => {
  */
 @unsafe
 provide let sub = (x: Float32, y: Float32) => {
-  let xv = WasmF32.load(WasmI32.fromGrain(x), 8n)
-  let yv = WasmF32.load(WasmI32.fromGrain(y), 8n)
+  let xv = WasmF32.load(WasmI32.fromGrain(x), _VALUE_OFFSET)
+  let yv = WasmF32.load(WasmI32.fromGrain(y), _VALUE_OFFSET)
   let ptr = newFloat32(WasmF32.sub(xv, yv))
   WasmI32.toGrain(ptr): Float32
 }
@@ -113,8 +116,8 @@ provide let sub = (x: Float32, y: Float32) => {
  */
 @unsafe
 provide let mul = (x: Float32, y: Float32) => {
-  let xv = WasmF32.load(WasmI32.fromGrain(x), 8n)
-  let yv = WasmF32.load(WasmI32.fromGrain(y), 8n)
+  let xv = WasmF32.load(WasmI32.fromGrain(x), _VALUE_OFFSET)
+  let yv = WasmF32.load(WasmI32.fromGrain(y), _VALUE_OFFSET)
   let ptr = newFloat32(WasmF32.mul(xv, yv))
   WasmI32.toGrain(ptr): Float32
 }
@@ -130,8 +133,8 @@ provide let mul = (x: Float32, y: Float32) => {
  */
 @unsafe
 provide let div = (x: Float32, y: Float32) => {
-  let xv = WasmF32.load(WasmI32.fromGrain(x), 8n)
-  let yv = WasmF32.load(WasmI32.fromGrain(y), 8n)
+  let xv = WasmF32.load(WasmI32.fromGrain(x), _VALUE_OFFSET)
+  let yv = WasmF32.load(WasmI32.fromGrain(y), _VALUE_OFFSET)
   let ptr = newFloat32(WasmF32.div(xv, yv))
   WasmI32.toGrain(ptr): Float32
 }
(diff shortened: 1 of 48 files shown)


stdlib/runtime/compare.gr
```diff
@@ -132,6 +133,18 @@ let rec heapCompareHelp = (heapTag, xptr, yptr) => {
...
+      let yval = WasmI32.load(yptr, 4n)
+      return if (WasmI32.ltS(xval, yval)) -1 else if (WasmI32.gtS(xval, yval)) 1
+      else 0
+    },
+    t when t == Tags._GRAIN_FLOAT32_HEAP_TAG => {
+      let xval = WasmF32.load(xptr, 4n)
+      let yval = WasmF32.load(yptr, 4n)
+      return if (WasmF32.lt(xval, yval)) -1 else if (WasmF32.gt(xval, yval)) 1
+      else 0
+    },
```
ospencer, line 147: This needs to be updated to handle NaN properly, with NaN being equal to itself (should already be correct here) and smaller than any other number (not handled here). Would be good to add a few tests as well.

stdlib/runtime/equal.gr
```diff
@@ -187,11 +188,14 @@ let rec heapEqualHelp = (heapTag, xptr, yptr) => {
...
     },
-    t when t == Tags._GRAIN_UINT32_HEAP_TAG => {
+    t when (
+      t == Tags._GRAIN_UINT32_HEAP_TAG || t == Tags._GRAIN_INT32_HEAP_TAG
+    ) => {
       let xval = WasmI32.load(xptr, 4n)
       let yval = WasmI32.load(yptr, 4n)
       WasmI32.eq(xval, yval)
     },
+    // Float32 would have already been checked by now
```
ospencer, line 198: This comment is a little confusing reading top-down, maybe something like this?
```suggestion
    // Float32 is handled by equalHelp directly
```

stdlib/runtime/numbers.gr
```diff
@@ -2465,11 +2030,22 @@ provide let coerceNumberToFloat64 = (x: Number) => {
   WasmI32.toGrain(result): Float64
 }
 
+@unsafe
+let _SIMPLE_NUMBER_MAX = 0x3FFFFFFFn
+@unsafe
+let _SIMPLE_NUMBER_MIN = 0xC0000001n
```
ospencer, line 2036: Since you added these, can you move them to the top and use them everywhere we do shifts on `I32_MAX` and `I32_MIN`?

stdlib/runtime/numbers.gr
```diff
@@ -2465,11 +2030,22 @@ provide let coerceNumberToFloat64 = (x: Number) => {
...
-    reducedInteger(WasmI64.extendI32S(boxedInt32Number(WasmI32.fromGrain(x))))
-  ): Number
+  let x = WasmI32.load(WasmI32.fromGrain(x), 4n)
+  let result = if (
+    WasmI32.gtS(x, _SIMPLE_NUMBER_MAX) || WasmI32.ltS(x, _SIMPLE_NUMBER_MIN)
+  ) {
+    newInt64(WasmI64.extendI32S(x))
+  } else {
+    tagSimple(x)
+  }
```
ospencer, line 2047: This mostly duplicates the logic of `reducedInteger`. Makes me feel like we should just use that with the extend.

ospencer (changes requested):
Looks pretty good overall, and I love the cleanup in the number runtime! Just a couple small fixes + tests are needed but otherwise this looks good to go. Looks like you'll also need to regen the docs.

stdlib/runtime/numbers.gr
ospencer: Oh, these are tagged, aren't they.

stdlib/runtime/numbers.gr
alex-snezhko: Good point; I think that would also remove the need for the simple number constants

stdlib/runtime/numbers.gr
alex-snezhko: No these are just the values; they should be equivalent to shifting I32_MAX

stdlib/runtime/numbers.gr
alex-snezhko: Actually now that you bring this up I think there may be a bug with the "minimum" values; `_SMIN32_I64` is defined as `-0x7FFFFFFFN` but it actually should be one lower than that (since twos complement signed ints can hold one more negative value than positive)

ospencer: That's true. And explains why I thought the simple min and max were tagged, since simple min should be `0xC0000000`. Good catch! And something hard to catch in tests.

stdlib/runtime/unsafe/constants.gr
```diff
@@ -2,17 +2,17 @@
...
 provide let _UMAX_I32 = 0xFFFFFFFFn
 
-provide let _SMIN_I64 = 0xFFFFFFFFFFFFFFFFN
+provide let _SMIN_I64 = 0x8000000000000000N
 provide let _SMAX_I64 = 0x7FFFFFFFFFFFFFFFN
 provide let _UMIN_I64 = 0x0N
 provide let _UMAX_I64 = 0xFFFFFFFFFFFFFFFFN
 
-provide let _SMIN32_I64 = -0x7FFFFFFFN
+provide let _SMIN32_I64 = -0x80000000N
```
ospencer, line 15: Can we write out the whole number here? This feels a tad confusing.

ospencer (approved):
Excellent, thank you! 🥳

stdlib/float32.gr
```diff
@@ -66,7 +66,21 @@ provide let tau = 6.2831853f
...
+ * @param number: The value to convert
+ * @returns The Number represented as a Float32
+ *
+ * @since v0.2.0
+ */
+@unsafe
+provide let fromNumber = (x: Number) => {
+  let result = newFloat32(coerceNumberToWasmF32(x))
+  WasmI32.toGrain(result): Float32
+}
```
phated, line 81: Can this be the `coerce` that is provided like other types?

stdlib/int32.gr
```diff
@@ -9,18 +9,33 @@
...
+ *
+ * @param number: The value to convert
+ * @returns The Number represented as an Int32
+ *
+ * @since v0.2.0
+ */
+@unsafe
+provide let fromNumber = (x: Number) => {
+  let result = newInt32(coerceNumberToWasmI32(x))
+  WasmI32.toGrain(result): Int32
```
phated, line 35: Can this be the `coerce` function that is exported, like the others?

phated (changes requested):
Just need the two coerce functions changes and resolution of the conflicts.

phated (approved):

PR #1684, opened 2023-02-20
feat(graindoc)!: Support docblocks on submodules

### LINKED ISSUE #1621: Graindoc: Support new module system

phated:
This updates graindoc to support submodules. Since modules need to be processed recursively, I needed to refactor most of the graindoc internals. This was a really good refactor because it makes a clear separation between the attribute parsing, the docblock data types, and the markdown generation phase. Both the docblock collection and markdown generation are recursive so we can start at the top of the program tree and call `Docblock.for_signature_items` and the `Docblock.to_markdown` on the results and everything gets printed.

Since we don't have any submodules in the stdlib at the moment, I've run this against #1479 and pushed the updated docs as https://github.com/grain-lang/grain/pull/1479/commits/4dfc7c56074cc929f3becad5db7a98e0d3e7174c

Closes #1621

### DIFF
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -3,19 +3,68 @@ open Grain_typed;
 open Grain_utils;
 open Grain_diagnostics;
 
-type t = {
-  module_name: string,
-  name: string,
-  type_sig: string,
-  description: option(string),
-  attributes: Comments.Attribute.attributes,
+type param = {
+  param_name: string,
+  param_type: string,
+  param_msg: string,
 };
+
+type since = {since_version: string};
+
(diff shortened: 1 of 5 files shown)


ospencer (approved):
This looks really good, and I appreciate the example. Great work!

PR #1687, opened 2023-02-20
feat(graindoc)!: Only use original doc blocks when re-providing

### LINKED ISSUE #1626: Graindoc: load docs from original implementation
In places like Pervasives, we re-export values defined in other modules. It'd be great if not overrided, Graindoc pulled the docs from the original implementation.

phated:
Graindoc will now pull docs from the original definition of a value/type/module. This was enabled by adding all of the location information to the CMI. Note that each file that provides a value needs to be recompiled through the parse phase so we can get the parsed comments.

This is a breaking change to graindoc since it will **only** pull from the original definition. You can see that I had to move definitions into the runtime to make this work. That results in the docs existing in the runtime and then in the stdlib places they are re-provided.

This should also solve the problem @alex-snezhko brought up in discord since each thing being documented will pull the exact definition.

Closes #1626

### DIFF
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -113,74 +113,6 @@ let () =
     }
   });
 
-let enumerate_provides = program => {
-  let id_tbl = ref(Ident.empty);
-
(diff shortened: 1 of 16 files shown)


ospencer (approved):
This is sweet!

PR #1688, opened 2023-02-21
feat(grainlsp): Provide hovers on use statements

ospencer:
image

cc @marcusroberts

### DIFF
--- a/compiler/src/diagnostics/modules.re
+++ b/compiler/src/diagnostics/modules.re
@@ -14,9 +14,8 @@ type export = {
   signature: string,
 };
 
-let rec get_exports = (~path, compiled_code: Typedtree.typed_program) =>
-  try({
-    let md = Env.find_module(path, None, compiled_code.env);
+let rec get_exports = (md: Types.module_declaration) =>
+  try(
     switch (md.md_type) {
     | TModSignature(sigs) =>
       List.filter_map(
@@ -63,9 +62,8 @@ let rec get_exports = (~path, compiled_code: Typedtree.typed_program) =>
         },
         sigs,
       )
-    | TModAlias(path) => get_exports(~path, compiled_code)
     | _ => []
-    };
-  }) {
+    }
+  ) {
   | e => []
   };
--- a/compiler/src/diagnostics/modules.re
+++ b/compiler/src/diagnostics/modules.re
@@ -14,7 +14,7 @@ type export = {
   signature: string,
 };
 
-let rec get_exports = (md: Types.module_declaration) =>
+let rec get_provides = (md: Types.module_declaration) =>
   try(
     switch (md.md_type) {
     | TModSignature(sigs) =>
--- a/compiler/src/language_server/hover.re
+++ b/compiler/src/language_server/hover.re
@@ -109,8 +109,8 @@ let print_type = (env, ty) => {
   };
 };
 
-let module_lens = (~program: Typedtree.typed_program, path: Path.t) => {
-  let vals = Modules.get_exports(~path, program);
+let module_lens = (decl: Types.module_declaration) => {
+  let vals = Modules.get_exports(decl);
   let signatures =
     List.map(
       (v: Modules.export) =>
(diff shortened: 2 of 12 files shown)


marcusroberts (approved):
Fantastic, unlocks go to definition too!

compiler/src/diagnostics/modules.re
```diff
@@ -14,9 +14,8 @@ type export = {
   signature: string,
 };
 
-let rec get_exports = (~path, compiled_code: Typedtree.typed_program) =>
-  try({
-    let md = Env.find_module(path, None, compiled_code.env);
+let rec get_exports = (md: Types.module_declaration) =>
```
phated, line 17:
```suggestion
let rec get_provides = (md: Types.module_declaration) =>
```
I'm going to be that guy 🙃

compiler/src/language_server/hover.re
```diff
@@ -109,8 +109,8 @@ let print_type = (env, ty) => {
   };
 };
 
-let module_lens = (~program: Typedtree.typed_program, path: Path.t) => {
-  let vals = Modules.get_exports(~path, program);
+let module_lens = (decl: Types.module_declaration) => {
+  let vals = Modules.get_exports(decl);
```
phated, line 113:
```suggestion
  let vals = Modules.get_provides(decl);
```

compiler/src/language_server/sourcetree.re
```diff
@@ -137,11 +137,11 @@ module Make = (Ord: OrderableSegment) => {
 module type Sourcetree = {
   include SegmentTree with type point = Protocol.position;
   type node =
-    | Expression(Typedtree.expression, option(Types.value_description))
+    | Value(Env.t, Types.type_expr, Location.t)
     | Type(Typedtree.core_type)
     | Pattern(Typedtree.pattern)
```
phated, line 142: Why don't these also have `Location.t`?

compiler/src/language_server/sourcetree.re
```diff
@@ -224,14 +224,14 @@ module Sourcetree: Sourcetree = {
...
+        let process_value_description = (id, instance_ty, ty) => {
           // Never consider special idents when deciding to display generalized
           // types, i.e. always display instance types for lists
           Parsetree.(
             Identifier.(
               switch (id) {
-              | {txt: IdentName({txt: "[]" | "[...]"})} => None
-              | _ => Some(desc)
+              | {txt: IdentName({txt: "[]" | "[...]"})} => instance_ty
+              | _ => ty
```
phated, line 234: Boo, this is going to conflict with List AST node and the formatter again.

compiler/src/language_server/sourcetree.re
```diff
@@ -259,14 +266,73 @@ module Sourcetree: Sourcetree = {
...
+                        module_.txt,
+                        Env.find_module(module_.txt, None, exp.exp_env),
+                        module_.loc,
+                      ),
+                    ),
+                    (
+                      loc_to_interval(exp.exp_loc),
+                      Value(exp.exp_env, exp.exp_type, exp.exp_loc),
+                    ),
+                    ...switch (items) {
```
phated, line 296: I'm very surprised reason doesn't wrap the switch in parens

compiler/src/middle_end/linearize.re
```diff
@@ -259,6 +259,7 @@ let rec transl_imm =
     | Left(imm) => (imm, [])
     | Right((name, cexprs)) => (Imm.id(~loc, ~env, name), cexprs)
     }
+  | TExpUse(_)
   | TExpNull => (Imm.const(~loc, ~env, Const_void), [])
```
phated, line 263: Why is this a void?

phated (changes requested):
A few questions and I think we should get rid of `exports` when we see it.

compiler/src/middle_end/linearize.re
ospencer: All expressions in Grain produce a value, so when something "doesn't" it's void.

compiler/src/language_server/sourcetree.re
ospencer: Patterns have a location on `pat_loc`, so we don't need to duplicate it. Since I changed `Expression` to `Value`, it doesn't have a location property on it so I just added it to the variant.

compiler/src/middle_end/linearize.re
ospencer: Speaking of which, I don't think `PExpNull` is used anymore, so we can remove it.

phated (approved):

PR #1689, opened 2023-02-21
fix(compiler)!: Disallow multiple `provide`s of the same value

### LINKED ISSUE #1686: Compiler: Add well-formedness check to ensure users don't provide the same values twice
In `runtime/numbers.gr` the `coerceNumberToBigInt` function appears (and is provided) twice. This shouldn't be allowed.

ospencer:
Closes #1686


phated (approved):
Thanks for fixing this so quickly!

PR #1690, opened 2023-02-21
feat(stdlib)!: Add `**` (pow) operator to Pervasives module

spotandjake:
This pr moves power to the runtime and adds pervasives as an operator. I didnt see anywhere to test the operator or functions in the runtime but power is well tested in the number test suite.

As I noted in discord I decided to just remake this pr as all the tests changed there were a ton of merge of conflicts and this made things easier. Replaces: #1538 

Closes: https://github.com/grain-lang/grain/issues/1476


stdlib/pervasives.gr
```diff
@@ -246,6 +247,17 @@ provide { (/) }
...
 
+/**
+ * Computes the exponentiation of the given base and power.
+ *
+ * @param base: The base number
+ * @param power: The exponent number
+ * @returns The base raised to the given power
+ *
+ * @since v0.6.0
+ */
```
phated, line 258:
```suggestion
```
These are pulled from the original definition location with #1687 (which also combined these separate provides, so we'll want to rebase and update when it lands shortly

spotandjake: Will do, uhm one question though on the number libary how do we handle the versioning given its in the number libary as pow since `v0.5.x` but only in the runtime and pervasive since `v0.6.0`? do we just go with since `v0.5.x` x being the version it was implemented in I don't know off the top of my head and then do we just detail it having moved around in the history section?

phated: Probably add a `@history` entry saying it existed in `number.gr` in whatever version it was added.

phated (approved):
Excellent work @spotandjake 🎉

stdlib/pervasives.md
```diff
@@ -482,6 +482,39 @@ Returns:
 |----|-----------|
 |`Number`|The modulus of its operands|
 
+### Pervasives.**(**)**
```
phated, line 485: This shouldn't be your fault, but this doesn't seem to be escaped correctly. Not sure what is happening.

phated: I've merged in main to see if the graindoc is out of date.

spotandjake: I specifically rebuilt the compiler and regen'd the docs before pushing the last region commit so that probably isn't the case.

phated: You are correct. Markdown.bold wasn't be used in all the cases where we were bolding things 🤦

phated: Fixed in main since #1695 landed

spotandjake:
Regenerated the docs, your fix looks to have worked now.

phated (approved):

PR #1694, opened 2023-02-21
Added 28 new examples for stdlib Array that had no examples

mahSource:
for void fn's used assert

### DIFF
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -711,7 +711,7 @@ provide let findIndex = (fn, array) => {
  * @returns The new array containing all pairs of `(a, b)`
  * 
  * @example
- * Array.product([> 1, 2], [> 11, 12]) // [> (1, 11), (1, 12), (2, 11), (2, 12)]
+ * Array.product([> 1, 2], [> 11, 12]) == [> (1, 11), (1, 12), (2, 11), (2, 12)]
  *
  * @since v0.2.0
  */
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -668,7 +668,7 @@ provide let contains = (search, array) => {
  * @returns `Some(element)` containing the first value found or `None` otherwise
  * 
  * @example
- * Array.find(a => a == 7, [> 5, 6, 7, 8, 9]) // Some(7)
+ * Array.find(a => a == 7, [> 5, 6, 7, 8, 9]) == Some(7)
  *
  * @since v0.2.0
  */
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -255,7 +255,6 @@ provide let copy = array => {
  * @param n: The number of times to iterate the given array
  * @param array: The array to iterate
  * 
- * @example
 * @example
 *  let arr = [> 'a', 'b', 'c', 'd', 'e' ]
 *  let mut str = ""
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -469,7 +469,7 @@ provide let flatMap = (fn, array) => {
  * @returns `true` if all elements satify the condition or `false` otherwise
  * 
  * @example
- * Array.every(a => a > 5, [> 2, 4, 6, 8]) // false
+ * Array.every(a => a > 5, [> 2, 4, 6, 8]) == false
  *
  * @since v0.3.0
  */
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -419,7 +419,7 @@ provide let reduceRight = (fn, initial, array) => {
  * @returns The final accumulator returned from `fn`
  * 
  * @example
- * Array.reducei((a, b, i) => a + b + i, 100, [> 1, 2, 3, 4]) // 116
+ * Array.reducei((a, b, i) => a + b + i, 100, [> 1, 2, 3, 4]) == 116
  *
  * @since v0.3.0
  */
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -839,7 +839,7 @@ provide let filteri = (fn, array) => {
  * @returns The new array with only unique values
  * 
  * @example
- * Array.unique([> 1, 1, 1, 2, 3, 3, 4, 5, 6]) // [> 1, 2, 3, 4, 5, 6]
+ * Array.unique([> 1, 1, 1, 2, 3, 3, 4, 5, 6]) == [> 1, 2, 3, 4, 5, 6]
  *
  * @since v0.3.0
  */
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -321,7 +321,7 @@ provide let forEachi = (fn, array) => {
  * @returns The new array with mapped values
  * 
  * @example
- * Array.map(a => a * a, [> 7, 8, 9]) // [> 49, 64, 81]
+ * Array.map(a => a * a, [> 7, 8, 9]) == [> 49, 64, 81]
  * 
  * @since v0.1.0
  * @history v0.2.0: Argument order changed to data-last
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -42,6 +42,7 @@ let checkLength = length => {
  * @returns The number of elements in the array
  *
  * @example Array.length([> 1, 2, 3, 4, 5]) == 5
+ * 
  * @since v0.1.0
  */
 @unsafe
@@ -129,6 +130,7 @@ provide let init: (Number, Number -> a) -> Array<a> =
  * @param index: The index to access
  * @param array: The array to access
  * @returns The element from the array
+ * 
  * @example Array.get(1,[> 1, 2, 3, 4, 5]) == 2
  *
  * @since v0.1.0
@@ -147,6 +149,7 @@ provide let get = (index, array) => {
  * @param index: The index to update
  * @param value: The value to store
  * @param array: The array to update
+ * 
  * @example Array.set(1, 9, [> 1, 2, 3, 4, 5]) == [> 1, 9, 3, 4, 5]
  *
  * @since v0.1.0
@@ -163,6 +166,7 @@ provide let set = (index, value, array) => {
  * @param array1: The array containing elements to appear first
  * @param array2: The array containing elements to appear second
  * @returns The new array containing elements from `array1` followed by elements from `array2`
+ * 
  * @example Array.append([> 1, 2], [> 3, 4, 5]) == [> 1, 2, 3, 4, 5]
  *
  * @since v0.1.0
@@ -186,6 +190,7 @@ provide let append = (array1, array2) => {
  *
  * @param arrays: A list containing all arrays to combine
  * @returns The new array
+ * 
  * @example Array.concat([[> 1, 2], [> 3, 4], [> 5, 6]]) == [> 1, 2, 3, 4, 5, 6]
  *
  * @since v0.1.0
@@ -229,7 +234,12 @@ provide let concat = arrays => {
  *
  * @param array: The array to copy
  * @returns The new array containing the elements from the input
- *
+ * 
+ * @example
+ * let array1 = [> 9, 8, 7, 6, 5]
+ * let array2 = Array.copy(array1) 
+ * assert array1 == array2
+ * 
  * @since v0.1.0
  */
 provide let copy = array => {
@@ -242,7 +252,10 @@ provide let copy = array => {
  * @param fn: The iterator function to call with each element
  * @param n: The number of times to iterate the given array
  * @param array: The array to iterate
- *
+ * 
+ * @example
+ * Array.cycle(a => print(a), 2, [> 1, 2, 3]) // sep lines: 1 2 3 1 2 3
+ * 
  * @since v0.4.4
  */
 provide let cycle = (fn, n, array) => {
@@ -259,7 +272,10 @@ provide let cycle = (fn, n, array) => {
  *
  * @param fn: The iterator function to call with each element
  * @param array: The array to iterate
- *
+ * 
+ * @example
+ * Array.forEach(a => print(a), [> 1, 2, 3]) // sep lines: 1 2 3
+ * 
  * @since v0.1.0
  * @history v0.2.0: Argument order changed to data-last
  */
@@ -276,7 +292,10 @@ provide let forEach = (fn, array) => {
  *
  * @param fn: The iterator function to call with each element
  * @param array: The array to iterate
- *
+ * 
+ * @example
+ * Array.forEachi((a, i) => print(a + i * 2), [> 1, 2, 3]) // sep lines: 1 4 7
+ * 
  * @since v0.1.0
  * @history v0.2.0: Argument order changed to data-last
  */
@@ -294,7 +313,10 @@ provide let forEachi = (fn, array) => {
  * @param fn: The mapper function to call on each element, where the value returned will be used to initialize the element in the new array
  * @param array: The array to iterate
  * @returns The new array with mapped values
- *
+ * 
+ * @example
+ * Array.map(a => a * a, [> 7, 8, 9]) // [> 49, 64, 81]
+ * 
  * @since v0.1.0
  * @history v0.2.0: Argument order changed to data-last
  */
@@ -312,7 +334,10 @@ provide let map = (fn, array) => {
  * @param fn: The mapper function to call on each element, where the value returned will be used to initialize the element in the new array
  * @param array: The array to iterate
  * @returns The new array with mapped values
- *
+ * 
+ * @example
+ * Array.mapi((a, i) => a * i, [> 10, 20, 30]) // [> 0, 20, 60]
+ * 
  * @since v0.1.0
  */
 provide let mapi = (fn, array) => {
@@ -386,6 +411,9 @@ provide let reduceRight = (fn, initial, array) => {
  * @param initial: The initial value to use for the accumulator on the first iteration
  * @param array: The array to iterate
  * @returns The final accumulator returned from `fn`
+ * 
+ * @example
+ * Array.reducei((a, b, i) => a + b + i, 100, [> 1, 2, 3, 4]) // 116
  *
  * @since v0.3.0
  */
@@ -404,6 +432,9 @@ provide let reducei = (fn, initial, array) => {
  * @param fn: The function to be called on each element, where the value returned will be an array that gets appended to the new array
  * @param array: The array to iterate
  * @returns The new array
+ * 
+ * @example
+ * Array.flatMap(b => [> b, 2 * b], [> 1, 2, 6, 7]) // [> 1, 2, 2, 4, 6, 12, 7, 14]
  *
  * @since v0.3.0
  */
@@ -430,6 +461,9 @@ provide let flatMap = (fn, array) => {
  * @param fn: The function to call on each element, where the returned value indicates if the element satisfies the condition
  * @param array: The array to check
  * @returns `true` if all elements satify the condition or `false` otherwise
+ * 
+ * @example
+ * Array.every(a => a > 5, [> 2, 4, 6, 8]) // false
  *
  * @since v0.3.0
  */
@@ -450,6 +484,9 @@ provide let every = (fn, array) => {
  * @param array: The array to iterate
  * @returns `true` if one or more elements satify the condition or `false` otherwise
  *
+ * @example
+ * Array.some(a => a > 5, [> 2, 4, 6, 8]) // true
+ * 
  * @since v0.3.0
  */
 provide let some = (fn, array) => {
@@ -466,6 +503,11 @@ provide let some = (fn, array) => {
  *
  * @param value: The value replacing each element
  * @param array: The array to update
+ * 
+ * @example
+ * let array = [> 1, 2, 3, 4]
+ * Array.fill(9, array)
+ * assert array == [> 9, 9, 9, 9]
  *
  * @since v0.2.0
  */
@@ -485,7 +527,12 @@ provide let fill = (value, array) => {
  * @param start: The index to begin replacement
  * @param stop: The (exclusive) index to end replacement
  * @param array: The array to update
- *
+ * 
+ * @example 
+ * let array = [> 1, 2, 3, 4]
+ * Array.fillRange(9, 1, 3, my_frarray_array)
+ * assert array == [> 1, 9, 9, 4]
+ * 
  * @since v0.2.0
  */
 provide let fillRange = (value, start, stop, array) => {
@@ -516,6 +563,9 @@ provide let fillRange = (value, start, stop, array) => {
  *
  * @param array: The array to reverse
  * @returns The new array
+ * 
+ * @example
+ * Array.reverse([> 1, 2, 3, 4]) // [> 4,3,2,1]
  *
  * @since v0.4.0
  */
@@ -532,6 +582,9 @@ provide let reverse = array => {
  *
  * @param array: The array to convert
  * @returns The list containing all elements from the array
+ * 
+ * @example
+ * Array.toList([> 1, 2, 3, 4]) // [> 1, 2, 3, 4] becomes [1, 2, 3, 4]
  *
  * @since v0.1.0
  */
@@ -552,6 +605,9 @@ provide let toList = array => {
  *
  * @param list: The list to convert
  * @returns The array containing all elements from the list
+ * 
+ * @example
+ * Array.fromList([1, 2, 3, 4]) // [1, 2, 3, 4] becomes [> 1, 2, 3, 4]
  *
  * @since v0.1.0
  */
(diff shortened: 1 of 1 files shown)


stdlib/array.gr
```diff
@@ -229,7 +234,12 @@ provide let concat = arrays => {
  *
  * @param array: The array to copy
  * @returns The new array containing the elements from the input
- *
+ * 
+ * @example
+ * let array1 = [> 9, 8, 7, 6, 5]
+ * let array2 = Array.copy(array1) 
+ * assert array1 == array2
```
spotandjake, line 241: You would often use copy for copying the exact contents rather then a reference to that so it might be worth chaning this to:
```suggestion
* array1[0] = 2
* assert array1 == [> 2, 8, 7, 6, 5]
* assert array2 == [> 9, 8, 7, 6, 5]
``` 
This shows that when you modify `array1` its not being copied by reference but by value.
But wait for feedback from the core team before you apply that.

stdlib/array.gr
```diff
@@ -242,7 +252,10 @@ provide let copy = array => {
  * @param fn: The iterator function to call with each element
  * @param n: The number of times to iterate the given array
  * @param array: The array to iterate
- *
+ * 
+ * @example
+ * Array.cycle(a => print(a), 2, [> 1, 2, 3]) // sep lines: 1 2 3 1 2 3
```
spotandjake, line 257: A good example that might alternativly work well for this is something like:
```suggestion
* @example
*  let arr = [> 'a', 'b', 'c', 'd', 'e' ]
*  let mut str = ""
 * Array.cycle(a =>str = str ++ a, 2, arr)
 * assert str == "abcdeabcde"
```

This shows well what cycle is doing I think though. This could also be used on forEach

Again wait for feedback from the core team before applying any of these, if you choose to apply them.

stdlib/array.gr
```diff
@@ -294,7 +313,10 @@ provide let forEachi = (fn, array) => {
  * @param fn: The mapper function to call on each element, where the value returned will be used to initialize the element in the new array
  * @param array: The array to iterate
  * @returns The new array with mapped values
- *
+ * 
+ * @example
+ * Array.map(a => a * a, [> 7, 8, 9]) // [> 49, 64, 81]
```
spotandjake, line 318: For this one it would probably make sense to change it to:
```suggestion
 * Array.map(a => a * a, [> 7, 8, 9]) == [> 49, 64, 81]
```

stdlib/array.gr
```diff
@@ -312,7 +334,10 @@ provide let map = (fn, array) => {
  * @param fn: The mapper function to call on each element, where the value returned will be used to initialize the element in the new array
  * @param array: The array to iterate
  * @returns The new array with mapped values
- *
+ * 
+ * @example
+ * Array.mapi((a, i) => a * i, [> 10, 20, 30]) // [> 0, 20, 60]
```
spotandjake, line 339: Same As Above

stdlib/array.gr
```diff
@@ -386,6 +411,9 @@ provide let reduceRight = (fn, initial, array) => {
  * @param initial: The initial value to use for the accumulator on the first iteration
  * @param array: The array to iterate
  * @returns The final accumulator returned from `fn`
+ * 
+ * @example
+ * Array.reducei((a, b, i) => a + b + i, 100, [> 1, 2, 3, 4]) // 116
```
spotandjake, line 416: It would again probably make sense to change this to use `==` given it doesnt return void, to keep things in line with the rest of the stdlib docs:
```suggestion
 * Array.reducei((a, b, i) => a + b + i, 100, [> 1, 2, 3, 4]) == 116
```

stdlib/array.gr
```diff
@@ -404,6 +432,9 @@ provide let reducei = (fn, initial, array) => {
  * @param fn: The function to be called on each element, where the value returned will be an array that gets appended to the new array
  * @param array: The array to iterate
  * @returns The new array
+ * 
+ * @example
+ * Array.flatMap(b => [> b, 2 * b], [> 1, 2, 6, 7]) // [> 1, 2, 2, 4, 6, 12, 7, 14]
```
spotandjake, line 437: Same As Above
```suggestion
 * Array.flatMap(b => [> b, 2 * b], [> 1, 2, 6, 7]) == [> 1, 2, 2, 4, 6, 12, 7, 14]
```

stdlib/array.gr
```diff
@@ -430,6 +461,9 @@ provide let flatMap = (fn, array) => {
  * @param fn: The function to call on each element, where the returned value indicates if the element satisfies the condition
  * @param array: The array to check
  * @returns `true` if all elements satify the condition or `false` otherwise
+ * 
+ * @example
+ * Array.every(a => a > 5, [> 2, 4, 6, 8]) // false
```
spotandjake, line 466: Same As Above
```suggestion
 * Array.every(a => a > 5, [> 2, 4, 6, 8]) == false
```

stdlib/array.gr
```diff
@@ -450,6 +484,9 @@ provide let every = (fn, array) => {
  * @param array: The array to iterate
  * @returns `true` if one or more elements satify the condition or `false` otherwise
  *
+ * @example
+ * Array.some(a => a > 5, [> 2, 4, 6, 8]) // true
```
spotandjake, line 488: Same As Above:
```suggestion
 * Array.some(a => a > 5, [> 2, 4, 6, 8]) == true
```

stdlib/array.gr
```diff
@@ -516,6 +563,9 @@ provide let fillRange = (value, start, stop, array) => {
  *
  * @param array: The array to reverse
  * @returns The new array
+ * 
+ * @example
+ * Array.reverse([> 1, 2, 3, 4]) // [> 4,3,2,1]
```
spotandjake, line 568: Same As Above:
```suggestion
 * Array.reverse([> 1, 2, 3, 4]) == [> 4,3,2,1]
```

stdlib/array.gr
```diff
@@ -532,6 +582,9 @@ provide let reverse = array => {
  *
  * @param array: The array to convert
  * @returns The list containing all elements from the array
+ * 
+ * @example
+ * Array.toList([> 1, 2, 3, 4]) // [> 1, 2, 3, 4] becomes [1, 2, 3, 4]
```
spotandjake, line 587: Same As Above:
```suggestion
 * Array.toList([> 1, 2, 3, 4]) == [1, 2, 3, 4]
```

stdlib/array.gr
```diff
@@ -552,6 +605,9 @@ provide let toList = array => {
  *
  * @param list: The list to convert
  * @returns The array containing all elements from the list
+ * 
+ * @example
+ * Array.fromList([1, 2, 3, 4]) // [1, 2, 3, 4] becomes [> 1, 2, 3, 4]
```
spotandjake, line 610: same As Above:
```suggestion
 * Array.fromList([1, 2, 3, 4]) == [> 1, 2, 3, 4]
```

stdlib/array.gr
```diff
@@ -583,6 +639,9 @@ provide let fromList = list => {
  * @param search: The value to compare
  * @param array: The array to inspect
  * @returns `true` if the value exists in the array or `false` otherwise
+ * 
+ * @example
+ * Array.contains(9, [> 5, 6, 7, 8, 9]) // true
```
spotandjake, line 644: Same As Above:
```suggestion
 * Array.contains(9, [> 5, 6, 7, 8, 9]) == true
```

stdlib/array.gr
```diff
@@ -601,6 +660,9 @@ provide let contains = (search, array) => {
  * @param fn: The function to call on each element, where the returned value indicates if the element satisfies the condition
  * @param array: The array to search
  * @returns `Some(element)` containing the first value found or `None` otherwise
+ * 
+ * @example
+ * Array.find(a => a == 7, [> 5, 6, 7, 8, 9]) // Some(7)
```
spotandjake, line 665: Same As Above:
```suggestion
 * Array.find(a => a == 7, [> 5, 6, 7, 8, 9]) == Some(7)
```

might also make sense to add a counter example here if it isnt found it returns `None`

stdlib/array.gr
```diff
@@ -620,6 +682,9 @@ provide let find = (fn, array) => {
  * @param fn: The function to call on each element, where the returned value indicates if the element satisfies the condition
  * @param array: The array to search
  * @returns `Some(index)` containing the index of the first element found or `None` otherwise
+ * 
+ * @example
+ * Array.findIndex(a => a == 7, [> 5, 6, 7, 8, 9]) // Some(2)
```
spotandjake, line 687: Same As Above:
```suggestion
 * Array.findIndex(a => a == 7, [> 5, 6, 7, 8, 9]) == Some(2)
```
Might also make sense to add a counter example

stdlib/array.gr
```diff
@@ -638,6 +703,9 @@ provide let findIndex = (fn, array) => {
  * @param array1: The array to provide values for the first tuple element
  * @param array2: The array to provide values for the second tuple element
  * @returns The new array containing all pairs of `(a, b)`
+ * 
+ * @example
+ * Array.product([> 1, 2], [> 11, 12]) // [> (1, 11), (1, 12), (2, 11), (2, 12)]
```
spotandjake, line 708: Same As Above:
```suggestion
 * Array.product([> 1, 2], [> 11, 12]) == [> (1, 11), (1, 12), (2, 11), (2, 12)]
```

stdlib/array.gr
```diff
@@ -660,6 +728,9 @@ provide let product = (array1: Array<a>, array2: Array<b>) => {
  * @param fn: The function to call on each element, where the returned value indicates if the element satisfies the condition
  * @param array: The array to iterate
  * @returns The total number of elements that satisfy the condition
+ * 
+ * @example
+ * Array.count(a => a >= 7, [> 5, 6, 7, 8, 9]) // 3
```
spotandjake, line 733: Same As Above:

```suggestion
 * Array.count(a => a >= 7, [> 5, 6, 7, 8, 9]) == 3
```

stdlib/array.gr
```diff
@@ -681,6 +752,9 @@ provide let count = (fn, array) => {
  * @param fn: The function to call on each element, where the returned value indicates if the element satisfies the condition
  * @param array: The array to iterate
  * @returns The total number of elements that satisfy the condition
+ * 
+ * @example
+ * Array.counti((a, i) => a >= 7 && i > 2, [> 5, 6, 7, 8, 9]) // 2
```
spotandjake, line 757: Same As Above:
```suggestion
 * Array.counti((a, i) => a >= 7 && i > 2, [> 5, 6, 7, 8, 9]) == 2
```

stdlib/array.gr
```diff
@@ -703,6 +777,9 @@ provide let counti = (fn, array) => {
  * @param fn: The function to call on each element, where the returned value indicates if the element satisfies the condition
  * @param array: The array to iterate
  * @returns The new array containing elements where `fn` returned `true`
+ * 
+ * @example
+ * Array.filter(a => a < 8, [> 5, 6, 7, 8, 9]) // [> 5, 6, 7]
```
spotandjake, line 782: Same As Above:
```suggestion
 * Array.filter(a => a < 8, [> 5, 6, 7, 8, 9]) == [> 5, 6, 7]
```

stdlib/array.gr
```diff
@@ -728,6 +805,9 @@ provide let filter = (fn, array) => {
  * @param fn: The function to call on each element, where the returned value indicates if the element satisfies the condition
  * @param array: The array to iterate
  * @returns The new array containing elements where `fn` returned `true`
+ * 
+ * @example
+ * Array.filteri((a, i) => a < 9 && i > 0, [> 5, 6, 7, 8, 9]) // [> 6, 7, 8]
```
spotandjake, line 810: Same As Above:
```suggestion
 * Array.filteri((a, i) => a < 9 && i > 0, [> 5, 6, 7, 8, 9]) == [> 6, 7, 8]
```

stdlib/array.gr
```diff
@@ -751,6 +831,9 @@ provide let filteri = (fn, array) => {
  *
  * @param array: The array to filter
  * @returns The new array with only unique values
+ * 
+ * @example
+ * Array.unique([> 1, 1, 1, 2, 3, 3, 4, 5, 6]) // [> 1, 2, 3, 4, 5, 6]
```
spotandjake, line 836: Same As Above:

```suggestion
 * Array.unique([> 1, 1, 1, 2, 3, 3, 4, 5, 6]) == [> 1, 2, 3, 4, 5, 6]
```

stdlib/array.gr
```diff
@@ -769,6 +852,9 @@ provide let unique = array => {
  * @returns The new array containing indexed pairs of `(a, b)`
  *
  * @throws Failure(String): When the arrays have different sizes
+ * 
+ * @example
+ * Array.zip([> 1, 2, 3], [> 11, 12, 13]) // [> (1, 11), (2, 12), (3, 13)]
```
spotandjake, line 857: Same As Above:
```suggestion
 * Array.zip([> 1, 2, 3], [> 11, 12, 13]) == [> (1, 11), (2, 12), (3, 13)]
```

stdlib/array.gr
```diff
@@ -816,6 +902,9 @@ provide let zipWith = (fn, array1: Array<a>, array2: Array<b>) => {
  *
  * @param array: The array of tuples to split
  * @returns An array containing all elements from the first tuple element, and an array containing all elements from the second tuple element
+ * 
+ * @example
+ * Array.unzip([> (1, 2), (3, 4), (5, 6)]) // ([> 1, 3, 5], [> 2, 4, 6])
```
spotandjake, line 907: Same As Above:

```suggestion
 * Array.unzip([> (1, 2), (3, 4), (5, 6)]) == ([> 1, 3, 5], [> 2, 4, 6])
```

stdlib/array.gr
```diff
@@ -841,6 +930,9 @@ provide let unzip = array => {
  * @param separator: The separator to insert between items in the string
  * @param items: The input strings
  * @returns The concatenated string
+ * 
+ * @example
+ * Array.join(" ", [> "The", "quick", "brown", "fox"]) // The quick brown fox
```
spotandjake, line 935: Same As Above:

```suggestion
 * Array.join(" ", [> "The", "quick", "brown", "fox"]) == "The quick brown fox"
```

stdlib/array.gr
```diff
@@ -880,6 +972,9 @@ let wrapNegativeIndex = (arrLen, idx) => {
  * @param endIndex: The index of the array where the slice will end (exclusive)
  * @param array: The array to be sliced
  * @returns The subset of the array that was sliced
+ * 
+ * @example
+ * Array.slice(1, 3, [> 5, 6, 7, 8, 9]) // [> 6, 7]
```
spotandjake, line 977: Same As Above:
```suggestion
 * Array.slice(1, 3, [> 5, 6, 7, 8, 9]) == [> 6, 7]
```

spotandjake (commented):
On The Premise of keeping in line with the rest of the pattern of the stdlib I think the pattern used here isnt quite right. It should be more like:
```
Returns a value -> uses `==`
Returns Void -> `comment`
Returns Void but modifies in place -> `multiline and assert`
```

I wouldnt merge any of these though before the core team just verifies this.

stdlib/array.gr
mahSource: Thanks Jake.  Again I think the single line with the embedded print(a) clearly shows to the person learning this function what is happening.  I did use sep lines: on I think three or so examples so I understand this might seem different.  But it shows that these are processed like a for loop when you cycle.  I believe this is an elegant way to demonstrate this in one line of code.

stdlib/array.gr
mahSource: Thanks Jake.  I specifically asked Oscar this yesterday and he said for these simple ones that an in line comment was preferred.  You have missed that discussion.  I personally do not like using the == without an assert comment and then an inline showing it is true.  I sometimes feel that the use of == alone is not good, what if it is false like when comparing strings in some languages.  The in line with the resultant Array<b> is super clear.

stdlib/array.gr
mahSource: Same as my comment on .map

stdlib/array.gr
mahSource: Same reply as on .map.  I believe the == is not a proper use of == unless you in line // true.  Why not just show the value 116 like Oscar recommended.

stdlib/array.gr
mahSource: Same counter comment as in .map, etc.

stdlib/array.gr
mahSource: Same counter comment as in .map et. al.

stdlib/array.gr
mahSource: Same counter comment as in .map et. al.

stdlib/array.gr
mahSource: Same counter comment as in .map et. al.

stdlib/array.gr
mahSource: As we discussed yesterday....the inline is better for the noob or someone new to Grain.  You can clearly see the array format [<...] side-by-side to the list format [...].  Also by just putting the == without an inline ' // true' you have not stated what the answer is.

stdlib/array.gr
mahSource: See my comment on .toList.  Prefer the as submitted format for clarity and completness.

stdlib/array.gr
mahSource: I would prefer the in line true.  It starts to get confusing when you are comparing a bool vs. bool with == and not providing the answer with another in line comment that true == true is true.  The as submitted with the in line is crystal clear and immediately you can see that the fn is returning a bool.

stdlib/array.gr
mahSource: See previous comment about discussion with Oscar and simplicity of in line.  Again if one is going to use a Boolean comparison in an example you must provide the answer and not infer it is true...or ad the assert.....

stdlib/array.gr
mahSource: See comment on .find I would recommend keeping the in line vs. Boolean without an assert or return from the == operation.

stdlib/array.gr
mahSource: Same comment as before.  This is not returning void so simply in line the result.....

stdlib/array.gr
mahSource: Same comment as before. This is not returning void so simply in line the result.....

stdlib/array.gr
mahSource: Same comment as before. This is not returning void so simply in line the result.....

stdlib/array.gr
mahSource: Same comment as before. This is not returning void so simply in line the result.....

stdlib/array.gr
mahSource: Same comment as before. This is not returning void so simply in line the result.....

stdlib/array.gr
mahSource: Same comment as before. This is not returning void so simply in line the result.....

stdlib/array.gr
mahSource: Same comment as before. This is not returning void so simply in line the result.....

stdlib/array.gr
mahSource: Same comment as before. This is not returning void so simply in line the result.....

stdlib/array.gr
mahSource: Same comment as before. This is not returning void so simply in line the result.....

stdlib/array.gr
phated: I'm going to have to **strongly** disagree with you here @mahSource. As we've just seen two examples above, you used a comment to show what is printed. In this example, these values aren't printed. This is confusing! In old versions of grain, we'd print the last line of your program, so it made sense, but we switch to using these equality comparisons so it was clear the value was being returned **plus** comments are never syntax highlighted, so they really just stink.

Please use comparisons, especially equality, wherever possible.

mahSource: Thanks @phated. The ones above have an embedded print() fn in the fn....so they print separate lines in reality as they functions through the cycle.  A different case for the ones that print on one line.  Is it common to use == without an assert or showing the result of the equality?

phated: Yes. That has been used in all our documentation. It's the same as `1 == 1` but with whatever (arrays in this case). Assert throws when the value is wrong and returns void, so isn't really helpful for real world examples.

mahSource: Hi @phated Here is another possible issue with using an operator vs. a method with == if the reader does not understand what == is implying in the example. https://www.geeksforgeeks.org/difference-between-and-equals-method-in-java/
I understand this is one case for Java but people coming to Grain might have different perceptions of what == really is from their experience.  Somehow it would be good to make it clear imho.

mahSource: I agree with your comment on assert...could add confusion to confusion.  @phated  If there is sunk cost in the rest of Docs with the use of == already then there would be consistency problem to change that.  So it sounds like when making up examples that we should use == and that (a) the reader should assume we are suggesting that the result of the operator is true and (b) that it will work with any type (i.e. not an issue with Java e.g. with strings) and is (c) immune from value vs any by reference artifacts (may not be an issue with Grain).  Also please confirm if you want any in line type outputs in any examples.  I am just trying to understand fully before spending a lot of time to create missing examples or help clean up ones that don't make sense, e.g. using == with a void return type.

mahSource: I just saw @spotandjake super helpful summary at the top of this PR (I pasted below too)....I agree with it and will use this unless @phated has an issue with it, please let me know when you can since I have some time today to work on the next stdlib.  Thanks!
From Jake:
-- Returns a value -> uses `==`
-- Returns Void -> `comment`
-- Returns Void but modifies in place -> `multiline and assert`
Nicely done Jake!

stdlib/array.gr
spotandjake: Yeha I see what you mean there maybe someone from the core team has a better example my only thought on that is the reason to use `copy` over just doing `let a = b` is that it clones the it. and your example doesnt really show that very well. There might be a better example though that shows its a clone rather then just a copy.

stdlib/array.gr
```diff
@@ -242,7 +254,14 @@ provide let copy = array => {
  * @param fn: The iterator function to call with each element
  * @param n: The number of times to iterate the given array
  * @param array: The array to iterate
- *
+ * 
+ * @example
```
spotandjake, line 258: This has two `@examples` which I dont think is intented
```suggestion
```

stdlib/array.gr
```diff
@@ -242,7 +254,14 @@ provide let copy = array => {
...
+* @example
```
spotandjake, line 259: It looks like the indenting here might have gotten messed up. It could just be github beind weird though too. IT looks like the `*`s are in a space which is different then the rest and also there is two spaces between some of the `*`s and the code which is different then the rest of the examples

stdlib/array.gr
```diff
@@ -386,6 +417,9 @@ provide let reduceRight = (fn, initial, array) => {
  * @param initial: The initial value to use for the accumulator on the first iteration
  * @param array: The array to iterate
  * @returns The final accumulator returned from `fn`
+ * 
+ * @example
+ * Array.reducei((a, b, i) => a + b + i, 100, [> 1, 2, 3, 4]) == 116
```
spotandjake, line 422: Just a thought here really up to you I find building strings makes things a lot more clear on the order this runs like if you did something like 
```
Array.reducei((a, b, i) => a ++ b ++ toString(i), "A", [>"B", "C", "D"])
```
it just makes the order a bit more clear while still being a relativly simple example `reduce` is hard though and some confusion can be left either way this might be best added as a second example on this as well, I would wait for @phated to comment on thougths on this though as it is slightly more complex.

stdlib/array.gr
```diff
@@ -485,7 +533,12 @@ provide let fill = (value, array) => {
  * @param start: The index to begin replacement
  * @param stop: The (exclusive) index to end replacement
  * @param array: The array to update
- *
+ * 
+ * @example 
+ * let array = [> 1, 2, 3, 4]
+ * Array.fillRange(9, 1, 3, my_frarray_array)
```
spotandjake, line 539:
```suggestion
 * Array.fillRange(9, 1, 3, array)
```
You forgot to change the array name here.

spotandjake (commented):
Just a few more little comments It looks good though in my opinion.

Just a note so its not forgotton about but someone needs to generate the `.md` files before this can be merged.

mahSource:
.

PR #1696, opened 2023-02-22
added 39 examples stdlib List that had no examples

mahSource:
All of these use the == 
I also added a couple examples for several of the fn's that showed off Some(..) vs. None and a manual fn vs. using compare, etc.

### DIFF
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -43,6 +43,9 @@ provide let init = (length, fn) => {
  * @param list: The list to inspect
  * @returns The number of elements in the list
  *
+ * @example
+ * List.length([1, 2, 3, 4]) == 4
+ *
  * @since v0.1.0
  * @history v0.2.0: Made the function tail-recursive
  */
@@ -61,6 +64,9 @@ provide let length = list => {
  *
  * @param list: The list to reverse
  * @returns The new list
+ *
+ * @example
+ * List.reverse([1, 2, 3, 4]) == [4, 3, 2, 1]
  *
  * @since v0.1.0
  */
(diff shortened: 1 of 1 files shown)


stdlib/list.gr
```diff
@@ -742,6 +826,11 @@ provide let rec takeWhile = (fn, list) => {
  * @param list: The list to search
  * @returns `Some(element)` containing the first value found or `None` otherwise
  *
+ * @example
+ * List.find(a => a < 7 && a > 4, [1, 2, 3, 8, 9, 5, 1, 8]) == Some(5)
```
spotandjake, line 830: This might be better off using chars. just so its substantially different then `findIndex` it just seems like needless confusion not that the example doesnt work chars would just be more clear. maybe something like: 

```
List.filter(a => a == 'a',  [ 'b', 'b', 'a', 'd', 'a']) == Some(2)
```

The simpler example also makes it more clear while the examples should not be the number one source of knowledge and if you read the description it is clear what it does the more info you can give with an example the better.

stdlib/list.gr
```diff
@@ -821,6 +921,9 @@ provide let sub = (start, length, list) => {
  * @param list: The list to combine
  * @returns The combined elements with the separator between each
  *
+ * @example
+ * List.join(" ", ["The", "quick", "brown", "fox"]) == The quick brown fox
```
spotandjake, line 925: This needs string qoutes
```suggestion
 * List.join(" ", ["The", "quick", "brown", "fox"]) == "The quick brown fox"
```

spotandjake (commented):
Just a few things that might make it slightly better it looks pretty good though imo.

mahSource:
Is there anything else I should do to move this one along.  Only my second PR ever so just double checking since I am hooked in via my GitHub UI (Desktop), the Web, and even my IDE and much is new.

ospencer:
Not at the moment! It can take us some time to get reviews in (we all work on this project in our spare time, after all) 🙂

mahSource:
.

Issue #303, opened 2020-09-12
How Should Float Equality Work?


belph:
At the moment, for our structural equality check, we simply do a `==` comparison on floating-point numbers. This is awkward, though, if we are, say, comparing a float32 and a float64, since a user might expect them to be equal. What's the correct course of action here? Some options:

- Leave it as-is, documenting it as "floating point equality is unstable". I think this is the approach most languages take, so I'm happy to take it too.
- Build in some tolerance threshold for equality checks. This will inevitably bite someone when they are working with tiny values, though.

indolering:
What about: `.1 + .2 ~== .3`?  You could also allow programmers to efficiency specify the level of equality that they care about: `.1 + .2 ==.prec(2) .3 `.

ospencer:
My vote would be to have `==` check if the values are strictly equal, but implement `Number.isClose`, `Float32.isClose`, and `Float64.isClose`, similar to Python's [`math.isclose`](https://docs.python.org/3/library/math.html#math.isclose) for checking if values are roughly similar.

Once we have that for a while and we're happy with it, I could see us adding some syntax similar to what @indolering posted.

PR #1699, opened 2023-02-23
feat(stdlib): Add `asin`, `acos`, `atan`, `isClose` to Number module

spotandjake:
This pr adds `Number.asin`, `Number.acos` and `Number.isClose` to the Number stdlib. I was just going to add `asin` and `acos` but given that the methods being used to calculate `asin` and `acos` are approximations I decided to also add `isClose` to make the testing more concise.

isClose is taken from https://peps.python.org/pep-0485/#proposed-implementation,
The license on it just says its in the public domain I think that means we can use it but I honestly dont know happy to find an alternative approach though if that isnt allowed.

The `asin` and `acos` functions are from:
https://git.musl-libc.org/cgit/musl/tree/src/math/asin.c
Just like `pow`.

This is a draft because `isClose` is blocked on optional Arguments for a nice api, once that gets merged I will go back and refactor the function to not use Options.

Blocked: #388 

Work for: #1017 

Note: if you guys are fine with merging in with a worse api based on options I am happy todo that too.

### DIFF
--- a/stdlib/number.gr
+++ b/stdlib/number.gr
@@ -314,9 +314,16 @@ provide let neg = (x: Number) => x * -1
  * @since v0.6.0
  */
 provide let isClose = (a, b, rel_tol, abs_tol) => {
-  if (a == Infinity && b == Infinity) {
+  if (a == Infinity && b == Infinity || a == -Infinity && b == -Infinity) {
     true
-  } else if (a == NaN || b == NaN || a == Infinity || b == Infinity) {
+  } else if (
+    a == NaN ||
+    b == NaN ||
+    a == Infinity ||
+    b == Infinity ||
+    a == -Infinity ||
+    b == -Infinity
+  ) {
     false
   } else {
     let rel_tol = match (rel_tol) {
--- a/stdlib/number.gr
+++ b/stdlib/number.gr
@@ -429,7 +429,6 @@ provide let isInfinite = (x: Number) => {
  * @param b: The second value
  * @param relativeTolerance: The maximum tolerance to use relative to the larger absolute value `a` or `b`
  * @param absoluteTolerance: The absolute tolerance to use, regardless of the values of `a` or `b`
- *
  * @returns `true` if the values are considered close to each other or `false` otherwise
  *
  * @since v0.6.0
@@ -633,10 +632,10 @@ let rf = z => {
   p / q
 }
 /**
- * Computes the inverse sine of the given angle
+ * Computes the inverse sine of the given angle.
  *
- * @param angle: A number between -1 and 1, representing the angle's sine value.
- * @returns The inverse sine (angle in radians between `-pi/2` and `pi/2`) of the given `angle`. The result will be `NaN` if the given `angle` is less than `-1` or greater than `1`
+ * @param angle: A number between -1 and 1, representing the angle's sine value
+ * @returns The inverse sine (angle in radians between `-pi/2` and `pi/2`) of the given `angle` or `NaN` if the given `angle` is not between`-1` and `1`
  *
  * @since v0.6.0
  */
@@ -697,10 +696,10 @@ provide let asin = angle => {
 }
 
 /**
- * Computes the inverse cosine of the given angle
+ * Computes the inverse cosine of the given angle.
  *
- * @param angle: A number between -1 and 1, representing the angle's cosine value.
- * @returnsThe inverse cosine (angle in radians between `-pi/2` and `pi/2`) of the given `angle`. The result will be `NaN` if the given `angle` is less than `-1` or greater than `1`
+ * @param angle: A number between -1 and 1, representing the angle's cosine value
+ * @returns The inverse cosine (angle in radians between `-pi/2` and `pi/2`) of the given `angle` or `NaN` if the given `angle` is not between`-1` and `1`
  *
  * @since v0.6.0
  */
@@ -772,10 +771,10 @@ provide let acos = angle => {
 }
 
 /**
- * Computes the inverse tangent of the given angle
+ * Computes the inverse tangent of the given angle.
  *
- * @param angle: A number between -1 and 1, representing the angle's tangent value.
- * @returns The inverse tangent (angle in radians between `-pi/2` and `pi/2`) of the given `angle`. The result will be `NaN` if the given `angle` is less than `-1` or greater than `1`
+ * @param angle: A number between -1 and 1, representing the angle's tangent value
+ * @returns The inverse tangent (angle in radians between `-pi/2` and `pi/2`) of the given `angle` or `NaN` if the given `angle` is not between`-1` and `1`
  *
  * @since v0.6.0
  */
--- a/stdlib/number.gr
+++ b/stdlib/number.gr
@@ -1,12 +1,13 @@
 /*
  * ====================================================
+ * Applies to all functions with a comment referring here.
+ *
  * Copyright (C) 2004 by Sun Microsystems, Inc. All rights reserved.
  *
  * Permission to use, copy, modify, and distribute this
  * software is freely granted, provided that this notice
  * is preserved.
  *
- * Applies to all functions with a comment referring here.
  * ====================================================
  */
 
@@ -306,31 +307,6 @@ provide let abs = (x: Number) => if (0 > x) x * -1 else x
 provide let neg = (x: Number) => x * -1
 
 /**
- * Computes the inverse sine of the given angle
- *
- * @param x: A number between -1 and 1, representing the angle's sine value.
- * @returns The inverse sine (angle in radians between `-pi/2` and `pi/2`) of x. If x is less than `-1` or greater than `1`, return NaN
- *
- * @since v0.6.0
- */
-provide let isClose = (a, b, rel_tol=1e-9, abs_tol=0.0) => {
-  if (a == Infinity && b == Infinity || a == -Infinity && b == -Infinity) {
-    true
-  } else if (
-    a == NaN ||
-    b == NaN ||
-    a == Infinity ||
-    b == Infinity ||
-    a == -Infinity ||
-    b == -Infinity
-  ) {
-    false
-  } else {
-    abs(a - b) <= max(rel_tol * max(abs(a), abs(b)), abs_tol)
-  }
-}
-
-/**
  * Checks if a number is a floating point value.
  *
  * @param x: The number to check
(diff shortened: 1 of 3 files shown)


stdlib/number.gr
```diff
@@ -287,6 +299,32 @@ provide let abs = (x: Number) => if (0 > x) x * -1 else x
...
+/**
+ * Computes the inverse sine of the given angle
+ *
+ * @param x: A number between -1 and 1, representing the angle's sine value.
+ * @returns The inverse sine (angle in radians between `-pi/2` and `pi/2`) of x. If x is less than `-1` or greater than `1`, return NaN
+ *
+ * @since v0.6.0
+ */
+provide let isClose = (a, b, rel_tol, abs_tol) => {
+  if (a == Infinity && b == Infinity) {
```
ospencer, line 311: What about `-Infinity`?

spotandjake: 🤦

spotandjake:
Note: refactor any tests that use `<>` to arround floating points to use `isClose` after we are unblocked.

stdlib/number.gr
```diff
@@ -292,6 +304,31 @@ provide let abs = (x: Number) => if (0 > x) x * -1 else x
...
+    b == NaN ||
+    a == Infinity ||
+    b == Infinity ||
+    a == -Infinity ||
+    b == -Infinity
+  ) {
+    false
+  } else {
+    abs(a - b) <= max(rel_tol * max(abs(a), abs(b)), abs_tol)
+  }
```
ospencer, line 329: Could maybe simplify this a bit, and short circuit too:
```suggestion
  if (a == b) {
    true
  } else if (isFinite(a) && isFinite(b)) {
    abs(a - b) <= max(rel_tol * max(abs(a), abs(b)), abs_tol)
  } else {
    // NaN and infinities which were not equal
    false
  }
```

stdlib/number.gr
```diff
@@ -404,6 +417,30 @@ provide let isInfinite = (x: Number) => {
   }
 }
 
+/**
+ * Determines weather two values are considered close to each other using a relative and absolute tolerance.
+ *
+ * @param a: The first value
+ * @param b: The second value
+ * @param relativeTolerance: The relative tolerance to use
+ * @param absoluteTolerance: The absolute tolerance to use
```
ospencer, line 426: These should probably be more descriptive because it's not clear what they do.

spotandjake: I tried to improve these and got them a bit better but i dont know if they are good enough yet,

https://docs.python.org/3/library/math.html I was looking at python's docs for the description.

spotandjake:
I think this might also close: #303 if we are deciding to with oscar's approach anyway.

spotandjake:
This has been unblocked and the docs have been regenerated

stdlib/number.gr
```diff
@@ -409,6 +422,30 @@ provide let isInfinite = (x: Number) => {
   }
 }
 
+/**
+ * Determines weather two values are considered close to each other using a relative and absolute tolerance.
```
ospencer, line 426:
```suggestion
 * Determines whether two values are considered close to each other using a relative and absolute tolerance.
```

ospencer (commented):
I love the new descriptions! I think they make a lot of sense now. Just one tiny change and this looks good from me.

stdlib/number.gr
spotandjake: Fixed that

ospencer (approved):
Stamp of approval from me! I'll let the stdlib expert @phated take a look and hit the merge button if he's happy with the docs.

phated:
@spotandjake I only see a copyright notice, but not a license. What's the license on the code you referred to?

spotandjake:
> @spotandjake I only see a copyright notice, but not a license. What's the license on the code you referred to? 

From my understanding as it's in the Public domain it doesn't need a license to be used. It seems to be unlicensed though.

I'm happy to use a different implementation if we need to though.

stdlib/number.gr
```diff
@@ -574,6 +611,260 @@ provide let tan = (radians: Number) => {
...
+  let c = (z - df * df) / (s + df)
+  let w = rf(z) * s + c
+  return WasmI32.toGrain(newFloat64(2.0W * (df + w))): Number
+}
+
+/**
+ * Computes the inverse tangent of the given angle
+ *
+ * @param angle: A number between -1 and 1, representing the angle's tangent value.
+ * @returns The inverse tangent (angle in radians between `-pi/2` and `pi/2`) of x. If x is less than `-1` or greater than `1`, returns NaN
```
phated, line 778: This is for all the returns in the file.

What is "x"? We usually say something like "of the given angle."

Also, the "if x is less" sentence needs to be something like "The result will be NaN if the given angle is not between -1 and 1." Or something similar

phated (changes requested):
Comment about docs. I still need to double check the musl libc license

stdlib/number.gr
spotandjake: I made that change

stdlib/number.gr
```diff
@@ -409,6 +422,30 @@ provide let isInfinite = (x: Number) => {
...
 }
 
+/**
+ * Determines whether two values are considered close to each other using a relative and absolute tolerance.
+ *
+ * @param a: The first value
+ * @param b: The second value
+ * @param relativeTolerance: The maximum tolerance to use relative to the larger absolute value `a` or `b`
+ * @param absoluteTolerance: The absolute tolerance to use, regardless of the values of `a` or `b`
+ *
```
phated, line 432:
```suggestion
```

stdlib/number.gr
```diff
@@ -574,6 +611,260 @@ provide let tan = (radians: Number) => {
...
+  let qS2 = 2.02094576023350569471e+00W /* 0x40002AE5, 0x9C598AC8 */
+  let qS3 = -6.88283971605453293030e-01W /* 0xBFE6066C, 0x1B8D0159 */
+  let qS4 = 7.70381505559019352791e-02W /* 0x3FB3B8C5, 0xB12E9282 */
+  // Calculations
+  let p = z * (pS0 + z * (pS1 + z * (pS2 + z * (pS3 + z * (pS4 + z * pS5)))))
+  let q = 1.0W + z * (qS1 + z * (qS2 + z * (qS3 + z * qS4)))
+  p / q
+}
+/**
+ * Computes the inverse sine of the given angle
```
phated, line 636:
```suggestion
 * Computes the inverse sine of the given angle.
```
Period on top level descriptions, no period on arg descriptions

stdlib/number.gr
```diff
@@ -574,6 +611,260 @@ provide let tan = (radians: Number) => {
...
+ *
+ * @param angle: A number between -1 and 1, representing the angle's sine value.
```
phated, line 638:
```suggestion
 * @param angle: A number between -1 and 1, representing the angle's sine value
```

stdlib/number.gr
```diff
@@ -574,6 +611,260 @@ provide let tan = (radians: Number) => {
...
+ * @returns The inverse sine (angle in radians between `-pi/2` and `pi/2`) of the given `angle`. The result will be `NaN` if the given `angle` is less than `-1` or greater than `1`
```
phated, line 639:
```suggestion
 * @returns The inverse sine (angle in radians between `-pi/2` and `pi/2`) of the given `angle` or `NaN` if the given `angle` is not between `-1` and `1`
```
How about this?

stdlib/number.gr
```diff
@@ -574,6 +611,260 @@ provide let tan = (radians: Number) => {
...
+          2.0W * c) -
+        (0.5W * pio2_hi -
+          2.0W * f))
+  }
+  x = WasmF64.copySign(x, origAngle)
+  return WasmI32.toGrain(newFloat64(x)): Number
+}
+
+/**
+ * Computes the inverse cosine of the given angle
```
phated, line 700:
```suggestion
 * Computes the inverse cosine of the given angle.
```

stdlib/number.gr
```diff
@@ -574,6 +611,260 @@ provide let tan = (radians: Number) => {
...
+ *
+ * @param angle: A number between -1 and 1, representing the angle's cosine value.
```
phated, line 702:
```suggestion
 * @param angle: A number between -1 and 1, representing the angle's cosine value
```

stdlib/number.gr
```diff
@@ -574,6 +611,260 @@ provide let tan = (radians: Number) => {
...
+ * @returnsThe inverse cosine (angle in radians between `-pi/2` and `pi/2`) of the given `angle`. The result will be `NaN` if the given `angle` is less than `-1` or greater than `1`
```
phated, line 703:
```suggestion
 * @returns The inverse cosine (angle in radians between `-pi/2` and `pi/2`) of the given `angle` or `NaN` if the given `angle` is not between `-1` and `1`
```

stdlib/number.gr
```diff
@@ -574,6 +611,260 @@ provide let tan = (radians: Number) => {
...
+  let df = WasmF64.reinterpretI64(
+    WasmI64.reinterpretF64(s) & 0xFFFFFFFF00000000N
+  )
+  let c = (z - df * df) / (s + df)
+  let w = rf(z) * s + c
+  return WasmI32.toGrain(newFloat64(2.0W * (df + w))): Number
+}
+
+/**
+ * Computes the inverse tangent of the given angle
```
phated, line 775:
```suggestion
 * Computes the inverse tangent of the given angle.
```

stdlib/number.gr
```diff
@@ -574,6 +611,260 @@ provide let tan = (radians: Number) => {
...
+ *
+ * @param angle: A number between -1 and 1, representing the angle's tangent value.
```
phated, line 777:
```suggestion
 * @param angle: A number between -1 and 1, representing the angle's tangent value
```

stdlib/number.gr
```diff
@@ -574,6 +611,260 @@ provide let tan = (radians: Number) => {
...
+ * @returns The inverse tangent (angle in radians between `-pi/2` and `pi/2`) of the given `angle`. The result will be `NaN` if the given `angle` is less than `-1` or greater than `1`
```
phated, line 778:
```suggestion
 * @returns The inverse tangent (angle in radians between `-pi/2` and `pi/2`) of the given `angle` or `NaN` if the given `angle` is not between `-1` and `1`
```

phated (changes requested):
More docs changes

stdlib/number.gr
spotandjake: Love It

spotandjake:
Made those changes

phated (approved):

PR #1705, opened 2023-02-28
fix: Fix String.ReplaceAll

### LINKED ISSUE #1700: `String.replaceAll` gives out of bounds error with recursive patterns
Hi,

this code:
```
include "string"

let input = "MeowMeowMeowMeowMeowMeowMeowMeow"
print("input: " ++ input)

let expected = "MeowMeowMeowMeow"
print("expected: " ++ expected)

let result = String.replaceAll("MeowMeow", "Meow", input)
print("result: " ++ result)
```

fails to run:
```
input: MeowMeowMeowMeowMeowMeowMeowMeow
expected: MeowMeowMeowMeow
wasm://wasm/001025ea:1


RuntimeError: memory access out of bounds
    at wasm://wasm/001025ea:wasm-function[601]:0x39739
    at wasm://wasm/001025ea:wasm-function[630]:0x40765
    at wasm://wasm/001025ea:wasm-function[632]:0x40922
    at t.start (/home/mars/Code/grain/js-runner/dist/grain-runner.js:1:18229)
    at pn.start (/home/mars/Code/grain/js-runner/dist/grain-runner.js:1:154634)
    at gn.runFile (/home/mars/Code/grain/js-runner/dist/grain-runner.js:1:158723)
    at async run (/home/mars/Code/grain/cli/bin/run.js:18:5)
```

My desired usecase is replacing backslash escape codes from string literals with backslashes in my [mal implementation](https://github.com/kanaka/mal).

spotandjake:
This pr fixes #1700 , The issue had nothing todo with the recursive pattern given how ReplaceAll only replaces the strings after it finds them all the real issue was that in the inital implementation it was never accounted for a scenario of the distance betweeen two matching patterns being smaller then the matching length this new pr changes that so when we find a match we increment by the length of the pattern.

Essentially what was happening is in the example below we were matching three sets of `Meow` when the second two `Meow`s should be eaten by the first match and the last match.
```
String.replaceAll("MeowMeow", "Meow", "MeowMeowMeowMeow")
```

### DIFF
--- a/compiler/test/stdlib/string.test.gr
+++ b/compiler/test/stdlib/string.test.gr
@@ -238,6 +238,7 @@ assert String.replaceAll("/", "\/", "/test/test/test/") ==
   "\/test\/test\/test\/"
 assert String.replaceAll(",", "|", "test,test,test") == "test|test|test"
 
+assert String.replaceAll("MeowMeow", "Meow", "MeowMeowMeowMeow") == "MeowMeow"
 let processBytes = b => {
   let ret = Array.make(Bytes.length(b), 0l)
   for (let mut i = 0; i < Bytes.length(b); i += 1) {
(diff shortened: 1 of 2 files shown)


phated (approved):
LGTM! Thanks for the regression test too

Issue #1149, opened 2022-03-03
Provide Obj.magic primitive for runtime use


ospencer:
From Discord: `primitive coerceFloat32ToNumber: Float32 -> Number = "@magic"`

PR #1706, opened 2023-02-28
chore: Make `pow` use early return

spotandjake:
This pr refactors `pow` to use early return. 

This also adds a small optimization to `Rational**Int` Cases.

### DIFF
--- a/stdlib/runtime/numbers.gr
+++ b/stdlib/runtime/numbers.gr
@@ -1139,10 +1139,10 @@ provide let cmpRationals = (x, y) => {
 
 /**
  * Finds the numerator of the rational number.
- * 
+ *
  * @param x: The rational number to inspect
  * @returns The numerator of the rational number
- * 
+ *
  * @since v0.6.0
  */
 @unsafe
@@ -1155,10 +1155,10 @@ provide let rationalNumerator = (x: Rational) => {
 
 /**
  * Finds the denominator of the rational number.
- * 
+ *
  * @param x: The rational number to inspect
  * @returns The denominator of the rational number
- * 
+ *
  * @since v0.6.0
  */
 @unsafe
@@ -2337,7 +2337,7 @@ provide let coerceNumberToBigInt = (x: Number) => {
  *
  * @param number: The value to convert
  * @returns The Number represented as a Rational
- * 
+ *
  * @since v0.6.0
  */
 @unsafe
@@ -2513,7 +2513,7 @@ provide let coerceBigIntToNumber = (x: BigInt) => {
  *
  * @param rational: The value to convert
  * @returns The Rational represented as a Number
- * 
+ *
  * @since v0.6.0
  */
 @unsafe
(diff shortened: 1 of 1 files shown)


stdlib/runtime/numbers.gr
```diff
@@ -2807,414 +2804,376 @@ let rec expBySquaring = (y, x, n) => {
...
-    if (power < 0) expBySquaring(1, 1 / base, power * -1)
-    else expBySquaring(1, base, power)
+    return 1
+  } else if (isInteger(basePtr) && isInteger(powerPtr)) {
+    if (power < 0) return expBySquaring(1, 1 / base, power * -1)
+    else return expBySquaring(1, base, power)
+  } else if (isRational(basePtr) && isInteger(powerPtr)) {
+    // Apply expBySquaring to numerator and denominator
+    let numerator = WasmI32.toGrain(boxedRationalNumerator(basePtr)): Number
+    let denominator = WasmI32.toGrain(boxedRationalDenominator(basePtr)): Number
```
ospencer, line 2817: `boxedRationalNumerator` and `boxedRationalDenominator` both just return raw bigint pointers, so these need to go through some transformation process to become proper numbers. You can use `reducedBigInteger`, though you'll need to `incRef` the numerator and denominator. It could be useful to have utilities for this. Could be good to check with @alex-snezhko to see if he added some when he did the rational work.

alex-snezhko: `rationalNumerator` and `rationalDenominator` in `runtime/numbers` seem to be about what you want, though they specifically take the `Rational` type as an argument. I guess you could "fake" the type of the argument with something like `let baseRational = WasmI32.toGrain(WasmI32.fromGrain(base)): Rational` or maybe extract the stuff after the `WasmI32.fromGrain` calls from these functions [(link)](https://github.com/grain-lang/grain/blob/main/stdlib/runtime/numbers.gr#L1151) into separate functions and reuse those function here

ospencer: Just chiming in that doing `WasmI32.toGrain(WasmI32.fromGrain(base))` is generally unsafe because it gives the GC an opportunity to collect your value. Even though you're making a new reference to it. We have #1149 though to make a utility to do this safely. (You can always add an incRef call, though.)

ospencer: But `rationalNumerator` and `rationalDenominator` sound really good 😛

spotandjake:
I went with the approach of just stealing the code out from `rationalDenominator` and `rationalNumerator` but I am not 100% sure I put the `incRef` call in the correct place.

ospencer:
Could you just call `rationalDenominator` and `rationalNumerator` directly?

spotandjake:
> Could you just call `rationalDenominator` and `rationalNumerator` directly?

I would have to do what alex was saying of `WasmI32.toGrain(WasmI32.fromGrain(base)): Rational` given it wants a Rational wouldnt I. and then that requires a `Memory.incRef` call right? would the `incRef`?

ospencer:
Yep, you'd need to do that. The incref just goes in the middle.

spotandjake:
Also is this breaking given `1/2^3` will no longer give `0.125` but will give `1/8`?

ospencer:
Yeah, certainly that too. It's probably worth it to separate them into two PRs, one that does early return, and the other that changes the squaring of rationals. Do you want to let this PR be the early return one and you can open another after with the behavior change?

spotandjake:
👍

spotandjake:
I reverted those commits after this gets merged I will add that on a separate pr.

ospencer (approved):

PR #1708, opened 2023-02-28
feat(stdlib): Add `chunk` function to Array module

spotandjake:
This pr adds `Array.chunk` to the array libary which allows for you to chunk arrays into arrays of given sizes as shown below:
```
assert Array.chunk(2, false, [> 1, 2, 3, 4]) == [> [> 1, 2], [> 3, 4]]
```


stdlib/array.gr
```diff
@@ -985,3 +985,38 @@ provide let rotate = (n, arr) => {
...
+ *
+ * @param chunkSize: The size of each chunk
+ * @param allowPartial: Whether to allow partial chunks at the end of the array
+ * @param arr: The array to chunk
+ * @returns An array of chunks
+ *
+ * @throws InvalidArgument(String): When `chunkSize` is not an integer
+ * @throws InvalidArgument(String): When `chunkSize` is negative
+ *
+ * @example chunk(2, false, [> 1, 2, 3, 4]) == [> [> 1, 2], [> 3, 4]]
```
alex-snezhko, line 1016: I would add an example example showing what happens when `allowPartial = false` and the array is not evenly splittable, and an example where `allowPartial = true`

alex-snezhko (changes requested):

spotandjake:
Applied those changes thanks for the suggestions alex.

stdlib/array.gr
```diff
@@ -985,3 +985,49 @@ provide let rotate = (n, arr) => {
...
+ * @throws InvalidArgument(String): When `chunkSize` is negative
+ *
+ * @example chunk(2, false, [> 1, 2, 3, 4]) == [> [> 1, 2], [> 3, 4] ]
+ * @example chunk(2, false, [> 1, 2, 3, 4, 5]) == [> [> 1, 2], [> 3, 4] ]
+ * @example chunk(2, true, [> 1, 2, 3, 4, 5]) == [> [> 1, 2], [> 3, 4], [> 5] ]
+ * @since v0.6.0
+ */
+provide let chunk = (chunkSize, allowPartial, arr) => {
+  if (chunkSize == 0) {
+    [>]
```
alex-snezhko, line 1026: Should this case throw an exception?

spotandjake: I think that having it return an empty array is fine behaviour certainly subjective though. I prefer not to throw exception where possible though and I think that having it return an empty array is behaviour that makes sense.

alex-snezhko: Having a chunk size of 0 doesn't really make sense to me though, in the same way as having a negative chunk size. On a related note calling the `windows` function on slices in Rust (basically the equivalent of the function you implemented) with an argument of 0 results in a panic.

spotandjake: That's fair enough. I can implement an exception there then.

stdlib/array.gr
```diff
@@ -1014,9 +1014,55 @@ provide let rotate = (n, arr) => {
...
+ *
+ * @throws InvalidArgument(String): When `chunkSize` is not an integer
+ * @throws InvalidArgument(String): When `chunkSize` is less than zero
+ *
+ * @example chunk(2, false, [> 1, 2, 3, 4]) == [> [> 1, 2], [> 3, 4] ]
+ * @example chunk(2, false, [> 1, 2, 3, 4, 5]) == [> [> 1, 2], [> 3, 4] ]
+ * @example chunk(2, true, [> 1, 2, 3, 4, 5]) == [> [> 1, 2], [> 3, 4], [> 5] ]
+ * @since v0.6.0
+ */
+provide let chunk = (chunkSize, allowPartial, arr) => {
```
phated, line 1052: I don't like this `allowPartial` argument

spotandjake: Really? I could for see cases where `allowPartial` could come in hand.

ospencer: It's not that it's not useful, but for a general purpose function in the standard library, it can get confusing seeing the behavior change via this boolean. If we wanted to support both behaviors, then it should be two separate functions. That said though, I think we can just have the partial behavior (doesn't lose data and is predictable).

spotandjake: Fair enough I will make that change.

stdlib/array.gr
```diff
@@ -1015,6 +1015,53 @@ provide let rotate = (n, arr) => {
   }
 }
 
+let rec _chunk = (chunkSize, allowPartial, arr, arrLen, chunkStartIndex) => {
```
phated, line 1018:
```suggestion
let rec _chunk = (chunkSize, arr, arrLen, chunkStartIndex) => {
```
It looks like you didn't clean up the helper

phated:
```suggestion
let rec chunkHelp = (chunkSize, allowPartial, arr, arrLen, chunkStartIndex) => {
```
I also think our general pattern for naming helpers is `somethingHelp`

stdlib/array.gr
```diff
@@ -1015,6 +1015,53 @@ provide let rotate = (n, arr) => {
...
+      arr,
+      arrLen - chunkSize,
+      chunkStartIndex + chunkSize
+    )
+    [firstChunk, ...restChunks]
+  }
+}
+
+/**
+ * Splits the given array into chunks of the provided size.
```
phated, line 1042:
```suggestion
 * Splits the given array into chunks of the maximum provided size.
```
I think it's worth noting that this is a maximum size

stdlib/array.gr
```diff
@@ -1015,6 +1015,53 @@ provide let rotate = (n, arr) => {
...
+ *
+ * @param chunkSize: The size of each chunk
```
phated, line 1044:
```suggestion
 * @param chunkSize: The maximum size of each chunk
```

stdlib/array.gr
```diff
@@ -1015,6 +1015,53 @@ provide let rotate = (n, arr) => {
...
+ * @param arr: The array to chunk
+ * @returns An array of chunks
+ *
+ * @throws InvalidArgument(String): When `chunkSize` is not an integer
+ * @throws InvalidArgument(String): When `chunkSize` is less than zero
```
phated, line 1049:
```suggestion
 * @throws InvalidArgument(String): When `chunkSize` is less than one
```
I'm not sure about this. You check explicitly for `0` but you then use slice, which accepts negative indexing. Do you need different checks?

phated (changes requested):
Thanks for cleaning this up! A few more things I noticed.

phated (approved):
Awesome! Thanks

PR #1709, opened 2023-02-28
fix(compiler): Inline record exception constructors

### LINKED ISSUE #1679: Exceptions should be able to carry a record, not just a tuple
It seems that we don't parse this code: `exception Foo{ bar: String, baz: Number }`

@ospencer said this is a bug that was probably missed when the record variant code was added.

Going to assign to @alex-snezhko since he's the most informed about this code.

alex-snezhko:
Closes #1679

### DIFF
--- a/compiler/src/typed/datarepr.re
+++ b/compiler/src/typed/datarepr.re
@@ -114,11 +114,7 @@ let constructor_descrs = (ty_path, decl, cstrs) => {
       };
     let cstr_name = Ident.name(cd_id);
     let (existentials, cstr_args, cstr_inlined) =
-      constructor_args(
-        cd_args,
-        cd_res,
-        Path.PExternal(ty_path, cstr_name, Path.nopos),
-      );
+      constructor_args(cd_args, cd_res, Path.PExternal(ty_path, cstr_name));
 
     let cstr = {
       cstr_name,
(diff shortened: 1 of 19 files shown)


compiler/src/typed/path.re
```diff
@@ -4,22 +4,22 @@ open Sexplib.Conv;
 [@deriving (sexp, yojson)]
 type t =
   | PIdent(Ident.t)
-  | PExternal(t, string, int);
```
alex-snezhko, line 7: The "pos" int value never ends up being used in the code so I just removed it

compiler/src/typed/datarepr.re
```diff
@@ -141,7 +137,11 @@ let extension_descr = (path_ext, ext) => {
   let ty_res = newgenconstr(ext.ext_type_path, ext.ext_type_params);
 
   let (existentials, cstr_args, cstr_inlined) =
-    constructor_args(ext.ext_args, Some(ty_res), path_ext);
+    constructor_args(
+      ext.ext_args,
+      Some(ty_res),
+      Path.PExternal(path_ext, "#extension#"),
```
alex-snezhko, line 143: Kind of a hack; the OCaml compiler extracts "extension" constructors into a separate variant but doing this would require changing a lot of other compiler code so I just use this string as a unique identifier for exceptions (but lmk if yall think it's still worth it to create a new variant)

ospencer (approved):
Looks good!

Issue #1371, opened 2022-07-26
Modulo Throws Misleading Error When Used On A Float


spotandjake:
![image](https://user-images.githubusercontent.com/40705786/181106936-43d57d97-5a2f-4945-bc26-8d5821b9426f.png)
In the image above i was using modulo and my Number value became a float, when i used modolo it gave me the error of `NumberNotIntlike: Can't coerce number to integer` it would be nice if either modulo worked on floats or an error was given stating something like `Modulo Doesn't Currently Support Floating Point Values` as the Number types goal is to be user firendly the NumbernotIntLike can be a little confusing and misleading to newer developers

ospencer:
Defining modulo for floats/rationals might be reasonable behavior here instead of throwing an error. Here's what Python does:
```
Python 3.10.5 (v3.10.5:f377153967, Jun  6 2022, 12:36:10) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 3.5 % 3
0.5
>>> 2/3 % 3
0.6666666666666666
```

cc @peblair

spotandjake:
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/40705786/181107906-b336f29e-55d9-4fe4-b9bc-342052736817.png">
js does the same thing

ospencer:
`%` in JS is remainder rather than modulo, so it makes a little more sense there.

spotandjake:
@ospencer as per your python behavour
```py
import math
def fmod(x, y):
    f = 0
    if (y < -1 and x > 0):
        f = abs(y) - abs(x)
    else:
        f = x - math.trunc(x / y) * y
    f = abs(f) * math.copysign(1, y)
    return f
def test(inA, inB, pA, pB):
    global failed
    global passed
    if (inA == inB):
        print(f"Pass: {inA} == {inB}, {pA} % {pB}")
    else:
        print(f"Fail: {inA} != {inB}, {pA} % {pB}")
_range = 100
for _y in range(_range):
    y = (_y / 5) - (_range / 2)
    for _x in range(_range):
        x = (_x / 8) - (_range / 2)
        test(fmod(x, y), x % y, x, y)
```
that function there seems to work for floats in python so i think it would be a matter of just porting the `fmod` there over to grain

PR #1710, opened 2023-02-28
feat(stdlib)!: Add an exception to `%` when used on floats

spotandjake:
This pr works on #1371 The ideal fix will eventually be to make `%` modulo work on floats but given that isnt staged for `v0.6.0` it makes sense to give it a nicer error until it is fixed. It would also make sense to rename that issue or open a new issue after this named `make modulo support floats`


ospencer:
We figured out what we wanted to do to make modulo support floats. Should we just implement it?

PR #1712, opened 2023-03-01
chore(compiler): Represent the else branch of PExpIf as an option instead of empty block

phated:
While working on the formatter stuff, I noticed that we represented the `else` branch as an empty block in the AST. This makes things like the formatter harder, so I think we should just change it to `option(expression)`.

### DIFF
--- a/compiler/test/suites/basic_functionality.re
+++ b/compiler/test/suites/basic_functionality.re
@@ -165,6 +165,11 @@ describe("basic functionality", ({test, testSkip}) => {
   assertSnapshot("if_one_sided5", "let mut x = 1; if (3 < 4) x = 2 + 3");
   assertSnapshot("if_one_sided6", "let mut x = 1; if (3 < 4) x = 2 + 3; x");
   assertCompileError(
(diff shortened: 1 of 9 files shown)


compiler/test/suites/basic_functionality.re
```diff
@@ -164,6 +164,11 @@ describe("basic functionality", ({test, testSkip}) => {
   assertSnapshot("if_one_sided4", "let mut x = 1; if (3 < 4) x = 2; x");
   assertSnapshot("if_one_sided5", "let mut x = 1; if (3 < 4) x = 2 + 3");
   assertSnapshot("if_one_sided6", "let mut x = 1; if (3 < 4) x = 2 + 3; x");
+  assertCompileError(
+    "if_empty_block",
+    "if (false) 1 else {}",
+    "Syntax error",
+  );
```
phated, line 171: I added this test because I was like "was an empty block after `else` ever valid syntax?" And the answer is nope, it errors with "Expected an expression or a record field."

ospencer (approved):

Issue #1620, opened 2023-01-24
Custom printing for boxes


phated:
As noted in #1608, we print boxes as single-argument tuples. This isn't ideal since that's not the syntax used to construct them. Ideally, we'd replace the printing with a custom string for boxes.

PR #1713, opened 2023-03-01
feat(compiler)!: Custom box printing

spotandjake:
This pr adds custom box printing.
Now single argument tuples will be wrinted as `Box(1)` intead of `(1,)` given the only safe way to create a single arugment tuple is by using the box primitive.

I am wondering where I would add a regression test for this if needed I couldn't see anywhere, that we were testing `print` or `toString`.

Closes: #1620

### DIFF
--- a/stdlib/runtime/string.gr
+++ b/stdlib/runtime/string.gr
@@ -28,7 +28,6 @@ include "runtime/unsafe/tags" as Tags
 include "runtime/numberUtils" as NumberUtils
 
 include "runtime/dataStructures" as DataStructures
-
 from DataStructures use { allocateString, allocateArray, untagSimpleNumber }
 
 foreign wasm fd_write: (
--- a/stdlib/runtime/string.gr
+++ b/stdlib/runtime/string.gr
@@ -28,6 +28,7 @@ include "runtime/unsafe/tags" as Tags
 include "runtime/numberUtils" as NumberUtils
 
 include "runtime/dataStructures" as DataStructures
+
 from DataStructures use { allocateString, allocateArray, untagSimpleNumber }
 
 foreign wasm fd_write: (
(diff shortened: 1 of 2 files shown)


stdlib/runtime/string.gr
```diff
@@ -512,13 +513,12 @@ let rec heapValueToString = (ptr, extraIndents, toplevel) => {
...
         let rparen = ")"
+        let mut lparen = "("
         let mut strings = [rparen]
         if (tupleLength <= 1n) {
           // Special case: unary tuple, which is not valid Grain syntax; however, boxed values
           // are stored as a unary tuple, so we keep this in case one gets printed
-          let comma = ","
-          strings = [comma, ...strings]
-          void
+          lparen = "Box("
```
ospencer, line 521: It'd be a bit better if in this case you appended `"box"` to the list that's all going to get concatenated, so just `strings = ["box", ...strings]`. You might need to rework things a bit and push this `if` til after the loop. We'll also want to do lowercase `"box"` 🙂

ospencer (changes requested):
Looking good! One small change, and then I think it'd be good to add one or two tests.

stdlib/runtime/string.gr
spotandjake: Done this change.

spotandjake:
> Looking good! One small change, and then I think it'd be good to add one or two tests.

I was wondering where I would put regression tests for these?

stdlib/runtime/string.gr
```diff
@@ -28,6 +28,7 @@ include "runtime/unsafe/tags" as Tags
 include "runtime/numberUtils" as NumberUtils
 
 include "runtime/dataStructures" as DataStructures
+
```
ospencer, line 31:
```suggestion
```

ospencer:
You can add tests to compiler/test/suites/strings.re!

ospencer (approved):
Sweet, thanks!

ospencer:
I suppose this is technically breaking.

PR #1714, opened 2023-03-01
feat(compiler): Streamlined reference counting

ospencer:
This PR is a reimplementation of Grain's application of reference counting instructions, namely to support #1092.

This new implementation does careful tracking of references to avoid many calls to `incRef` and `decRef`, especially in the case of tail calls, to ensure all live references are cleaned up before returning (or `return_call`ing) from the function.

For example, compiling `let rec loop = () => loop()` with tail calls enabled now results in a perfect loop, with no refcounting instructions needed:
```wat
 (func $loop_1113 (param $0 i32) (result i32)
  (local $1 i32)
  (local $2 i32)
  (local $3 i32)
  (local $4 i64)
  (local $5 f32)
  (local $6 f64)
  (return
   (block $compile_block.1
    (return_call $loop_1113
     (local.get $0)
    )
   )
  )
 )
 ```
This is in contrast to the old version, which leaks memory:
```wat
 (func $loop_1113 (param $0 i32) (result i32)
  (local $1 i32)
  (local $2 i32)
  (local $3 i32)
  (local $4 i64)
  (local $5 f32)
  (local $6 f64)
  (return
   (block $cleanup_locals.2 (result i32)
    (local.tee $1
     (tuple.extract 0
      (tuple.make
       (block $compile_block.1
        (return_call $loop_1113 ; everything after this instruction is not executed
         (call $incRef_0
          (global.get $GRAIN$EXPORT$incRef_0)
          (local.get $0)
         )
        )
       )
       (local.get $1)
      )
     )
    )
    (drop
     (call $decRef_0 ; dead code—the closure has leaked
      (global.get $GRAIN$EXPORT$decRef_0)
      (local.get $0)
     )
    )
    (local.get $1)
   )
  )
 )
 ```
This should also result in more efficient Grain programs overall. The test suite runs 15% faster after this change, and that time gained back should nearly all be within the execution of Grain code.

Other important notes: references are decRefed immediately after their last use, compared to the end of the function as it works today. This has a slight effect on users in that memory usage while in a function is much lower, but has a major effect on unsafe code in Grain.

In this example:
```
@unsafe
let foo = (s: String) => {
  let ptr = WasmI32.fromGrain(s)
  ...
}
```
`s` is no longer used, so `s` is a candidate for immediate `decref`. This would obviously present a large problem for unsafe code generally, so this PR aids the programmer in that any time `fromGrain` is used, `decref` of that value is delayed until the end of the function, similar to how it works now. There is one major difference though—the value is only held until the function returns, meaning that the reference will die prior to any tail call. Why this matters—take a look at the implementation of `toString`:
```
@unsafe
provide let toString = value => {
  let value = WasmI32.fromGrain(value)
  toStringHelp(value, 0n, true)
}
```

`toStringHelp` is a _tail call_, and thus the original value is decrefed _before_ the call to `toStringHelp`. This code was updated in this PR to prevent the tail call. Generally, we should refrain from passing raw pointers between functions, and we will need to refactor some unsafe code to ensure values aren't being cleaned up prematurely.

### CODE REVIEW

**Review by jozanza — APPROVED** [review]
lgtm

File: compiler/src/codegen/garbage_collection.re [hunk 1]
```diff
@@ -0,0 +1,781 @@
...
+            ((branch_id, branch)) => (branch_id, apply_gc(branch)),
+            branches,
+          ),
+          apply_gc(default),
+          alloc,
+        )
+      | MPrim0(_) => instr.instr_desc
+      | MPrim1(WasmFromGrain, imm) =>
+        // special case: we delay cleanup of any `fromGrain` to the end of the function to
+        // remove the headache of the value being GC'd before anything can be done with it
```
phated, line 453: Not that it should be done in this PR but what other options were considered for making `fromGrain` not a special case?

compiler/src/middle_end/linearize.re
```diff
@@ -554,11 +554,20 @@ let rec transl_imm =
...
-    let (fst_ans, fst_setup) = transl_comp_expression(fst);
-    let (rest_ans, rest_setup) =
-      transl_imm({...e, exp_desc: TExpBlock(rest)});
-    (rest_ans, fst_setup @ [BSeq(fst_ans)] @ rest_setup);
+  | TExpBlock(stmts) =>
+    let stmts = List.rev_map(transl_comp_expression, stmts);
+    // stmts is non-empty, so this cannot fail
+    let (last_ans, last_setup) = List.hd(stmts);
+    let stmts = List.tl(stmts);
+    let tmp = gensym("block_ans");
```
phated, line 562: What does "ans" stand for?

compiler/test/input/sinister-tail-call-gc.gr
```diff
@@ -1,33 +0,0 @@
-module SinisterTailCallGC
-
-// Contrived example to test *ADVANCED* tail calls
```
phated, line 3: Deleted because it works now?

compiler/test/stdlib/string.test.gr
```diff
@@ -315,119 +318,137 @@ assert String.decode(String.encode(emojis, String.UTF32_BE), String.UTF32_BE) ==
...
   emojis
 
+let buf = Bytes.make(500)
+// ensure buffer is clear for our assertions on contents
+Bytes.clear(buf)
 // decodeRange
 // 51 is chosen to stress-test these functions, since it's not an aligned offset
 assert String.decodeRange(
-  String.encodeAt(emojis, String.UTF8, Bytes.make(500), 51),
+  String.encodeAt(emojis, String.UTF8, buf, 51),
```
phated, line 327: Why were these changed to re-use a Bytes?

stdlib/runtime/string.gr
```diff
@@ -179,7 +180,9 @@ let rec writeStrings = (buf, list) => {
       Memory.copy(buf, hd + 8n, hdSize)
       writeStrings(buf + hdSize, tl)
     },
-    [] => void,
+    [] => {
+      void
+    },
```
phated, line 185: Why block?

phated (approved):
This PR is absolute 🔥  I love how many hacks got removed with the improvements. I had a couple questions but none of them stop my approval.

I also spot checked a couple snapshots and they look sweet but I'm not going to review all of them so 🤞 they are correct.

I noticed this isn't called breaking. Since we know there are some problems with `--experimental-tail-calls` with this being landed, I think we should call it breaking (even though we will fix those quickly).

compiler/src/codegen/garbage_collection.re
ospencer: The only other option would be to have us _always_ use `ignore` to hold onto the reference. We could maybe write some sort of well-formedness check to try to catch when we let something die too soon, but it's really tricky.

compiler/src/middle_end/linearize.re
ospencer: Answer. Can probably clean those up.

compiler/test/input/sinister-tail-call-gc.gr
ospencer: Yeah, it's essentially the same as the other tests.

compiler/test/stdlib/string.test.gr
ospencer: The GC actually cleans things up in a timely manner now, so if you allocate a `Bytes`, stop using it, and allocate another one of the same size immediately, you're pretty like to get the same memory region right back. It's definitely time to `calloc` these.

phated:
Not sure why the number suite is failing now

PR #1718, opened 2023-03-02
chore: Switch operator assignments to using

spotandjake:
switches stuff such as `let (+) = WasmI32.add` to `from Wasmi32 use { add as (+) }`.

this is marked as draft because the tests are not passing on bigint which isnt good but I have not been able to find out where I messed up.

### DIFF
--- a/stdlib/runtime/atof/parse.gr
+++ b/stdlib/runtime/atof/parse.gr
@@ -72,13 +72,15 @@ from Slow use { parseLongMantissa }
 // https://johnnylee-sde.github.io/Fast-numeric-string-to-int/
 @unsafe
 let parse8Digits = (digits: WasmI64) => {
-  let (+) = WasmI64.add
-  let (-) = WasmI64.sub
-  let (*) = WasmI64.mul
-  let (>>) = WasmI64.shrU
-  let (&) = WasmI64.and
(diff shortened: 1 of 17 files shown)


spotandjake (commented):
I looked over all my changes and I cant find any operators that I messed up on. I also tried copying your patch from `bigint` and that didnt fix it.

stdlib/runtime/atof/parse.gr
```diff
@@ -520,19 +528,17 @@ provide let isFastPath =
...
       let floatOpt = if (isFastPath(exponent, mantissa, negative, manyDigits)) {
-        let (<=) = WasmI32.leS
+        from WasmI32 use { leS as (<=) }
         if (exponent <= _MAX_EXPONENT_FAST_PATH) {
           // normal fast path
-          let (*) = WasmF64.mul
-          let (/) = WasmF64.div
-          let (<) = WasmI32.leS
+          from WasmF64 use { mul as (*), div as (/) }
+          from WasmI32 use { leS as (<) }
```
ospencer, line 541: I guess this isn't a bug given how it's used, but we should fix it:
```suggestion
          from WasmI32 use { ltS as (<) }
```

ospencer (approved):

PR #1720, opened 2023-03-04
chore(compiler): Escape `"` in char well formness error

spotandjake:
This pr escapes `"` in the error for chars that are meant to be strings.

This is because in a case like `'{"test":1}'` when the compiler asks you if you mean `"{"test":1}"` it would be an invalid string so instead the new error looks like `"{\"test\":1}'`

### DIFF
--- a/compiler/src/parsing/well_formedness.re
+++ b/compiler/src/parsing/well_formedness.re
@@ -33,7 +33,7 @@ let prepare_error =
           ~loc,
           "This character literal contains multiple characters: '%s'\nDid you mean to create the string \"%s\" instead?",
           cl,
-          cl,
+          Str.global_replace(Str.regexp({|"|}), {|\"|}, cl),
         )
       | ExternalAlias(name, loc) =>
         errorf(~loc, "Alias '%s' should be at most one level deep.", name)
--- a/compiler/test/suites/chars.re
+++ b/compiler/test/suites/chars.re
@@ -67,9 +67,15 @@ describe("chars", ({test, testSkip}) => {
   assertRun("char_toString_escape8", {|print(('\v', 1))|}, "('\\v', 1)\n");
   assertRun("char_toString_escape9", {|print(('\'', 1))|}, "('\\'', 1)\n");
   assertCompileError(
-    "char_illegal",
+    "char_illegal1",
     "'abc'",
     "This character literal contains multiple characters: 'abc'\nDid you mean to create the string \"abc\" instead?",
+  );
+  assertCompileError(
+    "char_illegal2",
+    {|'{"test": 1}'|},
+    {|This character literal contains multiple characters: '{"test": 1}'
+Did you mean to create the string "{\"test\": 1}" instead?|},
   );
   assertCompileError(
     "unicode_err1",
--- a/compiler/test/suites/chars.re
+++ b/compiler/test/suites/chars.re
@@ -74,8 +74,8 @@ describe("chars", ({test, testSkip}) => {
   assertCompileError(
     "char_illegal2",
     {|'{"test": 1}'|},
-    {|This character literal contains multiple characters: '{"test": 1}'
-Did you mean to create the string "{\"test\": 1}" instead?|},
+    {|This character literal contains multiple characters: '\{"test": 1\}'
+Did you mean to create the string "\{\\"test\\": 1\}" instead?|},
   );
   assertCompileError(
     "unicode_err1",


phated:
@spotandjake when would a record field be wrapped in a string? That's not grain code.

spotandjake:
> @spotandjake when would a record field be wrapped in a string? That's not grain code.

It isn't a record field imagine working with json doing `Json.parse('{"test":1}')` and the developer messed up the string and accidentally used a char because they are coming from another language and when the compiler tells you how to fix it, it gives you bad code because the new quotes are not escaped a common use case in js for alternative strings such as \`, `'` and `"` is so you don't have to escape inside the string.

ospencer:
That's maybe just a bad example. What'd be far more common is someone trying to use a single quote string _because_ they didn't want to escape the double quotes, something like `let hello = '"hello world!"'`.

spotandjake:
> That's maybe just a bad example. What'd be far more common is someone trying to use a single quote string _because_ they didn't want to escape the double quotes, something like `let hello = '"hello world!"'`.

yeah pretty much I was just using json because that's where I came across it while trying to help jarius with the json pr.

On another Note I can't for the life of me figure out why the test I added is failing I added some print statements to the test runner and the expected output was in the input, I tried using normal strings and escaping the values the special syntax, I tried just including the part that I wanted to match that in the output but couldn't figure it out.

alex-snezhko:
@spotandjake It looks like the failure is because `assertCompileError` does a regex match on the expected output so curly braces (and `\`) need to be escaped.

Try `{|This character literal contains multiple characters: '\{"test": 1\}'
Did you mean to create the string "\{\\"test\\": 1\}" instead?|},`

spotandjake:
> @spotandjake It looks like the failure is because `assertCompileError` does a regex match on the expected output so curly braces (and `\`) need to be escaped.
> 
> Try `{|This character literal contains multiple characters: '\{"test": 1\}' Did you mean to create the string "\{\\"test\\": 1\}" instead?|},`

That didn't seem to fix it.

alex-snezhko:
Whoops, I meant to put before "Did" on a new line. I just tried doing it with
```
{|This character literal contains multiple characters: '\{"test": 1\}'
Did you mean to create the string "\{\\"test\\": 1\}" instead?|}
```
And it worked for me

spotandjake:
> Whoops, I meant to put before "Did" on a new line. I just tried doing it with
> 
> ```
> {|This character literal contains multiple characters: '\{"test": 1\}'
> Did you mean to create the string "\{\\"test\\": 1\}" instead?|}
> ```
> 
> And it worked for me

It worked when I put Did on the newline I forget to get rid of the indentation. Thanks for the fix Alex.

compiler/src/parsing/well_formedness.re
```diff
@@ -33,7 +33,7 @@ let prepare_error =
           ~loc,
           "This character literal contains multiple characters: '%s'\nDid you mean to create the string \"%s\" instead?",
           cl,
-          cl,
+          Str.global_replace(Str.regexp({|"|}), {|\"|}, cl),
```
phated, line 36: Do we want to use [`String.escaped`](https://v2.ocaml.org/api/String.html) here? Maybe a question for @ospencer 

```suggestion
          String.escaped(cl),
```

phated (changes requested):
Need to know if we should use `String.escaped` to cover other characters.

compiler/src/parsing/well_formedness.re
ospencer: I don't think we want to use `escaped` here since the semantics are pretty different than Grain's. Whatever the user wrote (with escaped quotes) is fine.

compiler/test/suites/chars.re
```diff
@@ -67,10 +67,16 @@ describe("chars", ({test, testSkip}) => {
...
-    "char_illegal",
+    "char_illegal1",
     "'abc'",
     "This character literal contains multiple characters: 'abc'\nDid you mean to create the string \"abc\" instead?",
   );
+  assertCompileError(
+    "char_illegal2",
+    {|'{"test": 1}'|},
+    {|This character literal contains multiple characters: '\{"test": 1\}'
+Did you mean to create the string "\{\\"test\\": 1\}" instead?|},
```
ospencer, line 78: Do we really need to escape `{` and `\`?

spotandjake: Alex had noted the rest suite uses regex. And offered that solution and it worked so I took it. Everything else I tried didn't work so I think maybe.

ospencer: Ah, I see.

ospencer (approved):
Nice fix!

phated (approved):

Issue #1520, opened 2022-12-04
Stdlib: Add `@throws` graindoc to Array module


phated:
When the `@throws` syntax for graindoc was added in #1492, the Array module's docs weren't updated. Someone needs to go through this module and write up accurate docs for each `throw` or `fail`

PR #1721, opened 2023-03-04
chore(stdlib): Add `throws` documentation to array

spotandjake:
This pr adds `@throws` graindoc to the array module.

A question I have is I used `int32` over `interger` for most of the throws docs because an i64 isnt current supported I am wondering if this is good or if there is a better altnerative, or if we just want to say integer.

My other questions was related to `fromList` you could technically have a list that has a length that is out of the i32 range right, I dont know if you could fit that in memory or what. but if you were to call `fromList` that would throw an exception I am wondering if I should add a `@throws` to that or just leave it given its virtually impossible to construct.

closes: #1520

### DIFF
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -58,7 +58,7 @@ provide let length = array => {
  * @param item: The value to store at each index
  * @returns The new array
  *
- * @throws InvalidArgument(String): When `length` is not an integer
+ * @throws InvalidArgument(String): When `length` is not an int32
  * @throws InvalidArgument(String): When `length` is negative
  *
  * @example Array.make(5, "foo") // [> "foo", "foo", "foo", "foo", "foo"]
@@ -91,7 +91,7 @@ provide let make: (Number, a) -> Array<a> = (length: Number, item: a) => {
  * @param fn: The initializer function to call with each index, where the value returned will be used to initialize the element
  * @returns The new array
  *
- * @throws InvalidArgument(String): When `length` is not an integer
+ * @throws InvalidArgument(String): When `length` is not an int32
  * @throws InvalidArgument(String): When `length` is negative
  *
  * @example Array.init(5, n => n + 3) // [> 3, 4, 5, 6, 7]
@@ -129,6 +129,10 @@ provide let init: (Number, Number -> a) -> Array<a> =
  * @param index: The index to access
  * @param array: The array to access
  * @returns The element from the array
+ *
+ * @throws IndexOutOfBounds: When `index` is not an int32
+ * @throws IndexOutOfBounds: When `index` is out of bounds
+ *
  * @example Array.get(1,[> 1, 2, 3, 4, 5]) == 2
  *
  * @since v0.1.0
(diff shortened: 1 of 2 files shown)


stdlib/array.gr
```diff
@@ -58,7 +58,7 @@ provide let length = array => {
  * @param item: The value to store at each index
  * @returns The new array
  *
- * @throws InvalidArgument(String): When `length` is not an integer
+ * @throws InvalidArgument(String): When `length` is not an int32
```
phated, line 61: Let's keep these `integer`

stdlib/array.gr
```diff
@@ -186,6 +197,9 @@ provide let append = (array1, array2) => {
  *
  * @param arrays: A list containing all arrays to combine
  * @returns The new array
+ *
+ * @throws InvalidArgument(String): When the combined length of the all arrays is not an integer
```
phated, line 201:
```suggestion
 * @throws InvalidArgument(String): When the combined length of all arrays is not an integer
```

stdlib/array.gr
```diff
@@ -405,6 +419,8 @@ provide let reducei = (fn, initial, array) => {
  * @param array: The array to iterate
  * @returns The new array
  *
+ * @throws InvalidArgument(String): When the combined length of the all arrays is not an integer
```
phated, line 422:
```suggestion
 * @throws InvalidArgument(String): When the combined length of all arrays is not an integer
```

phated (changes requested):
Noticed some grammar issues

stdlib/array.gr
spotandjake: oops, Thanks for catching these fixed them.

stdlib/array.gr
```diff
@@ -768,7 +789,7 @@ provide let unique = array => {
  * @param array2: The array to provide values for the second tuple element
  * @returns The new array containing indexed pairs of `(a, b)`
  *
- * @throws Failure(String): When the arrays have different sizes
+ * @throws IndexOutOfBounds(String): When the arrays have different sizes
```
phated, line 792:
```suggestion
 * @throws IndexOutOfBounds: When the arrays have different sizes
```
IndexOutOfBounds doesn't carry data, right?

stdlib/array.gr
```diff
@@ -797,6 +818,8 @@ provide let zip = (array1: Array<a>, array2: Array<b>) => {
  * @param array2: The array whose elements will each be passed to the function as the second argument
  * @returns The new array containing elements derived from applying the function to pairs of input array elements
  *
+ * @throws IndexOutOfBounds(String): When the arrays have different sizes
```
phated, line 821:
```suggestion
 * @throws IndexOutOfBounds: When the arrays have different sizes
```
Same question as above

phated (changes requested):
I think there's an error for IndexOutOfBounds but need @spotandjake to confirm

stdlib/array.gr
spotandjake: That is true, this for catching that.

phated (approved):
Thanks

PR #1729, opened 2023-03-06
feat(stdlib)!: Provide empty byte array from Bytes.make

### LINKED ISSUE #1723: Provide clean `Bytes` from `Bytes.make`
The GC actually cleans things up in a timely manner now, so if you allocate a `Bytes`, stop using it, and allocate another one of the same size immediately, you're pretty like to get the same memory region right back. It's definitely time to `calloc` these.

_Originally posted by @ospencer in https://github.com/grain-lang/grain/pull/1714#discussion_r1125569745_

ospencer:
This change ensures we always provide a zeroed byte array from `Bytes.make`.

Closes #1723


spotandjake (approved):
Looks Good To Me

phated (approved):

Issue #1731, opened 2023-03-06
GrainFMT: Adding two comma's on trailing `rest


spotandjake:
```
module Test

let n = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
let x = match (n) {
  [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, ...rest] => void,
  _ => void,
}
```
When i run the above code through the grain formatter I get the invalid code below, The invalid part is it adding two `,` after the`...rest`, This happens with any literal it would seem though
```
module Test

let n = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
let x = match (n) {
  [
    1,
    2,
    3,
    4,
    5,
    6,
    7,
    8,
    9,
    10,
    11,
    12,
    13,
    14,
    15,
    16,
    17,
    18,
    19,
    20,
    ...rest,,
  ] =>
    void,
  _ => void,
}
```

PR #1732, opened 2023-03-07
chore(grainfmt): Fix double comma on list spread after break

spotandjake:
image
This line here seems to be whats causing the issue. by getting rid of that line the issue fixes itself. One comma is being appended there and one is being appended in `item_iterator` as per the trailing_comma.

From what I understand all that does is add a comma at the end of the patterns but I could be wrong given Im not overly familar with the code.

Closes: #1731


marcusroberts (approved):
Good fix, thank you!

phated:
This is going to conflict with my work but it's worth having just for the test.

PR #1738, opened 2023-03-07
chore(stdlib)!: Update Operator `uint` operator Names

spotandjake:
This changes `>>` to `>>>` in the `uint` libs to reflect that they are unsighned. this also changes `%` to rem in the unsighned libs to better reflect that it is just returning the remainder.

### DIFF
--- a/compiler/test/stdlib/uint16.test.gr
+++ b/compiler/test/stdlib/uint16.test.gr
@@ -12,12 +12,12 @@ from Uint16 use {
   (-),
   (*),
   (/),
-  (%),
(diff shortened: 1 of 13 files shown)


ospencer (approved):
Looks good! Just going to note that this is breaking.

spotandjake:
> Looks good! Just going to note that this is breaking.

A follow up question I had `@since` documentation isn't neccessairy here right because we never shipped a release with the old names?

ospencer:
That's correct.

Issue #1719, opened 2023-03-03
Remove PExpNull AST node


ospencer:
Speaking of which, I don't think `PExpNull` is used anymore, so we can remove it.

_Originally posted by @ospencer in https://github.com/grain-lang/grain/pull/1688#discussion_r1123965197_

PR #1740, opened 2023-03-08
chore(compiler): Remove `PExpNull` and `TExpNull`

spotandjake:
This pr removes `PEXPNULL` and `TEXPNULL` from the ast. given they are no longer used.

Closes: #1719


phated (approved):
I love to see these types of changes 🎉

PR #1741, opened 2023-03-09
chore(compiler)!: Optimize type metadata loading

### LINKED ISSUE #1619: Optimize loading of type metadata during printing
I would completely disallow modules complied in runtime mode from producing type metadata information (this may already be the case?) and implement a simple hash for quick lookup of type metadata. Probably a fixed size, either 256 or 512 buckets.

alex-snezhko:
Closes #1619

Summary of changes:
- Removed metadata-loading function; instead:
- Metadata gets loaded into cmi section of wasm module when compiled (before linking)
- At link time, the cmi is read and a hash table (with 256 buckets) containing the metadata is put into an active `data` section in the linked wasm module
- The beginning of the runtime heap is shifted to be after the metadata table


ospencer:
Looks like you need some snapshot updates!

stdlib/runtime/string.gr
```diff
@@ -44,25 +45,22 @@ primitive builtinId = "@builtin.id"
...
-      }
-      modData += WasmI32.load(modData, 0n)
+let findTypeMetadata = typeHash => {
+  let hashHash = typeHash % 256n
+  let bucketPtr = _RUNTIME_TYPE_METADATA + (hashHash << 3n) // 8 bytes/bucket
+  let bucketDataOffset = WasmI32.load(bucketPtr, 0n)
+  let bucketSize = WasmI32.load(bucketPtr, 4n)
+  let beginDataPtr = _RUNTIME_TYPE_METADATA + bucketDataOffset
+  let endDataPtr = beginDataPtr + (bucketSize << 3n)
+  // TODO double-check all values
```
ospencer, line 58: Shall we TODO?

ospencer (commented):
Super stoked that you got this working so quickly! I'll get you a proper review soon, but jotted down some thoughts as I skimmed through.

compiler/src/codegen/compcore.re
```diff
@@ -297,6 +305,21 @@ let get_wasm_imported_name = (~runtime_import=true, mod_, name) => {
   Ident.unique_name(name);
 };
 
+let get_runtime_heap_start = wasm_mod =>
```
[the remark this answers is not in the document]
alex-snezhko: On second thought I think the current global should just be replaced with the runtime heap start; I now realize that the only time the table size is used is to be added to the memory base to get the runtime heap start

compiler/src/linking/link.re
```diff
@@ -284,6 +284,93 @@ let rec globalize_names = (~function_names, ~global_names, ~label_names, expr) =
   };
 };
 
+let construct_type_metadata_table = metas => {
+  // Create a hash table mapping "type hashes" unique to each Grain type to
+  // offsets in the metadata section
+
+  // Structure:
+  // | Hash table (256 buckets): array of (offset into mappings data, bucket size) for each hash (type_hash % 256)
```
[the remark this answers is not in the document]
alex-snezhko: I ended up choosing the bucket size to be the greatest power of 2 <= the number of entries in the table as a reasonable tradeoff of memory usage and lookup speed; lmk if you think some other metric should be used

alex-snezhko:
Hmm, I think the snapshots are failing because of the way that the type hashes get computed (hashing the source file location object, which contains the file path where the type is defined). Depending on the file system path names where the program is compiled the hashes would get computed differently.

alex-snezhko:
Resolved by taking source of uniqueness to be a list of strings relevant to the type (concatenated constructor names + inline record field names for inline record cstrs) and field names for records

compiler/src/linking/link.re
```diff
@@ -284,6 +284,105 @@ let rec globalize_names = (~function_names, ~global_names, ~label_names, expr) =
...
+    );
+  let rec next_pow_of_2 = (num, res) =>
+    if (num <= 1) {
+      1 lsl res;
+    } else {
+      next_pow_of_2(num lsr 1, res + 1);
+    };
+  let num_entries = List.length(hash_to_offset);
+  // Choose number of buckets to be the greatest power of 2 <= # table entries
+  // (limit 512) as a memory/lookup speed tradeoff (expect 1-2 entries/bucket)
```
ospencer, line 338: This limit feels kind of small. Maybe limit to 2048 or 4096? These would still fit in one page even in 64-bit mode.

compiler/src/linking/link.re
```diff
@@ -284,6 +284,105 @@ let rec globalize_names = (~function_names, ~global_names, ~label_names, expr) =
...
+  let num_buckets = min(512, next_pow_of_2(num_entries, 0));
```
ospencer, line 339: Shouldn't this be `max`?

compiler/src/parsing/parsetree.re
```diff
@@ -327,7 +327,8 @@ type prim0 =
   | AllocateFloat64
   | AllocateRational
   | WasmMemorySize
-  | Unreachable;
+  | Unreachable
+  | HeapStart;
```
ospencer, line 331: It looks like we didn't actually need the changes to `HeapStart`, can we back those out?

compiler/src/linking/link.re
alex-snezhko: `max` would make it have at minimum 512 entries e.g. if you had 8 entries max would tell you to use 512 rather than 8

compiler/src/parsing/parsetree.re
alex-snezhko: How do you mean? How would you query where the start of the heap is then?

compiler/src/linking/link.re
ospencer: Yes. 😛

compiler/src/parsing/parsetree.re
ospencer: Oh not that, just that you changed the primitive to be a prim0 instead of a prim_constant, but I don't see anywhere why it needs to be a function.

alex-snezhko: Oh ok gotcha; in order to fit in with [the rest of the translprim code](https://github.com/grain-lang/grain/blob/242d845a91b0bb7769f7fc4b2a1ad7dfec41eec1/compiler/src/typed/translprim.re#L1503), should I create a new parsetree node like `PExpHeapBase` to be only used in this context?

ospencer: Nah. We kinda hate this pattern but haven't gotten around to changing it.

alex-snezhko: Ok how would you suggest that code segment be changed then?

ospencer: Just putting it back to the way it was—we didn't have this variant defined here because there aren't any constant prims in the parsetree.

alex-snezhko: Unless I'm missing something, it wouldn't work going back to the way it was would it? Because the heap start is dependent on the size of the table, which isn't known until link time, you can't know what number to plug in here when compiling a module

ospencer: Yeah, sorry for not being clear. Not the functionality being the same as it was before; just the primitive continuing to be a prim_constant, but the value is pulled from the imported global in `compcore` (instead of being set in `linearize` like it was before). Does that make sense?

alex-snezhko: Sorry, I still don't quite understand what you mean. If `HeapStart` is added back to `primitive_constant` that would necessitate the value being supplied for `@heap.start` [here](https://github.com/grain-lang/grain/blob/242d845a91b0bb7769f7fc4b2a1ad7dfec41eec1/compiler/src/typed/translprim.re#L1500). Without significant changes to how this segment of code in `translprim`, I'm not sure how you would do that. And the only thing you could really do in `compcore` w.r.t. getting a reference to the global is to create a `Global_get` binaryen `Expression` right? The current code is `translprim` is not set up to handle that and instead expects `Parsetree.expression`s

ospencer: I see what you mean. That's... annoying. I suppose we can keep it the way it is for now 😩

alex-snezhko: Yeah it is a bit annoying haha. Maybe an alternative here could be to move `prim_constant` into the parsetree and then handle compiling them in `compcore` in the same manner as the other `prim`s; that would require a bit of extra work but would add some consistency and also allow `@heap.start` to be a constant :shrug:

ospencer: I realized that, but it makes me sad.

ospencer: I'll leave it up to you.

alex-snezhko: Lol ok I think it's fine to have it as a prim0 for now; it could always be changed in 0.7 or later anyways. It doesn't look like this primitive gets used anywhere in grain stdlib or runtime code right now anyways

phated:
@alex-snezhko this looks like it is conflicting with a few files.

Issue #1715, opened 2023-03-02
compiler: Add Support For Decimal Scientific Notation


spotandjake:
A definition can be found here https://en.cppreference.com/w/cpp/language/floating_literal

Grain already supports this in `parseFloat` the implementation is rather small it can be found here [atof Implementation for it](https://github.com/grain-lang/grain/blob/84e6e5457c0a26622e4949bc66839c98c99d1b07/stdlib/runtime/atof/parse.gr#L94)

This isn't overly useful from a high level persepctive but given its a low maintence feature and it is useful for lower level programming it would be handy to have for use in the runtime.

PR #1743, opened 2023-03-10
feat(compiler): Add hex float syntax

spotandjake:
This adds the the following type of syntax to be supported for decimals in grain `0x1F.4p5` which is similar to the `1.1e1` syntax but its treated as hexadecimal instead of decimal.

Adds to the syntax to the lexer
Adds regression tests
Replaces places alternative syntax was used.

Documentation here for when looking over tests: https://en.cppreference.com/w/cpp/language/floating_literal

Closes: #1715


ospencer (approved):
Very cool! Can you add an issue to the GLS repo so we can add syntax highlighting?

phated:
Issue was opened at https://github.com/grain-lang/grain-language-server/issues/154

PR #1734, opened 2023-03-07
feat(stdlib)!: Convert unsafe `Wasm` functions to operators

spotandjake:
This pr changes the names of some of the Wasm Unsafe Runtime Operations such as `WasmI32.add` to be operators instead.

Currently Completed is `WasmF32` and `WasmF64`

Breaking: This is breaking because we are chaining the module apis.

This is a Draft because I still have todo `WasmI32` and `WasmI64`.

PR #1744, opened 2023-03-11
fea(stdlib): Add comparison operators to char stdlib

spotandjake:
This pr adds comparison operators to the `char` stdlib.

This is blocked on #1734 beacuse currently the imports from the wasm library are using the names rather then the operators.

### DIFF
--- a/stdlib/char.gr
+++ b/stdlib/char.gr
@@ -203,7 +203,7 @@ provide let toString = (char: Char) => {
  */
 @unsafe
 provide let (<) = (x: Char, y: Char) => {
-  from WasmI32 use { ltU as (<) }
+  from WasmI32 use { (<) }
   let x = WasmI32.fromGrain(x)
   let y = WasmI32.fromGrain(y)
   x < y
@@ -220,7 +220,7 @@ provide let (<) = (x: Char, y: Char) => {
  */
 @unsafe
 provide let (<=) = (x: Char, y: Char) => {
-  from WasmI32 use { leU as (<=) }
+  from WasmI32 use { (<=) }
   let x = WasmI32.fromGrain(x)
   let y = WasmI32.fromGrain(y)
   x <= y
@@ -237,7 +237,7 @@ provide let (<=) = (x: Char, y: Char) => {
  */
 @unsafe
 provide let (>) = (x: Char, y: Char) => {
-  from WasmI32 use { gtU as (>) }
+  from WasmI32 use { (>) }
   let x = WasmI32.fromGrain(x)
   let y = WasmI32.fromGrain(y)
   x > y
@@ -254,7 +254,7 @@ provide let (>) = (x: Char, y: Char) => {
  */
 @unsafe
 provide let (>=) = (x: Char, y: Char) => {
-  from WasmI32 use { geU as (>=) }
+  from WasmI32 use { (>=) }
   let x = WasmI32.fromGrain(x)
   let y = WasmI32.fromGrain(y)
   x >= y
--- a/stdlib/char.gr
+++ b/stdlib/char.gr
@@ -193,7 +193,7 @@ provide let toString = (char: Char) => {
 }
 
 /**
- * Checks if the first value is less than the second value by Unicode scalar value.
+ * Checks if the first character is less than the second character by Unicode scalar value.
  *
  * @param x: The first value
  * @param y: The second value
@@ -210,7 +210,7 @@ provide let (<) = (x: Char, y: Char) => {
 }
 
 /**
- * Checks if the first value is less than or equal to the second value by Unicode scalar value.
+ * Checks if the first character is less than or equal to the second character by Unicode scalar value.
  *
  * @param x: The first value
  * @param y: The second value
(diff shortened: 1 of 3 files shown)


ospencer:
I thought we were just going to reimplement `<` and `>` in Pervasives using `compare`?

spotandjake:
Discuessed a little bit more in discord, and it sounds like well the compare changes makes this extra. Adding these operators to the char lib makes their api's more similar to the other data types and these are also a lot faster and smaller then using the pervasives ones given they are data type specific so it probably makes sense to add these as well.

stdlib/char.gr
```diff
@@ -191,3 +191,71 @@ provide let toString = (char: Char) => {
 
   result
 }
+
+/**
+ * Checks if the first value is less than the second value.
```
ospencer, line 196: Probably want to mention that it's checking if the first char is less than/greater than the other by Unicode scalar value.

stdlib/char.gr
```diff
@@ -191,3 +191,71 @@ provide let toString = (char: Char) => {
 
   result
 }
+
+/**
+ * Checks if the first value is less than the second value by Unicode scalar value.
```
ospencer, line 196:
```suggestion
 * Checks if the first character is less than the second character by Unicode scalar value.
```

stdlib/char.gr
```diff
@@ -191,3 +191,71 @@ provide let toString = (char: Char) => {
...
+@unsafe
+provide let (<=) = (x: Char, y: Char) => {
+  from WasmI32 use { leU as (<=) }
+  let x = WasmI32.fromGrain(x)
+  let y = WasmI32.fromGrain(y)
+  x <= y
+}
+
+/**
+ * Checks if the first value is greater than the second value by Unicode scalar value.
```
ospencer, line 230:
```suggestion
 * Checks if the first character is greater than the second character by Unicode scalar value.
```

stdlib/char.gr
```diff
@@ -191,3 +191,71 @@ provide let toString = (char: Char) => {
...
+@unsafe
+provide let (>) = (x: Char, y: Char) => {
+  from WasmI32 use { gtU as (>) }
+  let x = WasmI32.fromGrain(x)
+  let y = WasmI32.fromGrain(y)
+  x > y
+}
+
+/**
+ * Checks if the first value is greater than or equal to the second value by Unicode scalar value.
```
ospencer, line 247:
```suggestion
 * Checks if the first character is greater than or equal to the second character by Unicode scalar value.
```

stdlib/char.gr
```diff
@@ -191,3 +191,71 @@ provide let toString = (char: Char) => {
 
   result
 }
+
+/**
+ * Checks if the first character is less than the second character by Unicode scalar value.
+ *
+ * @param x: The first value
+ * @param y: The second value
+ * @returns `true` if the first value is less than the second value or `false` otherwise
```
phated, line 200:
```suggestion
 * @param x: The first character
 * @param y: The second character
 * @returns `true` if the first character is less than the second character or `false` otherwise
```
You changed the description but not the docs on args

stdlib/char.gr
```diff
@@ -191,3 +191,71 @@ provide let toString = (char: Char) => {
...
+  let y = WasmI32.fromGrain(y)
+  x < y
+}
+
+/**
+ * Checks if the first character is less than or equal to the second character by Unicode scalar value.
+ *
+ * @param x: The first value
+ * @param y: The second value
+ * @returns `true` if the first value is less than or equal to the second value or `false` otherwise
```
phated, line 217:
```suggestion
 * @param x: The first character
 * @param y: The second character
 * @returns `true` if the first character is less than or equal to the second character or `false` otherwise
```

stdlib/char.gr
```diff
@@ -191,3 +191,71 @@ provide let toString = (char: Char) => {
...
+  let y = WasmI32.fromGrain(y)
+  x <= y
+}
+
+/**
+ * Checks if the first character is greater than the second character by Unicode scalar value.
+ *
+ * @param x: The first value
+ * @param y: The second value
+ * @returns `true` if the first value is greater than the second value or `false` otherwise
```
phated, line 234:
```suggestion
 * @param x: The first character
 * @param y: The second character
 * @returns `true` if the first character is greater than the second character or `false` otherwise
```

stdlib/char.gr
```diff
@@ -191,3 +191,71 @@ provide let toString = (char: Char) => {
...
+  let y = WasmI32.fromGrain(y)
+  x > y
+}
+
+/**
+ * Checks if the first character is greater than or equal to the second character by Unicode scalar value.
+ *
+ * @param x: The first value
+ * @param y: The second value
+ * @returns `true` if the first value is greater than or equal to the second value or `false` otherwise
```
phated, line 251:
```suggestion
 * @param x: The first character
 * @param y: The second character
 * @returns `true` if the first character is greater than or equal to the second character or `false` otherwise
```

phated (changes requested):
Need to update the docs and regen markdown, but otherwise looks good.

phated (approved):
Thanks!

PR #1745, opened 2023-03-12
chore(compiler): Update cli version in package lock

spotandjake:
For some reason the version of the cli in `package-lock` was back in `0.4` this pr just brings it up to `0.5.13`


phated:
This is unnecessary churn, as it won't be changed for 0.6. I'll look into a more long term solution when reworking the workflows.

PR #1748, opened 2023-03-14
feat(stdlib)!: Use String length as default end in `String.slice` function

### LINKED ISSUE #1566: Make it easier to `String.slice` to the end of a string

phated:
Closes #1566

This uses defaulted arguments to make the API of `String.slice` nicer. The `end` argument is now required to be used as a label or it is defaulted to the length of the input string.

### DIFF
--- a/compiler/test/stdlib/buffer.test.gr
+++ b/compiler/test/stdlib/buffer.test.gr
@@ -271,7 +271,7 @@ for (let mut i = 0; i < Array.length(slices); i += 1) {
   let (str, start, end) = slices[i]
   let buf = Buffer.make(0)
   Buffer.addStringSlice(start, end, str, buf)
-  assert Buffer.toString(buf) == String.slice(start, end, str)
+  assert Buffer.toString(buf) == String.slice(start, end=end, str)
 }
 
 // addBytesSlice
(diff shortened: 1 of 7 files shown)


spotandjake (commented):
These changes all look good, Glad that this will be easier now. It looks like you forgot to regenerate the graindoc though.

ospencer (approved):
Looks great!

PR #1749, opened 2023-03-14
fix(compiler): Link modules based on crc

ospencer:
This fixes an issue where putting a module on your include path that re-provides other modules fails to compile.


ospencer:
Closing in favor of #1756

Issue #1057, opened 2021-12-08
Use exceptions in the `Err` case of `Number.parseInt`


ospencer:
Right now the errors are just plain strings. It'd be more useful if they were exceptions we could pattern match on.

Blocked on #1053

phated:
This is unblocked because `Result` is available as a language-supplied type now.

spotandjake:
This has been blocked by #1755 or #1759 either one would allow this to move forward.

A decision was made to use enums in the case of exceptions the reason this is blocked is because you cannot re-export the enum to get this working properly.

spotandjake:
We are going to switch this back to using an enum instead of an exception which means this is blocked on #1886

Issue #1053, opened 2021-12-07
Cannot re-export an exception


ospencer:
It's not currently possible to re-export an exception in Grain as there is no syntax to do so. One might try this:

test.gr

```
export exception Foo
```

test2.gr

```
import { Foo } from "./test"

export Foo
```

and they'd be greeted with this unhelpful internal error message:

```
/Users/oscar/Development/grain/cli/bin/grainc.exe: internal error, uncaught exception:
                                                   Not_found
```

This happens because `export` followed by a capitalized identifier looks for a type to export.

PR #1755, opened 2023-03-17
chore(stdlib)!: Replace `parseInt` error strings with structured error enum

spotandjake:
This makes `parseInt` use an exception in the `err` case.

On a note though until #1053 is fixed I dont know if it makes sense to make this change as you lose the ability to pattern match on the individual cases without importing the exceptions directly from the runtime which we dont want users doing.

That could still make it in before 0.6 so this pr is up for that and also just in case we want to move forward on this anyways.

This has been blocked by: #1758   or #1759  either one would work to get this moving.

Closes: #1057

### DIFF
--- a/compiler/src/typed/oprint.re
+++ b/compiler/src/typed/oprint.re
@@ -673,8 +673,7 @@ and print_out_sig_item = ppf =>
         | Otyp_abstract
         | Otyp_tuple(_)
         | Otyp_constr(_, _) => "type"
-        | Otyp_manifest(_, _) =>
-          failwith("NYI: Otyp_manifest pretty-printer")
+        | Otyp_manifest(_, _) => ""
         | Otyp_object(_, _) => failwith("NYI: Otyp_object pretty-printer")
         | Otyp_stuff(_) => failwith("NYI: Otyp_stuff pretty-printer")
         | Otyp_var(_, _) => failwith("NYI: Otyp_var pretty-printer")
--- a/compiler/src/typed/oprint.re
+++ b/compiler/src/typed/oprint.re
@@ -660,8 +660,8 @@ and print_out_sig_item = ppf =>
       //   | Orec_first => "type"
       //   | Orec_next => "and"
       //   };
-      let kwd =
-        switch (td.otype_type) {
+      let rec resolve_kwd = out_type => {
+        switch (out_type) {
         | Otyp_record(_) => "record"
         | Otyp_sum(_) => "enum"
         | Otyp_variant(_, _, _, _) =>
@@ -673,7 +673,7 @@ and print_out_sig_item = ppf =>
         | Otyp_abstract
         | Otyp_tuple(_)
         | Otyp_constr(_, _) => "type"
-        | Otyp_manifest(_, _) => ""
+        | Otyp_manifest(_, original_type) => resolve_kwd(original_type)
         | Otyp_object(_, _) => failwith("NYI: Otyp_object pretty-printer")
         | Otyp_stuff(_) => failwith("NYI: Otyp_stuff pretty-printer")
         | Otyp_var(_, _) => failwith("NYI: Otyp_var pretty-printer")
@@ -682,6 +682,10 @@ and print_out_sig_item = ppf =>
         | Otyp_attribute(_, _) =>
           failwith("NYI: Otyp_attribute pretty-printer")
         };
+      };
+
+      let kwd = resolve_kwd(td.otype_type);
+
       print_out_type_decl(kwd, ppf, td);
     }
   | Osig_value(vd) => {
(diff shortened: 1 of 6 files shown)


spotandjake:
This is now blocked on #1886

phated:
@spotandjake is this unblocked now that the exception re-export PR has been merged?

spotandjake:
> @spotandjake is this unblocked now that the exception re-export PR has been merged?

No this is still blocked. We had decided on a conversation in discord, to use an enum instead of an exception.


#1886 is what is blocking this at the moment.

phated:
@spotandjake I don't remember that conversation exactly. We don't really want many enums in pervasives because they can cause ambiguous type inference with user code.

spotandjake:
> @spotandjake I don't remember that conversation exactly. We don't really want many enums in pervasives because they can cause ambiguous type inference with user code.

Switched back to an `exception`, looks like that did fix the problem I was running into. One note graindoc doesnt seem to be generating documentation for provided exceptions.

ospencer (approved):

compiler/test/stdlib/number.test.gr
```diff
@@ -639,9 +651,9 @@ assert Number.parse("-3/-9") == Ok(1/3)
 assert Number.parse("0x3/-9") == Ok(-1/3)
 assert Number.parse("3/-0x9") == Ok(-1/3)
 assert Number.parse("9223372036854775808/27_670_116_110_564_327_424") == Ok(1/3)
-assert Number.parse("1/2/") == Err("Invalid digit in input")
-assert Number.parse("1/") == Err("Invalid input")
-assert Number.parse("1//") == Err("Invalid digit in input")
+assert Result.isErr(Number.parse("1/2/"))
+assert Result.isErr(Number.parse("1/"))
+assert Result.isErr(Number.parse("1//"))
```
ospencer, line 656: Maybe we should check for the specific error here?

stdlib/number.gr
```diff
@@ -455,9 +466,10 @@ provide let isClose = (a, b, relativeTolerance=1e-9, absoluteTolerance=0.0) => {
  *
  * @param string: The string to parse
  * @param radix: The number system base to use when parsing the input string
- * @returns `Ok(value)` containing the parsed number on a successful parse or `Err(msg)` containing an error message string otherwise
+ * @returns `Ok(value)` containing the parsed number on a successful parse or `Err(msg)` containing a parseInt exception
  *
  * @since v0.4.5
+ * @history v0.6.0: Switched from a string based error message to a custom exception
```
ospencer, line 472:
```suggestion
 * @history v0.6.0: Switched from a string-based error message to a parseInt exception
```

phated:
```suggestion
 * @history v0.6.0: Switched from a string-based error message to a parseInt error type
```

stdlib/number.gr
```diff
@@ -42,6 +42,17 @@ from "runtime/atof/parse" include Parse as Atof
...
+  exception ParseIntEmptyString,
+  exception ParseIntInvalidDigit,
+  exception ParseIntInvalidRadix,
+}
+
+provide {
+  exception ParseIntEmptyString,
+  exception ParseIntInvalidDigit,
+  exception ParseIntInvalidRadix,
+}
```
phated, line 55: I though we decided these would be an enum

compiler/test/stdlib/number.test.gr
spotandjake: I test some of the cases after doing the generic error tests.

spotandjake:
When trying to graindoc this I get the error `Failure("NYI: Otyp_manifest pretty-printer")`, so the graindoc needs to be updated.

compiler/src/typed/oprint.re
```diff
@@ -673,8 +673,7 @@ and print_out_sig_item = ppf =>
         | Otyp_abstract
         | Otyp_tuple(_)
         | Otyp_constr(_, _) => "type"
-        | Otyp_manifest(_, _) =>
-          failwith("NYI: Otyp_manifest pretty-printer")
+        | Otyp_manifest(_, _) => ""
```
spotandjake, line 676: Small change to `oprint` to support docs generation

stdlib/number.md
```diff
@@ -13,6 +13,47 @@ No other changes yet.
...
+
+### Number.**ParseIntError**
+
+<details disabled>
+<summary tabindex="-1">Added in <code>next</code></summary>
+No other changes yet.
+</details>
+
+```grain
+ ParseIntError = Atoi.ParseIntError {
```
spotandjake, line 28: I think we might want to output something in oprint for a type manifest as this looks a little ugly.

spotandjake (commented):
This has been switched to an `enum` now.

stdlib/number.md
```diff
@@ -1119,7 +1168,7 @@ Returns:
 
 |type|description|
 |----|-----------|
-|`Result<Number, String>`|`Ok(value)` containing the parsed number on a successful parse or `Err(msg)` containing an error message string otherwise|
+|`Result<Number, Atoi.ParseIntError>`|`Ok(value)` containing the parsed number on a successful parse or `Err(err)` containing a variant of `ParseIntError`|
```
phated, line 1171: Shoot, this is still broken

phated: This is correct in all regards. It just sucks to see.

phated (approved):

PR #1756, opened 2023-03-17
fix(compiler): Provide full module interface in intermediate wasm files

ospencer:
This PR fixes an issue where a module on the include path that re-provided other included modules could not be compiled as a part of a project. 

Previously, a module re-export was just a pointer to the original module. Now, the intermediate module provides the full interface. This incurs no runtime cost, but compilation may be a tad slower due to the indirection.


phated (approved):
LGTM, thanks for creating the follow up issue.

Issue #1521, opened 2022-12-04
Stdlib: `Array.fillRange` doesn't properly respect negative indexing


phated:
When the index is given as a negative number, this function only subtracts the index from the length but if the `index` is larger than the length, the index is still a negative number. This will cause issues further in the function.

phated:
It also throws Failure instead of appropriate errors. I noticed this in #1721

PR #1761, opened 2023-03-18
fix(stdlib)!: Ensure `Array.fillRange` works with negative indexing & throws IndexOutOfBound

spotandjake:
This pr makes `Array.fillRange` work correctly with negative indexs.
This pr also corrects the `failure`s used to be actual exceptions.

Closes: #1521

### DIFF
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -40,6 +40,18 @@ let checkLength = length => {
 }
 
 /**
+ * A simple helper function to convert a negative array index
+ * number to its positive positional equivalent.
+ */
+let wrapNegativeIndex = (arrLen, idx) => {
+  if (idx >= 0) {
+    idx
+  } else {
+    arrLen + idx
+  }
+}
+
+/**
  * Provides the length of the input array.
  *
  * @param array: The array to inspect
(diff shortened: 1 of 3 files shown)


stdlib/array.gr
```diff
@@ -506,29 +518,31 @@ provide let fill = (value, array) => {
...
-  let startIndex = if (start < 0) length + start else start
-  let stopIndex = if (stop < 0) length + stop else stop
+  let startIndex = wrapNegativeIndex(length, start)
+  let stopIndex = wrapNegativeIndex(length, stop)
+  // Ensure we aren't working with an `stop` value that is too big
+  let stopIndex = if (stopIndex < 0 || stopIndex > length) {
+    length
+  } else {
+    stopIndex
+  }
```
phated, line 535: Just making sure we want this specific behavior. We don't want this to do "cycles" like Array.rotate, right? It's more like Array.slice?

phated (changes requested):
One question for @spotandjake or @grain-lang/core that has an opinion on this API.

stdlib/array.gr
phated: Talked about this on Community call and this looks correct. Will resolve this and approve later.

stdlib/array.gr
```diff
@@ -506,29 +518,31 @@ provide let fill = (value, array) => {
...
  *
  * @since v0.2.0
  */
 provide let fillRange = (value, start, stop, array) => {
   let length = length(array)
-  let startIndex = if (start < 0) length + start else start
-  let stopIndex = if (stop < 0) length + stop else stop
+  let startIndex = wrapNegativeIndex(length, start)
+  let stopIndex = wrapNegativeIndex(length, stop)
+  // Ensure we aren't working with an `stop` value that is too big
```
phated, line 530:
```suggestion
  // Ensure we aren't working with a `stop` value that is too big
```

phated (approved):
Awesome! Thanks 🙇

Issue #1576, opened 2023-01-05
Call Start Function On Each Export


spotandjake:
https://discord.com/channels/676306448594108416/1059596404898680862

After talking to @ospencer in the discord about the `_start` function. It was mentioned that `rust-wasm` calls the start function on each exported function if the module has not been initialized yet. This would allow for a better user experience when using with external hosts because the user could use:
```
wasmModule.add(a, b);
```
instead of having todo:
```
wasmModule._start();
wasmModule.add(a, b);
```

It would be nice if this behavouir could be brought over to grain given one of grains goals is towards developer experience.

I might be slightly underestimating this but I am pretty sure what needs to be done here is a wasm global needs to be added to the module indicating if the start function has been run. The start function needs to set that global to true(1) at the top of the start function and each exported function needs to get the code:
```
    global.get $ModuleInitialized
    (if
      (then
        call $_start
      )
    )
   ...functionContents
```

spotandjake:
Closing this issue as when we tried to implement it with #1765 we came to the conclussion that the behaviour of `--use-start-section` better fulfills the goals of this change.

PR #1765, opened 2023-03-18
Feat: Call Start Function On Export

spotandjake:
This is a developer experience improvment it allows a user to run a grain program in wasmtime for example by using `wasmtime index.gr.wasm --invoke randomExport` while still working.

This pr adds a new flag to the grain cli `--call-start` which causes any exported grain function to call the `_start` function when it is called this is wrapped with an if and the start function sets a boolean flag for weather it has run or not. 

Cons:
The way I implemented it is the export checks the if statement of weather or not the module has been initalized and if not it calls the `_start` function which sets a bool on start. This approach means slightly bigger module size but I was hoping it was better for performance as now you are not invoking a function.
A general con of this is any module using this is slightly larger and slightly slower its negligable though and well worth the developer experience benefit.

Pros:
This makes invoking exports from externally from a host runtime a lot more seamless.

Questions:
Does it make sense to prioritize size or speed? would it make sense to call the `_start` function always and check if it has been run before inside the `_start` function this would lower module size but I speculate it would be slower. or do we want todo it the way I have implemented it here.
Another small question is as this is a developer experience improvment currently I implemented it as a flag to toggle on but would it make sense as a flag to toggle off instead.

Reason for the tests changing:
This adds two new imports to the module before it is linked these get resolved during linking and if the flag isnt called they will get optimized out when running with `--release`. This isnt neccessairy but i was running into an issue where when checking the config vairable in the main scope it was false and as such the globals would not get appended to the import list.

The `cmi`s are different sizes on all the tests which is wierd I dont think i touched the cmi's in any way?

Closes: #1576


ospencer:
Great work getting started on this @spotandjake! We pretty much have to use an if statement to check if the module initialization has been run, because we can't initialize more than once! That'd be bad because all of the side effects would repeat, which could have surprising effects. I would probably do this during the link step, though. The linker creates one mega-start function, which is the only one you'd need to call.

spotandjake:
> Great work getting started on this @spotandjake! We pretty much have to use an if statement to check if the module initialization has been run, because we can't initialize more than once! That'd be bad because all of the side effects would repeat, which could have surprising effects. I would probably do this during the link step, though. The linker creates one mega-start function, which is the only one you'd need to call.

Alright then I already put the global change in the linker there so it shouldn't be that hard to just make each function call the initialization and then move the if statement.

ospencer:
For that, I think you just need to rebase.

spotandjake:
> Great work getting started on this @spotandjake! We pretty much have to use an if statement to check if the module initialization has been run, because we can't initialize more than once! That'd be bad because all of the side effects would repeat, which could have surprising effects. I would probably do this during the link step, though. The linker creates one mega-start function, which is the only one you'd need to call.

Switched to doing it that way is there a way to get `makeFileSnapshot` perform linking? I realized the test I added doesnt snapshot after linking just before linking?

ospencer:
There is, but we don't snapshot fully linked programs (too volatile). Not a great way to test this, but I'll let you know if I think of one.

spotandjake:
Also those cmi's still seem to differ from my local build even though I rebased.

ospencer:
I'd start with removing the changes to `compcore` just to eliminate any weirdness that could possibly be happening.

Yeah, `assertRun` won't work without a `--invoke` for `grain run`, but that wouldn't belong in this PR and will take a long time to get right.

As for a TODO, I would just make an issue.

spotandjake:
> I'd start with removing the changes to `compcore` just to eliminate any weirdness that could possibly be happening.
> 
> Yeah, `assertRun` won't work without a `--invoke` for `grain run`, but that wouldn't belong in this PR and will take a long time to get right.
> 
> As for a TODO, I would just make an issue.

Removing The Changes to compcore  would remove the feature. You still need to make the exported functions call the start function

ospencer (commented):
I'm not sure why the CMIs changed here.

I've thought about this a bit, and we actually already (kinda) have this feature in the form of `--use-start-section`. For the wasmtime example, if you pass that flag when compiling and then attempt to call some export, it would work.

This is not the default behavior, and thus a little less than ideal, but we'll no longer have a `_start` when we move to wasi preview 2, and thus this will become the default behavior in Grain 0.7, so I think we should just hold off until then.

What do you think?

spotandjake:
> I'm not sure why the CMIs changed here.
> 
> I've thought about this a bit, and we actually already (kinda) have this feature in the form of `--use-start-section`. For the wasmtime example, if you pass that flag when compiling and then attempt to call some export, it would work.
> 
> This is not the default behavior, and thus a little less than ideal, but we'll no longer have a `_start` when we move to wasi preview 2, and thus this will become the default behavior in Grain 0.7, so I think we should just hold off until then.
> 
> What do you think?

That is 100% fair, should i convert this to draft, in the meantime, or just close it?

phated:
Given Oscar's comments, I think it makes sense to close this. Sorry for the delayed feedback @spotandjake

PR #1767, opened 2023-03-18
feat(graindoc): Add test harness

spotandjake:
This pr adds a test suite for `graindoc` it has some basic tests so far but could use a lot more still. The testing was setup the same way the formatters testing was setup.


compiler/test/TestFramework.re
```diff
@@ -32,6 +32,9 @@ let test_snapshots_dir = Fp.At.(test_dir / "__snapshots__");
 let test_formatter_in_dir = Fp.At.(test_dir / "formatter_inputs");
 let test_formatter_out_dir = Fp.At.(test_dir / "formatter_outputs");
 
+let test_gaindoc_in_dir = Fp.At.(test_dir / "graindoc_inputs");
+let test_gaindoc_out_dir = Fp.At.(test_dir / "graindoc_outputs");
```
phated, line 36: @spotandjake Instead of having these as separate directories, can we have them in the same directory and name them `*.input.gr` and `*.expected.md`? I want to make this change to the formatter too because it's really easy to get lost/confused while working on the formatter stuff.

spotandjake: I think that's a good change I made it.

phated (approved):
This makes me sOoOoOo happy! Thank you @spotandjake 🤗

PR #1770, opened 2023-03-19
chore(ci): Convert CI workflows to reusable & apply various QoL changes

### LINKED ISSUE #1409: CI: Workflow refactor
The workflows should be refactored so each "step" is essentially its own workflow and then our primary workflows use https://github.com/grain-lang/workflow-dispatch-action to provide the necessary details to each workflow.

This would allow us to "recover" any failed steps (such as steps in our release workflow).

phated:
Closes #1409

This updates the CI to use reusable workflows, which will allow us to construct more complex workflows (such as a nightly build). Some of the splitting is unintuitive so I tried to document why specific workflows exist.

The best thing about these changes are that built artifacts are attached to the workflow run, such as the native build executables and the pkg binaries, so we can use them later (see the test-pkg workflow) or download them to debug, etc.

I also took the opportunity to refactor a bit so it's easier to figure out what failed without having to look at logs.

### DIFF
--- a/.github/workflows/build-js.yml
+++ b/.github/workflows/build-js.yml
@@ -1,0 +1,115 @@
+# This workflow exists to provide a way to dispatch a CI run for any
+# given ref on any of our OS targets. It can also be consumed in our
+# various other builds.
+name: (js) Build and test
+
+on:
+  workflow_dispatch:
+    inputs:
+      os:
+        description: "Operating system to run CI on"
+        required: true
+        default: "ubuntu-latest"
+        type: choice
+        options:
+          - ubuntu-latest
+          - macos-latest
+          - windows-latest
+  workflow_call:
+    inputs:
+      os:
+        type: string
+        required: true
+
+jobs:
+  build:
+    name: (js) Build and test
+    runs-on: ${{ inputs.os }}
+
+    steps:
+      - name: Checkout project
+        uses: actions/checkout@v3
+
+      - name: Setup node.js
+        uses: actions/setup-node@v3.6.0
+        with:
+          node-version: "16"
+          check-latest: true
+          cache: "npm"
+
+      - name: Install npm dependencies
+        run: |
+          npm ci
+
+      - name: Esy cache
+        id: esy-cache
+        uses: actions/cache/restore@v3
+        with:
+          path: compiler/_export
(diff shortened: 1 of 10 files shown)


.github/workflows/build-js.yml
```diff
@@ -0,0 +1,115 @@
...
+      # Don't build native executables, only the JS builds
+      - name: Build compiler
+        run: |
+          npm run compiler build:js
+
+      - name: Run tests
+        run: |
+          npm run compiler test:js
+
+      # This will log a warning because we removed don't have a native exe files
```
ospencer, line 67: Can you reword these?

ospencer (commented):
Looks great overall! Just the one nit.

.github/workflows/build-js.yml
phated:
```suggestion
      # This will log a warning because we didn't build the native exe files
```

.github/workflows/build-js.yml
```diff
@@ -0,0 +1,115 @@
...
+          tar -cvf grain.tar -C pkg grain.exe
+
+      - name: Upload pkg binary for Windows
+        if: inputs.os == 'ubuntu-latest'
+        uses: actions/upload-artifact@v3
+        with:
+          path: ./grain.tar
+          name: grain-win-x64
+
+      # This will log a warning because we removed don't have a native exe files
```
phated, line 81:
```suggestion
      # This will log a warning because we didn't build the native exe files
```

.github/workflows/build-js.yml
```diff
@@ -0,0 +1,115 @@
...
+          tar -cvf grain.tar -C pkg grain
+
+      - name: Upload pkg binary for Mac
+        if: inputs.os == 'ubuntu-latest'
+        uses: actions/upload-artifact@v3
+        with:
+          path: ./grain.tar
+          name: grain-mac-x64
+
+      # This will log a warning because we removed don't have a native exe files
```
phated, line 95:
```suggestion
      # This will log a warning because we didn't build the native exe files
```

ospencer (approved):

phated:
I just realized I didn't put concurrency groups on these. Converting to draft for now.

phated:
Added concurrency group to ci.yml, so this is promoted again.

ospencer (approved):

Issue #671, opened 2021-05-27
Wrong parameter ordering for List.map and similar functions


rkrisztian:
You have probably heard of the Yoda code style, or Yoda conditionals (_blue is the sky_, usually expressed as `"blue" == sky`). We don't like such style because they break the reader's natural flow or reading: they put the blue in focus, rather than the sky, which is the main subject of the comparison). What does that have anything to do with code like the following?

```none
List.map((n) => n + 3, [1, 2, 3])
```

Two things:

1. When I read this statement, I read it as _map to `n+3` the list `[1, 2, 3]`_. Rather than something like, _map `[1, 2, 3]` to `n+3`_. 
2. When the last parameter is the function expression, the function part can be naturally broken down into multiple lines. Consider the following bad design from JavaScript:
    ```javascript
    setTimeout(() => {
        alert("Hello");
    }, 3000);
    ```
    Arguably it can be read as _Do an alert 3 seconds later_, but it sounds closer to the original code if I say _set for this function a 3-second timeout_, rather than _set a 3-second timeout for this function_.
    In Groovy we know that if the closure is the last parameter, then we can write it in a more readable style (this is hypothetical code):
    ```groovy
    setTimeout(3000) {
        alert 'Hello'
    }
    ```
    Also consider Gradle task declarations. They would likely irritate the reader if they were created like this:
    ```groovy
    task({
         // configure this task...
    }, 'myTask')
    ```

So please reconsider the ordering of parameters. Make them more natural to readers. Sadly several languages don't get it right. And I didn't even mention that `[1, 2, 3].map((n) => n + 3)` would look even more natural to read. The static method call way is verbose. The best languages combine OOP and FP styles together. See Kotlin for example.

Conaclos:
I second this!

ReScript has made this swicth. An interesting article about that is available [here](https://www.javierchavarri.com/data-first-and-data-last-a-comparison/).

rkrisztian:
Thank you, @Conaclos, interesting indeed! I find the pros strong ones, but the cons weak ones, and here is why. Look at this currying:

```none
let add = (a, b) => a + b;
let addTwo = add(2);
```

A human can interpret `add(2)` both as a function call, and as a partial function reference. The only reason a compiler can tell which one it is, is that a function call like that would be a syntax error. To me such designs are unclean and confusing, and adds mental mapping (which can be a real problem in real applications).

Now if I look at the placeholder way:

```none
let addTwo = add(_, 2);
```

[...]

PR #1772, opened 2023-03-20
chore(stdlib)!: Use default arguments in more of stdlib

### LINKED ISSUE #697: BOM arguments to `String.encode` and `String.decode` should be optional
Blocked by #388, but it's rather frustrating to always need to specify these.

Further context: #683

alex-snezhko:
Makes use of default arguments in various places of the stdlib where I thought they made sense.

I was also considering combining `unwrap` and `expect` in `Option`/`Result` but was unsure if I should.

Closes #697


stdlib/map.gr
```diff
@@ -32,29 +32,25 @@ abstract record Map<k, v> {
...
 
 /**
- * Creates a new empty map with an initial storage of the given size. As values are added or removed, the internal storage may grow or shrink. Generally, you won't need to care about the storage size of your map and can use `Map.make()` instead.
+ * Creates a new empty map with an initial storage of the given size. As
+ * values are added or removed, the internal storage may grow or shrink.
+ * Generally, you won't need to care about the storage size of your map and
+ * can use the default size.
  *
- * @param size: The initial storage size of the map
+ * @param size: The initial storage size of the map (default is 16)
```
phated, line 40: Let's just fix the docgen so we don't need to specify this

stdlib/priorityqueue.gr
```diff
@@ -220,12 +207,13 @@ provide let drain = pq => {
...
  * @param array: An array of values used to initialize the priority queue
- * @param comp: A comparator function used to assign priority to elements
+ * @param comp: A comparator function used to assign priority to elements (default is `compare` pervasive)
  * @returns A priority queue containing the elements from the array
  *
  * @since v0.5.4
+ * @history v0.6.0: Made `comp` a default argument
  */
-provide let fromArray = (array, comp) => {
+provide let fromArray = (array, comp=compare) => {
```
phated, line 216: Let's rename all these "comp" to compare

phated (changes requested):
Quick skim, 2 things jumped out:
1. graindoc needs to be fixed and you shouldn't be embedding defaults into the descriptions
2. No shortened names for defaulted params, we want to be very clear with their names.

stdlib/priorityqueue.gr
alex-snezhko: You pointing this out also brought to my attention that the sort comparator could also be optional

spotandjake (commented):
I didnt look over any of the test changes just the code changes but I am liking the look of this pr.

phated:
blocked on #1776

alex-snezhko:
#1776 merged, no longer blocked

spotandjake (approved):
This looks good to me.

stdlib/array.gr
```diff
@@ -946,16 +946,18 @@ provide let slice = (start, end=length(array), array) => {
...
  * Ordering is calculated using a comparator function which takes two array elements and must return 0 if both are equal, a positive number if the first is greater, and a negative number if the first is smaller.
- * @param comp: The comparator function used to indicate sort order
  * @param array: The array to be sorted
+ * @param compare: The comparator function used to indicate sort order
+ * 
  * @since v0.4.5
+ * @history v0.6.0: Made `compare` a default argument
  */
-provide let sort = (comp, array) => {
+provide let sort = (array, compare=compare) => {
```
ospencer, line 955: Why did you change the argument order?

alex-snezhko: No reason in particular other that I prefer default arguments appearing after required ones in the definition. I can change it back though

ospencer: I don't know if it's the OCamler in me or what, but I prefer the default arguments before the required ones. It's probably the OCamler in me.

phated: I love them at the end because it sorts them to the bottom of the documentation. So yeah, I think it's the ocamler in you 😂

phated: Additionally, default arguments at the end are familiar to JavaScript/TypeScript folks.

alex-snezhko: I think as long as we're doing data-last in the rest of the stdlib then it makes sense to have the `compare` arg before the array/list. This is kinda the same issue/argument as https://github.com/grain-lang/grain/issues/671

stdlib/string.gr
```diff
@@ -1425,29 +1409,18 @@ let encodeHelp = (string: String, encoding: Encoding, includeBom: Bool) => {
 }
 
 /**
- * Encodes the given string using the given encoding scheme, excluding any byte-order marker.
+ * Encodes the given string using the given encoding scheme.
  *
  * @param string: The input string
  * @param encoding: The encoding to use
+ * @param includeBom: Whether or not to include a byte order marker (false by default)
```
phated, line 1416: We should add a `@default` attribute to graindoc that allows us to document defaults in plain language to avoid parentheticals like this. I'm not going to block on it though.

phated (approved):
This is an absolute delight! Thanks @alex-snezhko 🎉

phated:
@alex-snezhko can you resolve the conflicts please?

stdlib/string.gr
phated: #1991

Issue #1775, opened 2023-03-21
Inconsistent argument label erasure when function given explicit type without argument labels


alex-snezhko:
When a function is given an explicit type (without argument labels) but the function *does* have labels for its arguments, within some places of the compiler (e.g. `Types.value_description`), the function's argument labels get erased due to the explicit type taking precedence.

For example with the following code:
```
let f: Number -> Void = a => void
```
`f`'s argument is treated as unlabeled when I attempt to access it through a `value_description` instance because of the explicit type signature, when in reality it does have a label, `a`.

However, this issue does not exist everywhere; for example with the same code I get the correct label in the code lens
![image](https://user-images.githubusercontent.com/48104782/226696471-2ad94fa9-e438-48e2-9ea4-30a263eec7ed.png)

PR #1776, opened 2023-03-21
feat(graindoc)!: Improve docgen for labeled & default arguments

### LINKED ISSUE #1763: GrainDoc: Labeled Arguments Get The Type `Option<a>` in param list
Labeled arguments get given the type `Option<a>` in the param list the generated graindoc as can be seen in the image below this should probably instead just be `a` given its defaulted to the same type or `?a` to represent that it is defaulted.

image
I think this miht have something todo with the todo that was left after the module update in the image below.
image

alex-snezhko:
Several improvements to graindoc w.r.t. labeled & default arguments.

- `@param` will now read the labels of parameters rather than their positions, meaning they can be written out of order as long as the names match. `@param unlabeled <argName>` can be used instead for unlabeled arguments, which uses the order of the parameter as was the old behavior (placeholder syntax for now, open to other ideas)
- Default parameters no longer show as `Option<a>` in graindoc output (closes #1763)
- Default parameters now have their default value shown in a new markdown column of the graindoc output if there are any default parameters for a function

### DIFF
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -166,7 +166,7 @@ let output_for_history = (~current_version, {history_version, history_msg}) => {
   };
 };
 
-let output_for_params = params =>
+let output_for_params = params => {
   Markdown.table(
     ~headers=["param", "type", "description"],
     List.map(
@@ -176,6 +176,7 @@ let output_for_params = params =>
       params,
     ),
   );
+};
 
 let output_for_returns = ({returns_type, returns_msg}) => {
   Markdown.table(
@@ -223,7 +224,7 @@ let types_for_function = (~ident, vd: Types.value_description) => {
 let lookup_arg_by_label = (name, args_opt) => {
   Option.bind(args_opt, args =>
     List.find_opt(
-      ((label: Grain_parsing.Parsetree.argument_label, _)) =>
+      ((label: Grain_parsing.Asttypes.argument_label, _)) =>
         switch (label) {
         | Default(l)
         | Labeled(l) => l.txt == name
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -261,21 +261,38 @@ let lookup_type_expr = (~idx, type_exprs) => {
   Option.bind(type_exprs, te => List.nth_opt(te, idx));
 };
 
-let get_argument_default = (input_path, default_expr) => {
-  open Grain_formatting.Format;
+let get_original_code = (loc_start, loc_end, source: array(string)) => {
+  let (_, start_line, startc, _) =
+    Locations.get_raw_pos_info(loc_start);
+  let (_, end_line, endc, _) = Locations.get_raw_pos_info(loc_end);
+
+  if (start_line == end_line) {
+    String_utils.Utf8.sub(source[start_line - 1], startc, endc - startc);
+  } else {
+    let text = ref("");
+    for (line in start_line - 1 to end_line - 1) {
+      if (line + 1 == start_line) {
+        text :=
+          text^
+          ++ String_utils.Utf8.string_after(source[line], startc)
+          ++ "\n";
+      } else if (line + 1 == end_line) {
+        text := text^ ++ String_utils.Utf8.sub(source[line], 0, endc);
+      } else {
+        text := text^ ++ source[line] ++ "\n";
+      };
+    };
+    text^;
+  };
+};
+
+let get_argument_default = (input_path, default_expr: Grain_parsing.Parsetree.expression) => {
+  // We have already compiled this file so we know it exists
   let ic = open_in_bin(input_path);
   let source = really_input_string(ic, in_channel_length(ic));
   close_in(ic);
-  // We have already compiled this file so we know it exists
-  let (_, lines, _) = Result.get_ok(parse_source(source));
-  let expr_doc =
-    print_expression(
-      ~expression_parent=GenericExpression,
-      ~original_source=lines,
-      ~comments=[],
-      default_expr,
-    );
-  Doc.toString(~width=1000, ~eol=Fs_access.LF, expr_doc);
+  let lines = Array.of_list(String.split_on_char('\n', source));
+  get_original_code(default_expr.pexp_loc.loc_start, default_expr.pexp_loc.loc_end, lines);
 };
 
 let get_comments_from_loc = (loc: Grain_parsing.Location.t) => {
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -4,7 +4,7 @@ open Grain_utils;
 open Grain_diagnostics;
 
 type param = {
-  param_name: string,
+  param_id: string,
   param_type: string,
   param_msg: string,
   param_default_value: option(string) // For default arguments
@@ -79,7 +79,7 @@ exception
   });
 
 exception MissingLabeledParamType({name: string});
-exception MissingUnlabeledParamType({name: string});
+exception MissingUnlabeledParamType({idx: int});
 exception MissingReturnType;
 exception
   InvalidAttribute({
@@ -106,11 +106,11 @@ let () =
           name,
         );
       Some(msg);
-    | MissingUnlabeledParamType({name}) =>
+    | MissingUnlabeledParamType({idx}) =>
       let msg =
         Printf.sprintf(
-          "Unable to find a type for %s. Did you specify too many @param attributes?",
-          name,
+          "Unable to find a type for parameter at index %d. Make sure a parameter exists at this index in the parameter list.",
+          idx,
         );
       Some(msg);
     | MissingReturnType =>
@@ -170,19 +170,19 @@ let output_for_history = (~current_version, {history_version, history_msg}) => {
 let output_for_params = params =>
   if (List.exists(param => Option.is_some(param.param_default_value), params)) {
     Markdown.table(
-      ~headers=["param", "type", "description", "default value"],
+      ~headers=["param", "type", "default", "description"],
       List.map(
-        ({param_name, param_type, param_msg, param_default_value}) => {
+        ({param_id, param_type, param_msg, param_default_value}) => {
           let default =
             switch (param_default_value) {
             | Some(default) => Markdown.code(default)
-            | None => "N/A"
+            | None => ""
             };
           [
-            Markdown.code(param_name),
+            Markdown.code(param_id),
             Markdown.code(param_type),
-            param_msg,
             default,
+            param_msg,
           ];
         },
         params,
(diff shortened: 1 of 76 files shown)


compiler/graindoc/dune
```diff
@@ -2,7 +2,7 @@
  (name docblock)
  (public_name graindoc.docblock)
  (modules docblock)
- (libraries grain grain_diagnostics grain_utils))
+ (libraries grain grain_diagnostics grain_utils grain_formatting))
```
phated, line 5:
```suggestion
 (libraries grain grain_diagnostics grain_utils))
```
When we print with printtyp or oprint.

compiler/src/diagnostics/graindoc_lexer.re
```diff
@@ -52,6 +52,9 @@ let newline = [%sedlex.regexp? 0x0A | 0x0C | 0x0D | 0x85 | 0x2028 | 0x2029];
 // A colon eats whitespace after it
 let colon = [%sedlex.regexp? (':', Star(blank))];
 
+// "unlabeled" eats whitespace after it
+let unlabeled = [%sedlex.regexp? ("unlabeled", Star(blank))];
```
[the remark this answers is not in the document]
phated: Alternatively, we could use `0`, `1`, `2`, etc for positional params

stdlib/array.gr
```diff
@@ -70,7 +70,7 @@ provide let length = array => {
  * @since v0.1.0
  */
 @unsafe
-provide let make: (Number, a) -> Array<a> = (length: Number, item: a) => {
+provide let make = (length: Number, item: a) => {
```
phated, line 73: Why'd you need to delete these? Should we instead change them to the labeled signature?

phated (changes requested):
I'm SO excited for this! @alex-snezhko you have no idea how much I've dreamed about this graindoc implementation (probably since I created "optional arguments" issue). I love that we will look up docs by label and warn/error if they aren't correct ❤️ 

I have a few thoughts about the implementation but the big one is bringing in grain_formatting.

compiler/graindoc/docblock.re
```diff
@@ -207,18 +238,46 @@ let output_for_throws = throws => {
...
+  // We have already compiled this file so we know it exists
+  let (_, lines, _) = Result.get_ok(parse_source(source));
+  let expr_doc =
+    print_expression(
+      ~expression_parent=GenericExpression,
+      ~original_source=lines,
+      ~comments=[],
+      default_expr,
+    );
+  Doc.toString(~width=1000, ~eol=Fs_access.LF, expr_doc);
```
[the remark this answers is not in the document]
alex-snezhko: I used the printing code from the formatter as it seemed well-suited for the task of taking a `Parsetree.expression` and printing it out as a string with proper formatting

compiler/src/diagnostics/graindoc_lexer.re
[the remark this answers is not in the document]
alex-snezhko: I think I like the idea of using numbers for positional params

compiler/src/parsing/parser.mly
```diff
@@ -308,7 +308,7 @@ typ:
 
 arg_typ:
   | LIDENT colon typ { TypeArgument.mk ~loc:(to_loc $loc) (Labeled (mkstr $loc($1) $1)) $3 }
-  | QUESTION LIDENT colon typ { TypeArgument.mk ~loc:(to_loc $loc) (Default (mkstr $loc($2) $2)) $4 }
+  | QUESTION LIDENT colon typ { TypeArgument.mk ~loc:(to_loc $loc) (Default ((mkstr $loc($2) $2), (Expression.constant Constant.void) (* dummy value *))) $4 }
```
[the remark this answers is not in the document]
alex-snezhko: Just because it seemed to conform better for what I was trying to do. This `expression` is only read when printing the default value of a parameter, and this value would always be filled in properly when parsing a function declaration. If I were to use an option here instead there would still more or less be an `Option.get` used to get the value

stdlib/array.gr
alex-snezhko: Related to #1775; I could change them to the labeled signature but I figured it would just be extraneous information since the param types and return type are already specified in the implementation

compiler/graindoc/docblock.re
```diff
@@ -239,7 +298,12 @@ let get_comments_from_loc = (loc: Grain_parsing.Location.t) => {
 };
 
 let for_value_description =
-    (~module_namespace, ~ident: Ident.t, vd: Types.value_description) => {
+    (
+      ~module_namespace,
+      ~ident: Ident.t,
+      ~input_path: string,
```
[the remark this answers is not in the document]
phated: That's the problem 🙃

phated:
Blocked on the new formatter.

compiler/graindoc/docblock.re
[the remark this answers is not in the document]
alex-snezhko: Update: formatter dep removed

ospencer:
I think for now we should drop printing what the default value is—we chatted about this at the core team meeting and felt there were too many situations where the value provided wouldn't make sense, especially in the cases where the default references a binding or computes a value, where not having the surrounding context makes the default meaningless.

For now we can have a column that just reports a yes/no on if the value has a default, and maybe we can figure out something more clever in the future.

alex-snezhko:
Should the column just be omitted altogether in favor of prefixing the parameter names with ? like they are in other places?

ospencer:
Yeah, I think so.

compiler/graindoc/docblock.re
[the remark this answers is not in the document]
alex-snezhko: Removed as no longer needed

alex-snezhko:
Changed the format to include `?` as a prefix for parameters with default args rather than printing the default value

compiler/src/parsing/parsetree.re
```diff
@@ -19,107 +19,11 @@ type provide_flag =
...
-type data_declaration = {
-  pdata_name: loc(string),
-  pdata_params: list(parsed_type),
-  pdata_kind: data_kind,
-  pdata_manifest: option(parsed_type),
-  pdata_rec: rec_flag,
-  [@sexp_drop_if sexp_locs_disabled]
-  pdata_loc: Location.t,
-};
+type attributes = Asttypes.attributes;
```
phated, line 26: did this get duplicated in some rebase or merge?

stdlib/pervasives.gr
```diff
@@ -48,7 +48,7 @@ from String use { toString, print, concat as (++) }
  * Computes the logical NOT (`!`) of the given operand.
  * Inverts the given Boolean value.
  *
- * @param value: The operand
+ * @param 0: The operand
```
phated, line 51: Are these positional because we don't have labels on primitives? I believe @spotandjake is going to update all the primitives to have labels like he did in #1979

stdlib/runtime/dataStructures.gr
```diff
@@ -4,7 +4,7 @@ module DataStructures
 /**
  * Allocates a new Grain array.
  *
- * @param numElts: The number of elements to be contained in this array
+ * @param 0: The number of elements to be contained in this array
```
phated, line 7: Same question about primitives as previously

stdlib/runtime/wasi.gr
```diff
@@ -77,10 +77,10 @@ provide foreign wasm fd_prestat_dir_name: (
 /**
  * Invokes the `fd_write` system call.
  *
- * @param file_descriptor: The file descriptor to write to
- * @param iovs: The pointer to the array of iovs to write
- * @param iovs_len: The length of the array of iovs
- * @param nwritten: Where to store the number of bytes written
+ * @param 0: The file descriptor to write to
```
phated, line 80: Same question about primitives, does it also apply to foreigns?

compiler/src/typed/btype.re
```diff
@@ -408,27 +408,30 @@ let forget_abbrev = (mem, path) =>
 /*  Utilities for labels          */
 /**********************************/
 
+module Parsetree = Grain_parsing.Parsetree;
```
phated, line 411: This should not be using a Parsetree type. We alias the Asttypes and then use the alias in the `typed/` directory

compiler/src/parsing/asttypes.re
```diff
@@ -115,9 +115,3 @@ type attribute = (loc(string), list(loc(string)));
 
 [@deriving (sexp, yojson)]
 type attributes = list(attribute);
-
-[@deriving (sexp, yojson)]
-type argument_label =
-  | Unlabeled
-  | Labeled(loc(string))
-  | Default(loc(string));
```
phated, line 123: Why was this moved out of Asttypes instead of being aliased in `parsetree.re` like we do for `mut_flag`, `rec_flag`, etc?

compiler/src/parsing/parsetree.re
```diff
@@ -475,16 +351,146 @@ and use_item =
...
+  ptyp_arg_type: parsed_type,
+  ptyp_arg_loc: Location.t,
+}
 
 [@deriving (sexp, yojson)]
-type attributes = Asttypes.attributes;
+and argument_label =
+  | Unlabeled
+  | Labeled(loc(string))
+  | Default(loc(string))
```
phated, line 381: See above about aliasing Asttypes

compiler/src/typed/typedtree.rei
```diff
@@ -41,7 +41,7 @@ type provide_flag =
 type rec_flag = Asttypes.rec_flag = | Nonrecursive | Recursive;
 type mut_flag = Asttypes.mut_flag = | Mutable | Immutable;
 type argument_label =
-  Asttypes.argument_label =
+  Parsetree.argument_label =
     | Unlabeled | Labeled(loc(string)) | Default(loc(string));
```
phated, line 45: This looks to be the alias, so if you do the Asttypes aliasing in parsetree, you don't need to change anything in this directory

phated (changes requested):
This is really close! 💪 I think you can remove a bunch of your changes around moving the `Asttypes.argument_label` and that'll reduce the diff by a ton.

compiler/src/parsing/asttypes.re
alex-snezhko: I don't actually remember why I did this. I'll revert this back into asttypes

stdlib/pervasives.gr
alex-snezhko: Yep, at this point there aren't labels on primitives

stdlib/runtime/wasi.gr
alex-snezhko: Actually I don't think there's anything prohibiting using named arguments on foreigns, I think we just haven't gotten around to assigning argument names to all of the foreigns in the stdlib. I can do this in this PR if you'd like or alternatively in a followup PR

stdlib/pervasives.gr
phated: #1981

stdlib/runtime/wasi.gr
phated: #1982

compiler/src/parsing/asttypes.re
phated: Thanks, the diff looks great now!

phated (approved):

PR #1782, opened 2023-03-29
feat: Add A Special Error for includes ending in `.gr`

spotandjake:
This changes the error when an include is not found and ends with `.gr` to say `Missing file for module foo.gr did you mean foo` otherwise it is left the same in the case of you trying to include a file named `foo.gr.gr` it would not error, This is really just a slight dx improvement.
I think this may be breaking but I am not sure, might be breaking because it changes the behiavour of an error let me know.

### DIFF
--- a/compiler/src/typed/module_resolution.re
+++ b/compiler/src/typed/module_resolution.re
@@ -552,8 +552,15 @@ let report_error = ppf =>
         "was not found",
       );
     }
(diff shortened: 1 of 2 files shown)


phated:
I think we are going to forgo this error message and instead require extensions for local filepaths, see #1797

PR #1783, opened 2023-03-29
chore(compiler): Use `String.ends_with` to check last character

spotandjake:
This just handles a single `TODO` related to using `String.ends_with`


compiler/src/utils/fs_access.re
```diff
@@ -12,12 +12,11 @@ let determine_eol = line => {
   | Some(line) when String.length(line) > 0 =>
     // check what the last char was
     // TODO: Replace with `String.ends_with` in OCaml 4.13
```
phated, line 14:
```suggestion
```
Probably want to delete the TODO, right? 😛

phated (approved):

PR #1787, opened 2023-03-29
feat(grainlsp): Implement goto definition

### LINKED ISSUE #1360: LSP: Support for go to definition navigation
Currently cmd  clicking on function invocation does not go to function definition, similarly for module name as well. We could implement go to defination functionality for better developer experience

marcusroberts:
This PR extends the LSP support to add goto definition (and its companion peek definition)

image

Closes #1360


marcusroberts:
Much of the hard work on this came from @ospencer !

alex-snezhko (approved):
LGTM

compiler/src/language_server/protocol.rei
```diff
@@ -40,6 +41,19 @@ type location = {
...
+[@deriving yojson]
+type location_link = {
+  [@key "originSelectionRange"]
+  origin_selection_range: range,
+  [@key "targetUri"]
+  target_uri: uri,
+  [@key "targetRange"]
+  target_range: range,
+  [@key "targetSelectionRange"]
+  target_selection_range: range,
```
phated, line 54:
```suggestion
  origin_selection_range: range,
  target_uri: uri,
  target_range: range,
  target_selection_range: range,
```
I'm pretty sure that key attributes shouldn't go in the interface files.

compiler/src/language_server/protocol.rei
```diff
@@ -142,6 +156,13 @@ type notification_message = {
   params: Yojson.Safe.t,
 };
 
+//https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#definitionClientCapabilities
+[@deriving yojson({strict: false})]
+type definition_client_capabilities = {
+  [@key "linkSupport"]
```
phated, line 162:
```suggestion
```

compiler/src/language_server/protocol.rei
```diff
@@ -154,3 +175,5 @@ let notification: (~method: string, Yojson.Safe.t) => unit;
 
 let uri_to_filename: uri => string;
 let filename_to_uri: string => uri;
+
+let loc_to_range: Grain_parsing.Location.t => range;
```
phated, line 179: This isn't really the "protocol" but instead a grain-to-protocol translation. It's probably fine here for now but we'll need to refactor in the future.

phated (changes requested):
Looking good. I didn't try it out yet but I think we need a bit of cleanup.

compiler/src/language_server/protocol.rei
phated: Nevermind this one. I think I am wrong.

compiler/src/language_server/protocol.rei
phated: And ignore this.

compiler/src/language_server/protocol.rei
marcusroberts: Agreed, I moved it (and other utilities) to a Utils module.  Is that more what you were looking for?

esy.lock/index.json
phated, line 1: @marcusroberts it looks like you accidentally added some esy files. I think you accidentally ran esy in the root of the project. Please remove

compiler/src/language_server/protocol.rei
```diff
@@ -14,7 +15,7 @@ type version;
 type message_id;
 
 [@deriving yojson]
-type uri;
+type uri = Uri.t;
```
phated, line 18: This leaks the implementation of the type, but I believe it's the easier way to get this to work with the separate utils file.

compiler/src/language_server/protocol.re
```diff
@@ -1,4 +1,5 @@
 open Grain_utils;
+open Grain_diagnostics;
```
phated, line 2:
```suggestion
```
Can this be removed now?

phated (changes requested):
A couple things

compiler/src/language_server/protocol.rei
marcusroberts: Yes, it's the only way I could find, as the types didn't match otherwise when calling the Uri functions.

esy.lock/index.json
marcusroberts: Oops, removed.

compiler/src/language_server/protocol.rei
```diff
@@ -1,4 +1,5 @@
 open Grain_utils;
+open Grain_diagnostics;
```
phated, line 2:
```suggestion
```
And this?

compiler/src/language_server/protocol.re
marcusroberts: Yup!

compiler/src/language_server/protocol.rei
marcusroberts: And yup :)

phated (approved):
This is fantastic! I rebuilt the compiler on the branch and the LSP automatically had goto definition. Shift-clicking on stuff jumped me around so quickly! Awesome work 🎉 

The only nit I had was naming a file "utils" which isn't very descriptive, but we can change that as the file becomes more defined.

PR #1792, opened 2023-04-01
feat!: Grain produces JS output instead of WebAssembly

phated:
WebAssembly was an interesting compile target experiment for Grain while the language was in its infancy, but “Wasm” has proven itself to be a niche technology. Even the experts, such as [Fermyon](https://www.fermyon.com/blog/introducing-spin-v1) and [Suborbital](https://blog.suborbital.dev/bringing-javascript-and-typescript-to-suborbital), are recommending using better languages, such as JavaScript, for working with their application frameworks. To gain more mainstream adoption, I’ve modified the compiler to produce “ASM.js” output which will run on those platforms—and any other that provides a JavaScript runtime!

You can also see that the Grain CLI was modified to run the produced JavaScript directly, which is pretty slick because it already ran inside node.js :tada: I didn’t remove the “js-runner” package yet because I needed to check if we needed any functionality migrated over.

To verify everything is working, you can make a `main.gr` file such as:
```grain
module Main

print(“hello world”)
```
And then run `grain main.gr` which will print `hello world` in your terminal and you can inspect the `main.gr.mjs` file produced to verify it is runnable JavaScript.


spotandjake (approved):
Great pr Blaine this was a long time coming

marcusroberts (approved):
Given the wider availability of JS runtimes and the slow uptake of Wasm this looks like the better direction to move in to me.

aleksrutins (approved):
LGTM! This is an exciting step forward in Grain's development.

ospencer (changes requested):
Finally! I am so over "Web" "Assembly" and it's about time it got the axe. @phated You are a true visionary. This change produces asm.js though, which is still loosely related to wasm. Requesting that you output TypeScript by default instead, or if you must have a low-level format, maybe Java bytecode. It's my understanding that Java and JavaScript are the same which makes it a compatible change, and this way our users can write once and run anywhere.

willemneal:
So no more Wasm at all?

aleksrutins:
> So no more Wasm at all?

Nope! It's time to move forward into the new era of web development, and leave old and backwards technologies behind.

kevaundray (approved):

willemneal:
Ah yeah JS is the future!

spotandjake:
> Finally! I am so over "Web" "Assembly" and it's about time it got the axe. @phated You are a true visionary. This change produces asm.js though, which is still loosely related to wasm. Requesting that you output TypeScript by default instead, or if you must have a low-level format, maybe Java bytecode. It's my understanding that Java and JavaScript are the same which makes it a compatible change, and this way our users can write once and run anywhere.

To keep support for the web and reduce the number of output formats we need we could use java applets for web support, I am for the switch to java the decades java has spent proving itself on the market should lend perfectly towards making grain more stable and efficent

phated:
The suggestions by @ospencer were truly heinous so I believe we should abandon this plan for now.

Issue #1799, opened 2023-04-05
Runtime: Remove Trailing Space from Print Bytes


spotandjake:
![image](https://user-images.githubusercontent.com/40705786/230102729-ced2dc5d-0ce9-4b1c-92ab-cd762f22ae54.png)
Printing a bytes object has a trailing space which looks bad. we should remove this.

PR #1800, opened 2023-04-05
fix(stdlib)!: Remove extra space when converting Bytes to String

spotandjake:
Closes: #1799 
This pr removes the trailing space when printing bytes literals, it looked off with the trailing space before the space seemed to have come from when we caculate the string length we do `numBytes * 3` two chars for the hex and 1 char for the space and then we use memory.Fill to fill the string with spaces this by subtracting one from the `numBytes * 3` we get rid of the spot for the trailing space.

Breaking: This changes that behaviour of `toString` and `print` to a user.

Note: I did not see any test for the `...` behaviour or print should i add a test under the test for bytes_literals printing to test strings taht are too long?

### DIFF
--- a/compiler/test/suites/strings.re
+++ b/compiler/test/suites/strings.re
@@ -296,6 +296,11 @@ bar", 1))|},
 
   // Bytes literals
   assertRun("bytes_literal", {|print(b"abc")|}, "<bytes: 61 62 63>\n");
+  assertRun(
+    "bytes_literal",
+    {|print(b"111111111111111111111111111111111111111111111111111111111111111111111111")|},
+    "<bytes: 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31...>\n",
+  );
   assertCompileError(
     "bytes_literal_err1",
     {|print(b"abc\u1234")|},
--- a/compiler/test/suites/strings.re
+++ b/compiler/test/suites/strings.re
@@ -299,7 +299,7 @@ bar", 1))|},
   assertRun(
     "bytes_literal_long",
     {|print(b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefg")|},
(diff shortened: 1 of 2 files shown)


stdlib/runtime/string.gr
```diff
@@ -367,7 +367,7 @@ let rec heapValueToString = (ptr, extraIndents, toplevel) => {
         needsEllipsis = true
       }
       let headBytes = 8n // <bytes:
-      let hexBytes = numBytes * 3n
+      let hexBytes = numBytes * 3n - 1n
```
ospencer, line 370: Can you add a comment to this explaining that this is two digits + a space for each byte, minus one for the extra space at the end?

ospencer (commented):
Yeah, let's add a test for the `...`.

spotandjake:
Those changes have been made while making those changes i realized the byte count added for `...>` was 5 instead of 4 which lead to a trailing space when it was printed after the `>` which is not a behaviour we want either so i also set that to 4.

compiler/test/suites/strings.re
```diff
@@ -295,7 +295,12 @@ bar", 1))|},
   );
 
   // Bytes literals
-  assertRun("bytes_literal", {|print(b"abc")|}, "<bytes: 61 62 63 >\n");
+  assertRun("bytes_literal", {|print(b"abc")|}, "<bytes: 61 62 63>\n");
+  assertRun(
+    "bytes_literal",
+    {|print(b"111111111111111111111111111111111111111111111111111111111111111111111111")|},
+    "<bytes: 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31 31...>\n",
```
ospencer, line 302: Just so we have a slightly better test:
```suggestion
    "bytes_literal_long",
    {|print(b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefg")|},
    "<bytes: 41 42 43 44 45 46 47 48 49 4a 4b 4c 4d 4e 4f 50 51 52 53 54 55 56 57 58 59 5a 61 62 63 64 65 66 67...>\n",
```

ospencer (approved):
Thanks @spotandjake!

phated (approved):

PR #1804, opened 2023-04-06
feat(compiler)!: Remove arbitrary per-file compiler flags, add acceptable options as module attributes

### LINKED ISSUE #1625: grainc-flags as module attribute
Could look like this:
```
@flags("no-gc", "no-bulk-memory")
module NearTest
```

alex-snezhko:
```
@noPervasives
@runtimeMode
module Example
```
Closes #1625

### DIFF
--- a/compiler/src/compile.re
+++ b/compiler/src/compile.re
@@ -126,9 +126,14 @@ let next_state = (~is_root_file=false, {cstate_desc, cstate_filename} as cs) => 
       cleanup();
       Parsed(parsed);
     | Parsed(p) =>
+      let has_attr = name =>
+        List.exists(
+          attr => attr.Asttypes.attr_name.txt == name,
+          p.attributes,
+        );
       Grain_utils.Config.apply_attribute_flags(
-        ~no_pervasives=Option.is_some(p.attributes.no_pervasives),
-        ~runtime_mode=Option.is_some(p.attributes.runtime_mode),
+        ~no_pervasives=has_attr("noPervasives"),
+        ~runtime_mode=has_attr("runtimeMode"),
       );
       if (is_root_file) {
         Grain_utils.Config.set_root_config();
--- a/compiler/src/compile.re
+++ b/compiler/src/compile.re
@@ -88,25 +88,6 @@ let log_state = state =>
     prerr_string("\n\n");
   };
 
-let apply_inline_flags = (prog: Parsetree.parsed_program) => {
-  switch (prog.comments) {
-  | [Block({cmt_content, cmt_loc}), ..._] =>
-    Grain_utils.Config.apply_inline_flags(
-      ~on_error=
-        err => {
-          switch (err) {
-          | `Help =>
-            raise(InlineFlagsError(cmt_loc, Cannot_use_help_or_version))
-          | `Message(msg) =>
-            raise(InlineFlagsError(cmt_loc, Cannot_parse_inline_flags(msg)))
-          }
-        },
-      cmt_content,
-    )
-  | _ => ()
-  };
-};
-
 let next_state = (~is_root_file=false, {cstate_desc, cstate_filename} as cs) => {
   let cstate_desc =
     switch (cstate_desc) {
@@ -145,7 +126,10 @@ let next_state = (~is_root_file=false, {cstate_desc, cstate_filename} as cs) => 
       cleanup();
       Parsed(parsed);
     | Parsed(p) =>
-      apply_inline_flags(p);
+      Grain_utils.Config.apply_attribute_flags(
+        ~no_pervasives=Option.is_some(p.attributes.no_pervasives),
+        ~runtime_mode=Option.is_some(p.attributes.runtime_mode),
+      );
       if (is_root_file) {
         Grain_utils.Config.set_root_config();
       };
@@ -266,7 +250,7 @@ let compile_wasi_polyfill = () => {
   switch (Grain_utils.Config.wasi_polyfill^) {
   | Some(file) =>
     Grain_utils.Config.preserve_config(() => {
-      Grain_utils.Config.compilation_mode := Some("runtime");
+      Grain_utils.Config.compilation_mode := Grain_utils.Config.Runtime;
       let cstate = {
         cstate_desc: Initial(InputFile(file)),
         cstate_filename: Some(file),
--- a/compiler/src/formatting/fmt.re
+++ b/compiler/src/formatting/fmt.re
@@ -3928,12 +3928,8 @@ let print_comment_range =
 };
 
 let print_program = (fmt, parsed_program) => {
-  let attrs =
-    Option.to_list(parsed_program.attributes.no_pervasives)
-    @ Option.to_list(parsed_program.attributes.runtime_mode);
-
   let first_loc =
-    switch (attrs) {
+    switch (parsed_program.attributes) {
     | [fst, ..._] => fst.attr_loc
     | _ => parsed_program.prog_core_loc
     };
@@ -4002,38 +3998,37 @@ let print_program = (fmt, parsed_program) => {
     };
 
   group @@
-  fmt.print_comment_range(
-    fmt,
-    enclosing_start_location(parsed_program.prog_loc),
-    first_loc,
+  concat_map(
+    ~lead=
+      _ =>
+        fmt.print_comment_range(
+          fmt,
+          enclosing_start_location(parsed_program.prog_loc),
+          first_loc,
+        ),
+    ~sep=
+      (prev, next) =>
+        fmt.print_comment_range(
+          fmt,
+          ~none=hardline,
+          ~lead=space,
+          ~trail=hardline,
+          prev.Asttypes.attr_loc,
+          next.attr_loc,
+        ),
+    ~trail=
+      prev =>
+        fmt.print_comment_range(
+          fmt,
+          ~none=hardline,
+          ~lead=space,
+          ~trail=hardline,
+          prev.Asttypes.attr_loc,
+          parsed_program.prog_core_loc,
+        ),
+    ~f=(~final, a) => fmt.print_attribute(fmt, a),
+    parsed_program.attributes,
   )
-  ++ group(
-       concat_map(
-         ~lead=_ => empty,
-         ~sep=
-           (prev, next) =>
-             fmt.print_comment_range(
-               fmt,
-               ~none=hardline,
-               ~lead=space,
-               ~trail=hardline,
-               prev.Asttypes.attr_loc,
-               next.attr_loc,
-             ),
-         ~trail=
-           prev =>
-             fmt.print_comment_range(
-               fmt,
-               ~none=hardline,
-               ~lead=space,
-               ~trail=hardline,
-               prev.Asttypes.attr_loc,
-               parsed_program.prog_core_loc,
-             ),
-         ~f=(~final, a) => fmt.print_attribute(fmt, a),
-         attrs,
-       ),
-     )
   ++ string("module ")
   ++ string(parsed_program.module_name.txt)
   ++ toplevel;
--- a/compiler/src/formatting/fmt.re
+++ b/compiler/src/formatting/fmt.re
@@ -3928,12 +3928,6 @@ let print_comment_range =
 };
 
 let print_program = (fmt, parsed_program) => {
-  let first_loc =
-    switch (parsed_program.attributes) {
-    | [fst, ..._] => fst.attr_loc
-    | _ => parsed_program.prog_core_loc
-    };
-
   let toplevel =
     switch (parsed_program.statements) {
     | [] =>
@@ -4000,11 +3994,11 @@ let print_program = (fmt, parsed_program) => {
   group @@
   concat_map(
     ~lead=
-      _ =>
+      first =>
         fmt.print_comment_range(
           fmt,
           enclosing_start_location(parsed_program.prog_loc),
-          first_loc,
+          first.attr_loc,
         ),
     ~sep=
       (prev, next) =>
--- a/compiler/src/formatting/fmt.re
+++ b/compiler/src/formatting/fmt.re
@@ -3997,6 +3997,7 @@ let print_program = (fmt, parsed_program) => {
       first =>
         fmt.print_comment_range(
           fmt,
+          ~trail=hardline,
           enclosing_start_location(parsed_program.prog_loc),
           first.attr_loc,
         ),
--- a/compiler/src/formatting/fmt.re
+++ b/compiler/src/formatting/fmt.re
@@ -3991,39 +3991,53 @@ let print_program = (fmt, parsed_program) => {
       )
     };
 
-  group @@
-  concat_map(
-    ~lead=
-      first =>
+  let attributes =
+    switch (parsed_program.attributes) {
+    | [] =>
+      group(
         fmt.print_comment_range(
           fmt,
-          ~trail=hardline,
           enclosing_start_location(parsed_program.prog_loc),
-          first.attr_loc,
+          parsed_program.module_name.loc,
         ),
-    ~sep=
-      (prev, next) =>
-        fmt.print_comment_range(
-          fmt,
-          ~none=hardline,
-          ~lead=space,
-          ~trail=hardline,
-          prev.Asttypes.attr_loc,
-          next.attr_loc,
-        ),
-    ~trail=
-      prev =>
-        fmt.print_comment_range(
-          fmt,
-          ~none=hardline,
-          ~lead=space,
-          ~trail=hardline,
-          prev.Asttypes.attr_loc,
-          parsed_program.prog_core_loc,
-        ),
-    ~f=(~final, a) => fmt.print_attribute(fmt, a),
-    parsed_program.attributes,
-  )
+      )
+    | _ =>
+      group @@
+      concat_map(
+        ~lead=
+          first =>
+            fmt.print_comment_range(
+              fmt,
+              ~trail=hardline,
+              enclosing_start_location(parsed_program.prog_loc),
+              first.attr_loc,
+            ),
+        ~sep=
+          (prev, next) =>
+            fmt.print_comment_range(
+              fmt,
+              ~none=hardline,
+              ~lead=space,
+              ~trail=hardline,
+              prev.Asttypes.attr_loc,
+              next.attr_loc,
+            ),
+        ~trail=
+          prev =>
+            fmt.print_comment_range(
+              fmt,
+              ~none=hardline,
+              ~lead=space,
+              ~trail=hardline,
+              prev.Asttypes.attr_loc,
+              parsed_program.prog_core_loc,
+            ),
+        ~f=(~final, a) => fmt.print_attribute(fmt, a),
+        parsed_program.attributes,
+      )
+    };
+
+  attributes
   ++ string("module ")
   ++ string(parsed_program.module_name.txt)
   ++ toplevel;
--- a/compiler/src/formatting/fmt.re
+++ b/compiler/src/formatting/fmt.re
@@ -3928,6 +3928,16 @@ let print_comment_range =
 };
 
 let print_program = (fmt, parsed_program) => {
+  let attrs =
+    Option.to_list(parsed_program.attributes.no_pervasives)
+    @ Option.to_list(parsed_program.attributes.runtime_mode);
+
+  let first_loc =
+    switch (attrs) {
+    | [fst, ..._] => fst.attr_loc
+    | _ => parsed_program.prog_core_loc
+    };
+
   let toplevel =
     switch (parsed_program.statements) {
     | [] =>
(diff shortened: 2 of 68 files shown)


spotandjake (commented):
Im loving how much cleaner this looks, looks like you forgot to run the formatter thats why the tests are failing the attribute looks a lot cleaner then the comment though.

cli/__test__/index.gr
```diff
@@ -1,4 +1,4 @@
-/* grainc-flags --no-gc --no-bulk-memory --no-wasm-tail-call */
+@compileFlags("--no-gc", "--no-bulk-memory", "--no-wasm-tail-call")
```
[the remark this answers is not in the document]
alex-snezhko: I was considering having the `-`/`--` be included automatically but one problem I see with that is that if someone wanted to do something like `--memory-base 0x400` they may accidentally do `@compile("memory-base", "0x400")` which wouldn't work. Of course in that situation you _could_ allow `@compile("memory-base 0x400")` and then split all of the arguments by spaces to get `--memory-base` and `0x400` be separate arguments rather than one `'--memory-base 0x400'` argument, but that itself creates additional issues e.g. using directories with spaces in their names. What do you think?

compiler/src/compile.re
```diff
@@ -90,21 +90,28 @@ let log_state = state =>
   };
 
 let apply_inline_flags = (prog: Parsetree.parsed_program) => {
```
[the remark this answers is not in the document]
alex-snezhko: Currently only compiler flags are allowed per-file. Imo I also think this makes more sense than allowing submodules to be compiled differently.

compiler/test/suites/print.re
```diff
@@ -8,13 +8,15 @@ describe("print", ({test, testSkip}) => {
   let assertRun = makeRunner(test_or_skip);
 
   assertRun(
+    ~compile_flags=["--elide-type-info"],
```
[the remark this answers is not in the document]
alex-snezhko: Lol, not a fan of optional arguments?

cli/__test__/index.gr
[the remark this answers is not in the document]
phated: As I mentioned in a different comment, I think these should require `=` for compile options that need a value

compiler/src/compile.re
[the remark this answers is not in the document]
phated: Why? A module is a module is a module.

compiler/test/suites/print.re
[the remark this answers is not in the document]
phated: I'm not a fan of the auto-injecting of the module header in our tests. It was a shortcut for the already-large diff.

cli/__test__/index.gr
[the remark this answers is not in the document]
alex-snezhko: Ok so should a `-` be injected if whatever comes before the `=` character is only one character and `--` otherwise?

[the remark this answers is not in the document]
phated: Why are you "injecting"? Maybe you should consider trimming the flags coming from the CLI 🤔

compiler/src/compile.re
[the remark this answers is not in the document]
alex-snezhko: I suppose so; I'm having a bit of trouble understanding how some interactions would work with submodules compiled with different flags though; for example say you had:
```
module Main

let mut val = "a"

@compile("no-gc")
module Sub {
  provide let f = () => {
    val = "b"
  }
}

Sub.f()
```
Since `Sub` has gc disabled, would reassigning `val` cause `"a"` to be leaked? Also would `"b"` now be unmanaged? I suppose you have that same problem with file-level flags though and having a GC'd module include a non-GC'd module :thinking:

In any case, allowing submodules to be compiled different would require substantial changes to the existing codebase and I'm not sure I'm confident enough with the module system to be able to get this done before 0.6.0

cli/__test__/index.gr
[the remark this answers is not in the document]
alex-snezhko: True, just that prepending it into the string seems much easier to implement within the existing codebase

compiler/test/input/memoryBase/asserts.gr
```diff
@@ -1,4 +1,4 @@
-/* grainc-flags --memory-base 0x110000 */
+@compileFlags("--memory-base", "0x110000")
```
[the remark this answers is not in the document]
alex-snezhko: I agree, I generally prefer `=` for long option syntax. Are you suggesting to just replace the instances of `@compileFlags` where this is done or changing the CLI to completely prohibit long options without `=`? Because `--arg value` is technically legal as well according to `getopt_long`/the GNU long option convention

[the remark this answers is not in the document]
phated: I'm suggesting that command line flags are not the same as configuration passed via `@compile` attribute. CLI flags should be left alone and the attribute stuff should be adjusted to my suggestions and then they should be normalized.

compiler/test/suites/arrays.re
```diff
@@ -123,6 +123,7 @@ describe("arrays", ({test, testSkip}) => {
           state[0] =
             5",
         {
+          compile_flags: None,
           module_name: Location.mknoloc("Test"),
```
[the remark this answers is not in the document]
alex-snezhko: Coming back around to this, imo it's fine to keep them separate as this entire record contains different things relevant to a single module anyway, and a module's name and its compilation options seem distinct enough to separate

compiler/src/utils/config.rei
```diff
@@ -78,9 +78,9 @@ let maximum_memory_pages: ref(option(int));
 
 let import_memory: ref(bool);
 
-/** Compilation mode to use when compiling */
+/** Whether this module should be compiled in runtime mode */
 
-let compilation_mode: ref(option(string));
+let runtime_mode: ref(bool);
```
peblair, line 83: Almost wondering if this should be an enum rather than a boolean.

compiler/test/suites/parsing.re
```diff
@@ -39,6 +39,10 @@ describe("parsing", ({test, testSkip}) => {
       op,
       "module Test; a " ++ op ++ " b",
       {
+        compile_flags: {
```
peblair, line 42: Should this maybe be a shared `default_compile_flags` constant?

peblair (commented):
One comment and one question, but otherwise this looks good!

compiler/src/utils/config.re
```diff
@@ -449,15 +398,11 @@ let import_memory =
...
+  toggle_flag(
     ~names=["compilation-mode"],
-    ~conv=
-      option_conv(
-        Cmdliner.Arg.enum([("normal", "normal"), ("runtime", "runtime")]),
-      ),
     ~doc="Compilation mode (advanced use only)",
-    None,
+    false,
   );
```
phated, line 406: I don't understand what this is doing

alex-snezhko: You mean you don't understand why a flag is used or why this variable exists at all?

phated: I don't understand why this exist as a cmdliner flag at all considering this should only be used as an attribute now, right?

alex-snezhko: Yeah, I've been trying to get it to work without it being a CLI flag but it seems like the value is being reset at some point during compilation of multiple files that I haven't been able to pin down yet.

compiler/src/parsing/parsetree.re
```diff
@@ -658,8 +658,15 @@ type comment =
...
 
+[@deriving (sexp, yojson)]
+type compile_flags = {
+  no_pervasives: bool,
+  runtime_mode: bool,
+};
+
 [@deriving (sexp, yojson)]
 type parsed_program = {
+  compile_flags,
```
ospencer, line 669: I'm not a huge fan of this being called `compile_flags` since there are other attributes we may add in the future that may not be related to flags. Ideally these are just `attributes` like anything else.

stdlib/runtime/wasi.gr
```diff
@@ -217,120 +218,219 @@ provide foreign wasm path_rename: (
 ) => WasmI32 from "wasi_snapshot_preview1"
 
 // clocks
+@unsafe
```
spotandjake, line 221: Thoughts on allowing an entire module to be marked as unsafe?

spotandjake (commented):
Looks good to me :), one small question.

stdlib/runtime/wasi.gr
alex-snezhko: Hmm, I'm not a big fan of that personally. I think requiring that it be explicit helps add context that the thing you're doing is unsafe, and I think we generally do not want to encourage users to use unsafe at all which allowing this attribute at the module level would make easier

spotandjake: Fair enough

ospencer (approved):

compiler/graindoc/docblock.re
```diff
@@ -691,12 +691,14 @@ and for_signature_items =
...
       ~loc: Grain_parsing.Location.t,
+      ~has_attrs=false,
       signature_items,
     ) => {
   let comments = get_comments_from_loc(loc);
+  let addl_offset = if (has_attrs) {1} else {0};
   let comment =
     Comments.Doc.ending_on_including_attribute(
-      ~lnum=loc.loc_start.pos_lnum - 1,
+      ~lnum=loc.loc_start.pos_lnum - (1 + addl_offset),
```
phated, line 701: Shouldn't need this with the new formatter

phated (changes requested):
@alex-snezhko this needs to be updated for the new formatter. We'll probably need locations on the module attributes to handle comments between them too.

phated (changes requested):
Had a few questions, but overall looks great

compiler/test/grainfmt/comments.expected.gr
```diff
@@ -330,16 +343,15 @@ foo += /* foo */ /* foo */ 1
...
   let // foo
   _ = 1
   @attr(
     "foo", /* doo */
     "foo",
     // foo
     // goo
-    // bar
-  )
+  ) // bar
```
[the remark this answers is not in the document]
alex-snezhko: This appears to be a bug that existed before; this is the result of formatting
```
@attr("foo", /* doo */ "foo"
// foo
// goo
) // bar
```

compiler/test/grainfmt/comments.expected.gr
```diff
@@ -599,15 +611,14 @@ provide {
...
+@attr // foo
 let _ = 1
 @attr(
   "foo", /* doo */
   "foo",
   // foo
   // goo
-  // bar
-)
+) // bar
```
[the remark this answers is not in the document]
alex-snezhko: https://github.com/grain-lang/grain/pull/1804#discussion_r1501929251

compiler/test/input/memoryBase/sameAsDefault.gr
```diff
@@ -1,4 +1,3 @@
-/* grainc-flags --memory-base 0x400 */
```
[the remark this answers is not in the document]
alex-snezhko: I believe `noPervasives` and `runtimeMode` were the only ones we wanted to keep right @ospencer ?

compiler/test/grainfmt/comments.expected.gr
```diff
@@ -1,3 +1,9 @@
+// Comment before first attr
+/* Block comment before attr */@noPervasives // Line comment on attr
+// Comment after first attr
+@runtimeMode
+// Comment after second attr
+
```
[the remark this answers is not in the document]
alex-snezhko: This is the behavior for attributes on other items as well, maybe can address in a followup?

compiler/src/typed/env.rei
```diff
@@ -162,13 +162,8 @@ let add_local_type: (Path.t, type_declaration, t) => t;
...
-/* Remember the current compilation unit: modname * filename * compilation mode. */
-type compilation_mode =
-  | Normal
-  | Runtime;
-
-let set_unit: ((string, string, compilation_mode)) => unit;
-let get_unit: unit => (string, string, compilation_mode);
+let set_unit: ((string, string, Grain_utils.Config.compilation_mode)) => unit;
+let get_unit: unit => (string, string, Grain_utils.Config.compilation_mode);
 let is_runtime_mode: unit => bool;
```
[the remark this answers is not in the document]
alex-snezhko: Do you mean in `type_utils.re`? I think it'd be easier to keep in `env.re` since it uses a state variable defined in `env.re`

compiler/src/utils/config.re
```diff
@@ -576,24 +521,21 @@ let module_search_path = () => {
...
+  if (np) {
+    no_pervasives := true;
+  };
+  if (rm) {
+    compilation_mode := Runtime;
   };
+};
 
-let with_inline_flags = (~on_error, cmt_content, thunk) => {
+let with_attribute_flags = (~on_error, ~no_pervasives, ~runtime_mode, thunk) => {
```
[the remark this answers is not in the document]
alex-snezhko: Ah, no longer needed so I'll remove

compiler/src/typed/typed_well_formedness.re
```diff
@@ -337,11 +337,10 @@ module WellFormednessArg: TypedtreeIter.IteratorArgument = {
...
       // For now, we only raise the error inside of functions.
       if (exp_is_wasm_unsafe(exp)
-          && !(
-               Grain_utils.Config.no_gc^
-               || Grain_utils.Config.compilation_mode^ == Some("runtime")
-               || is_unsafe()
-             )) {
+          && !
+               Grain_utils.Config.(
+                 no_gc^ || compilation_mode^ == Runtime || is_unsafe()
```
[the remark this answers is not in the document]
alex-snezhko: This is the only time this logic is used so not sure if it's necessary

compiler/src/parsing/parser_header.re
```diff
@@ -116,13 +116,47 @@ let make_include_alias = ident => {
...
+          )
+        }
+      },
+      (None, None),
+      attributes,
+    );
 
-  fix_blocks({module_name, statements, comments: [], prog_loc});
+  fix_blocks({
+    attributes: {
```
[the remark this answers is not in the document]
alex-snezhko: Addressed

compiler/src/parsing/parser.messages
```diff
@@ -24,6 +24,42 @@ program: EOL YIELD
...
+## program -> option(attribute) . module_header eos toplevel_stmts EOF [ # ]
+## program -> option(attribute) . module_header option(eos) EOF [ # ]
+##
+## The known suffix of the stack is as follows:
+## option(attribute)
+##
+## WARNING: This example involves spurious reductions.
+## This implies that, although the LR(1) items shown above provide an
+## accurate view of the past (what has been recognized so far), they
+## may provide an INCOMPLETE view of the future (what was expected next).
```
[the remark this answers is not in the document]
ospencer: No, you'll see a ton of these throughout the messages file. The thing it's telling you to remember is that it's possible that there are many valid continuations from this point, not just an individual token, so your error message should be mindful of that.

compiler/test/grainfmt/comments.expected.gr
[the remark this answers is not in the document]
ospencer: https://github.com/grain-lang/grain/pull/2048

compiler/test/input/memoryBase/sameAsDefault.gr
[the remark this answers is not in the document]
ospencer: That's correct. Memory base doesn't make sense as a module attribute since it applies to your entire program, not just an individual file.

compiler/src/formatting/fmt.re
```diff
@@ -3985,10 +3991,36 @@ let print_program = (fmt, parsed_program) => {
...
-    fmt,
-    enclosing_start_location(parsed_program.prog_loc),
-    parsed_program.module_name.loc,
+  concat_map(
+    ~lead=
+      _ =>
+        fmt.print_comment_range(
+          fmt,
+          enclosing_start_location(parsed_program.prog_loc),
+          first_loc,
```
phated, line 4000: Shouldn't this just use the argument to the function instead of trying to calculate the first_loc outside?

compiler/src/parsing/parsetree_iter.re
```diff
@@ -3,8 +3,8 @@ open Parsetree;
 type hooks = {
   enter_location: Location.t => unit,
   leave_location: Location.t => unit,
-  enter_attribute: attribute => unit,
-  leave_attribute: attribute => unit,
+  enter_attribute: (attribute, attribute_context) => unit,
+  leave_attribute: (attribute, attribute_context) => unit,
```
phated, line 7: I don't like this attribute_context type. Why don't we actually change the type of attributes?

compiler/test/grainfmt/comments.expected.gr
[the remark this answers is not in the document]
phated: Please open an issue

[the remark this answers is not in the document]
phated: Wait, you've added the formatter changes. You can just change the line ending. Right? We still need a follow up to fix the other ones though.

compiler/test/suites/basic_functionality.re
```diff
@@ -244,7 +244,11 @@ describe("basic functionality", ({test, testSkip}) => {
     "Failure: boo",
   );
   assertSnapshotFile("toplevel_statements", "toplevelStatements");
-  assertSnapshotFile("unsafe_wasm_globals", "unsafeWasmGlobals");
+  assertSnapshotFile(
+    ~config_fn=() => {Grain_utils.Config.no_gc := true},
```
phated, line 248: why do we set the config_fn for things like this instead of changing the source file to add the disableGC attribute?

compiler/test/suites/functions.re
```diff
@@ -150,9 +150,15 @@ describe("functions", ({test, testSkip}) => {
...
   assertSnapshot("lambda_pat_any", "let x = (_) => 5; x(\"foo\")");
   assertCompileError(
-    "unknown_attribute",
+    "unknown_toplevel_attribute",
     "@unknown let x = () => 5",
-    "Unknown attribute",
+    "Unknown top-level attribute",
+  );
+  assertCompileError(
+    ~default_module_header=false,
```
phated, line 158: Instead of `~default_module_header` being a boolean, should we instead make it `~module_header` that is defaulted to the default module header but allows us to override?

phated (changes requested):
Few more things

compiler/src/parsing/parsetree_iter.re
alex-snezhko: Do you mean basically adding the `attribute_context` as a field on the `attribute` type itself? If so then should it be on the `attributes` type since everything in an attribute group should have the same "context"?

phated: Do we need this at all? Can the attribute wellformedness check be done inside of `enter_toplevel`, `enter_expression`, etc?

compiler/src/formatting/fmt.re
alex-snezhko: Sorry, not sure what you mean; are you referring to an argument to `print_comment_range`? Not sure which one I'd want here

phated: No, you calculate `first_loc` outside of this function, but you don't use the argument to `~lead` like you are supposed to.

compiler/test/suites/basic_functionality.re
alex-snezhko: I believe disabling GC is something we want to be doing on the level of the whole program rather than on a per-file basis right @ospencer ?

compiler/test/grainfmt/comments.expected.gr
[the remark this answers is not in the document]
alex-snezhko: Looked into this and I think it would require making some invasive changes to the `print_comment_range` function to disallow double-breaks. I think we can just defer to https://github.com/grain-lang/grain/issues/2055

compiler/src/parsing/parsetree_iter.re
```diff
@@ -3,8 +3,6 @@ open Parsetree;
 type hooks = {
   enter_location: Location.t => unit,
   leave_location: Location.t => unit,
-  enter_attribute: (attribute, attribute_context) => unit,
-  leave_attribute: (attribute, attribute_context) => unit,
```
phated: I don't like that you removed these. We might want to iter attributes at some point

compiler/src/formatting/fmt.re
```diff
@@ -3992,10 +3992,36 @@ let print_program = (fmt, parsed_program) => {
...
+          prev.Asttypes.attr_loc,
+          next.attr_loc,
+        ),
+    ~trail=
+      prev =>
+        fmt.print_comment_range(
+          fmt,
+          ~none=hardline,
+          ~lead=space,
+          ~trail=hardline,
```
phated, line 4019: I'm pretty sure that the `~none` and/or `~trail` being hardline here are causing your extra newline in the attributes.

compiler/src/parsing/parsetree_iter.re
```diff
@@ -3,8 +3,6 @@ open Parsetree;
 type hooks = {
   enter_location: Location.t => unit,
   leave_location: Location.t => unit,
-  enter_attribute: attribute => unit,
-  leave_attribute: attribute => unit,
```
phated, line 7: Can you put these back please? I don't see a reason to remove them

compiler/src/parsing/parsetree_iter.re
```diff
@@ -87,6 +83,7 @@ let iter_attributes = (hooks, attrs) => {
 
 let rec iter_parsed_program = (hooks, {statements} as program) => {
   hooks.enter_parsed_program(program);
+  iter_attributes(hooks, program.attributes);
```
phated, line 86: What's going on here?

compiler/test/grainfmt/comments.expected.gr
```diff
@@ -1,3 +1,9 @@
+// Comment before first attr
+/* Block comment before attr */@noPervasives // Line comment on attr
```
phated, line 2: This is wrong. It needs to have a space between `*/` and `@`

compiler/src/formatting/fmt.re
alex-snezhko: Doesn't look like that fixed it; I think it's the fact that `print_comment_range` inserts `phatom_hardline ++ phantom_hardline` since `line_delta` > 1 between the comment and module header

compiler/test/grainfmt/comments.expected.gr
alex-snezhko: Looks like in other places it actually puts the block comment on its own line, which I think I prefer here

compiler/src/parsing/parsetree_iter.re
alex-snezhko: What do you mean? It's iterating over the attributes in the same way done with expressions and toplevels

phated (approved):

PR #1807, opened 2023-04-07
feat(compiler)!: Reduce size of runtime heap

ospencer:
Since #1410 was landed, far less memory is used in programs run in runtime mode. I did an audit, and it seems that we now only use ~650 bytes, so I reduced the reserved runtime size from 16384 bytes down to 2048 bytes. This is helpful when running in some memory-constrained runtimes, like wasm4.

### DIFF
--- a/stdlib/runtime/malloc.gr
+++ b/stdlib/runtime/malloc.gr
@@ -76,7 +76,7 @@ let logMallocHeaderSize = 3n
  */
 let mut heapSize = 0n
 
-provide let _RESERVED_RUNTIME_SPACE = 0x4000n
+provide let _RESERVED_RUNTIME_SPACE = 0x800n
 
 /**
  * The base the heap. The block at this address will be size 0 and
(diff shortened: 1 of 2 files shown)


ospencer:
I considered going down to 1024 bytes, but I wanted to leave some breathing room.

spotandjake (approved):
Looks good less memory usage should defs help with targets like wasm4

stdlib/runtime/malloc.gr
```diff
@@ -76,7 +76,7 @@ let logMallocHeaderSize = 3n
  */
 let mut heapSize = 0n
 
-provide let _RESERVED_RUNTIME_SPACE = 0x4000n
+provide let _RESERVED_RUNTIME_SPACE = 0x800n
```
phated, line 79: I believe the docs need to be updated? They say we reserve a quarter page.

phated (changes requested):
Docs

peblair (approved):
Looks good, modulo @phated's comment about docs. Tossing an approval on here so that it's ready to go once that's addressed.

stdlib/runtime/malloc.gr
phated: I just did this

phated (approved):

PR #1817, opened 2023-04-18
fix(stdlib): Return early from `Process.argv()` if length is zero

phated:
This returns early from the `Process.argv()` function if the `argsLength` is zero. This avoids some differences between wasi implementations, and is also probably just a good idea to avoid extra sys calls.

Fixes an issue I was encountering in #1585

### DIFF
--- a/stdlib/sys/process.gr
+++ b/stdlib/sys/process.gr
@@ -94,6 +94,14 @@ provide let argv = () => {
   }
 
   let argc = WasmI32.load(argcPtr, 0n)
+
+  let argsLength = argc * 4n
+  let arr = allocateArray(argc)
+
+  if (WasmI32.eqz(argsLength)) {
+    return Ok(WasmI32.toGrain(arr): Array<String>)
+  }
+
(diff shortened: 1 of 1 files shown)


marcusroberts (approved):
Nice fix.

stdlib/sys/process.gr
```diff
@@ -94,6 +94,14 @@ provide let argv = () => {
   }
 
   let argc = WasmI32.load(argcPtr, 0n)
+
+  let argsLength = argc * 4n
+  let arr = allocateArray(argc)
+
+  if (WasmI32.eqz(argsLength)) {
+    return Ok(WasmI32.toGrain(arr): Array<String>)
```
ospencer, line 102:
```suggestion
    Memory.free(argcPtr)
    return Ok(WasmI32.toGrain(arr): Array<String>)
```
could also refactor a bit to do the free after we read `argc` so we don't need to do it in multiple places.

phated: I tried to do the refactor. Let me know if it is correct!

ospencer (approved):
Looks great!

PR #1821, opened 2023-04-22
feat(stdlib)!: Replace `Result`/`Option` `and`/`or` functions with operators

### LINKED ISSUE #1819: Convert `Option`/`Result` `and`/`or` to operators
Ref https://github.com/grain-lang/grain/pull/1791#discussion_r1167639031

With the addition of `and` as a keyword in #1791, the `and` functions in `Option` and `Result` were changed to `and_` as placeholders, which feels awkward. Since `and`/`or` from these modules behave akin to the short-circuiting `&&` and `||` pervasives they should be changed in these modules as well.

alex-snezhko:
Closes #1819

### DIFF
--- a/compiler/test/stdlib/option.test.gr
+++ b/compiler/test/stdlib/option.test.gr
@@ -163,17 +163,19 @@ let mut d = "🌾"
 assert Option.peek(x => fail "Shouldn't be called", None) == None
 assert d == "🌾"
 
-// Option.or
-assert Option.or(Some(1), Some(2)) == Some(1)
-assert Option.or(Some(1), None) == Some(1)
-assert Option.or(None, Some(2)) == Some(2)
-assert Option.or(None, None) == None
+from Option use { (||), (&&) }
 
-// Option.and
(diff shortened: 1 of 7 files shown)


compiler/test/stdlib/option.test.gr
```diff
@@ -163,17 +163,19 @@ let mut d = "🌾"
...
-
-// Option.and
-assert Option.and(Some(1), Some(2)) == Some(2)
-assert Option.and(Some(1), None) == None
-assert Option.and(None, Some(2)) == None
-assert Option.and(None, None) == None
+from Option use { (||), (&&) }
+
+// Option.(||)
+assert (Some(1) || Some(2)) == Some(1)
```
spotandjake, line 169: Is there a reason you left the parens here, I think this will fail the formatter test.
Same for all the other tests you replaced

phated (approved):

spotandjake (approved):
This looks good to me other than the extra parens.

Issue #1420, opened 2022-08-17
Lang: Types should not be recursive by default


phated:
Currently, when defining a `type` in Grain, it is recursive by default. I believe this is a holdover from ocaml and ocaml allows you to opt-out with the `type nonrec` syntax.

Grain should make `type` definitions non-recursive by default and allow them to be annotated with a `rec` keyword, like bindings.

renatoalencar:
I've made a proof of concept for that, but I think I'd be nice to discuss syntax before escalate this into a PR. Following OCaml's and the let syntax, I added the `rec` keyword just after the type declaration for `enum`, `record` and `type`.  So  a list should be something like this:

```
enum rec List<a> {
   Nil,
   Cons(List<a>)
}
```

Things start to get confusing when there's more than one type on the same top level statement, example:

```
record rec A {
  b: B
},
record B {
  a: A
}
```

For some reason that looks a little bit confusing to me.

https://github.com/renatoalencar/grain/commit/a2591efe2cd9e649f69a21aa7bb9a39b74b3e649

By the way, my parser changes are not very great as I had to add a few more duplicated (kind of) stuff in order for this to work. I'm accepting suggestions on how to simplify that.

phated:
@renatoalencar This is super cool! I'll poke the rest of the team to get some eyes on this for the syntax and your other question.

phated:
@renatoalencar We're still having discussions, but I wanted to give you an update.

The initial purpose of the issue was for self-recursive types; for example: We wanted to move `enum Range` into pervasives and then re-export it as `export type Range = Range` from the Range module. This isn't possible because `type Range` is self-recursive so you can't reference the `enum Range` in pervasives. 

The solution we are leaning towards is to make types non-self-recursive by default but still allow types to be mutually recursive by default (without the `rec` keyword), e.g. `type Foo = Bar, type Bar = Foo`. We may also want to reintroduce the `and` keyword for mutually recursive type definitions.

PR #1825, opened 2023-04-23
feat(compiler)!: Make types nonrecursive by default

alex-snezhko:
Requires that recursive type definitions use rec keyword
```
record rec T {
  x: T
}
```
Mutually recursive types are also self-recursive

```
record A {
  x: Number
},
enum B {
  B(A)
}
```
Closes https://github.com/grain-lang/grain/issues/1420


alex-snezhko:
Wrong branch name

PR #1839, opened 2023-05-11
feat(Compiler)!: Change Grain Artifact Name

spotandjake:
This change modifies the grain compiler to produce the artifact files with the extension `.grartifact` instead of `.gr.wasm` the output of the compiler can still be found at `.gr.wasm` though so it is just for the intermediate build files.

We should make this change before 0.6 given its breaking nature and the fact that without the ability to run grain programs in unlinked mode there is really no reason to give the false impression that grain libs are valid wasm modules.

This will need to be rewritten after #1842

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ _build/
 .installed-pkgs
 notes.org
 node_modules/
+*.grartifact
 *.wasm
 *.wasm.map
 *.wat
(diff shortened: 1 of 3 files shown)


phated:
I think we decided against this. @ospencer feel free to reopen if I'm wrong.

PR #1840, opened 2023-05-11
feat(cli)!: Allow specifying WASI environment variables and CLI args

### LINKED ISSUE #1822: Grain CLI should have a way to pass args to the WASI runner
We need to figure out a syntax, such as `-- --foo --bar --baz`, in which the extra args are passed to the WASI runner when using a command like `grain` or `grain run` from the grain CLI.

### LINKED ISSUE #1823: Grain CLI should have a way to pass env vars to the WASI runner
Suggested syntax is `--env=FOO=bar`

alex-snezhko:
Closes #1822 
Closes #1823 

Examples:
```
grain a.gr --env=FOO=bar
grain a.gr --env FOO=bar FOO2=baz -- arg1 arg2
grain a.gr arg1 arg2   # will not pass args to WASI, -- required
```

### DIFF
--- a/cli/bin/exec.js
+++ b/cli/bin/exec.js
@@ -129,13 +129,25 @@ function execGrainrun(file, options, program, execOpts = { stdio: "inherit" }) {
     preopens[guestDir] = hostDir;
   });
 
+  const rawArgs = process.argv;
+  const endOptsIndex = rawArgs.findIndex((x) => x === "--");
+  const args = rawArgs.slice(endOptsIndex === -1 ? Infinity : endOptsIndex + 1);
+
+  const cliEnv = {};
+  options.env?.forEach((env) => {
+    const [name, ...rest] = env.split("=");
+    const val = rest.join("=");
+    cliEnv[name] = val;
+  });
+
   const env = {
+    ...cliEnv,
     PREOPENS: JSON.stringify(preopens),
     NODE_OPTIONS: `--experimental-wasi-unstable-preview1 --no-warnings`,
   };
 
   try {
-    execSync(`${grainrun} ${file}`, { ...execOpts, env });
+    execSync(`${grainrun} ${file} ${args.join(" ")}`, { ...execOpts, env });
   } catch (e) {
     process.exit(e.status);
   }
--- a/cli/bin/exec.js
+++ b/cli/bin/exec.js
@@ -122,16 +122,18 @@ function getGrainrun() {
 
 const grainrun = getGrainrun();
 
-function execGrainrun(file, options, program, execOpts = { stdio: "inherit" }) {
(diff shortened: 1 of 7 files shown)


cli/bin/exec.js
```diff
@@ -129,13 +129,25 @@ function execGrainrun(file, options, program, execOpts = { stdio: "inherit" }) {
     preopens[guestDir] = hostDir;
   });
 
+  const rawArgs = process.argv;
+  const endOptsIndex = rawArgs.findIndex((x) => x === "--");
+  const args = rawArgs.slice(endOptsIndex === -1 ? Infinity : endOptsIndex + 1);
```
phated, line 134: This PR fails on Windows because you need to pre-process the `process.argv` instead of post-process it after commander messes around. I believe @spotandjake saw a way to do that.

spotandjake: Yes as shown by the types [here](https://github.com/tj/commander.js/blob/33195f189b1ffa568c232503fb6ac0cf5548eb74/typings/index.d.ts#LL674C6-L674C6), it looks as though we could get process the arguments and then just pass the pre proccessed arguments in.

stdlib/sys/process.gr
```diff
@@ -152,6 +152,11 @@ provide let env = () => {
   let envc = WasmI32.load(envcPtr, 0n)
   let envvBufSize = WasmI32.load(envvBufSizePtr, 0n)
 
+  if (envc == 0n) {
+    Memory.free(envcPtr)
+    return Ok(WasmI32.toGrain(allocateArray(0n)): Array<String>)
```
spotandjake, line 157: Just to double check this is just a fast path for if there are no args right?

and also is there a reason you chose to allocate an array using `allocateArray` over `[>]`?

```suggestion
    return Ok([>]: Array<String>)
```

spotandjake (commented):
Looks pretty much good, had two small questions about your changes to `process.gr`.

stdlib/sys/process.gr
alex-snezhko: > this is just a fast path for if there are no args right?

Yes, it also makes sure that the function doesn't break since a `Wasi.environ_get` is run in subsequent code, which wouldn't work with 0 arguments.

> and also is there a reason you chose to allocate an array using allocateArray over [>]?

Not particularly, I can change it

spotandjake (approved):

peblair (approved):

marcusroberts (approved):
I like how this works!

phated (approved):
I'm happy with the outcome here. I was thinking the escaped double-dash could be done in the tests themselves, but I think the transformation inside runner is fine.

PR #1842, opened 2023-05-19
feat(compiler)!: Require extension when including relative file paths

### LINKED ISSUE #1797: Compiler: Require extensions for local include paths
Currently in the module system when you write an `include` we append `.gr` onto it this behaviour works well but realistically file extensions have no meaning. We should have a discusion on weather it makes sense to require the `.gr` extension, and also if it makes sense to append it automatically a counter example to why we shouldnt auto append it is to allow for future expansion such as if we wanted to allow `include "./bindings.wit"`

alex-snezhko:
Requires local `include`d files to be specified with their file extension.
```
include "list"
include "list.gr" // error
include "./b.gr"
include "./b.whatever"
include "./b" // error if file is b.gr
```

Closes #1797


compiler/src/compile.re
```diff
@@ -41,8 +41,7 @@ type error =
 
 exception InlineFlagsError(Location.t, error);
 
-let default_output_filename = name =>
-  Filepath.String.remove_extension(name) ++ ".gr.wasm";
+let default_output_filename = name => name ++ ".wasm";
```
ospencer, line 44: We'll need to call this out in a changelog entry, since we'll consider this breaking.

compiler/src/typed/module_resolution.re
```diff
@@ -202,30 +202,65 @@ let locate_module = (~disable_relpath=false, base_dir, path, unit_name) => {
...
+    try(locate(name)) {
+    | Not_found =>
+      if (check_suffix(name, ".gr")) {
+        let no_extension = chop_suffix(name, ".gr");
+        switch (locate(no_extension)) {
+        | exception Not_found => error(No_module_file(loc, name, None))
+        | _ =>
+          let name = !is_relpath(name) ? no_extension : name;
+          error(
+            No_module_file(
```
peblair, line 240: Maybe we should have a test which throws this error?

compiler/src/compile.re
alex-snezhko: This actually isn't breaking, it's just that the input to this function now contains the `.gr` extension; the output filepaths will be the same for `.gr` files

ospencer: I gotcha.

spotandjake (approved):
Looks good to me

compiler/src/middle_end/analyze_inline_wasm.re
```diff
@@ -59,7 +59,7 @@ let analyze = ({imports, body, analyses}) => {
   mod_has_inlineable_wasm := false;
   let process_import = ({imp_use_id, imp_desc}) => {
     switch (imp_desc) {
-    | GrainValue("runtime/unsafe/memory", name) =>
+    | GrainValue("runtime/unsafe/memory.gr", name) =>
```
ospencer, line 62: Wouldn't this break the optimization for any unsafe code that doesn't exist within the stdlib / wouldn't we need to check for both to handle library code too?

alex-snezhko: I set it up so that the `.gr` extension would be injected automatically on the filepath so it would be applied, but you bringing this up is actually making me see that this logic is a little busted if you're using relative imports. For example if you're in the `runtime/unsafe` directory then doing `include "./memory.gr"` wouldn't trigger this optimization

alex-snezhko: https://github.com/grain-lang/grain/issues/1990

ospencer:
@alex-snezhko Can you resolve the conflicts here?

compiler/src/typed/module_resolution.re
phated: @alex-snezhko I saw the test for the extraneous extension, but not a test for the error message asking if they forgot an extension. Does that still need a test?

alex-snezhko: Ah right; added a test for this

compiler/src/typed/module_resolution.re
```diff
@@ -202,30 +202,65 @@ let locate_module = (~disable_relpath=false, base_dir, path, unit_name) => {
...
+              loc,
+              name,
+              Some("did you forget to remove the .gr extension?"),
```
ospencer, line 243: I don't love this error message because it implies that there was something they were supposed to do but forgot, when really there wasn't a local file matching the name, but there is one on the module search path, and maybe that's what they wanted to reference.

Maybe something like `` "path/to/module.gr" was not found locally. Did you mean "path/to/module"? ``

compiler/src/typed/module_resolution.re
```diff
@@ -202,30 +202,65 @@ let locate_module = (~disable_relpath=false, base_dir, path, unit_name) => {
...
+        };
+      } else {
+        switch (locate(name ++ ".gr")) {
+        | exception Not_found => error(No_module_file(loc, name, None))
+        | _ =>
+          error(
+            No_module_file(
+              loc,
+              name,
+              Some("did you forget to add the .gr extension?"),
```
ospencer, line 255: Similarly, we could say something here like `did you mean "path/to/module.gr"`.

ospencer (approved):

phated:
@alex-snezhko looks like the reason formatter needs to be run

phated (approved):

PR #1844, opened 2023-05-23
feat(stdlib)!: Handle printing of reference cycles

### LINKED ISSUE #1841: Handle printing of cyclic structures
Currently attempting to print data that contains cyclic references results in a stack overflow. A minimal example:
```
record Node {
  val: Number,
  mut next: Option<Node>
}

let a = { val: 1, next: None }
let aOpt = Some(a)
a.next = Some({ val: 2, next: aOpt })
// cycle between 1 and 2
print(a) // RangeError: Maximum call stack size exceeded
```
Perhaps an error value of `"<data with cyclic reference>"` (or some other more clever way of displaying this data) can be printed instead if a cycle is detected.

alex-snezhko:
Closes #1841 

I used a style of reporting cycles similar to Node.js where references are labeled with numbers whenever there is a cycle. For example:
```
record A {
  val: Number,
  mut next: Option<A>
}

let a = { val: 1, next: None }
let aOpt = Some(a)
a.next = aOpt
print(a)

// output:
<1>: {
  val: 1,
  next: Some(<cycle to <1>>)
}
```

### DIFF
--- a/compiler/test/suites/cycles.re
+++ b/compiler/test/suites/cycles.re
@@ -1,0 +1,136 @@
+open Grain_tests.TestFramework;
+open Grain_tests.Runner;
+
+describe("cyclic references", ({test, testSkip}) => {
+  let test_or_skip =
+    Sys.backend_type == Other("js_of_ocaml") ? testSkip : test;
+
+  let assertRun = makeRunner(test_or_skip);
+
+  assertRun(
+    "cycles1",
+    {|
+      record A {
+        mut a: Option<A>
+      }
+      let x = { a: None }
+      x.a = Some(x)
+      print(x)
+    |},
+    "<1>: {\n  a: Some(<cycle to <1>>)\n}\n",
+  );
+  assertRun(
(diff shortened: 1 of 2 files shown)


spotandjake (approved):
This looks awesome.

ospencer (commented):
Were you able to look into only allocating the vec if needed? If not, it'd probably be worth it to just allocate a vec and reuse it instead of making a new one each time.

peblair (approved):
LGTM, but see @ospencer's comment

alex-snezhko:
@ospencer I did end up doing something like that; initially the allocated vector box is empty and then the first time that `vecPush` is called it initializes the vector (though you still have to allocate the vector box each time). Reusing one vector through every call to `print` sounds like an interesting idea, though (and maybe I'm thinking too far ahead here) would this continue to work properly with `async`/parallelism gets implemented and you are trying to print in multiple threads?

phated (approved):

PR #1845, opened 2023-05-26
feat(stdlib): Add mutable doubly-linked list

### LINKED ISSUE #1815: Add doubly linked list.
Currently grian has linked lists which are awesome when you need them but have limitations such as arbitrary spread is not efficent and it isnt well suited for every situation such as if you need to remove an element from the middle of the list which could be a common task. 

A solution to this would be to add doubly linked lists.

The first question is do we think it makes sense to add them at all, adding another type of list data structure when there is already confusion for some people about when to use arrays or lists might just make things wort though i dont think it will really be  that big a deal.

The second is how niche / much use do we think it will get do we want this in the compiler or in the stdlib.

and thirdly if we choose the compiler route which has advantages such as syntax, easy spread and pattern matching what would syntax look like maybe something like `[|1,2,3|]` but that would need to be deteremined with arrays and lists `[]` is kind of already full more symbols might not help.

alex-snezhko:
Closes #1815 

Wherever I thought it made sense, I made the list functions mutate the given list rather than producing a new list e.g. `append`


stdlib/list.gr
```diff
@@ -901,3 +901,1012 @@ provide let sort = (comp, list) => {
...
+   */
+
+  provide let copy = list => {
+    map(identity, list)
+  }
+
+  /**
+   * Clears the list.
+   *
+   * @param queue: The list to clear
```
spotandjake, line 1902:
```suggestion
   * @param list: The list to clear
```

stdlib/list.gr
```diff
@@ -901,3 +901,1012 @@ provide let sort = (comp, list) => {
...
+   *
+   * @since v0.6.0
+   */
+
+  provide let copy = list => {
+    map(identity, list)
+  }
+
+  /**
+   * Clears the list.
```
spotandjake, line 1900:
```suggestion
   * Clears the provided list.
```

stdlib/list.gr
```diff
@@ -901,3 +901,1012 @@ provide let sort = (comp, list) => {
...
+      }
+    }
+    match (reduce(iter, None, list)) {
+      None => "",
+      Some(s) => s,
+    }
+  }
+
+  /**
+   * Produces a shallow copy of the input list. The new list contains the
```
spotandjake, line 1886: I dont know if we need to specify it contains the same elements, I feel the terminology shallow copy covers that.

stdlib/list.gr
```diff
@@ -901,3 +901,1012 @@ provide let sort = (comp, list) => {
...
+   * Sorts the list.
+   *
+   * Ordering is calculated using a comparator function which takes two array elements and must return 0 if both are equal, a positive number if the first is greater, and a negative number if the first is smaller.
+   * @param comp: The comparator function used to indicate sort order
+   * @param list: The list to be sorted
+   * 
+   * @since v0.6.0
+   */
+
+  provide let sort = (comp, list) => {
```
spotandjake, line 1853: Do we want this to use the default compare from pervasives maybe with an optional argument? might be a change for a seperate pr.

stdlib/list.gr
```diff
@@ -901,3 +901,1012 @@ provide let sort = (comp, list) => {
...
+    let { head, tail, length } = fromList(sort(comp, toList(list)))
```
spotandjake, line 1854: Im lost at this implementation does this use the sort function from the normal list type?

stdlib/list.gr
```diff
@@ -901,3 +901,1012 @@ provide let sort = (comp, list) => {
...
+   * Rotates list elements by the specified amount to the left, such that `n`th
+   * element becomes the first in the list.
+   *
+   * If value is negative, list elements will be rotated by the
+   * specified amount to the right. See examples.
+   *
+   * @param n: The number of elements to rotate by
+   * @param list: The list to be rotated
+   *
+   * @example let list = fromList([1, 2, 3, 4, 5]); List.Mutable.rotate(2, list); assert list == fromList([3, 4, 5, 1, 2])
```
spotandjake, line 1681: These should probably take advantage of proper multiline examples

stdlib/list.gr
```diff
@@ -901,3 +901,1012 @@ provide let sort = (comp, list) => {
...
+  }
+
+  /**
+   * Rotates list elements by the specified amount to the left, such that `n`th
+   * element becomes the first in the list.
+   *
+   * If value is negative, list elements will be rotated by the
+   * specified amount to the right. See examples.
+   *
+   * @param n: The number of elements to rotate by
```
spotandjake, line 1678:
```suggestion
   * @param count: The number of elements to rotate by
```

stdlib/list.gr
```diff
@@ -901,3 +901,1012 @@ provide let sort = (comp, list) => {
...
+    fromList(reduceRight((x, arr) => if (fn(x)) [x, ...arr] else arr, [], list))
+  }
+
+  /**
+   * Flattens nested lists.
+   *
+   * @param list: The list to flatten
+   * @returns A new list containing all nested list elements combined
+   *
+   * @example List.Mutable.flatten(fromList([fromList([1, 2]), fromList([3, 4])])) // fromList([1, 2, 3, 4])
```
spotandjake, line 1569: Multiline example

stdlib/list.gr
```diff
@@ -901,3 +901,1012 @@ provide let sort = (comp, list) => {
...
+  /**
+   * Creates a new list of the specified length where each element is
+   * initialized with the result of an initializer function. The initializer
+   * is called with the index of each list element.
+   *
+   * @param length: The length of the new list
+   * @param fn: The initializer function to call with each index, where the value returned will be used to initialize the element
+   * @returns The new list
+   *
+   * @example List.Mutable.init(5, n => n + 3) // fromList([3, 4, 5, 6, 7])
```
spotandjake, line 1045: Should we use an assert example here?

spotandjake (commented):
This implementation looks really solid. a few little documentaiton issues i noticed.

stdlib/list.gr
alex-snezhko: This will have to be rectified in a rebase regardless of whether this PR or the one making `List.sort`'s `comp` a default arg gets merged first so I just left it like this

spotandjake (approved):
Looks Good To Me.

marcusroberts (approved):
Amazing!     People can start to teach Computer Science using Grain now.

This would make a great example of some idiomatic Grain!

peblair (changes requested):
The implementation looks good here, but I have some nits with the high-level design (maybe I missed a conversation):
1. I think this module should be called `Deque`, not `Mutable`. For one, it's not just a mutable singly-linked list, which is what I would expect `List.Mutable` to be. Second, most languages (e.g. Java and Python) call these structures Deques.
2. With that change, I think it makes sense to pull this module out of `list.gr` and put it in its own file.

alex-snezhko:
@peblair True, but I think an argument in favor of keeping it as a member of the `List` module is that the interface for this type extends beyond the typical scope of a deque as an abstract data type, notably adding/removing/querying from only the beginning or end of the structure. There are indeed some languages like Python whose `collections.deque` has more capabilities than just these, but in my view that betrays the purpose of a deque.

I also think in this same way you can draw a parallel between Grain's `List` and `Stack`, where the only ways you can "modify" a `List` (adding to and removing from the front) capture the behavior of a stack, but the other functions made available to the `List` module like `map`, etc make it philosophically more general-purpose even though on its face a `List` is just a stack.

What do you think?

alex-snezhko:
@peblair wanted to bump the previous comment to see what your thoughts were?

PR #1849, opened 2023-05-28
feat(compiler): Providing, including, reproviding exceptions

### LINKED ISSUE #1053: Cannot re-export an exception

alex-snezhko:
Closes #1053

Syntax:
```
// provider.gr
provide exception MyException
// OR
exception MyException
provide { exception MyException }

// includer.gr
include "./provider"
from Provider use { exception MyException }
// can re-provide
provide { exception MyException }
```

### DIFF
--- a/compiler/test/grainfmt/includes.input.gr
+++ b/compiler/test/grainfmt/includes.input.gr
@@ -11,7 +11,7 @@ include /* special include */ "array"
 include "array" as  /* special include */ Foo
 from List use { length, map, forEach as each }
 from Opt use { module MutableOpt, module ImmutableOpt as Imm, type Opt, type Opt as OptAlias }
-from Opt use { module MutableOpt, /* comment1 */ module ImmutableOpt /* comment2 */ as /* comment3 */ Imm /* comment4 */, /* comment5 */ type /* comment6 */ Opt, type Opt as /* comment7 */ OptAlias, exception Exc as /* comment8 */ E }
+from Opt use { module MutableOpt, /* comment1 */ module ImmutableOpt /* comment2 */ as /* comment3 */ Imm /* comment4 */, /* comment5 */ type /* comment6 */ Opt, type Opt as /* comment7 */ OptAlias, exception Exc as /* comment8 */ E, exception Exc2 /* comment9 */ }
 
 include "runtime/unsafe/wasmi32"
 from WasmI32 use { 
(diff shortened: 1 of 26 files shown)


compiler/test/grainfmt/includes.input.gr
```diff
@@ -11,7 +11,7 @@ include /* special include */ "array"
 include "array" as  /* special include */ Foo
 from List use { length, map, forEach as each }
 from Opt use { MutableOpt, ImmutableOpt as Imm, type Opt, type Opt as OptAlias }
-from Opt use { MutableOpt, /* comment1 */ ImmutableOpt /* comment2 */ as /* comment3 */ Imm /* comment4 */, /* comment5 */ type /* comment6 */ Opt, type Opt as /* comment7 */ OptAlias }
+from Opt use { MutableOpt, /* comment1 */ ImmutableOpt /* comment2 */ as /* comment3 */ Imm /* comment4 */, /* comment5 */ type /* comment6 */ Opt, type Opt as /* comment7 */ OptAlias, exception Exc as /* comment8 */ E }
```
spotandjake, line 14: For completeness and to follow the rest it would probably make sense to add the test here without `as` as well to mirror the variable and type import.
```suggestion
from Opt use { MutableOpt, /* comment1 */ ImmutableOpt /* comment2 */ as /* comment3 */ Imm /* comment4 */, /* comment5 */ type /* comment6 */ Opt, type Opt as /* comment7 */ OptAlias, , exception /* comment8 */ Exc, exception Exc as /* comment9 */ E }
```

spotandjake (approved):
Looks good, just had one small nit.

ospencer (commented):
Can you add some tests where you provide an exception + a value of that exception and then successfully pattern match on it in the consuming module? Just want to make sure that works.

peblair (approved):
LGTM modulo @ospencer's test request

spotandjake:
as `exceptions` behave a little differently then most types. in that the value is based off the type. so if you did 
```
exception F
print(F) // F
```
when you use the from syntax with `as`. you could do 
```
from Module use { exception F as Y }

print(Y) // F
``` 
I feel like this might be a little confusing the closest analogy I can think of would be if we allowed enum aliasing, you could do something like `print(Enum.North)` and get `South` as the printed value. 

I wonder if we want to allow aliasing exceptions. 

Cut down version of message sent in discord.

marcusroberts (approved):
Looks useful to me and good syntax.

ospencer (approved):

PR #1852, opened 2023-06-03
feat(graindoc): Allow doc comments on variants and record fields

### LINKED ISSUE #992: GrainDoc: Allow documenting Enum variants
possibly something like such.
```
/*
 * Directions
 *
 * @variant NORTH: The north direction
 * @variant SOUTH: The south direction
 * @variant EAST: The east direction
 * @variant WEST: The west direction
 * /
enum DIRECTION {
  NORTH,
  SOUTH,
  EAST,
  WEST
}
```
and generated markdown
```
| Name    |        Description        |
| ------- | ----------------------    |
| North   | The north direction       |
| South   | The south direction       |
| East    | The east direction        |
| West    | The west direction        |
```

[Solves](https://github.com/grain-lang/grain/pull/945#discussion_r729356805_)

alex-snezhko:
Closes #992 

```
provide record R {
  /**
   * A very special number
   */
  x: Number
}

provide enum E {
  /**
   * A thing
   */
  Thing
}
```

### DIFF
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -241,9 +241,7 @@ let has_any_record_field_descrs = fields =>
   List.exists(f => Option.is_some(f.field_description), fields);
 
 let output_for_record_fields = (buf, fields) =>
-  if (!has_any_record_field_descrs(fields)) {
-    ();
-  } else {
+  if (has_any_record_field_descrs(fields)) {
     Buffer.add_string(buf, Markdown.paragraph("Fields:"));
     Buffer.add_string(
       buf,
@@ -274,9 +272,7 @@ let has_any_variant_descrs = variants =>
   List.exists(variant_has_desc, variants);
 
 let output_for_variants = (buf, variants) =>
-  if (!has_any_variant_descrs(variants)) {
-    ();
-  } else {
+  if (has_any_variant_descrs(variants)) {
     Buffer.add_string(buf, Markdown.paragraph("Variants:"));
     List.iter(
       variant => {
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -30,6 +30,11 @@ type throw = {
 
 type example = {example_txt: string};
 
+type compound_type_descrs =
+  | RecordFields(list((string, string)))
+  | Variants(list((string, string)))
+  | NonCompound;
+
 type t =
   | Type({
       module_namespace: option(string),
@@ -40,6 +45,7 @@ type t =
       since: option(since),
       history: list(history),
       examples: list(example),
+      compound_type_descrs,
     })
   | Value({
       module_namespace: option(string),
@@ -219,6 +225,26 @@ let output_for_throws = throws => {
   |> String.concat("");
 };
 
+let output_for_variants = variants => {
+  Markdown.table(
+    ~headers=["variant name", "description"],
+    List.map(
+      ((name, description)) => {[Markdown.code(name), description]},
+      variants,
+    ),
+  );
+};
+
+let output_for_record_fields = fields => {
+  Markdown.table(
+    ~headers=["field name", "description"],
+    List.map(
+      ((name, description)) => {[Markdown.code(name), description]},
+      fields,
+    ),
+  );
+};
+
 let types_for_function = (~ident, vd: Types.value_description) => {
   switch (Ctype.repr(vd.val_type).desc) {
   | TTyArrow(args, returns, _) => (Some(args), Some(returns))
@@ -261,6 +287,20 @@ let get_comments_from_loc = (loc: Grain_parsing.Location.t) => {
     };
 
   Comments.to_ordered(comments);
+};
+
+let attr_name = attr => {
+  Comment_attributes.(
+    switch (attr) {
+    | Deprecated(_) => "deprecated"
+    | Since(_) => "since"
+    | History(_) => "history"
+    | Param(_) => "param"
+    | Returns(_) => "returns"
+    | Throws(_) => "throws"
+    | Example(_) => "example"
+    }
+  );
 };
 
 let for_value_description =
@@ -422,6 +462,49 @@ let for_type_declaration =
   let comment =
     Comments.Doc.ending_on(~lnum=loc.loc_start.pos_lnum - 1, comments);
 
+  let extract_compound_type_descrs = data => {
+    List.filter_map(
+      ((loc: Warnings.loc, id)) => {
+        let comment =
+          Comments.Doc.ending_on(
+            ~lnum=loc.loc_start.pos_lnum - 1,
+            ~check_prev=false,
+            comments,
+          );
+        switch (comment) {
+        | Some((_, Some(description), [])) =>
+          Some((id.Ident.name, description))
+        | Some((_, _, [attr, ..._])) =>
+          raise(
+            InvalidAttribute({
+              name: Format.asprintf("%a", Printtyp.ident, id),
+              attr: attr_name(attr),
+            }),
+          )
+        | _ => None
+        };
+      },
+      data,
+    );
+  };
+
+  let compound_type_descrs =
+    switch (td.type_kind) {
+    | TDataVariant(cds) =>
+      Variants(
+        extract_compound_type_descrs(
+          List.map(cd => (cd.Types.cd_loc, cd.cd_id), cds),
+        ),
+      )
+    | TDataRecord(rfs) =>
+      RecordFields(
+        extract_compound_type_descrs(
+          List.map(rf => (rf.Types.rf_loc, rf.rf_name), rfs),
+        ),
+      )
+    | _ => NonCompound
+    };
+
   let (description, attributes) =
     switch (comment) {
     | Some((_, description, attributes)) => (description, attributes)
@@ -454,12 +537,10 @@ let for_type_declaration =
             [{history_version, history_msg}, ...history],
             examples,
           )
-        | Param({attr_id: param_id, attr_desc: param_msg}) =>
-          raise(InvalidAttribute({name, attr: "param"}))
-        | Returns({attr_desc: returns_msg}) =>
-          raise(InvalidAttribute({name, attr: "returns"}))
-        | Throws({attr_type: throw_type, attr_desc: throw_msg}) =>
-          raise(InvalidAttribute({name, attr: "throws"}))
+        | Param(_)
+        | Returns(_)
+        | Throws(_) =>
+          raise(InvalidAttribute({name, attr: attr_name(attr)}))
         | Example({attr_desc}) => (
             deprecations,
             since,
@@ -482,6 +563,7 @@ let for_type_declaration =
     since,
     history: List.rev(history),
     examples: List.rev(examples),
+    compound_type_descrs,
   });
 };
 
@@ -583,12 +665,10 @@ and for_signature_items =
             [{history_version, history_msg}, ...history],
             examples,
           )
-        | Param({attr_id: param_id, attr_desc: param_msg}) =>
-          raise(InvalidAttribute({name, attr: "param"}))
-        | Returns({attr_desc: returns_msg}) =>
-          raise(InvalidAttribute({name, attr: "returns"}))
-        | Throws({attr_type: throw_type, attr_desc: throw_msg}) =>
-          raise(InvalidAttribute({name, attr: "throws"}))
+        | Param(_)
+        | Returns(_)
+        | Throws(_) =>
+          raise(InvalidAttribute({name, attr: attr_name(attr)}))
         | Example({attr_desc}) => (
             deprecations,
             since,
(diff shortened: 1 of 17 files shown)


phated:
Excited to look at this! I'm pretty sure you need to add tests now that @spotandjake added the test harness

compiler/test/graindoc/types.expected.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+An enum
+
+Variants:
+
+|variant name|description|
+|------------|-----------|
+|`Variant1`|Enum variant|
+|`Variant2`|Another variant|
```
spotandjake, line 43: I question if this should contain the type info in here as well like 
```suggestion
|`Variant2`|(String, Number)|Another variant|
```
Similar to the docs for functions

compiler/test/graindoc/types.input.gr
```diff
@@ -0,0 +1,35 @@
...
+ */
+provide enum E {
+  /**
+   * Enum variant
+   */
+  Variant1,
+  /**
+   * Another variant
+   */
+  Variant2(String, Number),
```
spotandjake, line 29: Can we get a variant testing `Variant3{ t: Number }` it would be nice if we tested nested docs here?

After looking a little closer as your implementation I dot think you handle allowing docs on nested records we might want to handle that somehow. Idk how just yet maybe we append them to the table like `Variant.t` and the description, though that could be confusing.

spotandjake (commented):
This looks really good, I love how much it cleaner it makes the docs. I just had two little notes to keep the tables in line with the function tables but besides that it looks good.

Also after this is merged dont forgot to update the website docs on this feature.

compiler/test/graindoc/types.expected.md
alex-snezhko: Sounds like a good idea to have a separate type column for records, although for variants I'm not sure that this is the best way to display this info, perhaps it would be better to just display the whole variant description in the first column like `Variant2(String, Number)`?

spotandjake: Maybe, I think that it is worth thinking about consistency if we have functions with a separate type column and use a separate type column for records we should probably use a separate column for variants. I think that things might get a little messy with records attached to variants though so this might need some discussion or be something we come back to in another pr later on.

compiler/test/graindoc/types.input.gr
alex-snezhko: Yeah I thought about this case as well and am unsure of the best way to display it; including a nested table in a new column is one option, and although I initially discounted this option it may be better than separating it out into another top-level table as that would cause an awkward reading flow imo :thinking:

peblair (commented):
Looks good, but holding off on an approval until @spotandjake's comments are addressed

compiler/test/graindoc/types.input.gr
alex-snezhko: I tried playing around with different formatting styles and I think having inline record data in a nested table seems like an ok option; since there would only ever be one layer of nesting I think it wouldn't be unreasonable. For example the following code could generate this table:
```
provide enum E {
  /**
   * Singleton constr
   */
  S,
  /**
   * Tuple constr
   */
  T(Number),
  /**
   * Record constr
   */
  R {
    /**
     * A number
     */
    x: Number,
    /**
     * A function
     */
    f: Number -> Number
  }
}
```
image

The image was generated with https://markdownlivepreview.com/, but will probably look similar on the website. Thoughts on this display format?

spotandjake: It looks better than I was expecting but I wonder if anyone else has any better ideas. Before we implement.

alex-snezhko: I ended up going ahead and implementing it, if there are suggestions for better output styling I'll update

compiler/src/diagnostics/comments.re
```diff
@@ -236,7 +236,7 @@ module Doc = {
     };
   };
 
-  let ending_on = (~lnum, module C: OrderedComments) => {
+  let ending_on = (~lnum, ~check_prev=true, module C: OrderedComments) => {
```
phated, line 239: I don't like passing this through since it is leaking a super hacky workaround

compiler/src/utils/markdown.re
```diff
@@ -15,26 +15,60 @@ let code = code => {
   Format.sprintf("`%s`", code);
 };
 
-let table = (~headers: list(string), rows) => {
+let table = (~nested=false, ~headers: list(string), rows) => {
```
phated, line 18: I don't like boolean flags like this, we should think through a real API that is needed if we are putting tables inside

stdlib/sys/process.gr
```diff
@@ -19,61 +19,113 @@ from DataStructures use { tagSimpleNumber, allocateArray, allocateString }
...
   XFSZ,
-  // Virtual timer expired.
+  /**
+   * Virtual timer expired.
+   */
   VTALRM,
   PROF,
   WINCH,
   POLL,
   PWR,
```
phated, line 125: Can we get descriptions on these too?

compiler/test/graindoc/types.expected.md
```diff
@@ -0,0 +1,57 @@
...
+
+An enum
+
+Variants:
+
+|name|data|description|
+|----|----|-----------|
+|`Variant1`||Enum variant|
+|`Variant2`|`(String, Number)`|Another variant|
+|`Variant3`|<table><thead><tr><th>name</th><th>type</th><th>description</th></tr></thead><tbody><tr><td>`a`</td><td>`Number`</td><td>Record field a</td></tr><tr><td>`f`</td><td>`Number => Number`</td><td>Function</td></tr></tbody></table><br /><br />|Record variant|
```
phated, line 48: Sorry for the delay here. I don't really like this layout. How about we do this?

````suggestion
Variants:

```grain
Variant1
```
Enum variant

```grain
Variant2(String, Number)
```
Another variant

```grain
Variant3{
  a: Number,
  f: Number => Number,
}
```
Record Variant

Fields:

|name|type|description|
|----|----|-----------|
|`a`|`Number`|Record field a|
|`f`|`Number => Number`|Function|
````

I was also thinking about making these a bulleted list, but I believe that'd be hard while also making them grain-syntax blocks.

File: stdlib/sys/process.md [hunk 7]
```diff
@@ -53,6 +53,37 @@ enum Signal {
...
+|`CHLD`||Child process terminated, stopped, or continued.|
+|`CONT`||Continue executing, if stopped.|
+|`STOP`||Stop executing.|
+|`TSTP`||Terminal stop signal.|
+|`TTIN`||Background process attempting read.|
+|`TTOU`||Background process attempting write.|
+|`URG`||High bandwidth data is available at a socket.|
+|`XCPU`||CPU time limit exceeded.|
+|`XFSZ`||File size limit exceeded.|
+|`VTALRM`||Virtual timer expired.|
```
phated, line 84: It seems that the enum variants without a description aren't included

compiler/test/graindoc/types.expected.md
alex-snezhko: Assuming that record fields are being kept in a table I'm not a huge fan of having enum variants not be in a table. What about something like this?
```
|variant|description|
|------------|-----------|
|`Variant1`|Enum variant|
|`Variant2(String, Number)`|Another variant|
|<code>Variant3{<br>&nbsp;&nbsp;a: Number,<br>&nbsp;&nbsp;f: Number => Number<br>}</code>|Record variant<br><ul><li>`a`: Record field a</li><li>`f`: Function</li></ul>|
```

image

phated: I don't like them in a table. It's too dense

stdlib/sys/process.md
alex-snezhko: I did this intentionally. Since the full enum definition is already printed I didn't see much of a point to having an entry with just the variant name and no context

phated: That's fine when we remove the table for variants. If it were a table like you did originally, it should be the full enumeration.

alex-snezhko: This would also apply to record fields though. Unless we want to ditch the table for those as well?

phated: Looking at my layout suggestion, record fields remain in a table, which contains all fields.

stdlib/sys/process.gr
alex-snezhko: On second thought should these just be removed altogether? Looks like signal handling isn't a thing anymore https://github.com/WebAssembly/WASI/issues/166

compiler/test/graindoc/types.expected.md
alex-snezhko: Applied your suggested format

stdlib/sys/process.gr
alex-snezhko: I think maybe we can clean these up when we rewrite everything to use wasi preview 2?

compiler/graindoc/docblock.re
```diff
@@ -219,6 +237,67 @@ let output_for_throws = throws => {
   |> String.concat("");
 };
 
+let has_any_record_field_descrs = fields =>
+  List.exists(f => Option.is_some(f.field_description), fields);
+
+let output_for_record_fields = (buf, fields) =>
+  if (!has_any_record_field_descrs(fields)) {
+    ();
+  } else {
```
phated, line 246: Can you flip this logic so we don't need the `!`?

compiler/graindoc/docblock.re
```diff
@@ -219,6 +237,67 @@ let output_for_throws = throws => {
...
+  || Option.map(has_any_record_field_descrs, variant.variant_record_fields)
+  == Some(true);
+
+let has_any_variant_descrs = variants =>
+  List.exists(variant_has_desc, variants);
+
+let output_for_variants = (buf, variants) =>
+  if (!has_any_variant_descrs(variants)) {
+    ();
+  } else {
```
phated, line 279: Can you flip the if-else to avoid the `!`?

compiler/graindoc/docblock.re
```diff
@@ -420,7 +516,62 @@ let for_type_declaration =
...
+        let comment =
+          Comments.Doc.ending_on(~lnum=loc.loc_start.pos_lnum - 1, comments);
+        switch (comment) {
+        | Some((_, _, [attr, ..._])) =>
+          raise(
+            InvalidAttribute({
+              name: Format.asprintf("%a", Printtyp.ident, id),
+              attr: attr_name(attr),
+            }),
+          )
```
phated, line 536: Does this need the logic for non-duplicate `@since` that @spotandjake added in the other PR?

compiler/src/typed/printtyp.re
```diff
@@ -986,6 +986,15 @@ let constructor_arguments = (ppf, a) => {
   Oprint.out_type^(ppf, Otyp_tuple(tys));
 };
 
+let constructor = (ppf, cstr) =>
+  Oprint.out_constr^(ppf, tree_of_constructor(cstr));
+
+let string_of_constructor = cstr => {
+  let str = asprintf("%a", constructor, cstr);
+  // Hacky workaround to avoid having to make invasive changes in Oprint
+  Str.global_replace(Str.regexp("^  "), "", str);
```
phated, line 995: Perhaps open an issue to clean this up when we drop oprint for the formatter code?

phated (changes requested):
I like the output a lot! I had just a couple comments on the implementation

compiler/src/typed/printtyp.re
alex-snezhko: https://github.com/grain-lang/grain/issues/1992

compiler/graindoc/docblock.re
alex-snezhko: This would raise the exception if there are any attributes on the docblock at all, so deduplicating isn't necessary here

phated: Hm, I'm pretty sure the JSON PR uses the since attribute on each variant so we'll want to support that

alex-snezhko: Hmm ok, should `@deprecated` and `@history` be supported as well?

phated: Yeah, I think those would be supported

alex-snezhko: What about record fields? And if so then how would they be displayed in the table?

phated: It's probably fine to only support them on variants. When thinking about exhaustive matching, I think about variants more than record fields and you might add variants over time.

alex-snezhko: Hmm, I think it'd be confusing if they were supported on variants but not record fields; for instance if you had a field that changed its type or name or something. What if it's just inserted below the field's description in the table cell? It wouldn't take up much space by default since it'd be collapsed in the `<details>` tag

phated: I don't want it in the table. We can open a follow up issue to evaluate for records

phated: Merging without attribute support and @spotandjake can remove them from the JSON PR

phated: #1998 and #1999

phated (approved):

PR #1854, opened 2023-06-10
chore(grainfmt): Removed all the comments from tests

marcusroberts:
Simplified the formatter tests by removing comments


phated (changes requested):
I think it's valuable to have a test suite "with comments" and "without comments". It seems like a waste to lose some of the tests that only existed to test comments

phated:
Oscar did this and pushed even though I asked him not to. So we can close this.

PR #1855, opened 2023-06-11
feat(compiler)!: Change `->` to `=>` in type signatures

alex-snezhko:
Closes #1179 (supersedes)

### DIFF
--- a/compiler/src/formatting/format.re
+++ b/compiler/src/formatting/format.re
@@ -1792,7 +1792,7 @@ and print_type =
         },
       ),
       Doc.space,
-      Doc.text("->"),
+      Doc.text("=>"),
       Doc.space,
       print_type(~original_source, ~comments, parsed_type),
     ])
--- a/compiler/src/parsing/lexer.re
+++ b/compiler/src/parsing/lexer.re
@@ -309,7 +309,6 @@ let rec token = lexbuf => {
   | "/" => positioned(SLASH)
   | "|" => positioned(PIPE)
   | "-" => positioned(DASH)
-  | "->" => positioned(ARROW)
   | "=>" => positioned(THICKARROW)
   | "type" => positioned(TYPE)
   | "enum" => positioned(ENUM)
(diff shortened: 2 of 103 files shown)


spotandjake (approved):
This looks good to me. 

I don't fully understand the code in the lever hack that checks the data types but what I did understand looks good.

ospencer (approved):
Fantastic work here!

marcusroberts (approved):
If Oscar's happy with the parsing, this looks great.

One question, can we use the formatter or provide a utility to help people bulk change their code to the new style?

peblair (commented):
Two comments:
1. I would love to see some stress tests/weird cases which should not parse (`let f = (foo, bar) => Number`) in the test suite
2. Extremely minor concern: just want to double-check other Core Team members (cc @ospencer and @phated especially) are cool with this possibly introducing some potential confusion for new users (e.g. `let lexer : (stream) => List<Token>` might look to a new user like `stream` is supposed to be the name of the function argument and not a placeholder type).

spotandjake:
> 2. Extremely minor concern: just want to double-check other Core Team members (cc @ospencer and @phated especially) are cool with this possibly introducing some potential confusion for new users (e.g. `let lexer : (stream) => List<Token>` might look to a new user like `stream` is supposed to be the name of the function argument and not a placeholder type).

I feel like the issue of the confusion was already there before the arrow change.

peblair:
> I feel like the issue of the confusion was already there before the arrow change.

Yes, true, but the `->` syntax made it a bit more clear when reading quickly that one was looking at a type, not a function definition. Maybe I am wording this scenario wrong, but I am just envisioning some potential weird interaction with the unified `=>` syntax and placeholder type names

alex-snezhko:
@peblair I added some parsing tests to `types.re`; are those the sort of cases you had in mind?

phated (approved):

phated:
Going to merge this, we can always add more tests later 🎉

Issue #1836, opened 2023-05-02
Docs: Change Griandoc wording in bytes


spotandjake:
https://github.com/grain-lang/grain/blob/618c1ba309b19409ba57f2e2b610a55883a03a43/stdlib/bytes.gr#L251

In the `bytes` module for `bytes.move` we use the term `from the source buffer` given we have a seperate buffer type I think it would be good if we found another word for this.

PR #1858, opened 2023-06-12
chore(stdlib): Clarify `Byte.move` docs

spotandjake:
This is a small pr changing the wording used in `Bytes.move`, it makes sense to stay away from the wording `Buffer` given grain has a specific `Buffer` type which is different than the `Bytes` type.

Closes: #1836


marcusroberts (approved):
Good spot!

peblair (approved):

phated (approved):

Issue #696, opened 2021-06-04
Clean up regular expression code with or-patterns


peblair:
The match expressions in `regex.gr` contain a variety of duplicated branches which differ only by the variant being matched. We should clean these up using or-patterns.

ospencer:
Unblocked by #1173!

PR #1859, opened 2023-06-13
chore(stdlib): Use `|` patterns to cleanup pattern matches

spotandjake:
This pr converts some of the matches in the `Regex` lib to use or patterns.

Closes: #696 

One other bit of cleanup I wonder if it is worth it is to merge these to use a variable f, and then just pass that to the functions:
image
@peblair is there a reason you didnt use a variable here.


phated (approved):
Looks great! Thanks @spotandjake

Issue #1784, opened 2023-03-29
Add List `isEmpty`


spotandjake:
So this might sound useless because you could just do `length(lst) == 0` but this would have an advantage in a case where you do not care about the `length` which needs to be computed by checking every element in the list and just care about if the list has contents.

TLDR: This would be a lot faster then doing `length(lst) == 0` for lists that are not empty.

Discussion: It wouldnt really speed up `array`, or `string`, or any of the other data types maybe `map`, `set` and `queue` im not quite sure how they calculate the size but it might make sense to add to those libs as well just to keep the apis the same.

alex-snezhko:
Couldn't you just do `lst == []` for checking emptiness?

spotandjake:
> Couldn't you just do `lst == []` for checking emptiness efficiently?

🤦 that is a really good point sometimes you overlook some really obvious stuff.

PR #1860, opened 2023-06-14
feat(stdlib): Add `isEmpty` to `List` module

spotandjake:
This pr adds `List.isEmpty` bringing the api inline with the other data types. This is beneficial for `List` especially as the common way of checking if a `List` is empty is `List.length(i) == 0` which in the case of list would have to iterate the entire list to calculate the length.

closes: #1784


jozanza (approved):

phated (approved):
Perfect PR on first try. Delightful 🎉

PR #1861, opened 2023-06-14
feat(stdlib): add `isEmpty` to `String` module

spotandjake:
This pr adds `String.isEmpty` to the string library to bring it in line with other api's from different modules, off of oscars in discord stdlib notes.

Benefit of adding this is making the api the same as other data types and providing more clarity in code.

Note: empty means no characters are included in the module so `bytelength` is equivalent to 0.


jozanza (approved):

phated (approved):
💯

PR #1864, opened 2023-06-14
feat(stdlib): Add `forEachChar`, `forEachCharI`, `map` and `mapi` to String module

spotandjake:
This pr adds `forEachChar`, `forEachCharI`, `map` and `mapi` to the String library.

### DIFF
--- a/stdlib/string.gr
+++ b/stdlib/string.gr
@@ -1971,7 +1971,7 @@ provide let forEachCodePointi = (fn: (Number, Number) => Void, str: String) => {
  * @since v0.6.0
  */
 @unsafe
-provide let forEachChar = (fn: Char -> Void, str: String) => {
+provide let forEachChar = (fn: Char => Void, str: String) => {
   from WasmI32 use { (+), (-), (&), (>>>), ltU as (<), leU as (<=), (==) }
 
   let strPtr = WasmI32.fromGrain(str)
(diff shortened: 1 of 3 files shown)


stdlib/string.gr
```diff
@@ -2016,6 +2016,148 @@ provide let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+  from WasmI32 use { (+), (-), (&), (>>>), ltU as (<), leU as (<=), (==) }
+
+  let strPtr = WasmI32.fromGrain(str)
+
+  let byteSize = WasmI32.load(strPtr, 4n)
+
+  let mut ptr = strPtr + 8n
+  let end = ptr + byteSize
+
+  let mut idx = 0n
```
alex-snezhko, line 2040: `idx` is never used

stdlib/string.gr
```diff
@@ -2016,6 +2016,148 @@ provide let forEachCodePointi = (fn: (Number, Number) -> Void, str: String) => {
...
+ *
+ * @param fn: The iterator function
+ * @param str: The string to iterate
+ *
+ * @example String.forEachChar(print, "Hello world")
+ *
+ * @since v0.6.0
+ */
+@unsafe
+provide let forEachChar = (fn: Char -> Void, str: String) => {
```
alex-snezhko, line 2030: You could probably share some code between this function and `forEachCodePoint`; same with `forEachChari`

spotandjake: I made the `idx` change, but I dont think it is optimial to share code here, the type signatures of `fn` are different so its not as simple as just having a bool that says to use char or code point. meaning you would need to use a wrapper function into the helper that tags the char, which it invokes an extra function call. On another note if we do want to go this route we might as well justhave a `forEachCodePointIHelper` which returns an unsafe code point. and then each function just becomes `forEachCodePointIHelper((c, _) => fn(tagChar(c)), str)`.

alex-snezhko (approved):

ospencer (approved):

PR #1865, opened 2023-06-14
fix(stdlib): Properly resize empty Queue on push

spotandjake:
This fixes a tiny bug where if you did:
```
let q = Queue.makeSized(0)
Queue.push(1, q)
```
you would get an index out of range error.

Breaking because you will no longer recieve an exception.

### DIFF
--- a/stdlib/queue.gr
+++ b/stdlib/queue.gr
@@ -90,7 +90,6 @@ provide let peek = queue => {
  * @param queue: The queue being updated
  *
  * @since v0.6.0
- * @history v0.6.0: No longer errors when pushing a new item
  */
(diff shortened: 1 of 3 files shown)


alex-snezhko:
Could you add this fix for `Stack` too?

stdlib/queue.gr
```diff
@@ -90,11 +90,14 @@ provide let peek = queue => {
  * @param queue: The queue being updated
  *
  * @since v0.6.0
+ * @history v0.6.0: No longer errors when pushing a new item
```
phated, line 93:
```suggestion
```
Same as stack

phated (approved):

PR #1866, opened 2023-06-14
feat(stdlib): Add `toList` and `fromList` functions to `Queue` module

spotandjake:
adds `toList` and `fromList` to the queue and Immutable queue modules.

Blocked by: #1865

### DIFF
--- a/stdlib/queue.gr
+++ b/stdlib/queue.gr
@@ -146,10 +146,12 @@ provide let pop = queue => {
  * @since v0.6.0
  */
 provide let toList = queue => {
-  let lst = List.init(queue.size, i => match (queue.array[queue.headIndex +
-    i]) {
+  let lst = List.init(queue.size, i => match (queue.array[(queue.headIndex +
+    i) %
+    Array.length(queue.array)]) {
     Some(n) => n,
-    None => fail "Impossible: Attempted to access non-existent bucket in Queue.toList",
+    None =>
+      fail "Impossible: Attempted to access non-existent bucket in Queue.toList",
   })
   lst
 }
--- a/stdlib/queue.gr
+++ b/stdlib/queue.gr
@@ -149,7 +149,7 @@ provide let toList = queue => {
   let lst = List.init(queue.size, i => match (queue.array[queue.headIndex +
     i]) {
     Some(n) => n,
-    None => fail "Impossible",
+    None => fail "Impossible: Attempted to access non-existent bucket in Queue.toList",
   })
   lst
 }
--- a/stdlib/queue.gr
+++ b/stdlib/queue.gr
@@ -90,6 +90,7 @@ provide let peek = queue => {
  * @param queue: The queue being updated
  *
  * @since v0.6.0
+ * @history v0.6.0: No longer errors when pushing a new item
  */
 provide let push = (value, queue) => {
   let arrLen = Array.length(queue.array)
--- a/stdlib/queue.gr
+++ b/stdlib/queue.gr
@@ -139,6 +139,37 @@ provide let pop = queue => {
 }
 
 /**
+ * Converts a queue into a list of its elements.
+ *
+ * @param queue: The queue to convert
+ * @returns A list containing all queue values
+ *
+ * @since v0.6.0
+ */
+provide let toList = queue => {
+  let lst = List.init(size(queue), i => match (queue.array[queue.headIndex +
+    i]) {
+    Some(n) => n,
+    None => fail "Impossible",
+  })
+  lst
+}
+
+/**
+ * Creates a queue from a list.
+ *
+ * @param list: The list to convert
+ * @returns A queue containing all list values
+ *
+ * @since v0.6.0
+ */
+provide let fromList = list => {
+  let queue = makeSized(List.length(list))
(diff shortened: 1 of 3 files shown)


spotandjake:
This commit has been unblocked

stdlib/queue.gr
```diff
@@ -90,6 +90,7 @@ provide let peek = queue => {
  * @param queue: The queue being updated
  *
  * @since v0.6.0
+ * @history v0.6.0: No longer errors when pushing a new item
```
phated, line 93:
```suggestion
```
Accidentally re-added by your merge

stdlib/queue.gr
```diff
@@ -137,6 +138,37 @@ provide let pop = queue => {
...
+/**
+ * Converts a queue into a list of its elements.
+ *
+ * @param queue: The queue to convert
+ * @returns A list containing all queue values
+ *
+ * @since v0.6.0
+ */
+provide let toList = queue => {
+  let lst = List.init(size(queue), i => match (queue.array[queue.headIndex +
```
phated, line 150: since you are using internals to implement this, why wouldn't you just use `queue.size` instead of calling the function?

stdlib/queue.gr
```diff
@@ -137,6 +138,37 @@ provide let pop = queue => {
...
+    i]) {
+    Some(n) => n,
+    None => fail "Impossible",
```
phated, line 153: Needs a better impossible message

stdlib/queue.gr
```diff
@@ -295,4 +327,43 @@ provide module Immutable {
...
+  provide let toList = queue => {
+    let mut queue = queue
+    let mut lst = []
+    while (!isEmpty(queue)) {
+      match (peek(queue)) {
+        Some(value) => {
+          queue = pop(queue)
+          lst = [value, ...lst]
+        },
+        None => break,
```
phated, line 349: Doesn't a `None` on peek indicate an empty queue? Why even do the `isEmpty` check?

But also, since you have access to the internals, can't this be implemented something like `List.concat(queue.forwards, List.reverse(queue.backwards))`?

stdlib/queue.gr
```diff
@@ -295,4 +327,43 @@ provide module Immutable {
...
+   * @param list: The list to convert
+   * @returns A queue containing all list values
+   *
+   * @since v0.6.0
+   */
+
+  provide let fromList = list => {
+    let mut queue = empty
+    List.forEach(e => queue = push(e, queue), list)
+    queue
```
phated, line 367: Given the above, can't this also be implemented by just constructing the record with the list as `forwards` ?

phated (changes requested):
I think we can tighten up some of these implementations.

stdlib/queue.gr
spotandjake: How about `"Impossible: Empty Bucket In Queue.toList`?

spotandjake:
Made those changes, those suggestions made things quite a bit simpler :)

stdlib/queue.gr
phated: Isn't it more about trying to access a bucket that doesn't exist?

spotandjake: fair, so how about: `Impossible: Attempted To Access Non Existant Bucket In Queue.toList`

phated:
```suggestion
    None => fail "Impossible:  Attempted to access non-existent bucket in Queue.toList",
```
Sounds good to me, fixed spelling and capitalization in suggestion

phated (approved):
This looks good to me but I'd love a review from @alex-snezhko too

stdlib/queue.gr
```diff
@@ -137,6 +137,37 @@ provide let pop = queue => {
...
+ * Converts a queue into a list of its elements.
+ *
+ * @param queue: The queue to convert
+ * @returns A list containing all queue values
+ *
+ * @since v0.6.0
+ */
+provide let toList = queue => {
+  let lst = List.init(queue.size, i => match (queue.array[queue.headIndex +
+    i]) {
```
alex-snezhko, line 150: You'd have to mod this by the queue size since the elements in the queue could wrap around past the end and back to the beginning of the array

alex-snezhko (dismissed):
LGTM except for the one change

stdlib/queue.gr
spotandjake: so I would have todo `[(queue.headIndex + i) % queue.size`?

alex-snezhko: It would be the length of `queue.array`. You can take a look at some of the other functions for an idea of what is done

spotandjake: Made that change

phated: Needs a test where they'd wrap back around

peblair (approved):

stdlib/queue.gr
phated: @spotandjake I'm not sure if this got lost but it still needs the test I mentioned.

spotandjake: Added that test

phated (approved):

PR #1867, opened 2023-06-14
fix(stdlib): Properly resize empty Stack on push

spotandjake:
Applies the fix made on queue in #1865 to the `Stack` Module

### DIFF
--- a/stdlib/stack.gr
+++ b/stdlib/stack.gr
@@ -86,7 +86,6 @@ provide let peek = stack => {
  * @param stack: The stack being updated
  *
  * @since v0.6.0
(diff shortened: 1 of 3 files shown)


stdlib/stack.gr
```diff
@@ -86,10 +86,13 @@ provide let peek = stack => {
  * @param stack: The stack being updated
  *
  * @since v0.6.0
+ * @history v0.6.0: No longer errors when stack size is `0`
```
phated, line 89:
```suggestion
```
This is a bug, so we don't need history and we don't need to mark breaking.

phated (approved):

PR #1870, opened 2023-06-21
feat(compiler): Allow exporting the function table

spotandjake:
While trying to play arround with https://github.com/contextfreeinfo/taca/tree/main I noticed that it uses the function table for callbacks which grain didnt support. This pr adds the ability to run the compiler with the `--export-function-table` flag which takes a `name` as a string. which causes the function table to be exported under the given name during linking. This is primarily useful for writing foreign bindings.

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -147,6 +147,10 @@ class GrainCommand extends commander.Command {
       "replaces the _start export with a start section during linking"
     );
     cmd.forwardOption(
+      "--export-function-table",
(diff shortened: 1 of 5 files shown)


spotandjake:
Converting to draft as this might need some more thought. given that grain closures make functions in the function table a little harder.

spotandjake:
I am closing this pr, I think the feature would still be beneficial eventually but more thought needs to go into it and as we removed the linker with the object file change the implementation here won't be much help in the future.

Issue #1847, opened 2023-05-27
Grainrun: Errors when compiling with use-start-section


spotandjake:
```
TypeError [ERR_INVALID_ARG_TYPE]: The "instance.exports._start" property must be of type function. Received undefined
    at WASI.start (node:wasi:110:5)
    at run (/Users/spotandjake/Documents/Github/grain/cli/bin/grainrun.js:37:10)
```
The grainrun produces the above error when running `grain index.gr --use-start-section` we should either fix the runner to properly handle this or give a warning and not run your grain program if you compile using `use-start-section`

PR #1871, opened 2023-06-23
fix(cli)!: Ensure `--use-start-section` can only be used with compile command

spotandjake:
This pr changes the cli to make it so `grain main.gr --use-start-section` no longer works and you have to use `grain compile main.gr --use-start-section` when using the `--use-start-section` flag.

This was done because in the move to the new wasi runner. the way the runner loads the wasm module changed meaning that if you tried to use the grain runner with `--use-start-section` you would get an error as wasi is not intialized when node calls the start function.

This is marked breaking because the flag can no longer be used with the runner.

This should be able to be reverted when grain switches to using wamr as a wasm runner.

closes: #1847

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -143,10 +143,6 @@ class GrainCommand extends commander.Command {
       "path to custom WASI implementation"
     );
     cmd.forwardOption(
-      "--use-start-section",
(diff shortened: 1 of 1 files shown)


phated (approved):
Hope this doesn't conflict with the other PR I just queued

PR #1876, opened 2023-07-03
chore(compiler): Update ocaml to upstream release

phated:
This updates ocaml in our esy lock to rely on upstream 4.14.1 as published by the esy maintainers. The released version has my patches for flexdll so we don't need my fork anymore.

I also updated dune since we don't pin every dep that comes out of the dune monorepo and they were out of sync with the lockfile updates.


compiler/src/diagnostics/comments.re
```diff
@@ -158,7 +158,9 @@ module type OrderedComments = {
 };
 
 module MakeOrderedComments =
-       (Raw: {let comments: list(Typedtree.comment);})
+       (Raw: {
```
spotandjake, line 161: looks good, just a nit but is there a reason this changed? (same as `anf_mapper.re`)?

spotandjake (approved):

compiler/src/diagnostics/comments.re
phated: Reason updated.

phated:
I figured out how to sync the reason dependency so we aren't building 2, which also dropped the utop dep and all its deps 🎉

ospencer (approved):

phated:
Reverted the dune updates until they release https://github.com/ocaml/dune/pull/8100

phated:
I vaguely remember have a problem with esy on windows when swapping out the reason version in a different PR. It seems I might be hitting that again.

phated:
This did seem to be an issue, so I've reverted the resolutions on reason

phated:
Dune released 3.9.1 so I updated that again. However, the newest utop doesn't build on windows so I pinned it. This looks good now.

Issue #1879, opened 2023-07-05
stdlib: Optimize Buffer.addChar


spotandjake:
I was looking over the code in buffer the other day and noticed that `addChar` converts to a string and then calls `addString` this is wasteful and creates an extra allocation for no reason. When you are using chars and buffers it is often for more performance, so we should probably rectify this.

PR #1133, opened 2022-02-06
feat(stdlib): Add `Json` module

cician:
This PR aims to include my JSON parser/printer (https://github.com/cician/grain-json) into stdlib. Hopefully for 0.5 release.

I'm opening this PR as a draft for now since I expect to still make some important changes once stack allocated chars and thick numbers are in. Strictly speaking no changes are really required after stack chars PR is merged, but code can be improved for clarity by using chars instead of code points as raw numbers. Thick numbers on the other hand are going to affect the behavior of number parsing.

PR #1883, opened 2023-07-15
feat(stdlib): Add `addCharFromCodePoint` to Buffer module

spotandjake:
This add `Buffer.addCharFromCodePoint` to the `Buffer` module, this is used in #1133. This is very similar to doing:
```
Buffer.addUint32(Char.code('a'))
```
but it handles the correct byte count of chars.

Internally I chose to just call `addChar` this leaves less surface area for bugs to crop up, and given char is a stack type we are not giving up a ton of perf. Well once #1879 is completed.


stdlib/buffer.gr
```diff
@@ -322,6 +322,18 @@ provide let addChar = (char, buffer) => {
   addString(Char.toString(char), buffer)
 }
 
+/**
+ * Appends a char from a code point to a buffer.
+ *
+ * @param codePoint: The codePoint to append to the buffer
```
ospencer, line 328:
```suggestion
 * @param codePoint: The code point to append to the buffer
```

spotandjake: Done

jozanza (approved):

peblair (approved):

stdlib/buffer.gr
```diff
@@ -322,6 +322,18 @@ provide let addChar = (char, buffer) => {
   addString(Char.toString(char), buffer)
 }
 
+/**
+ * Appends a char from a code point to a buffer.
```
phated, line 326:
```suggestion
 * Appends a character represented by a code point a buffer.
```

phated (changes requested):
Doc could use some work

stdlib/buffer.gr
spotandjake: I made this change, one note is should we refer to the `char` as a `character` in `addchar`? 
currently we say `Appends the bytes of a char to a buffer.`

phated: Sure, we use full words for descriptions, but that is unrelated to this PR.

stdlib/buffer.gr
```diff
@@ -322,6 +322,18 @@ provide let addChar = (char, buffer) => {
   addString(Char.toString(char), buffer)
 }
 
+/**
+ * Appends a character represented by a code point a buffer.
```
phated, line 326:
```suggestion
 * Appends a character represented by a code point to a buffer.
```
Looks like we lost this word

phated (approved):

PR #1887, opened 2023-07-16
fix(compiler): Type aliasing in `use` statements

### LINKED ISSUE #1869: Lang: Re-providing aliased types fails
Attempting to re-provide a type from one module after aliasing it with a `use` alias causes a compilation failure. Example that causes a failure:
```
//a.gr
module A
provide enum T {
  Val
}

// b.gr
module B
include "./a"
from A use { type T as AliasedT }
provide { type AliasedT }
```

alex-snezhko:
Closes #1869

### DIFF
--- a/compiler/src/typed/env.re
+++ b/compiler/src/typed/env.re
@@ -329,6 +329,8 @@ module IdTbl = {
     root: Path.t,
     /** Components from the opened module. */
     components: Tbl.t(string, ('a, int)),
+    /** Aliased names in the module. */
+    aliases: Tbl.t(string, string),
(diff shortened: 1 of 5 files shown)


spotandjake (approved):
Looks great 👍

ospencer (commented):
Could you add tests where you annotate using the alias?

peblair (approved):
LGTM (with the test that @ospencer requested)

alex-snezhko:
Added a test with annotation

ospencer (approved):
Excellent, great work!

PR #1892, opened 2023-07-19
chore(compiler): Improve unsafe equality warning

spotandjake:
Small Improvement to the `FuncWasmUnsafe` warning, to make it slightly more helpful.


compiler/src/utils/warnings.re
```diff
@@ -119,10 +119,12 @@ let message =
...
   | NonClosedRecordPattern(s) =>
     "the following fields are missing from the record pattern: " ++ s
-  | FuncWasmUnsafe(func) =>
+  | FuncWasmUnsafe(func, m) =>
     "it looks like you are using "
     ++ func
-    ++ " on two unsafe Wasm values here.\nThis is generally unsafe and will cause errors. Use one of the equivalent functions in `WasmI32`, `WasmI64`, `WasmF32`, or `WasmF64` instead."
+    ++ " on two unsafe Wasm values here.\nThis is generally unsafe and will cause errors. Use "
+    ++ m
+    ++ " instead."
```
ospencer, line 127: You still need to tell them that they need to use one of the functions in that library. I think this phrasing would be pretty confusing if you don't have the context that libraries exist for these types.

spotandjake: Agreed, now it says the module as well

jozanza:
The message phrasing seems pretty good with the module name included 😎

jozanza (approved):

peblair (approved):
🚀 Great!

compiler/src/typed/typed_well_formedness.re
```diff
@@ -228,9 +245,18 @@ module WellFormednessArg: TypedtreeIter.IteratorArgument = {
         )
           when func == "==" || func == "!=" =>
         if (List.exists(((_, arg)) => exp_is_wasm_unsafe(arg), args)) {
+          let (_, arg) = List.nth(args, 0);
```
phated, line 248:
```suggestion
          let (_, arg) = List.hd(args);
```
this should either use `List.hd` or a pattern match. Note that both `nth` and `hd` raise on an empty list. How are we certain this list contains an element?

compiler/src/typed/typemod.re
```diff
@@ -1042,7 +1042,7 @@ let type_implementation = prog => {
   let initenv = initial_env();
   let (statements, sg, finalenv) = type_module(initenv, prog.statements);
   let simple_sg = simplify_signature(sg);
-  let filename = sourcefile; // TODO(1396): Don't use filepath as filename
+  let filename = sourcefile; // TODO(#1396): Don't use filepath as filename
```
phated, line 1045:
```suggestion
  let filename = sourcefile; // TODO(1396): Don't use filepath as filename
```
Unrelated to this PR

stdlib/set.gr
phated, line 1: Formatting changes in this file are unrelated to this PR

phated (changes requested):
Some issues

compiler/src/typed/typed_well_formedness.re
spotandjake: I went with the pattern match, and have a fallback where we show all the possible modules because we dont know the module in that case. It is worth noting I dont think that we can hit here without having args as the typechecker will throw a warning that you did not include enough arguments.

spotandjake:
Fixed those and reverted the changes

compiler/src/typed/typed_well_formedness.re
```diff
@@ -228,9 +245,21 @@ module WellFormednessArg: TypedtreeIter.IteratorArgument = {
         )
           when func == "==" || func == "!=" =>
         if (List.exists(((_, arg)) => exp_is_wasm_unsafe(arg), args)) {
+          let typeName =
+            switch (args) {
+            | [(_, arg), _] =>
+              switch (resolve_unsafe_type(arg)) {
+              | Some(typeName) => typeName
+              // This should never be hit
+              | None => failwith("Impossible: exp_is_wasm_unsafe > typeName")
```
phated, line 254: You can move this into the resolve_unsafe_type function and avoid working with options

compiler/src/typed/typed_well_formedness.re
```diff
@@ -37,6 +37,23 @@ let rec exp_is_wasm_unsafe = ({exp_type}) => {
...
+    | TTyConstr(path, _, _) =>
+      switch (path) {
+      | t when t == Builtin_types.path_wasmi32 => Some("WasmI32")
+      | t when t == Builtin_types.path_wasmi64 => Some("WasmI64")
+      | t when t == Builtin_types.path_wasmf32 => Some("WasmF32")
+      | t when t == Builtin_types.path_wasmf64 => Some("WasmI64")
+      | _ => None
+      }
+    | TTyLink(t) => type_is_wasm_unsafe(t)
+    | _ => None
```
phated, line 52:
```suggestion
      switch (path) {
      | t when t == Builtin_types.path_wasmi32 => "WasmI32"
      | t when t == Builtin_types.path_wasmi64 => "WasmI64"
      | t when t == Builtin_types.path_wasmf32 => "WasmF32"
      | t when t == Builtin_types.path_wasmf64 => "WasmI64"
      | _ => None
      }
    | TTyLink(t) => type_is_wasm_unsafe(t)
    | _ => failwith("Impossible: type cannot be a wasm unsafe value")
```

compiler/src/typed/typed_well_formedness.re
phated: See above

phated (approved):
LGTM now, let's get @ospencer to sound off too

ospencer (approved):
This looks great! Thanks @spotandjake :)

PR #1893, opened 2023-07-20
fix(lsp): Improve lsp behaviour on goto definition

spotandjake:
The previous behaviour of the lsp with goto definition failed when you selected a variable from `left to right`, we now try and account for this by trying to find again, but one char over if we fail the first check.

### DIFF
--- a/compiler/src/language_server/definition.re
+++ b/compiler/src/language_server/definition.re
@@ -84,7 +84,7 @@ let rec find_definition =
   | None =>
     if (check_position == Forward && position.character > 0) {
       // If a user selects from left to right, their pointer ends up after the identifier
-      // this trys to check if the identifier was selected.
+      // this tries to check if the identifier was selected.
       find_definition(
         ~check_position=Backward,
         sourcetree,
--- a/compiler/src/language_server/definition.re
+++ b/compiler/src/language_server/definition.re
@@ -53,9 +53,12 @@ let send_definition =
     }),
   );
 };
+type check_position =
+  | Forward
+  | Backward;
 let rec find_definition =
         (
-          ~try_last_char=false,
+          ~check_position=Forward,
           sourcetree: Sourcetree.sourcetree,
           position: Protocol.position,
         ) => {
@@ -79,11 +82,11 @@ let rec find_definition =
     };
   switch (result) {
   | None =>
-    if (!try_last_char && position.character > 0) {
+    if (check_position == Forward && position.character > 0) {
       // If a user selects from left to right, their pointer ends up after the identifier
       // this trys to check if the identifier was selected.
       find_definition(
-        ~try_last_char=true,
+        ~check_position=Backward,
         sourcetree,
         {line: position.line, character: position.character - 1},
       );
--- a/compiler/src/language_server/definition.re
+++ b/compiler/src/language_server/definition.re
@@ -53,10 +53,9 @@ let send_definition =
     }),
   );
 };
-
 let rec find_definition =
         (
-          ~depth=0,
+          ~try_last_char=false,
           sourcetree: Sourcetree.sourcetree,
           position: Protocol.position,
         ) => {
@@ -80,9 +79,11 @@ let rec find_definition =
     };
   switch (result) {
   | None =>
-    if (depth == 0 && position.character > 0) {
+    if (!try_last_char && position.character > 0) {
+      // If a user selects from left to right, their pointer ends up after the identifier
+      // this trys to check if the identifier was selected.
       find_definition(
-        ~depth=1,
+        ~try_last_char=true,
         sourcetree,
         {line: position.line, character: position.character - 1},
       );
--- a/compiler/src/language_server/definition.re
+++ b/compiler/src/language_server/definition.re
@@ -54,6 +54,45 @@ let send_definition =
   );
 };
 
+let rec find_definition =
+        (
+          ~depth=0,
+          sourcetree: Sourcetree.sourcetree,
+          position: Protocol.position,
+        ) => {
+  let results = Sourcetree.query(position, sourcetree);
+
+  let result =
+    switch (results) {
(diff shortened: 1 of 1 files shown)


marcusroberts (approved):
Well spotted and nice fix!

compiler/src/language_server/definition.re
```diff
@@ -54,37 +54,64 @@ let send_definition =
...
+        Some((loc, uri));
       }
-    | _ => send_no_result(~id)
+    | _ => None
+    };
+  switch (result) {
+  | None =>
+    if (depth == 0 && position.character > 0) {
+      find_definition(
+        ~depth=1,
```
phated, line 84: You are using `~depth` with an integer that is only ever `1` or `0`. That seems feels like it means it should be `bool` but that is probably not clear enough to explain the purpose so maybe you should make variants for the states this function can handle.

phated (changes requested):
Need to clean up the API design on your function

compiler/src/language_server/definition.re
spotandjake: I fixed this by using a better variable name and adding a comment. as I didnt feel that a variant provided enough context as to what we were doing.

phated: Reading your comment helps, but I think this should be variants that wrap the `Position.position` and probably called `Forwards` and `Backwards` or something. Bool flags are almost always bad design.

spotandjake: I switched to a variant, named `Forwards, Backwards`

compiler/src/language_server/definition.re
```diff
@@ -53,38 +53,69 @@ let send_definition =
...
+  switch (result) {
+  | None =>
+    if (check_position == Forward && position.character > 0) {
+      // If a user selects from left to right, their pointer ends up after the identifier
+      // this trys to check if the identifier was selected.
+      find_definition(
+        ~check_position=Backward,
+        sourcetree,
+        {line: position.line, character: position.character - 1},
+      );
```
phated, line 91: I wonder if this is actually just a "bug" in the sourcetree, as such the sourcetree seems to be position exclusive when it probably should be location *inclusive*. Is this a correct assessment?

spotandjake: I don't think so the issue really seems to be from highlighting and where your cursor is. we do not know which way the user is highlighting left to right or right to left. as such we don't know where the cursor will be.

phated: The sourcetree is supposed to provide a correct node, no matter the position of the cursor. Whether the cursor is like: `|foobar` or `foobar|` or `foo|bar` or anything. And it seems that it doesn't.

spotandjake: > The sourcetree is supposed to provide a correct node, no matter the position of the cursor. Whether the cursor is like: `|foobar` or `foobar|` or `foo|bar` or anything. And it seems that it doesn't.

Digging into this a bit more If we want to fix this in `sourceTree` then I think we should just add `1` to `pos_end.charater` in `loc_to_interval`. I still think this is an issue in the lsp spec. mainly because of the case of `foo|+1` with the cursor location being represented by `|` its ambigous if the user is selecting over `foo` from left to right or selecting `+` from right to left.

phated: @ospencer does Jake's thoughts above sound correct here?

ospencer: Jake is correct. The LSP spec doesn't account for this situation pretty much at all—if your cursor is on the edges of the value, it's not overlapping the range.

phated: If the LSP spec doesn't say that it's in the range, then I think we should close this PR.

ospencer: It's up to our interpretation. The editor will send the request with whatever the cursor position is. We don't consider a cursor after something as overlapping that range. That's the correct behavior, but in the case Jake is trying to address—highlighting an identifier from the beginning to the end—your cursor ends up after, and the goto fails.

We have to decide if we care. I think it's reasonable enough to try to improve the user experience because I imagine some users do this. I don't think we should change the sourcetree though, because it's correct.

spotandjake: I agree about not changing the sourcetree. For context as to when this was a problem for me. I was double clicking on an identifier to select it and then clicking goto definition which I feel is common behavior.

phated: > We don't consider a cursor after something as overlapping that range.

If *we don't* consider the cursor in the range, then **one** of our tools shouldn't either.

spotandjake: I agree with the statement and when considering the cursor position I agree with the viewpoint. The issue in this case isn't the cursor position though it is the selection range. And the fact that when you double click on an identifier to select it, it puts the cursor at the end. Which seems like reasonable user behaviour.

compiler/src/language_server/definition.re
```diff
@@ -53,38 +53,69 @@ let send_definition =
...
+        Some((loc, uri));
       }
-    | _ => send_no_result(~id)
+    | _ => None
+    };
+  switch (result) {
+  | None =>
+    if (check_position == Forward && position.character > 0) {
+      // If a user selects from left to right, their pointer ends up after the identifier
+      // this trys to check if the identifier was selected.
```
ospencer, line 86:
```suggestion
      // this tries to check if the identifier was selected.
```

ospencer (commented):
While I believe that from a technical standpoint our current behavior is correct, from a user experience perspective, this change makes sense to me and is the right way forward.

I have one small nit, but otherwise this LGTM.

ospencer (approved):

alex-snezhko (approved):
This change makes sense. As we eventually implement other "go to ..." functionality we should also carry this logic forward. We can take an opportunity to generalize this behavior when that time comes though

PR #1895, opened 2023-07-20
feat(lsp): Support goto definition on exceptions

spotandjake:
Adds support for goto definition on `exceptions`.


marcusroberts (approved):
Nice one!

compiler/src/language_server/sourcetree.re
```diff
@@ -428,7 +442,10 @@ module Sourcetree: Sourcetree = {
             | Some(path) =>
               let decl = Env.find_type(path, pat.pat_env);
               if (decl.type_loc == Location.dummy_loc) {
-                None;
+                switch (pat.pat_desc) {
+                | TPatConstruct(_, desc, _) => Some(desc.cstr_loc)
+                | _ => None
+                };
```
phated, line 448: Why is this only done when the `type_loc` is a dummy_loc?

phated (changes requested):
One question

compiler/src/language_server/sourcetree.re
spotandjake: I think it was just a backup, for if it had a dummy loc because I was running into issues.

phated: That doesn't help. I want to know why there was a dummy_loc... what causes this? does a larger fix need to be investigated? Is it unrelated to the PR?

spotandjake: This wasnt neccessairy I think i was just being precautious when i added that. so i removed it

compiler/src/language_server/sourcetree.re
```diff
@@ -427,11 +441,7 @@ module Sourcetree: Sourcetree = {
             switch (get_type_path(pat.pat_type)) {
             | Some(path) =>
               let decl = Env.find_type(path, pat.pat_env);
-              if (decl.type_loc == Location.dummy_loc) {
-                None;
-              } else {
-                Some(decl.type_loc);
-              };
+              Some(decl.type_loc);
```
phated, line 444: We need feedback from @ospencer because I'm pretty sure this had some purpose that you are missing.

ospencer: Yeah, I'd likely put this back. It's an option just to be specific about if there's a location or not. Changing this back shouldn't affect anything.

spotandjake: I reapplied that code

compiler/src/language_server/sourcetree.re
phated, line 448: You reverted incorrectly. We want it to be original code.
```suggestion
                None
```

phated (approved):
Tried it in the stdlib and it jumps correctly 🎉

Issue #1894, opened 2023-07-20
Oprint NYI error on exceptions 


spotandjake:
![image](https://github.com/grain-lang/grain/assets/40705786/4f4fe40d-b71d-475b-b72e-c092373777de)
In the lsp for hover implementations when you have exceptions Oprint gives an `NYI` error for the open type.

PR #1896, opened 2023-07-21
feat: Add `exception` opentype implementation to `Oprint`

spotandjake:
Fixes: #1894

Technically we may want to use another prefix compared to `exception` but given they are the only open type in grain. this should be good for now.

### DIFF
--- a/compiler/src/typed/oprint.re
+++ b/compiler/src/typed/oprint.re
@@ -666,7 +666,7 @@ and print_out_sig_item = ppf =>
         | Otyp_sum(_) => "enum"
         | Otyp_variant(_, _, _, _) =>
           failwith("NYI: Otyp_variant pretty-printer")
-        | Otyp_open => failwith("NYI: Otyp_open pretty-printer")
+        | Otyp_open => "exception"
         | Otyp_alias(_, _) => failwith("NYI: Otyp_alias pretty-printer")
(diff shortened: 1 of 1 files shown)


jozanza (approved):

peblair (commented):
Code changes look good here, but I am missing the necessary context to know whether `exception` is what we want to print here. I'll let others take a peek at it before I approve.

compiler/src/typed/oprint.re
```diff
@@ -767,7 +767,7 @@ and print_out_type_decl = (kwd, ppf, td) => {
         print_list(print_out_constr, ppf => fprintf(ppf, ",@;<0 2>")),
         constrs,
       )
-    | Otyp_open => fprintf(ppf, " = ..")
+    | Otyp_open => fprintf(ppf, " = ...")
```
phated, line 770: Why did you change this to `...` instead of `..`?

phated (changes requested):
Exception is our only open type, so that's correct. I'm not sure why you changed `..` to `...`

compiler/src/typed/oprint.re
spotandjake: I think that was a mistake. Its been corrected

phated (approved):

PR #1897, opened 2023-07-22
fix(compiler): Allow providing values of types provided later

### LINKED ISSUE #1884: Lang: allow providing values of types provided by `provide { ... }` statements
When compiling the following code:
```
module Main

enum T {
  A
}
let a = A
provide { type T }
provide { a }
// provide { type T, a } should probably be allowed as well
```
It fails to compile with "This value is provided but contains type T, which is neither provided nor abstract". This scenario only works if you do `provide enum T { ... }`, but it should probably be made to work if you `provide` later on in the code as well

alex-snezhko:
Closes #1884

### DIFF
--- a/compiler/test/suites/provides.re
+++ b/compiler/test/suites/provides.re
@@ -189,4 +189,19 @@ describe("provides", ({test, testSkip}) => {
       ("GRAIN$EXPORT$y", Binaryen.Export.external_global),
     ],
   );
+  assertHasWasmExport(
+    "issue_1884_type_provided_later1",
+    "module Test; enum T { A }; let a = A; provide { type T }; provide { a }",
+    ("GRAIN$EXPORT$a", Binaryen.Export.external_global),
+  );
+  assertHasWasmExport(
+    "issue_1884_type_provided_later2",
+    "module Test; enum T { A }; let a = A; provide { a, type T }",
+    ("GRAIN$EXPORT$a", Binaryen.Export.external_global),
+  );
+  assertHasWasmExport(
+    "issue_1884_type_provided_later3",
+    "module Test; enum T { A }; let a = A; provide { a }; provide { type T }",
+    ("GRAIN$EXPORT$a", Binaryen.Export.external_global),
+  );
 });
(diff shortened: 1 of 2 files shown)


compiler/test/suites/provides.re
```diff
@@ -181,4 +181,19 @@ describe("provides", ({test, testSkip}) => {
...
+  assertHasWasmExport(
+    "issue_1884_type_provided_later2",
+    "module Test; enum T { A }; let a = A; provide { a, type T }",
+    ("GRAIN$EXPORT$a", Binaryen.Export.external_global),
+  );
+  assertHasWasmExport(
+    "issue_1884_type_provided_later3",
+    "module Test; enum T { A }; let a = A; provide { a }; provide { type T }",
+    ("GRAIN$EXPORT$a", Binaryen.Export.external_global),
+  );
```
alex-snezhko, line 198: This is an unusual-looking case at first but still makes sense to allow imo

ospencer (approved):

compiler/test/suites/provides.re
spotandjake: Im not a major fan of allowing this. but I dont have a major problem with it either.

spotandjake (approved):
Code looks good to me.

compiler/test/suites/provides.re
phated: This should 100% be allowed. Thanks for adding a test for it 🤗

phated, line 198:
```suggestion
  );
   assertHasWasmExport(
    "issue_1884_type_provided_later3",
    "module Test; enum T { A }; provide let a = A; provide { type T }",
    ("GRAIN$EXPORT$a", Binaryen.Export.external_global),
  );
```
This is likely a more common way that you'd provide a value and the type after. I assume it works, but let's add a test to be certain

phated (changes requested):
One test I'd like to add

ospencer (approved):

phated (approved):

PR #1900, opened 2023-07-23
feat(lsp): Add autocomplete

spotandjake:
This pr adds autocomplete to the grain language. I think we may want to expand this in the future. Possibly with a better parser to improve finding the current context of completion. Right now we just treat `=` as indicating you want value completions and `:` as indicating you want type completions There are a few issues with this approach though, such as ew do not currently trigger completions `1 + <here>` in the position of `<here>` and other similar cases. we also mistreat `type T =` as value completions currently, so some more consideration may need to be given on parsing rules. Im opening the pr now though as I am hoping to get feedback on parsing rules here.

Another possible addition would be to try and use the `sourceTree` to give context, I didnt take this approach as in most cases `sourceTree` is very far behind where your autocompleting and not very helpful, it would have to be used in addition to the half parser though rather than as an alternative.

Another possible future expansion is adding autocompletes to `include`s that scan the directory for `.gr` files and suggest them. 

This pr replaces: #1262


marcusroberts (approved):
Some excellent work here @spotandjake  and so fantastic to get this into the LSP!

ospencer:
I think we need to go ahead and handle the cases for completion a little better—going ahead and looking for the `let ... =` for value completions and `type ... =` for type completions. We can fall back to just `=` but we should look for the `let` first.

spotandjake:
I made that change and improved the way detection is done.

compiler/src/language_server/doc.re
phated, line 1: Should we name this file `document.re`? I don't want to get it confused with `doc.re` in the formatter.

spotandjake: Im a fan of that change

spotandjake: Made that change

compiler/src/language_server/document.re
```diff
@@ -0,0 +1,68 @@
...
+  // Horizonal rules between code blocks render a little funky
+  // so we manually add linebreaks
+  Printf.sprintf(
+    "%s\n---\n<br><br>\n%s",
+    a,
+    b,
+  );
+};
+
+let supressed_types = [Builtin_types.path_void, Builtin_types.path_bool];
```
ospencer, line 24: Why is this not all builtin types?

compiler/src/language_server/completion.re
```diff
@@ -0,0 +1,547 @@
...
+  );
+};
+let find_completable_state = (documents, uri, position: Protocol.position) => {
+  // try and find the code we are completing in the original source
+  switch (Hashtbl.find_opt(documents, uri)) {
+  | None => None
+  | Some(source_code) =>
+    // Get Document
+    let source =
+      Str.global_replace(Str.regexp({|\r\n|}), {|\n|}, source_code);
```
ospencer, line 139: Is this necessary? This is extremely expensive.

compiler/src/language_server/completion.re
```diff
@@ -0,0 +1,547 @@
...
+    let offset = convert_position_to_offset(source, position);
+    let source_chars =
+      List.rev(
+        String_utils.explode(String_utils.slice(~last=offset, source)),
+      );
```
ospencer, line 144: This is really expensive too. I imagine it's cheaper just to check string prefixes directly.

compiler/src/language_server/completion.re
```diff
@@ -0,0 +1,547 @@
...
+
+let rec get_completions =
+        (
+          ~include_values,
+          ~include_types,
+          env,
+          root_decl: Types.module_declaration,
+          path: list(string),
+        ) => {
+  // Get The Modules Exports
```
ospencer, line 272:
```suggestion
  // Get the module's exports
```

compiler/src/language_server/completion.re
```diff
@@ -0,0 +1,547 @@
...
+        (s: Types.signature_item) => {
+          switch (s) {
+          // Enabled
+          | TSigValue(ident, decl) when include_values =>
+            Some(build_value_completion(~env, (ident.name, decl)))
+          | TSigType(ident, decl, _) when include_types =>
+            Some(build_type_completion(~env, (ident.name, decl)))
+          | TSigModule(ident, decl, _) =>
+            Some(build_module_completion(~env, (ident.name, decl)))
+          // Dissabled
```
ospencer, line 286:
```suggestion
          // Disabled
```

compiler/src/language_server/completion.re
```diff
@@ -0,0 +1,547 @@
...
+      ~include_values,
+      ~include_types,
+      program: Typedtree.typed_program,
+      compiled_code: Hashtbl.t(Protocol.uri, Lsp_types.code),
+      search: list(string),
+    ) => {
+  // Keyword Completions
+  let keyword_completions =
+    if (include_keywords) {
+      let keywords = [
```
ospencer, line 336: It'd be good to have this list moved globally so it's easier to maintain.

compiler/src/language_server/completion.re
```diff
@@ -0,0 +1,547 @@
...
+    List.map(build_module_completion(~env=program.env), modules);
+  // Merge Completions
+  let completions =
+    List.concat([
+      keyword_completions,
+      value_completions,
+      type_completions,
+      module_completions,
+    ]);
+  // Find The Top Level Modules
```
ospencer, line 436:
```suggestion
  // Find all top-level modules
```

compiler/src/language_server/completion.re
```diff
@@ -0,0 +1,547 @@
...
+  let completions =
+    switch (search) {
+    // User Wrote Nothing
```
ospencer, line 439:
```suggestion
    // User wrote nothing
```

compiler/src/language_server/completion.re
```diff
@@ -0,0 +1,547 @@
...
+    | []
+    // User Wrote A Single Word, i.e Top Level Search
```
ospencer, line 441:
```suggestion
    // User wrote a single word, i.e top-level search
```

compiler/src/language_server/completion.re
```diff
@@ -0,0 +1,547 @@
...
+    | [_] => completions
+    // User Wrote A Path
```
ospencer, line 443:
```suggestion
    // User wrote a path
```

compiler/src/language_server/completion.re
```diff
@@ -0,0 +1,547 @@
...
+    | [search, ...path] =>
+      // Find Module Root
```
ospencer, line 445:
```suggestion
      // Find module root
```

compiler/src/language_server/completion.re
```diff
@@ -0,0 +1,547 @@
...
+          root_decl,
+          path,
+          ~include_values,
+          ~include_types,
+        )
+      | None => []
+      }
+    };
+  // Remove Operators
+  let operators = [
```
ospencer, line 459: Would also be good to have this at the top level.

ospencer (changes requested):
This looks pretty reasonable, but we'll need to do some work on how we find completeable states.

spotandjake:
@ospencer if you get a chance would you be willing to review the new implementation. It needs more work before its good so im going to convert this to a draft.

ospencer (commented):
In general the new approach seems really smart!

spotandjake:
mentioned in the grain meeting, but if anyone wants to take over this pr feel free ive gotten a little burnt out from it,My current approach was to run the lexer and get the lexing token and then try to recover the current state from the cursor to get some context for what were working on from there I was using the cached program to get varaibles and stuff this could be combined with sourcetree to get additonal information such as some program scope. 

Things I was considering: I was trying to leave room in my implementation so that we could look into recovering some parse state from the parser in the future this could be combined with the lexing approach to get a more accurate state of whats going on, addittionally I was trying to make sure adding new states didnt require touching the parsing logic instead it just required defining the state, handling the parsing seperatly from our completion states also would let us fall back to a simple parser like what reasonml does in their autocomplete engine for cases where we lose complete context.

If anyone decides to take this up feel free to send me a message here or on discord and I give a better explanation of the technique I was trying to use if needed.

PR #1910, opened 2023-09-12
chore(compiler): Upgrade to binaryen 112

phated:
This updates to binaryen.ml 0.22 which are our bindings for the binaryen 112 release


ospencer (approved):
Excellent!

Issue #1912, opened 2023-09-14
Separate wasm compilation in CLI runner and add specific error message


ShalokShalom:
`grain compiler file` actually runs fine on this, and doesn't complain. 

Running it, then, produces some of the most confusing error messages I have seen yet. 🗡️ 

![Screenshot_20230914_234856](https://github.com/grain-lang/grain/assets/6344099/c1e2a9eb-0d49-48c7-98c8-23cb29867b31)

phated:
A grain file is not a wasm file. This is a WebAssembly error because you passed a non-wasm module to the WebAssembly constructor.

phated:
Re-opening because I think we can split the wasm instantiation into a separate try/catch and then say something like "Invalid WebAssembly module. Make sure you are running a `.wasm` file"

ShalokShalom:
o.O

ShalokShalom:
Let me get this straight. You are providing me with exactly one way to run your code, and that is via wasm. The pure compilation from grain, as said, yielded no error. 

I expect this to work, just as I did. As in any other language. This sounds like the guts of the languages are exposed to the user, and that one is currently expected to understand the inner workings, just to get the info on a super simple function, and why it failed. 

Like...

ospencer:
`grain run` is a command that runs wasm files, not Grain programs. The command that both compiles and then runs the resulting wasm from a Grain program is just the top level `grain` command, `grain test.gr`.

It's a little different than other languages because interpreted languages like JavaScript (running in node.js) don't have a compile step—you just run the JavaScript. Compiled languages typically produce an executable that you run yourself. Since we compile to WebAssembly, which can't be run natively on your machine, we also provide a utility to run wasm files.

phated:
The `grain run` command clearly states it takes a wasm module:
```
$ grain run --help
Usage: grain run [options] <file>

run a wasm file via grain's WASI runner
```

While the `grain` command says it compiles and runs a program:

```
$ grain --help
Usage: grain [options] <file>

Compile and run Grain programs. 🌾
```

These are not the same and `grain run` can run any wasm file, not just wasm output by grain.

ShalokShalom:
I would expect `grain run file` to run a grain file. 

Both due to the name of it, and because its like this in other languages. 
But just `grain file` compiles, seemingly successful and gives me no output: 

[...]

PR #1913, opened 2023-09-18
feat(cli): Improve error handling around `grain run`

spotandjake:
This improves the error handling arround loading wasm files with `grain run`.

Closes: #1912


cli/bin/grainrun.js
```diff
@@ -31,7 +31,18 @@ const importObject = { wasi_snapshot_preview1: wasi.wasiImport };
...
+        if (filename.endsWith(".gr")) {
+          console.log(
+            `Grain run is for compiled wasm files, to compile a grain file use grain ${filename}`
+          );
+        } else {
+          console.log("Invalid wasm file detected");
+        }
+        process.exit();
+      }
+    );
```
phated, line 45: I think I want to use try/catch with the awaits instead of `.catch` so we need to restructure this a big.

spotandjake: I made that change if we want as well we could make a prettier error for a file that doesn't exist. and we can try and make runtime errors a little prettier might be a job for a different pr though.

phated: I've updated further

cli/bin/grainrun.js
```diff
@@ -30,12 +30,37 @@ const wasi = new WASI({
...
+  } catch (err) {
+    if (filename.endsWith(".gr")) {
+      console.error(
+        `The \`grain run\` command is used on compiled \`.wasm\` files.`
+      );
+      console.error(
+        `To compile and run your \`.gr\` file, use \`grain ${filename}\``
+      );
+    } else {
+      console.error(`Unable to compile WASM module.`);
```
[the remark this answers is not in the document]
ospencer: The appropriate shorthand is `Wasm`, though I think we can just write it out here.

ospencer (approved):

PR #1914, opened 2023-09-18
feat(runtime): Allow modulo on floating point numbers

spotandjake:
This pr adds floating point modulo support closing #1371

### DIFF
--- a/stdlib/runtime/numbers.gr
+++ b/stdlib/runtime/numbers.gr
@@ -1585,6 +1585,7 @@ let i64abs = x => {
   from WasmI64 use { (-), (>=) }
   if (x >= 0N) x else 0N - x
 }
+
 @unsafe
 let numberMod = (x, y) => {
   from WasmI64 use { (!=), (-), (*), (<), (>) }
@@ -1596,7 +1597,7 @@ let numberMod = (x, y) => {
     let xval = coerceNumberToWasmF64(WasmI32.toGrain(x): Number)
     let yval = coerceNumberToWasmF64(WasmI32.toGrain(y): Number)
     let yInfinite = yval == InfinityW || yval == -InfinityW
-    let out = if (
+    if (
       yval == 0.0W || yInfinite && (xval == InfinityW || xval == -InfinityW)
     ) {
       newFloat64(NaNW)
@@ -1605,7 +1606,6 @@ let numberMod = (x, y) => {
     } else {
       newFloat64(xval - WasmF64.trunc(xval / yval) * yval)
     }
-    out
   } else {
     let xval = coerceNumberToWasmI64(WasmI32.toGrain(x): Number)
     let yval = coerceNumberToWasmI64(WasmI32.toGrain(y): Number)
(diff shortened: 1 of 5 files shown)


stdlib/runtime/numbers.gr
```diff
@@ -1585,29 +1585,46 @@ let i64abs = x => {
...
+    ) {
+      newFloat64(NaNW)
+    } else if (yInfinite) {
+      newFloat64(xval)
     } else {
-      modval
+      newFloat64(xval - WasmF64.trunc(xval / yval) * yval)
     }
-    reducedInteger(result)
+    Memory.incRef(out)
```
peblair, line 1608: Do we need this `incRef`?

spotandjake: When i didnt have it i was getting `decRef` errors. so i am pretty sure we need it.

peblair: @ospencer Does that make sense to you? I'm a bit confused, since it looks like `reducedInteger` doesn't require an `incRef`. It's been a minute since I grokked this, though.

ospencer: No, this shouldn't be necessary. If there are decRef errors, the bug is somewhere else, possibly where this function is being called.

spotandjake: Would someone who understands how the gc works a bit better be willing to take a look into why there is a `decRef` error when this doesnt exist?

phated: @ospencer can you please help Jake debug this so we can land it?

ospencer: Way ahead of you

ospencer:
@spotandjake #1993 was the major reason you were seeing the DecRefError. While debugging, I noticed some places we're a bit sloppy when we need to hold on to references, but I'll address those separately.

stdlib/runtime/numbers.gr
```diff
@@ -1585,29 +1585,45 @@ let i64abs = x => {
   from WasmI64 use { (-), (>=) }
   if (x >= 0N) x else 0N - x
 }
-
```
ospencer, line 1588: I liked this space!

stdlib/runtime/numbers.gr
```diff
@@ -1585,29 +1585,45 @@ let i64abs = x => {
...
-  if (xval < 0N && yval > 0N || xval > 0N && yval < 0N) {
-    let modval = WasmI64.remS(i64abs(xval), i64abs(yval))
-    let result = if (modval != 0N) {
-      i64abs(yval) - modval * (if (yval < 0N) -1N else 1N)
+  if (isFloat(x) || isFloat(y) || isRational(x) || isRational(y)) {
+    from WasmF64 use { (==), (/), (*), (-) }
+    let xval = coerceNumberToWasmF64(WasmI32.toGrain(x): Number)
+    let yval = coerceNumberToWasmF64(WasmI32.toGrain(y): Number)
+    let yInfinite = yval == InfinityW || yval == -InfinityW
+    let out = if (
```
ospencer, line 1599: Don't need this  binding anymore

spotandjake: Fixed that.

ospencer (approved):

PR #1922, opened 2023-09-27
chore(compiler): Build JS version with 2 jobs on Windows

phated:
Trying to build our JS version with 2 jobs on Windows in the hopes we stop OoM flakes.


phated:
It's green. We can probably try this out.

ospencer (approved):

PR #1923, opened 2023-10-09
feat: Generate tree-sitter grammar from parser

phated:
This is my initial experimentation to figure out if we can use `MenhirSdk` and the `cmly` file it can generate to automatically output a tree-sitter grammar.

It's a little annoying because you need to specify the RegExp pattern for each token in the parser, as that logic is actually done by sedlex_ppx in the lexer, but tree-sitter generates a combined lexer/parser given a JavaScript RegExp 😦.

If we can actually get this working, we could possibly throw away all our formatter code and just rely on [Topiary](https://github.com/tweag/topiary).

### DIFF
--- a/compiler/esy.json
+++ b/compiler/esy.json
@@ -3,7 +3,7 @@
   "version": "0.5.13",
   "esy": {
     "build": [
-      "dune build @native --no-buffer"
+      "dune build @native @install --no-buffer"
     ],
     "buildEnv": {
       "DUNE_BUILD_DIR": "#{self.target_dir}",
--- a/compiler/src/dune
+++ b/compiler/src/dune
@@ -1,11 +1,31 @@
 (library
  (name grain)
  (public_name grain)
+ (modules compile)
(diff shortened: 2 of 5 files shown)


spotandjake:
I was looking into this a bit more, and there are two problems with this it seems as though it is really hard to get the input grammar from the cmly file instead we get something more akin to the built tables which are a lot messier this isn't a major road block but certainly makes generation a lot harder. A more pressing issue is that when it comes to syntax highlighting the grammar we currently have isn't really optimal. Additionally dealing with precedence during generation seems to be rather challanging with this approach. Some of my work on this can be found [here](https://github.com/spotandjake/grain/tree/spotandjake-tree-sitter) if it is of interest. 

Interestingly a better approach here might to be follow [Obelisk](https://github.com/Lelio-Brun/Obelisk/blob/master/src/parser.mly) and parse the menhir syntax ourself and generate the corresponding tree sitter syntax, we could probably use the parser from obelisk for this.

PR #1909, opened 2023-09-04
update jsoo for testing

phated:
opening this draft for windows CI testing purposes

PR #1926, opened 2023-10-30
chore(compiler): Avoid round trip between string & bytes when reading cmi

phated:
This pulls @hhugo's change that we included in #1909 into a separate commit so we can land the improvement without the JSOO upgrade while we figure out the performance degradation.


spotandjake (commented):
looks like this needs the formatter to be run. otherwise looks great.

ospencer (approved):

PR #1930, opened 2023-11-30
feat(stdlib): Add `toArray` to `Queue`

spotandjake:
Adds `Queue.toArray` to the `Queue` module.


alex-snezhko:
I think we should make the equality function `(==)` and not equal

jozanza (approved):

ospencer (approved):

phated (approved):

PR #1931, opened 2023-12-04
feat(stdlib): Add `==` operator to `Queue` module

spotandjake:
This pr adds `Queue.equal` which provides a more useful equality check that cares about the `queue` contents rather than internal structure.


ospencer:
I agree with @alex-snezhko that we should name it `(==)`.

spotandjake:
> I agree with @alex-snezhko that we should name it `(==)`.

Made that change

stdlib/queue.gr
```diff
@@ -164,6 +164,33 @@ provide let copy = queue => {
   { size, array: Array.copy(array), headIndex, tailIndex }
 }
 
+/**
+ * Checks if two queues are equivalent by value.
+ *
+ * @param queue1: The first queue to compare
+ * @param queue2: The second queue to compare
+ *
```
phated, line 172:
```suggestion
```

spotandjake:
Applied that suggestion

phated:
@spotandjake please resolve the conflicts 🙏

phated (approved):

phated:
@spotandjake it looks like this conflicts again. Can you update?

spotandjake:
Updated

PR #1933, opened 2023-12-04
chore(stdlib): Optimize `Buffer.addChar`

spotandjake:
This optimizes `Buffer.addChar` which previously relied on `Char.toString`. This new method eliminates the need for a needless string allocation on the heap.

Closes: #1879


jozanza (approved):
Nice!

spotandjake:
> Is there any utility we have that could possibly be shared here?



> Is there any utility we have that could possibly be shared here?

I dont really feel like it, I used the source from `Char.toString`, but I dont know if it can be shared with while avoiding the extra heap allocations, that this optimizes away in the first place.

phated (approved):

Issue #1106, opened 2022-01-17
Compiler: Ast_helper locations should not be optional


ospencer:
It's too easy to accidentally forget to pass locations to parsetree nodes because they're optional.

(They're currently optional to make writing tests a little easier, but we no longer feel it's worth it.)

tyrone-wu:
I can take this up if no one is working on this? ʕ•́ᴥ•̀ʔっ

ospencer:
Go for it!

PR #1934, opened 2023-12-07
chore(compiler): Make Ast_helper locs non optional

spotandjake:
This pr makes locatons on ast nodes not optional.

Question should we also refactor the anf_helper nodes to have non optional locs in this pr or should we do that in a seperate pr.

Closes: #1106


ospencer (approved):
This looks great! I agree that we should do anf_helper in another PR.

phated (approved):
Awesome work. Thanks!

PR #1935, opened 2023-12-07
feat(stdlib)!: Replace `Int64` arithmatic functions with operators

spotandjake:
This pr brings the `Int64` library inline with the changes made in #1742 and #1734 

Marked as breaking beecause the external api changes.


phated (approved):
Nice!

PR #1942, opened 2023-12-30
chore(compiler): Make `locs` on `anf_helper` non optional

spotandjake:
This pr makes locations non optional on `anf_helper`.

I think after this is merged we might want to open an issue to go through and evaluate the use of `Location.dummy_loc` throughout the compiler.

### DIFF
--- a/compiler/src/middle_end/anf_helper.re
+++ b/compiler/src/middle_end/anf_helper.re
@@ -9,223 +9,193 @@ type env = Env.t;
 type ident = Ident.t;
 type attributes = Typedtree.attributes;
 
-let default_loc = Location.dummy_loc;
 let default_env = Env.empty;
 let default_attributes = [];
 let default_allocation_type = Managed;
 
-let or_default_loc = Option.value(~default=default_loc);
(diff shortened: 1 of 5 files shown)


phated (dismissed):
Thanks for doing this @spotandjake! I wonder if it is actually better to use this diff as a way for @ospencer and @peblair to inventory the usage of `dummy_loc` 🤔

spotandjake:
> Thanks for doing this @spotandjake! I wonder if it is actually better to use this diff as a way for @ospencer and @peblair to inventory the usage of `dummy_loc` 🤔 

There are other cases of it being used in the compiler that I think might be worth inventorying as well.

alex-snezhko (approved):

ospencer (approved):
Did inventory of the usage of `dummy_loc` here. Caught a couple, though by nature of what's happening in the linearization/matchcomp phase we're generating a number of expressions that don't have corresponding locations in the source tree. In the future we may decide to handle those differently (maybe just trying to attribute it to the closest node or something).

Issue #787, opened 2021-08-05
Graindoc: Raise exception on multiple attributes that only allow one


phated:
Currently, graindoc grabs the first attribute for cases like `@module` or `@since` but it should actually fail if there are more than 1 of those attributes.

_Originally posted by @ospencer in https://github.com/grain-lang/grain/pull/785#discussion_r683460671_

PR #1946, opened 2024-01-02
feat(graindoc)!: Allow `@since` and `@returns` once per export

spotandjake:
This pr makes it so that you can only have one occurance of `@since` and `@returns` per provided export. 

Closes: #787


alex-snezhko (commented):
A test would be nice

spotandjake:
Added tests for those.

alex-snezhko (approved):

phated (approved):

Issue #728, opened 2021-06-11
Rewrite List.join and Array.join to use low-level instructions to do joins faster


phated:
Ref https://github.com/grain-lang/grain/pull/722#pullrequestreview-681373223

Ask @ospencer for guidance on the actual implementation.

PR #1948, opened 2024-01-02
chore(stdlib): Optimize `Array.join`

spotandjake:
This pr optimizes `Array.join` using low level instructions.

Work for: #728

### DIFF
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -889,17 +889,30 @@ provide let unzip = array => {
  *
  * @since v0.4.0
  */
+@unsafe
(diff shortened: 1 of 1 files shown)


alex-snezhko (approved):
Looks good, do you want to also apply this optimization for `List` in this PR? The code should be mostly the same

spotandjake:
> Looks good, do you want to also apply this optimization for `List` in this PR? The code should be mostly the same

I actually wanna do that work in a seperate pr as I thinkthe implementation will be decently different. similar steps but, I think it needs some related to iterating the lists efficenctly.

phated (approved):

PR #1950, opened 2024-01-02
fix(stdlib): Properly handle extremely large integer bases in `Number.(**)`

spotandjake:
This pr cleans up a tiny bit in `Number.pow` it also fixes a small edge case mistake, and adds a regression test.

edit:
The bug was introduced during the refactor to use early return, for extremely high powers done on floating point bases we were previously returning the `z` variable which is an internal variable rather than output which caused the awnsers to be incorrect.


phated:
@spotandjake can you please update the PR title to be a `fix(stdlib)` and explain the bug you fixed. Bug fixes **never** get hidden in a `chore`

spotandjake:
> @spotandjake can you please update the PR title to be a `fix(stdlib)` and explain the bug you fixed. Bug fixes **never** get hidden in a `chore`

made that change

alex-snezhko (approved):

phated (approved):

PR #1951, opened 2024-01-02
chore(stdlib): Add examples to `Char` module

spotandjake:
This pr improves the docs for `Char` by adding examples. It also switches the `semvar` format to use the `v` as the other stdlibs do.

### DIFF
--- a/stdlib/char.gr
+++ b/stdlib/char.gr
@@ -5,7 +5,12 @@
  *
  * @example include "char"
  *
- * @since 0.3.0
+ * @example 'a'
+ * @example '1'
+ * @example '🌾'
+ *
+ *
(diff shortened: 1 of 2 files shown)


alex-snezhko (approved):

stdlib/char.gr
```diff
@@ -5,7 +5,12 @@
  *
  * @example include "char"
  *
- * @since 0.3.0
+ * @example 'a'
+ * @example '1'
+ * @example '🌾'
+ *
```
phated, line 11:
```suggestion
```

spotandjake: Made that change

phated (approved):

PR #1952, opened 2024-01-02
feat(stdlib): Add ability to set seed on `Hash.hash`

spotandjake:
This pr adds the ability to set the seed of the hashing function.

I think it might be worth pinning the seed when this is used in `Map` and `Set`, which would make marshalling the values work between two separate runs.

### DIFF
--- a/stdlib/hash.gr
+++ b/stdlib/hash.gr
@@ -4,11 +4,12 @@
  * @example include "hash"
  *
  * @since v0.1.0
+ * @history v0.6.0: Added SeededHash submodule
  */
 
 module Hash
 
-/**
+/*
   This module implements MurmurHash3 for Grain data types.
   https://en.wikipedia.org/wiki/MurmurHash
 */
@@ -41,11 +42,6 @@ include "runtime/bigint" as BI
 include "sys/random"
 include "result"
 
-let seed = {
-  let random = Random.random()
-  Result.unwrap(random)
-}
-
 @unsafe
 let _MAX_HASH_DEPTH = 31n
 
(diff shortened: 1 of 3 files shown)


alex-snezhko (approved):

stdlib/hash.gr
```diff
@@ -240,13 +239,18 @@ let rec hashOne = (val, depth) => {
...
+ * @example Hash.hash(a) == Hash.hash(a)
+ * @example Hash.hash(a, seed=1) == -1015171190
+ *
  * @since v0.1.0
+ * @history v0.6.0: Added ability to set the seed
  */
 @unsafe
-provide let hash = anything => {
-  h = seed
+provide let hash = (anything, seed=seed) => {
```
phated, line 252: I don't think we want to allow setting a seed on each hash. Instead, I think we want to add a submodule with a Hash record that has the seed attached and then create a singleton at the top-level module that passes it to the submodule `hash`. This way people can do `Hash.SeededHash.make()` and then `Hash.SeededHash.hash(hasher, anything)`

phated (changes requested):
Whoops. Submitted that as a comment.

stdlib/hash.gr
spotandjake: I think this would be more elegant if `Hash.SeededHash.make()` returned. Function.

`Hash.SeededHash.make(seed: Number) -> ((val: a) -> Number)` thoughts?

phated: I don't hate it! Let's get some other feedback before doing the work.

alex-snezhko: I like Jake's idea; however in that case I'm not sure it needs to be in a submodule if it's just one function; maybe could just make a `makeSeeded` function?

spotandjake: I am good with that approach.

phated: In the case of the make function, we'd want seed to be defaulted and use that to construct our global hasher, right?

spotandjake: I dont think we should make it defaulted I think the api should look like.
```
module Hash

provide let makeSeeded = (seed) => ((value: a) => Number)
let hash = makeSeeded(currentSeedGenCode)
provide let hash = (value) => hash
```
we could just directly provide `hash` but then I think we miss out on the for good graindoc we currently have. Thoughts?

spotandjake: I don't actually think we can use the approach I suggested. It doesnt look like the grain type system supports it.
image
I seem to be getting an error because it cannot generalize the param type in the hash function.

spotandjake:
I updated this, due to the issue I ran into above I decided to go with the approach @phated had suggested at first.


I also rewrote the hashing function to avoid a mutable global. while this wasn't necessary I think it reduces the surface area for bugs to pop up.

phated:
This should be done in a non-breaking way so I'm updating the title

stdlib/hash.gr
ospencer: Yeah, our type system can't really support that right now.

I don't love the idea of the top module and the seeded module having such different APIs. I think the `SeededHash` module would make more sense if we had functors, e.g. `SeededHash.Make()` which would give you just a module with a `hash` function on it.

It'd be breaking, but I'd like to see that same `hash(hasher, anything)` on both versions, with `make` and `makeSeeded` functions to make the hasher.

spotandjake:
Closing in favour of #2170

PR #1954, opened 2024-01-02
feat(stdlib)!: Replace `Float32` arithmatic/comparison functions with operators

spotandjake:
This pr switches `Float32` from using functions to operators


phated (approved):
`NaNf` and `Infinityf` are hilarious.

We should remember to remove any `let NaN` and `let Infinity` bindings throughout the stdlib.

phated:
@spotandjake please remember to flag breaking changes as breaking. I almost missed this one.

PR #1955, opened 2024-01-02
chore(stdlib): Simplify `Float32.Infinity` and `Float32.NaN` definitions

spotandjake:
Repurposed this pr to remove `Float32.Infinity` and `Float32.NaN` in favour of the `Infinityf` and `NaNf` syntax

### DIFF
--- a/stdlib/float32.gr
+++ b/stdlib/float32.gr
@@ -29,12 +29,7 @@ let _VALUE_OFFSET = 4n
  * @since v0.4.0
  */
 @unsafe
-provide let infinity = {
-  let ptr = newFloat32(
-    WasmF32.reinterpretI32(0b01111111100000000000000000000000n)
-  )
-  WasmI32.toGrain(ptr): Float32
-}
+provide let infinity = Infinityf
 
 /**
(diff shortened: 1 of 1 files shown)


phated (changes requested):
Shall we just remove these or do we like `Float32.infinity` instead of `Infinityf`?

compiler/test/stdlib/float32.test.gr
```diff
@@ -6,11 +6,11 @@ from Pervasives use { (-) as numberSub }
 
 // Constants Test
 // smoke test:
-assert infinity > 100000000.0f
+assert Infinityf > 100000000.0f
```
phated, line 9: @ospencer Thoughts?

ospencer: I would prefer to keep the bindings... the syntax is not obvious, and having the binding gets us documentation in the library that is very searchable. If we want, we can mention the syntax in the documentation. We should, of course, remove the old implementations in favor of just `let infinity = Infinityf`.

phated (approved):

PR #1960, opened 2024-01-03
chore(stdlib): Add examples to `Bytes` module

spotandjake:
This pr adds examples to the docs for `Bytes`

### DIFF
--- a/stdlib/bytes.gr
+++ b/stdlib/bytes.gr
@@ -91,7 +91,7 @@ provide let make = (size: Number) => {
 /**
  * An empty byte sequence.
  *
- * @example Bytes.empty() == b""
+ * @example Bytes.empty == b""
  *
  * @since v0.3.2
  */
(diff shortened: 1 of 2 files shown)


stdlib/bytes.gr
```diff
@@ -85,6 +91,8 @@ provide let make = (size: Number) => {
 /**
  * An empty byte sequence.
  *
+ * @example Bytes.empty() == b""
```
alex-snezhko, line 94: No parentheses; this is a value

stdlib/bytes.gr
```diff
@@ -113,6 +123,8 @@ provide let fromString = (string: String) => {
  * @param bytes: The source byte sequence
  * @returns The string representation of the bytes
  *
+ * @example Bytes.toString(b"\x48\x65\x6c\x6c\x6f\x20\x57\x6f\x72\x6c\x64") == "Hello World"
```
alex-snezhko, line 126: Maybe also add the same example but like `Bytes.toString(b"Hello World")` so that it doesn't make people think that the bytes sntax is only usable with hex codes

spotandjake:
Made those changes

phated (approved):

PR #1961, opened 2024-01-03
chore(stdlib): Add examples to `Exception` module

spotandjake:
This pr improves the docs for exceptions, by adding examples. Changes the since attribute to use the `v` pattern used by the rest of the stdlib. and fixes a small spelling mistake.

### DIFF
--- a/stdlib/exception.gr
+++ b/stdlib/exception.gr
@@ -1,11 +1,14 @@
 /**
  * Utilities for working with the Exception type.
  *
- * The Exception type represents an error that has occured during computation.
+ * The Exception type represents an error that has occurred during computation.
  *
  * @example include "exception"
  *
- * @since 0.3.0
+ * @example exception ExampleError(Number)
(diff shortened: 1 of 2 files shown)


stdlib/exception.gr
```diff
@@ -1,11 +1,14 @@
...
  * Utilities for working with the Exception type.
  *
- * The Exception type represents an error that has occured during computation.
+ * The Exception type represents an error that has occurred during computation.
  *
  * @example include "exception"
  *
- * @since 0.3.0
+ * @example exception ExampleError(Number)
+ * @example exception ExampleError
```
alex-snezhko, line 9: Were these two lines a mistake

spotandjake: No I started adding an example of how to use the data type to the top of the module. I think it's a good place to provide the syntax and maybe in the future it could be replaced with a link to documentation on the actual data type.

phated (approved):

PR #1963, opened 2024-01-03
feat(lsp): Support hover on `include` statements

spotandjake:
This pr adds a hover item on `Include` statements, an example can be seen below:
image

Reminder to myself: After this is merged implement goto definition on `include` that will take you to the source module.


alex-snezhko (approved):

phated (approved):
Absolute hero 💪

Issue #982, opened 2021-10-11
Stdlib: Avoid double iteration on String.trim


ospencer:
There are some heuristics we can use to improve performance. See this discussion: https://github.com/grain-lang/grain/pull/951#discussion_r721777742

phated:
Updated the title because the discussion in #951 points out that `trimStart` and `trimEnd` are both called upon the string which causes two iterations. This is different that #1442 which only uses low-level instructions to operate on bytes, but it still does two iterations.

spotandjake:
I could be mistaken but the only logical way I see to use one iteration on `trimString` is to check each side of the string in one loop, This would likely create a bunch of code duplication though and wouldn't greatly reduce the number of instructions. I could be wrong though.

spotandjake:
Thinking over this a bit more. I dont think the optimal solution is to use one loop. I think it would make sense to have the `trimString` function just return the byteIndex of the slice locations. and then we can do a single `newString` allocation after, and single `Memory.copy`.

PR #1965, opened 2024-01-03
chore(stdlib): Optimize `String.trim`

spotandjake:
This optimizes `String.trim` by avoiding the double allocation of a new string after each separate string. Now we only allocate a new string once weve trimmed both sides.

I think this optimization can also close: #982 even though we still do two seperate iterations combing them would likely lead to unmaintainable code with a minimal performance increase.


phated (approved):
I don't have the energy to debate about a single loop 😛

PR #1966, opened 2024-01-14
feat(stdlib): Add user-friendly file system module

### LINKED ISSUE #211: Stdlib: user-friendly filesystem API

alex-snezhko:
Add a file system module that is more high-level and user-friendly than the current `sys/file`

Closes #211


stdlib/sys/fs.gr
```diff
@@ -0,0 +1,1068 @@
...
+include "int64"
+include "string"
+include "runtime/wasi"
+include "runtime/unsafe/wasmi32"
+include "runtime/dataStructures"
+
+/**
+ * Potential errors that can be raised from system interactions.
+ */
+provide enum FileError {
```
spotandjake, line 18: Do we want to pre-emptively add documentation for these, so its done when #992 is merged?

stdlib/sys/fs.gr
```diff
@@ -0,0 +1,1068 @@
...
+    75 => CrossDeviceLink,
+    76 => ExtensionCapabilitiesInsufficient,
+    _ => UnknownFileError(errno),
+  }
+}
+
+/**
+ * Represents non-standard system file types.
+ */
+provide enum SpecialFileType {
```
spotandjake, line 174: Same as above?

stdlib/sys/fs.gr
```diff
@@ -0,0 +1,1068 @@
...
+  CharacterDevice,
+  SocketDatagram,
+  SocketStream,
+}
+
+/**
+ * Represents different system file types.
+ */
+provide enum FileType {
+  File,
```
spotandjake, line 186: Same as above?

stdlib/sys/fs.gr
```diff
@@ -0,0 +1,1068 @@
...
+  SymbolicLink,
+  Special(SpecialFileType),
+}
+
+/**
+ * Represents metadata about a file.
+ */
+provide record Stats {
+  fileType: FileType,
+  size: Number,
```
spotandjake, line 197: can we specify that size is in I assume bytes? idk if we want to document it using the proposed syntax in #1852 or just add the documentation for it to the record doc.

stdlib/path.gr
```diff
@@ -597,7 +597,7 @@ provide let relativeTo = (source, dest) => {
   let pathInfo2 = pathInfo(dest)
   let (base2, _, _) = pathInfo2
   match ((base1, base2)) {
-    (Abs(_), Rel(_)) | (Abs(_), Rel(_)) => Err(Incompatible(DifferentBases)),
+    (Abs(_), Rel(_)) | (Rel(_), Abs(_)) => Err(Incompatible(DifferentBases)),
```
spotandjake, line 600: We might wanna split this fix into a seperate pr and add a regression test.

stdlib/sys/fs.gr
```diff
@@ -0,0 +1,1068 @@
...
+  accessedTimestamp: Number,
+  modifiedTimestamp: Number,
+  changedTimestamp: Number,
+}
+
+provide record DirectoryEntry {
```
spotandjake, line 203: We should probably add a bit of graindoc here even though its rather self explanatory.

spotandjake (changes requested):
The code looks good to me, Most of my comments were related to graindoc. 

Two other things:
* Currently the `sys/` path is used for more low level `wasi` interactions. I think we should either move these wasi interactions elsewhere or put this module elsewhere.
* Are we currently able to add some better streaming functions such as `File.open`, `File.readLine` and `File.close`?

ospencer (commented):
Overall, I love the API design! My major piece of feedback here is that I don't think this module belongs in the `sys` subdirectory, but in the regular `stdlib` folder. The `sys` folder was really just meant for the low-level wasi stuff (and probably should have been named just `wasi`).

I envision that moving forward to Preview 2, the `wasi` folder will just have the generated bindings to each of the wasi standards, and we'll have high-level libraries in the regular standard library. We can note in the module doc that use of that library incurs a dependency on a particular wasi world. For now, we can say that use of the module requires WASI Preview 1.

marcusroberts (approved):
I read all the way through and didn't spot anything, looks like a nice usable interface too.

spotandjake:
> Overall, I love the API design! My major piece of feedback here is that I don't think this module belongs in the `sys` subdirectory, but in the regular `stdlib` folder. The `sys` folder was really just meant for the low-level wasi stuff (and probably should have been named just `wasi`).
> 
> I envision that moving forward to Preview 2, the `wasi` folder will just have the generated bindings to each of the wasi standards, and we'll have high-level libraries in the regular standard library. We can note in the module doc that use of that library incurs a dependency on a particular wasi world. For now, we can say that use of the module requires WASI Preview 1.

Thoughts on maybe in 0.6 we rename the `sys` folder to `wasi` and keep it in the `sys` subdirectory? I do kind of like the separation especially when considering grain for web development or embedded development where you would not have file system access.

stdlib/sys/fs.gr
alex-snezhko: I think most of the values from the types defined are self-explanatory so not sure if needed; I'll add doc comments where things may not be entirely clear though.

stdlib/path.gr
alex-snezhko: https://github.com/grain-lang/grain/pull/2054

ospencer:
> Thoughts on maybe in 0.6 we rename the sys folder to wasi and keep it in the sys subdirectory? I do kind of like the separation especially when considering grain for web development or embedded development where you would not have file system access.

We did make that rename happen. I do see a world in which it makes sense to still have this separated, though in environments where you're not allowed a filesystem or otherwise, WASI makes it pretty easy to virtualize.

ospencer:
@alex-snezhko Do you mind rebasing this?

alex-snezhko:
@ospencer rebased

stdlib/fs.gr
```diff
@@ -0,0 +1,1045 @@
+/**
+ * High-level file system interactions. Utilizes WASI Preview 1 for underlying API
+ * 
+ * @example from "fs" include Fs
+ *
```
spotandjake, line 5: I think it would be beneficial to add some quick usage examples to the top of the library for what will probably be the most commonly used functions, similary to what we do in [The has module](https://grain-lang.org/docs/stdlib/hash)
```suggestion
 * @example Fs.Utf8.readFile(Path.fromString("baz.txt"))
 * @example Fs.Utf8.writeFile(Path.fromString(("baz.txt"), "Hello Grain File's")
```

stdlib/fs.gr
```diff
@@ -0,0 +1,1045 @@
+/**
+ * High-level file system interactions. Utilizes WASI Preview 1 for underlying API
```
spotandjake, line 2: Should this be changed to match the wording of most of the other docs?
```suggestion
 * Utilities for high-level file system interactions. Utilizes WASI Preview 1 for underlying API
```
I noticed most of the time we say `Utilities` for x.

spotandjake (commented):
Everything looks great, I've been using this library for months locally and haven't run into a single problem.

Should we provide a `Utf8.readLine`? in cases with big files having a streamed api seems helpful, maybe this is a future addition when we have scoped resouces so the user doesn't need to worry about `File.open` and `File.close`.

alex-snezhko:
I was thinking we could build out a more comprehensive streaming API in the future, I feel like there are a few different directions we could go

spotandjake (approved):

stdlib/fs.gr
```diff
@@ -0,0 +1,1048 @@
...
+  /** Last accessed timestamp in nanoseconds */
+  accessedTimestamp: Number,
+  /** Last modified timestamp in nanoseconds */
+  modifiedTimestamp: Number,
+  /** Last file status change timestamp in nanoseconds */
+  changedTimestamp: Number,
+}
+
+/**
+ * Represents information about an item in a directory
```
ospencer, line 218:
```suggestion
 * Represents information about an item in a directory.
```

stdlib/fs.gr
```diff
@@ -0,0 +1,1048 @@
...
+ * @param removeMode: The type of removal to perform; `RemoveFile` by default
+ * @returns `Ok(void)` if the operation succeeds, `Err(err)` if a file system error is encountered
+ * 
+ * @example Fs.remove(Path.fromString("file.txt")) // removes a file
+ * @example Fs.remove(Path.fromString("dir"), removeMode=Fs.RemoveEmptyDirectory) // removes an empty directory
+ * @example Fs.remove(Path.fromString("dir"), removeMode=Fs.RemoveRecursive) // removes the directory and its contents
+ * 
+ * @since v0.6.5
+ */
+provide let remove = (baseDirPath=None, path, removeMode=RemoveFile) => {
```
ospencer, line 542: Could we put the path as the last argument?

stdlib/fs.gr
```diff
@@ -0,0 +1,1048 @@
...
+ * Creates a new symbolic link with the given contents.
+ * 
+ * @param linkContents: The path to store into the link
+ * @param targetBaseDirPath: The path to the directory in which the target path resolution starts
+ * @param targetPath: The path to the target of the link
+ * @returns `Ok(void)` if the operation succeeds, `Err(err)` if a file system error or relativization error is encountered
+ * 
+ * @since v0.6.5
+ */
+provide let makeSymlink = (linkContents, targetBaseDirPath=None, targetPath) => {
```
ospencer, line 616: I think this should be called `createSymlink`.

stdlib/fs.gr
```diff
@@ -0,0 +1,1048 @@
...
+ * Queries information about a file.
+ * 
+ * @param baseDirPath: The path to the directory in which the path resolution starts
+ * @param path: The path of the file to query
+ * @param followSymlink: Whether to follow symlinks or not; if `true` then the stats of a valid symlink's underlying file will be returned. `true` by default
+ * @returns `Ok(stats)` containing metadata or `Err(err)` if a file system error is encountered
+ * 
+ * @since v0.6.5
+ */
+provide let stats = (baseDirPath=None, path, followSymlink=true) => {
```
ospencer, line 643: Path last?

stdlib/fs.gr
```diff
@@ -0,0 +1,1048 @@
...
+ * Copies a file or directory.
+ * 
+ * @param sourceBaseDirPath: The path to the directory in which the source path resolution starts
+ * @param sourcePath: The path of the file or directory to copy
+ * @param targetBaseDirPath: The path to the directory in which the target path resolution starts
+ * @param targetPath: The path to copy the file or directory to
+ * @param copyMode: The type of copy to perform; `CopyFile` by default
+ * @returns `Ok(void)` if the operation succeeds, `Err(err)` if a file system error is encountered
+ */
+provide let copy = (
```
ospencer, line 876: Can we put `copyMode` and `followSymlink` to be the first arguments?

stdlib/fs.gr
```diff
@@ -0,0 +1,1048 @@
...
+   * 
+   * @param baseDirPath: The path to the directory to begin path resolution
+   * @param path: The file path to write to
+   * @param data: The bytes to write to the file
+   * @param writeMode: The type of write operation to perform; `Truncate` by default
+   * @returns `Ok(void)` if the operation is successful or `Err(err)` if a file system error is encountered
+   * 
+   * @since v0.6.5
+   */
+  provide let writeFile = (baseDirPath=None, path, data, writeMode=Truncate) => {
```
ospencer, line 989: Can we move `writeMode` so path and data are the final arguments?

stdlib/fs.gr
```diff
@@ -0,0 +1,1048 @@
...
+   * 
+   * @param baseDirPath: The path to the directory to begin path resolution
+   * @param path: The file path to write to
+   * @param data: The string to write to the file
+   * @param writeMode: The type of write operation to perform; `Truncate` by default
+   * @returns `Ok(void)` if the operation is successful or `Err(err)` if a file system error is encountered
+   * 
+   * @since v0.6.5
+   */
+  provide let writeFile = (baseDirPath=None, path, data, writeMode=Truncate) => {
```
ospencer, line 1044: Can we move `writeMode` to before path and data?

stdlib/fs.gr
alex-snezhko: Ok going to rename `makeDir` to `createDir` as well for consistency

spotandjake:
Is this currently being blocked by anything?

stdlib/fs.gr
```diff
@@ -0,0 +1,1049 @@
...
+    RenameSource => ([File.Directory], [File.PathRenameSource], [], []),
+    RenameTarget =>
+      ([File.Directory], [File.PathRenameTarget, File.PathFilestats], [], []),
+    ReadDir => ([File.Directory], [File.FdReaddir], [], []),
+    MakeDir => ([File.Directory], [File.PathCreateDirectory], [], []),
+    MakeSymlink => ([File.Directory], [File.PathSymlink], [], []),
+    ReadLink =>
+      ([File.Directory], [File.PathFilestats, File.PathReadlink], [], []),
+    Stats => ([File.Directory], [File.PathFilestats], [], []),
+    Read => ([], [File.FdRead, File.FdFilestats], [], [File.Rsync]),
```
spotandjake, line 332: The `File.Rsync` seems to be causing `NotSupportedOrOperationNotSupportedOnSocket` to be thrown when running with wasmtime.

ospencer: It'll really depend on the runtime implementation.

ospencer (approved):
This looks good to go! @alex-snezhko Can you update the `@since` versions and then merge?

stdlib/fs.gr
spotandjake: > It'll really depend on the runtime implementation.

I understand that but I still think we could improve the experience a little by having a way to disable the flag maybe something like `Fs.disableRsync`, or providing a way for users to override stuff like this, the workaround is directly using `Wasi/File` which doesn't seem ideal especially given how popular wasmtime is.

alex-snezhko:
Aren't the versions already up-to-date @ospencer ?

stdlib/fs.gr
alex-snezhko: Added a flag as a default argument to the read and write functions

spotandjake (approved):
Looks like the tests just need to be for after otherwise ltgm.

ospencer:
> Aren't the versions already up-to-date @ospencer ?

Everything is marked as since 0.6.7, but the next version is 0.7.0: https://github.com/grain-lang/grain/pull/2147

Issue #1967, opened 2024-01-14
LSP: crash when hovering on function definition with default arguments


alex-snezhko:
When hovering over a function definition with default arguments, the LSP encounters an error:
![image](https://github.com/grain-lang/grain/assets/48104782/db1826cc-fe7c-4fd4-a881-377b386fe80f)

PR #1969, opened 2024-01-15
fix(lsp): Fix lsp crash when hovering on function with default arg

spotandjake:
Fixes: #1967

### DIFF
--- a/compiler/src/typed/printtyp.re
+++ b/compiler/src/typed/printtyp.re
@@ -730,7 +730,8 @@ and tree_of_argtyplist = (sch, al) =>
         switch (l) {
         | Default(_) =>
           switch (ty.desc) {
-          | TTyConstr(_, [ty], _) => ty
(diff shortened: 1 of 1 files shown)


compiler/src/typed/printtyp.re
```diff
@@ -730,7 +730,8 @@ and tree_of_argtyplist = (sch, al) =>
         switch (l) {
         | Default(_) =>
           switch (ty.desc) {
-          | TTyConstr(_, [ty], _) => ty
+          | TTyConstr(_, [ty], _)
+          | TTyLink({desc: TTyConstr(_, [ty], _)}) => ty
```
ospencer, line 734: You probably want to recurse on this case in the event there are next TTyLinks.

spotandjake: Done

ospencer (approved):

PR #1970, opened 2024-01-16
feat(stdlib): Add `Uri` module

alex-snezhko:
URI parsing and utility module, mostly conforming to RFC 3986.

```
module Main

include "uri"

let uri = Result.unwrap(Uri.parse("https://grain-lang.org"))
print(Uri.scheme(uri)) // https
print(Uri.host(uri)) // grain-lang.org

Uri.encode("a:b") // a%3Ab
```

### DIFF
--- a/stdlib/uri.gr
+++ b/stdlib/uri.gr
@@ -7,17 +7,16 @@
  */
 module Uri
 
-from "string" include String
-from "char" include Char
-from "uint8" include Uint8
-from "number" include Number
-from "bytes" include Bytes
-from "buffer" include Buffer
-from "list" include List
 from "array" include Array
-from "map" include Map
+from "buffer" include Buffer
+from "bytes" include Bytes
+from "char" include Char
+from "list" include List
+from "number" include Number
 from "option" include Option
 from "result" include Result
+from "string" include String
+from "uint8" include Uint8
 
 /**
  * Represents a parsed RFC 3986 URI.
@@ -64,14 +63,14 @@ provide enum ResolveReferenceError {
 /**
  * Represents an error encountered while attempting to percent-decode a string.
  */
-provide enum PercentDecodingError {
-  InvalidPercentEncoding,
+provide enum DecodingError {
+  InvalidEncoding,
 }
 
 /**
  * Used to specify which characters to percent-encode from a string.
  */
-provide enum PercentEncodeSet {
+provide enum EncodeSet {
   EncodeNonUnreserved,
   EncodeUserinfo,
   EncodeRegisteredHost,
@@ -108,7 +107,7 @@ let isPchar = char => {
   isUnreservedChar(char) || isSubDelim(char) || char == ':' || char == '@'
 }
 
-let makePercentEncoder = (encodeSet: PercentEncodeSet) => {
+let makeEncoder = (encodeSet: EncodeSet) => {
   let shouldEncodeForNonUnreserved = char => !isUnreservedChar(char)
 
   let shouldEncodeForUserinfo = char => {
@@ -159,18 +158,18 @@ let hexValueToChar = val => {
   }
 }
 
-let percentDecodeValid = (str, onlyUnreserved=false) => {
+let decodeValid = (str, onlyUnreserved=false) => {
   let bytes = String.encode(str, String.UTF8)
   let len = Bytes.length(bytes)
   let out = Buffer.make(len)
-  let cAt = i => Char.fromCode(Uint8.toNumber(Bytes.getUint8(i, bytes)))
+  let charAt = i => Char.fromCode(Uint8.toNumber(Bytes.getUint8(i, bytes)))
   for (let mut i = 0; i < len; i += 1) {
-    if (i >= len - 2 || cAt(i) != '%') {
+    if (i >= len - 2 || charAt(i) != '%') {
       let byte = Bytes.getUint8(i, bytes)
       Buffer.addUint8(byte, out)
     } else {
-      let next = cAt(i + 1)
-      let nextNext = cAt(i + 2)
+      let next = charAt(i + 1)
+      let nextNext = charAt(i + 2)
       let pctDecodedVal = charToHexValue(next) * 16 + charToHexValue(nextNext)
       if (onlyUnreserved && !isUnreservedChar(Char.fromCode(pctDecodedVal))) {
         Buffer.addChar('%', out)
@@ -185,7 +184,7 @@ let percentDecodeValid = (str, onlyUnreserved=false) => {
   Buffer.toString(out)
 }
 
-let isValidPercentEncoding = str => {
+let isValidEncoding = str => {
   let chars = String.explode(str)
   let len = Array.length(chars)
   for (let mut i = 0; i < len; i += 1) {
@@ -201,7 +200,7 @@ let isValidPercentEncoding = str => {
 
 // Lowercase all non-percent-encoded alphabetical characters
 let normalizeHost = str => {
-  let str = percentDecodeValid(str, onlyUnreserved=true)
+  let str = decodeValid(str, onlyUnreserved=true)
 
   let chars = String.explode(str)
   let rec getChars = (i, acc) => {
@@ -259,7 +258,7 @@ let removeDotSegments = path => {
 }
 
 /**
- * Percent-encodes characters in a string based on the specified `PercentEncodeSet`.
+ * Percent-encodes characters in a string based on the specified `EncodeSet`.
  * 
  * @param str: The string to encode
  * @param encodeSet: An indication for which characters to percent-encode. `EncodeNonUnreserved` by default
@@ -272,7 +271,8 @@ let removeDotSegments = path => {
  * @since v0.6.0
  */
 provide let encode = (str, encodeSet=EncodeNonUnreserved) => {
-  let shouldEncode = makePercentEncoder(encodeSet)
+  let shouldEncode = makeEncoder(encodeSet)
+  // TODO(#2053): use String.map when implemented
   let chars = String.explode(str)
   let rec getChars = (i, acc) => {
     if (i < 0) {
@@ -303,15 +303,15 @@ provide let encode = (str, encodeSet=EncodeNonUnreserved) => {
  * Decodes any percent-encoded characters in a string.
  * 
  * @param str: The string to decode
- * @returns `Ok(decoded)` containing a the decoded string or `Err(err)` if the decoding failed
+ * @returns `Ok(decoded)` containing the decoded string or `Err(err)` if the decoding failed
  * 
  * @since v0.6.0
  */
 provide let decode = str => {
-  if (!isValidPercentEncoding(str)) {
-    Err(InvalidPercentEncoding)
+  if (isValidEncoding(str)) {
+    Ok(decodeValid(str))
   } else {
-    Ok(percentDecodeValid(str))
+    Err(InvalidEncoding)
   }
 }
 
@@ -325,9 +325,7 @@ provide let decode = str => {
  */
 provide let encodeQuery = (urlVals, encodeSet=EncodeNonUnreserved) => {
   let parts = List.map(((key, val)) => {
-    encode(key, encodeSet=encodeSet) ++
-      "=" ++
-      encode(val, encodeSet=encodeSet)
+    encode(key, encodeSet=encodeSet) ++ "=" ++ encode(val, encodeSet=encodeSet)
   }, urlVals)
 
   List.join("&", parts)
@@ -337,14 +335,12 @@ provide let encodeQuery = (urlVals, encodeSet=EncodeNonUnreserved) => {
  * Decodes a query string into a list of pairs.
  * 
  * @param str: A query string
- * @returns A list of key-value pairs containing the values of the encoded string
+ * @returns `Ok(decoded)` containing a list of key-value pairs from the decoded string or `Err(err)` if the decoding failed
  * 
  * @since v0.6.0
  */
 provide let decodeQuery = str => {
-  if (!isValidPercentEncoding(str)) {
-    Err(InvalidPercentEncoding)
-  } else {
+  if (isValidEncoding(str)) {
     let parts = Array.toList(String.split("&", str))
     Ok(List.map(part => {
       match (String.indexOf("=", part)) {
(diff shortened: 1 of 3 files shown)


stdlib/uri.gr
```diff
@@ -0,0 +1,1189 @@
...
+  EncodeNonUnreserved,
+  EncodeUserinfo,
+  EncodeRegisteredHost,
+  EncodePath,
+  EncodePathSegment,
+  EncodeQueryOrFragment,
+  EncodeCustom(Char => Bool),
+}
+
+let isDigit = char => {
```
spotandjake, line 85: Thoughts on moving these into the char lib? This can be changed as well to instead use the operators from the `char` lib:
```
from Char use { (>=), (<=) }
char >= 0 && char <= 9
```

spotandjake (commented):
This looks great.

marcusroberts:
Alex, this looks fantastic and really useful.  I also learnt a lot about what are valid URLs (I swore you could only include? once!).   My only comment before I press approve is that pretty much every language calls urlEncode what you call percentageEncode, and I wondered if we should standardize on that?

ospencer (commented):
What made you return an abstract record and have access methods rather than just return the record?

alex-snezhko:
@ospencer This is following the pattern I've seen used in OCaml libraries for encapsulating data/providing a uniform interface for simple & computed data. In this particular instance not much is really being encapsulated but I think having these access methods still provides a degree of uniformity. Kind of like the idea behind properties in C# or getters in Java, where the interface to a data type can remain disconnected from the implementation.

I guess just providing the record directly here would be simpler though and would make the API smaller. I'm not heavily set in either direction so I'm open to arguments in either direction

alex-snezhko:
Yeah that makes sense. I can change it to a provided record

ospencer:
I'd wait to hear what the stdlib experts say :P

alex-snezhko:
@marcusroberts good point, I also just did a survey of a handful of languages and they all seem to prefer the terminology of "URL encoding". Though RFC 3986 uses the terminology of "percent encoding" so I guess it's a decision on whether we want to use the official terminology or the common lingo :shrug:

phated:
> it made me think some kind of computation was happening 

I think I agree here. Usually if you are calling a function on some abstract type, it'd be due to delayed processing. For example, a Rust library I was using was storing the full cookie `key=value` string as `&'a str` and then storing the key and value range as integers. When you want to get the key, you'd call `cookie.key()` and `cookie.value()` which would do the processing. Since you are immediately doing processing in `parse`, it's probably good to just return the record.

The caveat to the above is returning records with functions attached; instead, I would say should be functions that take the record (such as our Queue, Map, etc data structures).

spotandjake:
> @marcusroberts good point, I also just did a survey of a handful of languages and they all seem to prefer the terminology of "URL encoding". Though RFC 3986 uses the terminology of "percent encoding" so I guess it's a decision on whether we want to use the official terminology or the common lingo 🤷

I think we should use the common terminology but keep our docs searchable by including the `percent encoding` wording somewhere in the description.

alex-snezhko:
@spotandjake @marcusroberts What about just `encode` and `decode`? The fact that it's "URL" encoding should be implicit since it's in this module

ospencer (commented):
Thoughts on changing all of the places we say `percentEncode`/`percentDecode` to just `encode`/`decode`?

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+from "char" include Char
+from "uint8" include Uint8
+from "number" include Number
+from "bytes" include Bytes
+from "buffer" include Buffer
+from "list" include List
+from "array" include Array
+from "map" include Map
+from "option" include Option
+from "result" include Result
```
ospencer, line 20:
```suggestion
from "array" include Array
from "bytes" include Bytes
from "buffer" include Buffer
from "char" include Char
from "list" include List
from "map" include Map
from "number" include Number
from "option" include Option
from "result" include Result
from "string" include String
from "uint8" include Uint8
```

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+  } else {
+    Char.fromCode(val + 0x40 - 9)
+  }
+}
+
+let percentDecodeValid = (str, onlyUnreserved=false) => {
+  let bytes = String.encode(str, String.UTF8)
+  let len = Bytes.length(bytes)
+  let out = Buffer.make(len)
+  let cAt = i => Char.fromCode(Uint8.toNumber(Bytes.getUint8(i, bytes)))
```
ospencer, line 166: Could we call this `charAt`?

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+      chars[i] == '%' &&
+      (i >= len - 2 || !isHexDigit(chars[i + 1]) || !isHexDigit(chars[i + 2]))
+    ) {
+      return false
+    }
+  }
+  return true
+}
+
+// Lowercase all non-percent-encoded alphabetical characters
```
ospencer, line 202: I think we should make issues for `String.lowercaseAscii` and `String.uppercaseAscii` and todos that reference it here.

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+ * 
+ * @example Uri.encode("h3ll0_.w ?o+rld", encodeSet=Uri.EncodeNonUnreserved) // "h3ll0_.w%20%3Fo%2Brld"
+ * @example Uri.encode("d+om@i:n.com", encodeSet=Uri.EncodeRegisteredHost) // "d+om%40i%3An.com"
+ * @example Uri.encode("word", encodeSet=Uri.EncodeCustom(c => c == 'o')) // "w%6Frd"
+ * 
+ * @since v0.6.0
+ */
+provide let encode = (str, encodeSet=EncodeNonUnreserved) => {
+  let shouldEncode = makePercentEncoder(encodeSet)
+  let chars = String.explode(str)
```
ospencer, line 276: Can we make an issue for `String.map` and `String.flatMap` and reference it here? We can make this more efficient by not allocating an array for the explode and using a buffer to build the string from the characters after.

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+ * @returns `Ok(decoded)` containing a the decoded string or `Err(err)` if the decoding failed
+ * 
+ * @since v0.6.0
+ */
+provide let decode = str => {
+  if (!isValidPercentEncoding(str)) {
+    Err(InvalidPercentEncoding)
+  } else {
+    Ok(percentDecodeValid(str))
+  }
```
ospencer, line 315: Could we reverse this to avoid the `!`?

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+/**
+ * Decodes a query string into a list of pairs.
+ * 
+ * @param str: A query string
+ * @returns A list of key-value pairs containing the values of the encoded string
+ * 
+ * @since v0.6.0
+ */
+provide let decodeQuery = str => {
+  if (!isValidPercentEncoding(str)) {
```
ospencer, line 345: Same deal with avoiding `!`.

stdlib/uri.gr
alex-snezhko: What do you mean here? This function uses the char `toAsciiLowercase` function

stdlib/uri.gr
alex-snezhko: https://github.com/grain-lang/grain/issues/2053

stdlib/uri.gr
ospencer: Instead of exploding and iterating the Char version, it'd be handy to just call `lowercase` on the string. Unless I didn't read this hard enough.

ospencer (approved):
Phenomenal work on this! I learned a ton about URIs just from reading your implementation. Excited to use this soon.

phated:
Looks like this needs docs regenerated

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+  }, urlVals)
+
+  List.join("&", parts)
+}
+
+/**
+ * Decodes a query string into a list of pairs.
+ * 
+ * @param str: A query string
+ * @returns A list of key-value pairs containing the values of the encoded string
```
phated, line 340:
```suggestion
 * @returns A list of key-value pairs containing the values of the encoded string or `Err(err)` if the decoding failed
```

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+  }
+  let chars = getChars(String.length(str) - 1, [])
+  String.implode(Array.fromList(chars))
+}
+
+/**
+ * Decodes any percent-encoded characters in a string.
+ * 
+ * @param str: The string to decode
+ * @returns `Ok(decoded)` containing a the decoded string or `Err(err)` if the decoding failed
```
phated, line 306:
```suggestion
 * @returns `Ok(decoded)` containing the decoded string or `Err(err)` if the decoding failed
```

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+    [plus(any([unreserved, pctEncoded, subDelims, char('@')])), pathAbempty]
+  )
+
+  provide let pathRootless = seq([plus(pchar), pathAbempty])
+
+  provide let query = star(any([pchar, char('/'), char('?')]))
+
+  provide let fragment = query
+}
+use Matchers.*
```
phated, line 567: Why do the `use` here instead of inside the functions?

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+  }
+}
+
+/**
+ * Parses a string into a `Uri` according to RFC 3986. If the URI string has a
+ * path it will be automatically normalized, removing unnecessary `.` and `..`
+ * segments.
+ * 
+ * @param str: The RFC 3986 URI string to parse
+ * @returns `Ok(uri)` containing a `Uri` if the given string is a valid URI, `Err(ParseError)` otherwise
```
phated, line 700:
```suggestion
 * @returns `Ok(uri)` containing a `Uri` if the given string is a valid URI or `Err(ParseError)` otherwise
```

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+    return Ok({ scheme, userinfo, host, port, path, query, fragment })
+  }
+}
+
+/**
+ * Transforms a base URI and a URI reference into a target URI
+ * 
+ * @param base: The base URI to resolve a URI reference on
+ * @param ref: The URI reference to apply onto the base
+ * @returns `Ok(uri)` containing the target `Uri`, or `Err(err)` if input is malformed
```
phated, line 730:
```suggestion
 * @returns `Ok(uri)` containing the target `Uri` or `Err(err)` if input is malformed
```

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+  UpdateParseError,
+}
+
+/**
+ * Constructs a new `Uri` from a base `Uri` and components to update. The
+ * pattern used to update each component is that `None` means the base URI's
+ * component should be used and `Some(val)` means that a new value should be
+ * used for that component.
+ * 
+ * @param uri: The base `Uri` to apply updates on top of
```
phated, line 910:
```suggestion
 * @param uri: The `Uri` to update
```

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+  }
+
+  return Ok({ scheme, userinfo, host, port, path, query, fragment })
+}
+
+/**
+ * Determines whether or not a `Uri` has an authority (i.e. has a host component)
+ * 
+ * @param uri: The `Uri` to consider
+ * @returns `true` if the `Uri` has an authority component, `false` otherwise
```
phated, line 1055:
```suggestion
 * @returns `true` if the `Uri` has an authority component or `false` otherwise
```

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+    KeepOriginal => uri.fragment,
+    UpdateTo(x) => x,
+    UpdateParseError => return Err(InvalidFragmentError),
+  }
+
+  return Ok({ scheme, userinfo, host, port, path, query, fragment })
+}
+
+/**
+ * Determines whether or not a `Uri` has an authority (i.e. has a host component)
```
phated, line 1052:
```suggestion
 * Determines whether or not a `Uri` has an authority (i.e. has a host component).
```

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+ * 
+ * @param uri: The `Uri` to consider
+ * @returns `true` if the `Uri` has an authority component, `false` otherwise
+ * 
+ * @since v0.6.0
+ */
+provide let hasAuthority = uri => uri.host != None
+
+/**
+ * Determines whether or not a `Uri` is an absolute URI (has a scheme component)
```
phated, line 1062:
```suggestion
 * Determines whether or not a `Uri` is an absolute URI (has a scheme component).
```

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+ * 
+ * @param uri: The `Uri` to consider
+ * @returns `true` if the `Uri` is absolute (has a scheme component), `false` otherwise
```
phated, line 1065:
```suggestion
 * @returns `true` if the `Uri` is absolute (has a scheme component) or `false` otherwise
```

stdlib/uri.gr
```diff
@@ -0,0 +1,1090 @@
...
+ * 
+ * @since v0.6.0
+ */
+provide let isAbsolute = uri => uri.scheme != None
+
+/**
+ * Converts the given `Uri` into a string.
+ * 
+ * @param uri: The `Uri` to convert to a string
```
phated, line 1074:
```suggestion
 * @param uri: The `Uri` to convert
```

phated (changes requested):
Sorry I did those as individuals. I always thought "surely this will be the last one" until the actual last one.

stdlib/uri.gr
alex-snezhko: I think you'd still need some custom logic to only lowercase the characters that aren't percent-encode triples so not sure if there'd be a super huge advantage to using lowercase on the whole string

phated (approved):

PR #1972, opened 2024-01-22
feat(stdlib): Simplify `equal` using `Memory.compare`

spotandjake:
This replaces the logic for `String` equality and `Bytes` equality using the `Memory.compare` function. which should help reduce the size of the generated function and increase performance a tiny bit.

### DIFF
--- a/stdlib/runtime/equal.gr
+++ b/stdlib/runtime/equal.gr
@@ -8,6 +8,7 @@ include "runtime/unsafe/wasmf32" as WasmF32
 include "runtime/unsafe/tags" as Tags
 include "runtime/numbers" as Numbers
 from Numbers use { isNumber, numberEqual }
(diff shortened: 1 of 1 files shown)


stdlib/runtime/equal.gr
```diff
@@ -8,6 +8,7 @@ include "runtime/unsafe/wasmf32" as WasmF32
 include "runtime/unsafe/tags" as Tags
 include "runtime/numbers" as Numbers
 from Numbers use { isNumber, numberEqual }
+include "runtime/unsafe/memory"
```
ospencer, line 11: Could you group the includes a little better? Otherwise looks great.

spotandjake: Done

ospencer (approved):

PR #1973, opened 2024-01-22
chore(stdlib): Export `@array.length` primitive directly

spotandjake:
This is a really small pr that changes the implementation of `Array.length` to use the internal instruction.

### DIFF
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -23,9 +23,9 @@ include "runtime/numbers"
 from Numbers use { coerceNumberToWasmI32 }
 
 @unsafe
-let mut _ARRAY_LENGTH_OFFSET = 4n
-@unsafe
 let mut _ARRAY_START_OFFSET = 8n
+
+primitive length = "@array.length"
 
 @unsafe
 let checkLength = length => {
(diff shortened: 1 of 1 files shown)


stdlib/array.gr
```diff
@@ -62,10 +62,7 @@ let wrapNegativeIndex = (arrLen, idx) => {
  * @since v0.1.0
  */
 @unsafe
-provide let length = array => {
-  let ptr = WasmI32.fromGrain(array: Array<a>)
-  tagSimpleNumber(WasmI32.load(ptr, _ARRAY_LENGTH_OFFSET))
-}
+provide let length = array => length(array)
```
ospencer, line 65: Can't we just `provide { length }`?

spotandjake: So we could do a `provide primitive length`
But then we don't get the argument name in the type signature.

phated: @spotandjake please open an issue about not having the argument name in the type signature and mentioned we can clean this up when that is fixed.

spotandjake: This is now blocked on #1979

phated: @spotandjake this should be ready to rebase

spotandjake:
This has been updated

phated (approved):

PR #1978, opened 2024-01-23
chore: Update README copyright year

ospencer:


spotandjake (approved):
Looks good.

we should probably also update the `stdlib` license copyright year as well?

phated (approved):

PR #1979, opened 2024-01-25
chore(compiler): Add arg label to `ArrayLength` primitive

spotandjake:
This adds a label to the `ArrayLength` primitive. 

Neccessairy for #1973.

Question, do we want to add labels to all the primitive functions?


phated:
This is great. We probably want to use this for all of them but let's double check with @ospencer

ospencer:
I suppose so.

phated (approved):
Thanks Jake!

PR #1980, opened 2024-01-25
chore(ci): Improve reliability of docker builds on AWS

ospencer:
I believe the flakiness happening with the docker builds is that one job starts the machine and the other sees that it's already running and decides to move forward, even if the machine isn't truly ready yet. This makes the CI wait until the machine reports OK status rather than just running.


phated (approved):
Let's try it!

PR #1987, opened 2024-01-26
chore(cli): Hide node's stack trace

spotandjake:
This pr hides the node stack trace when an error is thrown from the cli. 

This should help improve user experience as the stack trace was irrelevant to the error previously.

### DIFF
--- a/cli/bin/exec.js
+++ b/cli/bin/exec.js
@@ -2,6 +2,14 @@ const path = require("path");
 const { execSync } = require("child_process");
 const fs = require("fs");
 
+function grainExec(command, execOpts) {
+  try {
+    return execSync(command, execOpts);
+  } catch (err) {
+    return "";
+  }
+}
+
 function flagsFromOptions(program, options) {
   const flags = [];
   program.options.forEach((option) => {
@@ -35,7 +43,7 @@ function execGrainc(
 ) {
   const flags = flagsFromOptions(program, options);
 
-  return execSync(`${grainc} ${flags.join(" ")} ${commandOrFile}`, execOpts);
+  return grainExec(`${grainc} ${flags.join(" ")} ${commandOrFile}`, execOpts);
 }
 
 function getGraindoc() {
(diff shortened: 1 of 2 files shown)


cli/bin/exec.js
```diff
@@ -2,6 +2,14 @@ const path = require("path");
 const { execSync } = require("child_process");
 const fs = require("fs");
 
+function grainExec(command, execOpts) {
```
ospencer, line 5: Maybe just call this `exec`?

spotandjake: Made that change

cli/bin/exec.js
```diff
@@ -2,6 +2,14 @@ const path = require("path");
 const { execSync } = require("child_process");
 const fs = require("fs");
 
+function exec(command, execOpts) {
+  try {
+    return execSync(command, execOpts);
+  } catch (err) {
+    process.exit(err.status);
```
phated, line 9: Never force exit in node.

phated (changes requested):
We should not be forcing an exit

cli/bin/exec.js
ospencer: Why?

ospencer: I found good information in https://nodejs.org/api/process.html#processexitcode.

ospencer: @spotandjake Looks like you can just set `process.exitCode` and that code will be used when the process exits.

spotandjake:
Switched to using `process.exitCode`

ospencer:
Looks like you need a `npm run cli format`

spotandjake:
Done

cli/bin/grain.js
```diff
@@ -181,9 +181,11 @@ program
   .version(pkgJson.version, "-v, --version", "output the current version")
   .forwardOption("-o <filename>", "output filename")
   .action(function (file, options, program) {
-    exec.grainc(file, options, program);
-    const outFile = options.o ?? file.replace(/\.gr$/, ".gr.wasm");
-    exec.grainrun(unprocessedArgs, outFile, options, program);
+    const compiled = exec.grainc(file, options, program);
```
phated, line 184: Probably name this variable "success" ?

phated (changes requested):

cli/bin/grain.js
spotandjake: Made that change

ospencer (approved):

cli/bin/exec.js
phated: Yeah. You want to submit another PR once I merge this?

phated (approved):

PR #1993, opened 2024-01-29
fix(compiler): Ensure refcounts are maintained when tail calls use arguments multiple times

ospencer:
In a tail call such as `let isNaN = x => x != x` where `x` is passed to `!=` twice, we would over-optimize by not `incRef`ing `x` since it would be `decRef`ed by `!=`; however as the argument is passed twice, it is `decRef`ed twice. This PR ensures we handle this case properly by only doing this optimization on the _last_ time the argument is passed to the tail call.

### DIFF
--- a/compiler/src/codegen/garbage_collection.re
+++ b/compiler/src/codegen/garbage_collection.re
@@ -91,16 +91,33 @@ let rec analyze_usage = instrs => {
     };
   };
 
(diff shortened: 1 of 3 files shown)


spotandjake (approved):
looks good

phated (approved):

PR #2003, opened 2024-02-01
fix(lsp): Prevent lsp crash when module cannot be found

spotandjake:
When #1963 was added we added support for hovering over the `include` tag to see the module signature. It looks like the function used for getting the signature can sometimes throw an exception this prevents the lsp from completely crashing in that case.

This has changed the env to be the program env as `include` is always going to be on the top level and a module is always going to imported into the programs env this makes it more reliable. In the case of failure we just provide the `module Header` so its still nice.

### DIFF
--- a/compiler/src/language_server/hover.re
+++ b/compiler/src/language_server/hover.re
@@ -185,11 +185,15 @@ let process =
     | [Module({decl, loc}), ..._] =>
       send_hover(~id, ~range=Utils.loc_to_range(loc), module_lens(decl))
     | [Include({env, path, loc}), ..._] =>
-      send_hover(
-        ~id,
-        ~range=Utils.loc_to_range(loc),
-        include_lens(env, path),
-      )
+      try(
+        send_hover(
+          ~id,
+          ~range=Utils.loc_to_range(loc),
+          include_lens(env, path),
+        )
+      ) {
+      | exn => send_no_result(~id)
+      }
     | _ => send_no_result(~id)
     };
   };
--- a/compiler/src/language_server/hover.re
+++ b/compiler/src/language_server/hover.re
@@ -126,11 +126,21 @@ let declaration_lens = (ident: Ident.t, decl: Types.type_declaration) => {
 };
 
 let include_lens = (env: Env.t, path: Path.t) => {
-  let module_decl = Env.find_module(path, None, env);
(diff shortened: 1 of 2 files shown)


compiler/src/language_server/hover.re
```diff
@@ -126,11 +126,21 @@ let declaration_lens = (ident: Ident.t, decl: Types.type_declaration) => {
...
+      switch (Modules.get_provides(decl)) {
+      | [_, ..._] => Some(module_lens(decl))
+      | [] => None
+      };
+    }) {
+    | Not_found => None
+    };
+  switch (module_decl) {
+  | Some(mod_sig) => markdown_join(header, mod_sig)
+  | None => header
```
alex-snezhko, line 142: Should this give a warning?

spotandjake: I don't think its necessary.

ospencer (commented):
Are we sure we want to show the module header when it crashes? That typically means that module is broken somehow, and an empty module is still a valid module signature, so this could be pretty misleading.

spotandjake:
> Are we sure we want to show the module header when it crashes? That typically means that module is broken somehow, and an empty module is still a valid module signature, so this could be pretty misleading.

I don't like the idea of an error message though. any thoughts on how we could handle it?

spotandjake:
This should be good now

compiler/src/language_server/hover.re
```diff
@@ -126,11 +126,17 @@ let declaration_lens = (ident: Ident.t, decl: Types.type_declaration) => {
...
+  let header = grain_code_block("module " ++ Path.name(path));
+  let decl = Env.find_module(path, None, env);
+  let module_decl =
+    switch (Modules.get_provides(decl)) {
+    | [_, ..._] => Some(module_lens(decl))
+    | [] => None
+    };
+  switch (module_decl) {
+  | Some(mod_sig) => markdown_join(header, mod_sig)
+  | None => header
```
ospencer, line 138: If there's no module declaration should we be showing the header?

spotandjake: I think so. I think of a module that exists but doesn't have any exports as empty.

ospencer (approved):

Issue #1996, opened 2024-01-30
Graindoc: positional arguments should start at 1


spotandjake:
Currently positional arguments numbering starts at 0. But I think we should start at 1, as when writing definitions you would never say `adds the zeroith element to the second element` it would be `adds the first element to the second`

alex-snezhko:
Going to close this since the PR was closed

PR #2004, opened 2024-02-02
chore(graindoc): Start param count at 1

spotandjake:
This pr makes the `graindoc` indexed parameter count start at `1`. I think this will make the documentation more constistent compared to the labels. Such as in a case like:
```
/**
  * Adds the first parameter to the second parameter.
  * @param 1: The first param
  * @param 2: The second param
  */
let add = (a, b) => a + b
```
When refereing to the params by number you wouldnt say the `zeroith param to the first param` you would say the `first param to the second param`.

This closes: #1996

### DIFF
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -126,10 +126,17 @@ let () =
       Some(msg);
     | MissingUnlabeledParamType({idx}) =>
       let msg =
-        Printf.sprintf(
-          "Unable to find a type for parameter at index %d. Make sure a parameter exists at this index in the parameter list.",
-          idx,
-        );
+        if (idx == 0) {
+          Printf.sprintf(
+            "Unable to find a type for parameter at index %d. Please note that parameter count starts at 1.",
+            idx,
+          );
+        } else {
+          Printf.sprintf(
+            "Unable to find a type for parameter at index %d. Make sure a parameter exists at this index in the parameter list.",
+            idx,
+          );
+        };
       Some(msg);
     | MissingReturnType =>
       let msg = "Unable to find a return type. Please file an issue!";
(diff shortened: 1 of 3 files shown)


compiler/graindoc/docblock.re
```diff
@@ -126,10 +126,17 @@ let () =
       Some(msg);
     | MissingUnlabeledParamType({idx}) =>
       let msg =
-        Printf.sprintf(
-          "Unable to find a type for parameter at index %d. Make sure a parameter exists at this index in the parameter list.",
-          idx,
-        );
+        if (idx == 0) {
+          Printf.sprintf(
+            "Unable to find a type for parameter at index %d. Please note that parameter count starts at 1.",
```
alex-snezhko, line 131: If we're starting it from 1 I think we should do away with the wording "index" as that almost unanimously implies "starting from 0" in programming languages. Maybe "position"?

alex-snezhko (approved):

compiler/graindoc/docblock.re
spotandjake: Thoughts on just saying `Unable to find a type for parameter %d`?

alex-snezhko: Sure sounds good to me

stdlib/runtime/wasi.gr
```diff
@@ -77,10 +77,10 @@ provide foreign wasm fd_prestat_dir_name: (
...
  * Invokes the `fd_write` system call.
  *
- * @param 0: The file descriptor to write to
- * @param 1: The pointer to the array of iovs to write
- * @param 2: The length of the array of iovs
- * @param 3: Where to store the number of bytes written
+ * @param 1: The file descriptor to write to
+ * @param 2: The pointer to the array of iovs to write
+ * @param 3: The length of the array of iovs
+ * @param 4: Where to store the number of bytes written
```
ospencer, line 83: The general idea of this change sounds reasonable, but when I read this it makes me feel like it should remain 0-based. What are other cases where this will show up?

phated:
This seems like an arbitrary change in search of a problem. I'm going to close this.

PR #2005, opened 2024-02-02
feat(lsp): Add Inlay types

### LINKED ISSUE #1359: LSP: Add support for inlay hints
Many languages like Javascript/ typescript, rust(through rust-analyzer VS code extension) provide support for inlay hint in VS Code as shown below.inlay hint provides hints for type of variable defined in a different colored text as shown below:
image
The inlay hints are optional can can be enabled by user. If grain lang could provide similar support, it could improve the developer experience greatly.

spotandjake:
This pr adds inlay types which closes #1359.

This pr is marked as a draft as there is still some work that needs to be done:
* We probably want to simplify some of the type signatures (i.e function types can get very complex, might be better to just say `Function`, and show the return type and param types on the actual lambda)
* We probably want to add a config to the vscode extension to toggle inlay types, as you might want inlay hints like the module hints but not the types.
* We should probably decide if we want to enable inlays on `Tuple patterns`, `record patterns`, `constructor patterns`

Current Demo Below:
image

### DIFF
--- a/compiler/src/language_server/inlayhint.re
+++ b/compiler/src/language_server/inlayhint.re
@@ -33,6 +33,11 @@ let send_no_result = (~id: Protocol.message_id) => {
   Protocol.response(~id, `Null);
 };
 
+let build_hint =
+    (position: Protocol.position, message: string): ResponseResult.inlay_hint => {
+  {label: ": " ++ message, position};
+};
+
 let find_hints = program => {
   let hints = ref([]);
   open Typedtree;
@@ -52,12 +57,26 @@ let find_hints = program => {
             line: stmt_end.pos_lnum - 1,
             character: stmt_end.pos_cnum - stmt_end.pos_bol + 1 + 1,
           };
+          hints := [build_hint(p, name), ...hints^];
+        | _ => ()
+        };
+      };
 
-          let r: ResponseResult.inlay_hint = {
-            label: ": " ++ name,
-            position: p,
-          };
-          hints := [r, ...hints^];
+      let enter_binding = ({vb_pat}: value_binding) => {
+        switch (vb_pat.pat_extra) {
+        | [] =>
+          switch (vb_pat.pat_desc) {
+          | TPatVar(_, {loc}) =>
+            let bind_end = loc.loc_end;
+            let p: Protocol.position = {
+              line: bind_end.pos_lnum - 1,
+              character: bind_end.pos_cnum - bind_end.pos_bol,
+            };
+            let typeSignature =
+              Printtyp.string_of_type_scheme(vb_pat.pat_type);
+            hints := [build_hint(p, typeSignature), ...hints^];
+          | _ => ()
+          }
         | _ => ()
         };
       };


spotandjake:
image
This pr should be ready for review now.

ospencer:
Here I'd want us to hide hints for function definitions—they're long and get in the way of the function arguments, which already have hints on them/show much of the same information.

It'd also be cool to have hints on destructures, but that could maybe come in a later PR.

marcusroberts (approved):
I'm going to approve because the code looks good, but subject to the request to be able to hide/show them as a setting

spotandjake:
> Here I'd want us to hide hints for function definitions—they're long and get in the way of the function arguments, which already have hints on them/show much of the same information.
> 
> It'd also be cool to have hints on destructures, but that could maybe come in a later PR.

I made that change, I also implemented a cli flag `--disable-inlay-types` which can be used to dissable the type hints when we release we can add a setting to the vscode extension that injects the flag.

alex-snezhko:
@spotandjake can't you just disable inlay hints in the client, and then users can decide if they want them or not? What is the purpose of the flag?

spotandjake:
> @spotandjake can't you just disable inlay hints in the client, and then users can decide if they want them or not? What is the purpose of the flag?

As we add more inlays, such as parameter hints. and other inlays I can see cases where the user might want to dissable the type hints but keep the others.

compiler/src/language_server/inlayhint.re
```diff
@@ -33,14 +33,113 @@ let send_no_result = (~id: Protocol.message_id) => {
...
+                              [build_hint(p, typeSignature), ...hints^];
+                          | _ => ()
+                          }
+                        },
+                        List.combine(args, argument_typs),
+                      );
+                    };
+                  | _ => ()
+                  }
+                | _ => ()
```
alex-snezhko, line 111: Would this ever be possible? If not should we do `failwith("Impossible: ...")`?

compiler/src/language_server/inlayhint.re
```diff
@@ -33,14 +33,113 @@ let send_no_result = (~id: Protocol.message_id) => {
...
+                            let typeSignature = string_of_typ(typ);
+                            hints :=
+                              [build_hint(p, typeSignature), ...hints^];
+                          | _ => ()
+                          }
+                        },
+                        List.combine(args, argument_typs),
+                      );
+                    };
+                  | _ => ()
```
alex-snezhko, line 109: I think this would never be possible. Should we do `failwith("Impossible: ...")`?

compiler/src/language_server/inlayhint.re
```diff
@@ -33,14 +33,113 @@ let send_no_result = (~id: Protocol.message_id) => {
...
+                      List.map(
+                        ((arg, typ: Types.type_expr)) =>
+                          switch (arg) {
+                          | Default(_) => None
+                          | _ => Some(typ)
+                          },
+                        typ_args,
+                      );
+                    // Make Hints
+                    if (List.length(argument_typs) == List.length(args)) {
```
alex-snezhko, line 90: Would this ever be false?

compiler/src/language_server/inlayhint.re
```diff
@@ -33,14 +33,113 @@ let send_no_result = (~id: Protocol.message_id) => {
...
+                switch (mb_pat.pat_desc) {
+                | TPatTuple(args) =>
+                  switch (resolve_typ(exp_type).desc) {
+                  | TTyArrow(typ_args, _, _) =>
+                    // Iterate For Types
+                    let argument_typs =
+                      List.map(
+                        ((arg, typ: Types.type_expr)) =>
+                          switch (arg) {
+                          | Default(_) => None
```
alex-snezhko, line 84: Could you expand on this? I'm seeing in your screenshot that the default argument has an inline hint on it

compiler/src/language_server/inlayhint.re
spotandjake: I don't think so but I also think it makes sense to ensure we don't make a change somewhere else in the compiler and break this. I am going to add a `failwith("Impossible")` though if it is the not equal.

compiler/src/language_server/inlayhint.re
spotandjake: The default hint comes from the binding pass, if this is enabled it shows `: Option<_>` after the value.

compiler/src/language_server/inlayhint.re
spotandjake: That case can be hit if we are omitting the annotation because `typ` is `None`

compiler/src/language_server/inlayhint.re
spotandjake: It seems as though there is a case where it can be a constructor.

alex-snezhko (approved):

compiler/src/language_server/inlayhint.re
```diff
@@ -33,14 +33,107 @@ let send_no_result = (~id: Protocol.message_id) => {
...
   let hints = ref([]);
   open Typedtree;
   open Protocol;
   module Iterator =
     TypedtreeIter.MakeIterator({
       include TypedtreeIter.DefaultIteratorArgument;
-      // Inlay hints for various expressions can be included here.
+
+      let enter_expression = ({exp_desc, exp_type}: expression) =>
+        // Function parameter type hints
```
ospencer, line 68:
```suggestion
```

compiler/src/language_server/inlayhint.re
```diff
@@ -33,14 +33,107 @@ let send_no_result = (~id: Protocol.message_id) => {
...
+        if (toggle_type_hints) {
+          switch (exp_desc) {
+          | TExpLambda(bindings, _) =>
+            List.iter(
+              ({mb_pat, mb_loc}: match_branch) => {
+                // Get Parameters
```
ospencer, line 74:
```suggestion
```

compiler/src/language_server/inlayhint.re
```diff
@@ -33,14 +33,107 @@ let send_no_result = (~id: Protocol.message_id) => {
...
+                switch (mb_pat.pat_desc) {
+                | TPatTuple(args) =>
+                  switch (resolve_typ(exp_type).desc) {
+                  | TTyArrow(typ_args, _, _) =>
+                    // Iterate For Types
```
ospencer, line 79:
```suggestion
```

compiler/src/language_server/inlayhint.re
```diff
@@ -33,14 +33,107 @@ let send_no_result = (~id: Protocol.message_id) => {
...
+                    let argument_typs =
+                      List.map(
+                        ((arg, typ: Types.type_expr)) =>
+                          switch (arg) {
+                          | Default(_) => None
+                          | _ => Some(typ)
+                          },
+                        typ_args,
+                      );
+                    // Make Hints
```
ospencer, line 89:
```suggestion
```

ospencer (changes requested):
Played with this locally and I love how this looks and feels! I understand your point about having the command line flag, but I feel like we should wait to add that when we actually are adding more hints and want them to be controlled separately and not commit to having the flag right now.

spotandjake:
After discussion in discord instead of the command line flag it would make more sense to use: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize but I think we can defer it to a separate pr when we are going to add more inlay hint types.

The final thing this pr needs is test but I don't quite understand how the new lsp tests work yet.

spotandjake:
Tests have been added so this should be ready for a review.

ospencer (approved):

PR #2006, opened 2024-02-05
Add comment nodes into the Doc and remove group_breaker node

phated:
This creates nodes in Doc for various comment types. This allows us to remove the `group_breaker` which was only used in comments.

I also think this will allow us to do smarter things in Doc since we'll know when we are working with comments.


spotandjake (approved):
looks good to me.

ospencer (commented):
I personally am not a fan of adding comment nodes to `Doc`. I wanted to keep Doc as a generic pretty printer that wasn't necessarily for just code (or Grain code for that matter), and potentially break it out into its own project one day. You mentioned that this would let us do special things because we know they're comments. What did you have in mind?

phated:
This needs to be further explored as part of #2038 so I'm closing this for now.

PR #2009, opened 2024-02-10
Print constants when we are able

phated:
This changes `print_constant` to print constants via the formatter when we have values that are easy to print. Huge shoutout to @ospencer on the guidance on making this great 🎉


compiler/test/grainfmt/binops.expected.gr
```diff
@@ -29,11 +29,11 @@ let p = 1 + 2 - 3
 
 let p2 = 1 - 2 + 3
 
-let y = 1 / 2 + 3
+let y = 1/2 + 3
 
 let ya = 1 / (2 + 3)
 
-let y1 = 1 + 2 /3
+let y1 = 1 + 2/3
```
phated, line 36: These expected were the primary goal of the change.

compiler/src/formatting/fmt.re
```diff
@@ -314,8 +314,44 @@ let build_document = (~original_source, parsed_program) => {
...
-  and print_constant = (~loc, constant) => {
-    string(get_original_code(loc));
+  and print_constant = constant => {
+    // TODO: Add originally written values for constants to the parsetree
+    // TODO: Add locations for constants into parsetree
+    switch (constant) {
+    | PConstNumber(PConstNumberInt(value)) => string(value)
+    | PConstNumber(PConstNumberFloat(value)) => string(value)
+    | PConstNumber(PConstNumberRational(numerator, denominator)) =>
+      string(numerator) ++ string("/") ++ string(denominator)
```
ospencer, line 324: This does not handle comments. You either need full locations on the individual components to be able to capture any comments (and determine if they're before or after the slash) or you need the original code.

ospencer (changes requested):

ospencer (approved):
Great work! This looks good to go to me.

PR #2010, opened 2024-02-10
Remove print_assigment for some utilities and less dry code

phated:
This removes the `print_assignment` function, as we don't have the concept of an assignment in the parsetree.

As we saw with the previous formatter, having a bunch of flags that you pass to a function is a code smell that makes it harder to tweak specific behavior on specific nodes of the tree. Instead, I've inlined the Doc creation code to the match branches, which is a little bit of duplication right now but will make our lives easier if one branch needs to change.

I've also made helpers to encapsulate the collapsible logic, which already lead to a question during the formatter code review.

### DIFF
--- a/compiler/src/formatting/fmt.re
+++ b/compiler/src/formatting/fmt.re
@@ -281,7 +281,7 @@ let get_op_and_assignment = new_value => {
   };
 };
 
-let is_collapsible_record = (record, elem, new_value) => {
+let is_collapsible_record_assignment = (record, elem, new_value) => {
   switch (record, elem, new_value) {
   | (
       {pexp_desc: PExpId({txt: IdentName({txt: name})})},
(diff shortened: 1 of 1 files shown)


compiler/src/formatting/fmt.re
```diff
@@ -260,6 +260,94 @@ let needs_grouping = (~parent, ~side: infix_side, expr) => {
   };
 };
 
+let get_op_and_assignment = new_value => {
```
ospencer, line 263: This name feels confusing to me out of context. I want a name that communicates that it gets the operator and assignment value from an infix assignment expression, but obviously that seems unnecessarily wordy with a name like `get_infix_assignment_op_and_value`.

spotandjake: What about `get_infix_parts`?

PR #2011, opened 2024-02-13
chore(docs): Add examples to `Hash` module

spotandjake:
This adds examples to the `Hash` module.

### DIFF
--- a/stdlib/hash.gr
+++ b/stdlib/hash.gr
@@ -3,6 +3,9 @@
  *
  * @example include "hash"
  *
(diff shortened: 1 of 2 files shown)


alex-snezhko:
Should we hold off until https://github.com/grain-lang/grain/pull/1952 before making this change?

spotandjake:
> Should we hold off until #1952 before making this change?

Probably

phated (approved):
I don't think seeded hash is landing before 0.6 so I'll merge this

PR #2014, opened 2024-02-13
chore(docs): Add examples to `Marshal` module

spotandjake:
This pr adds examples to the `Marshal` module.


stdlib/marshal.gr
```diff
@@ -430,6 +434,9 @@ let marshal = (value, buf) => {
  * @param value: The value to serialize
  * @returns A byte-based representation of the value
  *
+ * @example Marshal.marshal(1) == b"\x03\x00\x00\x00"
+ * @example Marshal.marshal("🌾") == Marshal.marshal('🌾')
```
alex-snezhko, line 438: I just tried this myself and it didn't seem to work. Is this a bug in our marshal implementation?

`Marshal.marshal("🌾") == Marshal.marshal("🌾")` also returned false when I tried it

spotandjake: image
I am thinking it may be an issue with how our byte comparison works just given the printed byte values

ospencer: The biggest problem with the example is that we're comparing the string `"🌾"` to the char `'🌾'`.
@alex-snezhko What you're seeing is differences in uninitialized bytes that we get back from `malloc`. I addressed this in #2045.

stdlib/marshal.gr
```diff
@@ -419,6 +423,9 @@ let marshal = (value, buf) => {
...
```
phated, line 427:
```suggestion
 * @example Marshal.marshal("🌾") == Marshal.marshal("🌾")
```

phated (approved):
This is unblocked by @ospencer's fix

Issue #1478, opened 2022-11-12
Stdlib: Reintroduce `gamma`, `factorial` functions


spotandjake:
Currently our trig functions in the module library such as `Number.sin`, `Number.cos` and `Number.tan` have some issues. That we should correct, we should make an attempt either fully improve the implementation for a more accurate function or create fast paths for special angles that would allow for more accuracy. Another current problem is with how we map values into the a `-pi` and `pi` range. as we dont have arbitrary precision floats we lose accuracy which makes the trig functions almost useless with bigints.

Edit Note: The original issue misidentified the problem we were running into, Upon furthur investigation there seems to be some other issues, so I hijacked this issue for that.

phated:
I updated the title since we're removing these in #2046 until the correct fix is made.

PR #2015, opened 2024-02-13
chore(docs): Add examples to `Number` module

spotandjake:
This pr improves the documentation on the `Number` module, by adding examples to most functions. It also adds a missing `since` property to `Number.sign`.

Notes:
* I omitted examples on the functions that take a `Range` until we have proper syntax.
* Our trig functions are currently not precise enough for good examples, I think we might need to look into rewriting them #1478 had been an issue to make `Number.sin` unbounded, but I noticed during this pr that was not the actual problem, the problem is how we map the input range and also with our implementation.

### DIFF
--- a/stdlib/number.gr
+++ b/stdlib/number.gr
@@ -16,6 +16,13 @@
  *
  * @example from "number" include Number
  *
+ * @example 1
+ * @example -1
+ * @example 0.5
+ * @example 1/2
+ * @example Infinity
+ * @example NaN
+ *
  * @since v0.4.0
  */
 module Number
@@ -70,6 +77,10 @@ provide let e = 2.718281828459045
  * @param num2: The second operand
  * @returns The sum of the two operands
  *
+ * @example
+ * from Number use { (+) }
+ * assert 1 + 2 == 3
+ *
  * @since v0.6.0
  * @history v0.4.0: Originally named `add`
  */
@@ -81,6 +92,10 @@ provide let (+) = (+)
  * @param num1: The first operand
  * @param num2: The second operand
  * @returns The difference of the two operands
+ *
+ * @example
+ * from Number use { (-) }
+ * assert 5 - 2 == 3
  *
  * @since v0.6.0
  * @history v0.4.0: Originally named `sub`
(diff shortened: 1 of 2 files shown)


stdlib/number.gr
```diff
@@ -577,12 +705,16 @@ provide let sin = (radians: Number) => {
...
 }
 
+// TODO(#1478): Improve the examples with more accurate values
 /**
  * Computes the cosine of a number (in radians) using Chebyshev polynomials.
  *
  * @param radians: The input in radians
  * @returns The computed cosine
  *
+ * @example Number.cos(Number.pi) == -0.9999998749960952
```
alex-snezhko, line 715: I don't know if we should be showcasing that our `cos` implementation has flaws :laughing:

alex-snezhko (approved):

stdlib/number.gr
spotandjake: I agree, I think the best option is to merge: #2032 as a temporary solution to our inaccuracies, and then fix this example.

ospencer:
Let's make sure something akin to #2032 is merged before we land this.

phated:
@spotandjake more conflicts (due to atan2?)

phated (approved):
LGTM but getting @ospencer or @alex-snezhko review

stdlib/number.gr
```diff
@@ -337,6 +409,13 @@ provide let isInteger = (x: Number) => {
  * @param x: The number to check
  * @returns `true` if the value is a non-integer rational number or `false` otherwise
  *
+ * @example Number.isRational(1/2)
```
ospencer, line 412: No == true here?

stdlib/number.gr
```diff
@@ -351,6 +430,14 @@ provide let isRational = (x: Number) => {
  * @param x: The number to check
  * @returns `true` if the value is finite or `false` otherwise
  *
+ * @example Number.isFinite(1/2)
+ * @example Number.isFinite(0.5)
+ * @example Number.isFinite(1.0)
+ * @example Number.isFinite(1)
```
ospencer, line 436: And these?

ospencer: Or is this our style? Just feels a little weird since it's not an assertion.

phated: I haven't been paying that close of attention. I know jake was doing assert in some cases.

ospencer (approved):

stdlib/number.gr
spotandjake: We have been using `assert` semantics for single line matches throughout the stdlib.

phated: @spotandjake why don't these contain `assert` then?

PR #2016, opened 2024-02-13
feat(stdlib): Add `atan2` to the `Number` module

spotandjake:
This pr adds `Number.atan2` which is useful in game development and when working with angles as it is a quick way to compute the angle of a ray from its coordinates.

The implementation is taken from the [here](https://en.wikipedia.org/wiki/Atan2#Definition_and_computation) with a slight modification to follow `c`, `python` and `js` where `x==0` and `y==0` results in an output of `0` instead of `NaN`.

Work for: #1017 

Other Documentation:
MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/atan2
Wiki Reference: https://en.wikipedia.org/wiki/Atan2

### DIFF
--- a/stdlib/number.gr
+++ b/stdlib/number.gr
@@ -796,6 +796,33 @@ provide let atan = angle => {
 }
 
 /**
+ * Computes the angle between the positive x-axis and the ray from the origin to the point (x, y).
+ *
+ * @param y: The given y coordinate
+ * @param x: The given x coordinate
+ * @returns The angle in radians between the positive x-axis and the point (x, y)
+ *
+ * @example Number.atan2(0, 1) == Number.pi
+ *
+ * @since v0.6.0
+ */
+provide let atan2 = (y, x) => {
+  if (x > 0) {
+    atan(y / x)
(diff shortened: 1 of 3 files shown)


stdlib/number.gr
```diff
@@ -864,6 +864,33 @@ provide let atan = angle => {
...
+ */
+provide let atan2 = (y, x) => {
+  if (x > 0) {
+    atan(y / x)
+  } else if (x < 0 && y >= 0) {
+    atan(y / x) + pi
+  } else if (x < 0 && y < 0) {
+    atan(y / x) - pi
+  } else if (x == 0 && y > 0) {
+    2 / pi
```
alex-snezhko, line 886: According to wikipedia this should be `pi/2`; the one below should also be `-pi/2`

spotandjake: Thanks for catching that.

ospencer (approved):

phated (changes requested):
Needs conflicts resolved

phated (approved):

PR #2018, opened 2024-02-13
chore(compiler): Add warning for `print` and `toString` with unsafe values

spotandjake:
This adds a warning when we detect a user trying to `print` an unsafe wasm value. This is similar to how we warn about `==` and `!=` on unsafe types.

I question if we should actually warn about this or just error as code such as `print(0.0W)` results in bad wasm generation which creates a nasty compiler error. 

### Counter Idea to improve dev experience
A counter-proposal to this pr, which might require some more complexity is we allow you to `print` unsafe types and just replace the call to `Pervasives.print` when we recognize it with the coresponding `DebugPrint` call. I think this would improve the developer experience when working with unsafe types and speed up development using them. 

### Future Safety Warning
We might also want to open an issue to block the usage of unsafe types on variables that cannot be inferred to prevent bad code generation. 

Would love to hear people's thoughts on this.


compiler/src/typed/typed_well_formedness.re
```diff
@@ -276,6 +277,29 @@ module WellFormednessArg: TypedtreeIter.IteratorArgument = {
             Grain_parsing.Location.prerr_warning(exp_loc, warning);
           };
         }
+      // Check: Warn if using Pervasives print on WasmXX types
+      | TExpApp(
```
alex-snezhko, line 281: This should also be done for `toString`

compiler/src/utils/warnings.re
```diff
@@ -157,7 +159,11 @@ let message =
       "it looks like you are calling Float64.fromNumber() with a constant number. Try using the literal syntax (e.g. `%sd`) instead.",
       String.contains(n, '.') ? n : n ++ ".",
     )
-  | UselessRecordSpread => "this record spread is useless as all of the record's fields are overridden.";
+  | UselessRecordSpread => "this record spread is useless as all of the record's fields are overridden."
+  | PrintUnsafe(typ) =>
+    "it looks like you are using print on an unsafe Wasm value here.\nThis is generally unsafe and will cause errors. Use Debug.print"
```
alex-snezhko, line 164: `DebugPrint.print...`

compiler/src/typed/typed_well_formedness.re
spotandjake: This now works on `toString`, I added stringification functions to the `DebugPrint` library as well so we could tell the user what to use.

compiler/src/utils/warnings.re
```diff
@@ -157,7 +161,15 @@ let message =
       "it looks like you are calling Float64.fromNumber() with a constant number. Try using the literal syntax (e.g. `%sd`) instead.",
       String.contains(n, '.') ? n : n ++ ".",
     )
-  | UselessRecordSpread => "this record spread is useless as all of the record's fields are overridden.";
+  | UselessRecordSpread => "this record spread is useless as all of the record's fields are overridden."
+  | PrintUnsafe(typ) =>
+    "it looks like you are using print on an unsafe Wasm value here.\nThis is generally unsafe and will cause errors. Use DebugPrint.print"
```
ospencer, line 166:
```suggestion
    "it looks like you are using `print` on an unsafe Wasm value here.\nThis is generally unsafe and will cause errors. Use `DebugPrint.print`"
```

compiler/src/utils/warnings.re
```diff
@@ -157,7 +161,15 @@ let message =
...
+    ++ typ
+    ++ " from the `runtime/debugPrint` module instead."
+  | ToStringUnsafe(typ) =>
+    "it looks like you are using toString on an unsafe Wasm value here.\nThis is generally unsafe and will cause errors. Use DebugPrint.toString"
```
ospencer, line 170:
```suggestion
    "it looks like you are using `toString` on an unsafe Wasm value here.\nThis is generally unsafe and will cause errors. Use `DebugPrint.toString`"
```

compiler/src/typed/typed_well_formedness.re
```diff
@@ -276,6 +277,33 @@ module WellFormednessArg: TypedtreeIter.IteratorArgument = {
             Grain_parsing.Location.prerr_warning(exp_loc, warning);
           };
         }
+      // Check: Warn if using Pervasives print or toString on WasmXX types
+      | TExpApp(
+          {
+            exp_desc:
+              TExpIdent(
+                Path.PExternal(Path.PIdent({name: "Pervasives"}), func),
```
ospencer, line 285: I would move the function names into this pattern rather than use a `when`.

compiler/src/typed/typed_well_formedness.re
```diff
@@ -276,6 +277,53 @@ module WellFormednessArg: TypedtreeIter.IteratorArgument = {
             Grain_parsing.Location.prerr_warning(exp_loc, warning);
           };
         }
+      // Check: Warn if using Pervasives print on WasmXX types
+      | TExpApp(
+          {
+            exp_desc:
+              TExpIdent(
+                Path.PExternal(Path.PIdent({name: "Pervasives"}), "print"),
```
phated, line 285: Why didn't you just do "print" | "toString"

phated (changes requested):

compiler/src/typed/typed_well_formedness.re
spotandjake: Because the error is different. and I wouldn't have had a reference to what the given function was to know which error to give.

phated (approved):
LGTM. Waiting for @ospencer to review

ospencer (approved):

Issue #2019, opened 2024-02-13
Compiler: Improve `FromNumberLiteral` warnings.


spotandjake:
Currently, we warn if we notice a user trying todo `Int64.fromNumber(1)` along with the same operation on `Float32`, `Float64`, `Int32`, `Int64`, `Uint32`, `Uint64`. But we have since added more types such as `Int8` we should add the same error for consistency and to promote good code.

PR #2024, opened 2024-02-14
chore(compiler): Add `fromNumber(<literal>)` warning for `short` types

spotandjake:
This pr makes the `fromNumber(<literal>)` warning work on all types.

On a side note I think we broke the implementation of this check when we added default args and this pr also fixes that issue.

I merged the `FromNumberLiteralXXX` warnings into a single warning. I don't know if we want to make this a breaking change as the warning numbers have changed (I feel like we should consider it as one). 

closes: #2019

### DIFF
--- a/compiler/src/typed/typed_well_formedness.re
+++ b/compiler/src/typed/typed_well_formedness.re
@@ -294,7 +294,11 @@ module WellFormednessArg: TypedtreeIter.IteratorArgument = {
                 exp_desc:
                   TExpConstant(
                     Const_number(
-                      (Const_number_int(_) | Const_number_float(_)) as n,
+                      (
+                        Const_number_int(_) | Const_number_float(_) |
+                        Const_number_rational(_) |
+                        Const_number_bigint(_)
+                      ) as n,
                     ),
                   ),
               },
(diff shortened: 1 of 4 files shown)


compiler/src/typed/typed_well_formedness.re
```diff
@@ -302,8 +302,12 @@ module WellFormednessArg: TypedtreeIter.IteratorArgument = {
           ],
         )
           when
-            modname == "Int32"
+            modname == "Int8"
```
alex-snezhko, line 305: Should we apply these to `BigInt` and `Rational` as well?

spotandjake: Yeah in that case I'm going to make an `isNumber` function

spotandjake: Made that change

ospencer (commented):
Can you add tests for the short nums?

compiler/src/typed/typed_well_formedness.re
```diff
@@ -289,42 +289,62 @@ module WellFormednessArg: TypedtreeIter.IteratorArgument = {
...
+            || modname == "BigInt" =>
         // NOTE: Due to type-checking, we shouldn't need to worry about ending up with a FloatXX value and a Const_number_float
         let n_str =
           switch (n) {
           | Const_number_int(nint) => Int64.to_string(nint)
           | Const_number_float(nfloat) => Float.to_string(nfloat)
-          | _ => failwith("Impossible")
+          | Const_number_rational({rational_num_rep, rational_den_rep}) =>
+            Printf.sprintf("%s/%s", rational_num_rep, rational_den_rep)
+          | Const_number_bigint({bigint_rep}) => bigint_rep
```
phated, line 328: Is this exhaustive? If so, you can streamline the first pattern match

phated (changes requested):
One thing

compiler/src/typed/typed_well_formedness.re
spotandjake: What do you mean by that?

phated: On lines 297-301 you are pattern matching and then assigning with `as n` but if this is exhaustive, those can just use `n` in the match. I'll make a suggestion above

compiler/src/typed/typed_well_formedness.re
```diff
@@ -289,42 +289,62 @@ module WellFormednessArg: TypedtreeIter.IteratorArgument = {
...
               {
                 exp_desc:
                   TExpConstant(
                     Const_number(
-                      (Const_number_int(_) | Const_number_float(_)) as n,
+                      (
+                        Const_number_int(_) | Const_number_float(_) |
+                        Const_number_rational(_) |
+                        Const_number_bigint(_)
+                      ) as n,
```
phated, line 301:
```suggestion
                      n
```
Suggestion for discussion below

ospencer (approved):

phated (approved):

Issue #1624, opened 2023-01-29
Grainlsp: Improvements for new module system


phated:
The LSP needs to be updated to provide a good experience for the new module system being introduced in #1584 

For example, something we must get done before the 0.6 release is inlay hints that show the module being introduced when you use `include "foo"`

spotandjake:
The three things that are left todo here are:

* Go to definition on `include`
  * When you click on go to definition while over an `Include` it would be extremely helpful if it took you to the module.
* Add autocomplete for the new module syntax once #2001 is figured out.
  * We may want to consider adding autocomplete for modules in the current path, and stdlib.
  * We for sure would want to be able to autocomplete the module name.
* Add autocomplete for the module values once #2002 is figured out. 
  * we would want to give you a list of values, types and submodules from the current opened module.

PR #2027, opened 2024-02-15
feat(lsp): Allow goto definition on `include` statements

spotandjake:
This pr allows for you to use `go to definition` on `Include` statements.

Work for: #1624

Note: 
As we did not reference the source module from the include statement before this change required adding that `attr`, The best way I could find to do it was when we got the module name. An alternative approach would have been to pass the `sod.pinc_path.txt` on in `include_module` but this would lead to inconsistencies with the `include` format.

### DIFF
--- a/compiler/src/typed/typemod.re
+++ b/compiler/src/typed/typemod.re
@@ -76,7 +76,8 @@ let extract_sig_open = (env, loc, mty) =>
 
 let include_module = (env, sod) => {
   let include_path = sod.pinc_path.txt;
-  let mod_name = Env.load_pers_struct(~loc=sod.pinc_loc, include_path);
+  let (mod_name, mod_file) =
+    Env.load_pers_struct(~loc=sod.pinc_loc, include_path);
   let mod_name =
     switch (sod.pinc_alias) {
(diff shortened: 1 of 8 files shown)


ospencer:
Is there not a valid location for the included module available from the environment?

spotandjake:
> Is there not a valid location for the included module available from the environment?

I looked for one but couldn't find it.

compiler/src/typed/typemod.re
```diff
@@ -104,7 +105,15 @@ let include_module = (env, sod) => {
     );
   let newenv = Env.include_module(mod_name, sod, env);
 
-  let od = {tinc_path: path, tinc_loc: sod.pinc_loc};
+  let mod_file =
```
marcusroberts, line 108: Just one question before I press approve.    When would the module file end in .wasm ?

ospencer:
I spent some time looking into this and I don't see an easy path forward right now. The method you're using right now to get the source is a bit fragile because it's looking at where the object file is rather than where the source is. We may want to just add a field to the objects that points to the source so we can read it from there.

PR #2032, opened 2024-02-18
fix(stdlib)!: Add a lookup table to `Numbr.sin` to improve trig accuracy

spotandjake:
This pr adds a lookup table to the `sin` function which increases the accuracy of `sin`, `cos` and `tan` technically this is breaking as our results are going to be different.

This isn't a long term solution as if we are not using a special angle there is still some inaccuracy, especially as numbers get larger. 

work for: #1478


ospencer (commented):
Why is this inaccurate in the first place?

phated (changes requested):
I don't like this hack. @spotandjake can you fix these correctly by the next community meeting? Otherwise I think we back out the broken math functions for the release

phated:
Closing this as @spotandjake says he won't be able to implement the full fix before v0.6. Instead, we are removing the functions (which is breaking) so we can re-add them as a non-breaking feature with the full fix.

PR #2041, opened 2024-02-20
feat(compiler)!: Update use syntax

### LINKED ISSUE #2002: Updated scoping operator syntax
With the changes introduced by #2001, this issue contains the new proposed syntax for the scoping operator.

# Old syntax
```
from Module use { foo, type Foo, module Foo }
from Module.Submodule use *
```

# New proposed syntax
```
use Module.{ foo, type Foo, module Foo }
use Module.Submodule.*
```

The `use` keyword moves to the front and the module name and values are separated by a dot. There are no changes to the values included within the curly braces.

ospencer:
Closes #2002

### DIFF
--- a/compiler/src/parsing/parser.messages
+++ b/compiler/src/parsing/parser.messages
@@ -36,7 +36,7 @@ program: MODULE YIELD
 ## The known suffix of the stack is as follows:
 ## MODULE
 ##
-program: MODULE UIDENT EOL FROM YIELD
+program: MODULE UIDENT EOL USE YIELD
 ##
 ## Ends in an error in state: 40.
 ##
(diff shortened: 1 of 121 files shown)


compiler/src/parsing/parser.messages
```diff
@@ -404,7 +404,7 @@ program: MODULE UIDENT EOL FROM UIDENT YIELD
 
 Expected the keyword `use` followed by items to use surrounded by `{}` or `*` to use all bindings.
```
phated, line 405: @ospencer don't the error messages need to be updated?

phated (changes requested):
Just the parser messages need updating. Otherwise looks good! It also made me wonder if we have any other error messages for the `from/use` syntax that need to be updated 🤔

compiler/src/parsing/parser.messages
ospencer: I thought I had checked all of them, and I did, but I missed just this one. Good catch!

phated (approved):

PR #2043, opened 2024-02-20
feat(compiler)!: Update include syntax

### LINKED ISSUE #2001: Reconsider module `include` syntax
# The current syntax

With the addition of the new module system, we defined the module include syntax as
```
include "list"
```

This would bring whatever module is defined in that file into scope. Optionally, an alias can be given for the module to give it a different name:
```
include "list" as StdList
```

# Reason for a syntax change

It's not clear what module name is brought into scope. By convention, the module name is _typically_ the Pascal-cased version of the name included in the file path. This already requires a user or module readers to assume the names of modules in scope, which isn't a good experience, but sometimes the convention is just not desired. When packages are introduced, the author may not want the module name to match the package name. For example, imagine a package named `pirate-tools`:
```
include "pirate-tools"
```
Is the module name `PirateTools`, or just `Pirate`? As a reader, you have no idea until you come across what you assume is a module matching this include statement somewhere within the file. Additionally, as a brand-new Grain user, you're likely not familiar with this convention, which makes onboarding far more difficult.

We wrote an inlay hint for the LSP that displays the name of the module next to the include statement to help mitigate this issue. We have a rule that the LSP may not be used as a crutch for bad language design—this feels like a place where we're blatantly violating that rule.

In some cases, we've always been adding an alias for modules to make it clear what names are in scope. This is a good practice, but it should be baked into the language.

# New proposed syntax

This proposed syntax always includes the module name.
```
from "list" include List
```
Aliasing is still allowed, and feels more consistent with how most aliasing happens in Grain:
```
from "list" include List as StdList
```

It's now completely clear what name is brought into scope without looking at the rest of the file:
```
from "pirate-tools" include Pirate
```

It also leaves the door open for packages to provide several modules together in the future, if we were to make that decision: (not included/associated with this issue, but just showing an example of what could be allowed one day):
```
from "stdlib" include List
from "stdlib" include Number
```

While there is no technical reason why this syntax could not coexist with the `from`/`use` scoping syntax, I think changing that syntax to not use the `from` keyword would be wise to always have it associated with including files.

ospencer:
Closes #2001


compiler/test/grainfmt/application.input.gr
```diff
@@ -49,4 +49,4 @@ let preEmatch = Map.make(): (Map.Map<Number, (Number,Number,Number,Number)>)
 
 let preExistingObjectsch2 = Map.make(): (Map.Map<Number, (Number,Number)>)
 
-let preExistingObjectsWithRefCountMismatch2 = Map.make(): (Map.Map<Number, (Number,Number,Number,Number,Number,Number,Number,Number,Number,Number,Number)>)
+let preExistingObjectsWithRefCountMismatch2 = Map.make(): (Map.Map<Number, (Number,Number,Number,Number,Number,Number,Number,Number,Number,Number,Number)>)
```
phated, line 52: why did this change?

compiler/src/typed/typemod.re
```diff
@@ -1081,6 +1092,13 @@ let report_error = ppf =>
       Includemod.report_error,
       errs,
     )
+  | Include_module_name_mismatch(provided_name, actual_name) =>
+    fprintf(
+      ppf,
+      "This statement includes module %s, but the file at the path defines module %s",
```
phated, line 1098: Do you think they'd be trying to define an alias? Perhaps we should add "Did you mean to write `include %s as %s"?

compiler/test/grainfmt/matches.input.gr
```diff
@@ -36,4 +36,4 @@ let testFormatting = charlist => {
       ['f', 'o', 'r', 'm', 'a', 't', ...rest] => true,
       _ => false
     }
-}
+}
```
phated, line 39: Why did this change?

stdlib/runtime/equal.gr
```diff
@@ -1,13 +1,13 @@
...
+from "runtime/unsafe/wasmi32" include WasmI32
 use WasmI32.{ (==), (!=), (&), (^), (+), (-), (*), (<), remS as (%), (<<) }
-include "runtime/unsafe/wasmi64" as WasmI64
-include "runtime/unsafe/wasmf32" as WasmF32
-include "runtime/unsafe/tags" as Tags
-include "runtime/numbers" as Numbers
+from "runtime/unsafe/wasmi64" include WasmI64
+from "runtime/unsafe/wasmf32" include WasmF32
+from "runtime/unsafe/tags" include Tags
+from "runtime/numbers" include Numbers as Numbers
```
phated, line 10: Duplicated "Numbers". Do we want the formatter to pun this?

stdlib/runtime/string.gr
```diff
@@ -20,15 +20,15 @@ use WasmI32.{
...
-include "runtime/unsafe/memory" as Memory
-include "runtime/unsafe/tags" as Tags
-include "runtime/numberUtils" as NumberUtils
-
-include "runtime/dataStructures" as DataStructures
+from "runtime/unsafe/wasmi64" include WasmI64
+from "runtime/unsafe/wasmf32" include WasmF32
+from "runtime/unsafe/wasmf64" include WasmF64
+from "runtime/bigint" include Bigint as BI
+from "runtime/unsafe/memory" include Memory as Memory
```
phated, line 27: Duplicated

stdlib/runtime/string.gr
```diff
@@ -20,15 +20,15 @@ use WasmI32.{
...
+from "runtime/unsafe/tags" include Tags
+from "runtime/numberUtils" include NumberUtils as NumberUtils
```
phated, line 29: Duplicated

phated (changes requested):
Found a few things. The big one is non-punning of `as` when the `include` is the same.

@ospencer Also, in the root issue you mentioned the LSP does an inlay for the module name, so you'll need to fix that as well. I think it just needs to be changed from `: ModuleName` to `include ModuleName`

compiler/test/grainfmt/matches.input.gr
ospencer: I accidentally formatted some of the inputs and thought I undid them properly but I suppose I didn't.

compiler/src/typed/typemod.re
ospencer: It's possible... I'll add it.

stdlib/runtime/equal.gr
phated: It looks like we **only** pun record fields, so we'll avoid punning this right now. @ospencer is going to open an issue to format away all duplicated aliases.

phated (approved):

PR #2045, opened 2024-02-21
fix(stdlib): Ensure consistent marshal representation

ospencer:
This makes sure that the marshaled representation is always consistent.

### DIFF
--- a/stdlib/marshal.gr
+++ b/stdlib/marshal.gr
@@ -24,6 +24,7 @@ module Marshal
 include "runtime/unsafe/wasmi32"
 use WasmI32.{
   (+),
+  (-),
   (*),
   (&),
   (==),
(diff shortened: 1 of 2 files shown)


phated (dismissed):

stdlib/marshal.gr
```diff
@@ -423,8 +423,11 @@ let marshal = (value, buf) => {
 @unsafe
 provide let marshal = value => {
   let valuePtr = fromGrain(value)
-  let buf = allocateBytes(size(valuePtr))
+  let size = size(valuePtr)
+  let buf = allocateBytes(size)
+  Memory.fill(buf + 8n, 0n, size)
```
phated, line 428: Remember when we introduced bytes/buffer and I asked "shouldn't we always zero it out"? 😜

phated (approved):

stdlib/marshal.gr
ospencer: We do, actually! This is just the low-level case where that doesn't happen.

PR #2046, opened 2024-02-22
chore(stdlib)!: Remove `sin`, `cos`, `tan`, `gamma`, `factorial` from `Number` module

phated:
This removes `sin` and everything that relies on it from our stdlib. As mentioned in #1478, our trig functions are wrong. We are removing them from the stdlib for v0.6 so the fix will not be breaking when we reintroduce them with proper accuracy.


ospencer (approved):
Sadge.

PR #2051, opened 2024-03-02
fix(compiler): Ensure reprovided types unify

### LINKED ISSUE #1886: Lang: re-provided type from one module incompatible with origin module
```
// c.gr
module C
provide record T {
  x: Number
}

// b.gr
module B
include "./c"
from C use { type T }
let a = { x: 1 }
provide { type T, a }

// a.gr
module A
include "./b"
from B use { type T, a }
print(a == { x: 1 })
```
The above code fails to compile with the error
```
Error: This expression has type B.T but an expression was expected of type
         C.T
C::T is abstract because no corresponding cmi file was found in path.
```

ospencer:
Closes #1886 

Shoutout to @peblair for the type manifest idea.

### DIFF
--- a/compiler/src/typed/typemod.re
+++ b/compiler/src/typed/typemod.re
@@ -674,6 +674,15 @@ let rec type_module = (~toplevel=false, anchor, env, statements) => {
             | Some(_) => failwith("Impossible: invalid alias")
             | None => Ident.create(name.txt)
             };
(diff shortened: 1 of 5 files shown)


peblair (approved):
The power of manifestation

phated (approved):

PR #2052, opened 2024-03-02
chore(compiler): Remove unnecessary fix_blocks in parser

ospencer:


phated (approved):
What was this for in the past?

ospencer:
The `fix_tyvar_mapper` was from before we used Menhir and couldn't properly distinguish between constructors and other values. The `fix_blocks` construct was from the very early days of Grain when we had `let foo = ... in` syntax...

PR #2054, opened 2024-03-02
fix(stdlib): Error when `relativeTo` used on relative source and absolute dest

alex-snezhko:
Properly handles case where first argument is absolute and second argument is relative to the `Path.relativeTo` function


spotandjake (approved):
Looks good

phated (approved):

PR #2064, opened 2024-03-10
feat(compiler): Allow non-block bodies with loops

alex-snezhko:
Allows `for` and `while` loops to be used without block bodies
```grain
for (let mut i = 0; i < 5; i += 1)
  print("Hello")
```


spotandjake (commented):
I think this looks great, should we add some comment formatting tests like:
```
for (let mut i = 0; i < 10; i += 1) /* Comment */ print("Here")
```

ospencer (approved):

PR #2068, opened 2024-03-13
fix(compiler): Allow proper disambiguation of enum variants

ospencer:
When multiple variants with the same name are in scope and one is used in a pattern match, the compiler would warn that the user should disambiguate, but when doing so for variants you'd get a type error. This resolves that issue.

### DIFF
--- a/compiler/src/typed/typepat.re
+++ b/compiler/src/typed/typepat.re
@@ -819,10 +819,12 @@ and type_pat_aux =
         ([{ppat_desc: desc, ppat_loc: loc}], true);
       };
     let opath =
-      /*try
-          let (p0, p, _) = extract_concrete_variant !env expected_ty in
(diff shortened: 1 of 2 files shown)


spotandjake (commented):
I love the look of this change, I assume that the compiler is using the newest occurance of the type? Could we add a third test with explit typing to make sure it is choosing the right one?

Otherwise looks great.

ospencer:
> I assume that the compiler is using the newest occurance of the type? Could we add a third test with explit typing to make sure it is choosing the right one?

I'm not sure what you mean. There's an annotation on the pattern for each test that forces the compiler to use the version annotated. Without the annotation it would guess.

spotandjake (approved):
ah ok, looks great than

phated (approved):

PR #1751, opened 2023-03-16
chore(compiler)!: Remove recursive_types and principal config

phated:
Closes #612

Deleting all the cruft.

PR #2072, opened 2024-03-16
fix(compiler): Fix disambiguation of record label access

ospencer:


compiler/src/typed/typecore.re
```diff
@@ -2592,7 +2592,7 @@ and type_label_access = (env, srecord, lid) => {
   let opath =
     try({
       let (p0, p, _) = extract_concrete_record(env, ty_exp);
-      Some((p0, p, repr(ty_exp).level == generic_level));
```
ospencer, line 2595: This has been broken for quite a while, long enough that I'm too lazy to find out. We did some cleanup in https://github.com/grain-lang/grain/pull/1751 where you can see what the code was before. Most of the logic around principal types was commented out, including a condition that would consider this to be true if principal types were not enabled. I imagine we accidentally commented that out way back in the day.

phated (approved):

PR #2073, opened 2024-03-22
chore(stdlib): Make `Array.contains` use recursion

### LINKED ISSUE #189: Stdlib: Improve Array.contains with recursion
When we can pattern match on arrays, we need to rewrite `Array.contains` with recursion instead of a while-loop.

Ref #188

spotandjake:
This rewrites `Array.contains` to use a recursive approach, I know the original idea was to use pattern matching similar to how it is done in the `list` library. but if we want a recursive approach this one would work well.

Some Notes:
* If we decide to merge this pr it would probably make sense to perform similar changes to `findIndex` and other iterative functions.
* Thinking about this from a performance standpoint, I don't know if the recursive approach is better here
* If we don't decide to merge this pr, I think we should close #189 as any indirection layer on pattern matching would likely be slower than the initial iterative approach or this split recursive approach.

Closes: #189

### DIFF
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -659,6 +659,16 @@ provide let fromList = list => {
   init(listLength(list, 0), next)
 }
 
+let rec containsHelp = (search, i, len, array) => {
+  if (array[i] == search) {
+    true
+  } else if (i == len) {
+    false
+  } else {
+    containsHelp(search, i + 1, len, array)
+  }
+}
+
 /**
  * Checks if the value is an element of the input array.
  * Uses the generic `==` structural equality operator.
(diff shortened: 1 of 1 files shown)


ospencer:
This implementation doesn't feel better/more clear to me than the current implementation.

spotandjake:
> This implementation doesn't feel better/more clear to me than the current implementation.

I do agree with that (I also don't think it's worse). I think given the constraints with array pattern matching though, there is no performance to be gained by switching to a recursive approach here in the first place and we should probably close the issue if we don't want to merge this pr.

ospencer:
If we did have the pattern matching on arrays, then that would probably be the more idiomatic way to write it, so I don't think we can close that issue until we close the discussion about array patterns.

Issue #1988, opened 2024-01-27
Add formatting support for array infix assignment


ospencer:
This functionality was introduced in #1928

PR #2075, opened 2024-03-22
feat(grainfmt): Support converting `binops` to `infix` on array assignments

spotandjake:
This pr makes the formatter convert `arr[0] = arr[0] + 1` to `arr[0] += 1` as is done with normal and record assignments.

This supports `constants` for the elem index and `idents` I didn't add support for `arr[test.index] = arr[test.index] + 1` as from my understanding that would be rather complex, I can try and add support if we feel its neccessairy though.

I hope I got the formatting functions correct and handled comments correctly I used the `RecordSet` as my template when building this.

closes: #1988

### DIFF
--- a/compiler/src/formatting/fmt.re
+++ b/compiler/src/formatting/fmt.re
@@ -1999,6 +1999,58 @@ let print_expression = (fmt, ~infix_wrap=d => group(indent(d)), expr) => {
            )
            ++ break,
          )
+    | PExpArraySet(
+        {pexp_desc: PExpId({txt: IdentName({txt: name})})} as arr,
+        {
+          pexp_desc:
+            PExpConstant(PConstNumber(PConstNumberInt({txt: elem_val}))) |
+            PExpId({txt: IdentName({txt: elem_val})}),
+        } as elem,
+        {
+          pexp_desc:
+            PExpApp(
+              _,
+              [
+                {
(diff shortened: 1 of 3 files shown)


ospencer:
We may need to figure out how to support the more complex case, since if I write code like `arr[test.index] += 1` it'll get unformatted to `arr[test.index] = arr[test.index] + 1`.

spotandjake:
> We may need to figure out how to support the more complex case, since if I write code like `arr[test.index] += 1` it'll get unformatted to `arr[test.index] = arr[test.index] + 1`.

Yeah that's less than ideal, I can look into it:
* `arr[test.index]`  - Records
* `arr1[arr2[0]]` - Arrays
I don't think im forgetting anything here, there are obviously cases like complex expressions, `are[test()]` and `arr[1 + 1]` but given those involve functions, unless we can prove the functions are pure first those are not reliable to return the same result so we can ignore them. I don't think we have any quick way of testing for the record or array cases though. for the record case we can match the pattern chain individually, I don't know a good way of handling the array index case.

In relation to unformatting, if I'm not mistaken the last nodes between the two would be the exact same so we could use a direct `=` right, that wouldn't work in the case of converting it to a binop in the first place though?

spotandjake:
Closing in favour of #2075

PR #2077, opened 2024-03-22
feat(compiler): Add warning for non-integer array indices

spotandjake:
This pr adds a basic warning for `arr[1.5]` and `arr[1.5] = 1` along with `rational` use cases. The purpose of this is to better catch instances where a user is indexing with floats ahead of time and prevent more runtime bugs. You can still achieve the runtime `Index not an integer` error by doing something like `let i = 1.5; arr[i]` though. The biggest place thing this helps to prevent is `arr[1.0]` where a user might assume we cast that to an `int like` number.

### DIFF
--- a/compiler/src/utils/warnings.re
+++ b/compiler/src/utils/warnings.re
@@ -173,7 +173,7 @@ let message =
     ++ typ
     ++ " from the `runtime/debugPrint` module instead."
   | ArrayIndexNonInteger(idx) =>
-    "Array index must be an integer, but found `" ++ idx ++ "`.";
+    "Array index should be an integer, but found `" ++ idx ++ "`.";
 
 let sub_locs =
   fun
--- a/compiler/src/utils/warnings.re
+++ b/compiler/src/utils/warnings.re
@@ -41,9 +41,10 @@ type t =
   | FromNumberLiteral(number_type, string, string)
   | UselessRecordSpread
   | PrintUnsafe(string)
-  | ToStringUnsafe(string);
+  | ToStringUnsafe(string)
+  | ArrayIndexNonInteger(string);
 
-let last_warning_number = 21;
+let last_warning_number = 22;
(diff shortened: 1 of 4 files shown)


ospencer:
Because our number system _does_ allow users to end up in situations where they might try to index by a non-integer, I don't think this should be a well-formedness error, but rather a warning. Making `arr[1.0]` fail to compile introduces some confusion about what the compiler is going to guarantee you.

spotandjake:
> Because our number system _does_ allow users to end up in situations where they might try to index by a non-integer, I don't think this should be a well-formedness error, but rather a warning. Making `arr[1.0]` fail to compile introduces some confusion about what the compiler is going to guarantee you.

do we currently support raising warnings in normal `well_formdness`? I can't see this functionality anywhere or would this need to be implemented?

ospencer:
It's supported, we just don't do it. You can see how to do it in typed_well_formedness.re: https://github.com/grain-lang/grain/blob/main/compiler/src/typed/typed_well_formedness.re#L270-L278

compiler/src/utils/warnings.re
```diff
@@ -169,7 +171,9 @@ let message =
   | ToStringUnsafe(typ) =>
     "it looks like you are using `toString` on an unsafe Wasm value here.\nThis is generally unsafe and will cause errors. Use `DebugPrint.toString`"
     ++ typ
-    ++ " from the `runtime/debugPrint` module instead.";
+    ++ " from the `runtime/debugPrint` module instead."
+  | ArrayIndexNonInteger(idx) =>
+    "Array index must be an integer, but found `" ++ idx ++ "`.";
```
ospencer, line 176:
```suggestion
    "Array index should be an integer, but found `" ++ idx ++ "`.";
```

ospencer (approved):

PR #2078, opened 2024-03-25
fix(stdlib): Avoid WASI random_get in Hash stdlib during module startup

ospencer:


stdlib/hash.gr
```diff
@@ -247,7 +253,11 @@ let rec hashOne = (val, depth) => {
  */
```
spotandjake, line 253: We should add the exception that `Result.unwrap` would throw to the grain doc to make sure it's known about.

spotandjake (changes requested):
Looks great, I think we need to add some the exception that could occur to grain doc but otherwise lets get this merged.

ospencer:
@spotandjake done.

spotandjake (approved):
Looks great.

spotandjake:
It does look like before this gets merged the grain doc is going to need to be regenerated.

marcusroberts (approved):
Looks good to me!

PR #2079, opened 2024-03-26
feat(compiler): Improve warning when omitting fields in record pattern

spotandjake:
When you omit record fields in a record pattern we currently just tell you, that you need them. The warning has been improved to let you know that if you meant to omit the fields you can use `, _` syntax, This helps the compiler better document and inform users about our language features.

### DIFF
--- a/compiler/src/utils/warnings.re
+++ b/compiler/src/utils/warnings.re
@@ -126,7 +126,9 @@ let message =
       msg,
     )
   | NonClosedRecordPattern(s) =>
-    "the following fields are missing from the record pattern: " ++ s
+    "the following fields are missing from the record pattern: "
+    ++ s
+    ++ ", if this is intentional you can use `, _` to ignore them."
   | FuncWasmUnsafe(func, f, m) =>
     "it looks like you are using "
     ++ func
--- a/compiler/src/utils/warnings.re
+++ b/compiler/src/utils/warnings.re
@@ -128,7 +128,7 @@ let message =
   | NonClosedRecordPattern(s) =>
     "the following fields are missing from the record pattern: "
     ++ s
-    ++ ", if this is intentional you can use `, _` to ignore them."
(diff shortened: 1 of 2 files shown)


ospencer:
This would be a feature rather than a chore. Tests?

spotandjake:
> This would be a feature rather than a chore. Tests?

I noticed we don't have tests for the field ommission error in the first place so added that. Given when we test warnings though we test them against their enum variants like:
```
assertWarning(
    "record_field_omit_2",
    "record Rec {foo: Number, bar: Number, bar2: Number}; let a = {foo: 1, bar: 2, bar2: 3}; match (a) { { foo } => void, _ => void }",
    Warnings.NonClosedRecordPattern("bar, bar2"),
  );
```
These tests wouldn't handle the change I made. I don't know if this would be considered a testing methodology error though, especially as we don't give any guarantee about the wording of compiler warnings.

compiler/src/utils/warnings.re
```diff
@@ -126,7 +126,9 @@ let message =
       msg,
     )
   | NonClosedRecordPattern(s) =>
-    "the following fields are missing from the record pattern: " ++ s
+    "the following fields are missing from the record pattern: "
+    ++ s
+    ++ ", if this is intentional you can use `, _` to ignore them."
```
ospencer, line 131:
```suggestion
    ++ "\nUse `_` to ignore unused fields."
```

compiler/test/suites/records.re
```diff
@@ -256,4 +255,15 @@ describe("records", ({test, testSkip}) => {
       (x: B) => x.field
     |},
   );
+  // well_formdness field omission warning
```
ospencer, line 258:
```suggestion
  // well_formedness field omission warning
```

ospencer (commented):
You're right. Thanks for adding those tests.

ospencer (approved):

PR #2080, opened 2024-03-29
fix(compiler): Fix array infix assign semantics

### LINKED ISSUE #1988: Add formatting support for array infix assignment

ospencer:
Closes #1988

### DIFF
--- a/compiler/src/typed/typecore.re
+++ b/compiler/src/typed/typecore.re
@@ -941,7 +941,7 @@ and type_expect_ =
       exp_type: instance(env, array_type),
       exp_env: env,
     });
-  | PExpArraySet(sarrexp, sidx, se) =>
+  | PExpArraySet({array: sarrexp, index: sidx, value: se, infix_op: None}) =>
     let array_type = newvar(~name="a", ());
     let arrexp =
       type_expect(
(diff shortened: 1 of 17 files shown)


spotandjake (approved):
This looks great, as a note my pr that implemented infix formatting still serves a role as in cases like:
```
arr[1] = arr[1] + 1
```
We are still going to want the formatter to convert those to infix automatically.

ospencer:
I didn't implement that because of the `arr[foo()]` case. I'm not sure if we want to do that change when it's semantically correct, because it's pretty annoying to detect.

spotandjake:
> I didn't implement that because of the `arr[foo()]` case. I'm not sure if we want to do that change when it's semantically correct, because it's pretty annoying to detect.

Detecting the simple case of an identifier or constant is easy I just don't think we should go beyond that .

ospencer:
I'll do that in a follow-on PR.

spotandjake:
Sounds good

ospencer:
I'll also do a follow-on PR to use this same format for regular infix assignment and record infix assignment.

compiler/src/typed/typecore.re
```diff
@@ -963,7 +963,88 @@ and type_expect_ =
...
+      };
+    let (ty_arg1, ty_arg2) =
+      switch (ty_args) {
+      | [(_, arg1), (_, arg2)] => (arg1, arg2)
+      | _ =>
+        raise(
+          Error(
+            infix.exp_loc,
+            env,
+            Arity_mismatch(expand_head(env, infix.exp_type), None),
```
alex-snezhko, line 1031: Shouldn't this be an `Impossible`?

alex-snezhko (approved):

compiler/src/typed/typecore.re
ospencer: It's not, because you can redefine what the infix operators do, and we don't stop you from making one take a number of arguments other than two. (You wouldn't be able to call it with the infix syntax, but you could call it as a regular function.)

marcusroberts (approved):
LGTM

peblair (commented):
Already merged, but also looks good to me

PR #2082, opened 2024-03-29
chore(stdlib): Improve `Invalid offset` warning in `charAt`

spotandjake:
This pr improves the warning for `Invalid offset` in `charAtHelp` which makes debugging a lot easier.

I'm not the biggest fan of the way I ended up laying out the error message, but I didn't have any better idea's so would love any suggestions.

### DIFF
--- a/stdlib/string.gr
+++ b/stdlib/string.gr
@@ -260,9 +260,12 @@ let getCodePoint = (ptr: WasmI32) => {
 }
 
 @unsafe
-let charAtHelp = (position, string: String) => {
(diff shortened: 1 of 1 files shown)


spotandjake:
> I don't see too much value being added here in the message. It seems like the issue is a broader debugging story issue of not being able to trace where the failure happened.

I agree with the issue being broader I think stack traces is the real solution which there is an issue open for. I think in the interim having error messages that mention which function they relate to helps a lot.

spotandjake:
Closing in favour of #2086

PR #2084, opened 2024-04-01
fix(grainlsp): Prevent LSP crash on startup when running in JS

ospencer:


marcusroberts (approved):
Looks like a good fix

PR #2091, opened 2024-04-07
feat(compiler): Partial function application

### LINKED ISSUE #402: Lang: `partial` operator
The `partial` operator allows you to partially apply a function.

For example, look at this implementation of an add function:

```grain
let add = (a, b) => a + b
```

In `let add2 = partial add(2)`, `add2` is a function that accepts a single argument and adds two to it.

Arguments may also be skipped and provided later:

```grain
let map123 = partial List.map(_, [1, 2, 3])

map123(a => a * 4) // [4, 8, 12]
```

alex-snezhko:
Partial function application

```grain
let add = (x, y) => x + y
let add1 = partial add(1, _)
assert add1(2) == 3

let printValues = (x, y, z=3) => {
  print(x)
  print(y)
  print(z)
}
let printSome = partial printValues(y=2, _, z=_)
printSome(1) // prints 1, 2, 3
```

Closes #402


alex-snezhko:
TODO:
- [ ] Make precedence of `partial` tighter to allow `x |> partial add(_, y)` without parentheses around the `partial` expression
- [ ] Decide if this feature should be supported for infix functions

compiler/src/parsing/ast_helper.re
```diff
@@ -319,6 +319,31 @@ module Expression = {
...
+          {
+            paa_loc: arg.ppaa_loc,
+            paa_expr:
+              switch (arg.ppaa_expr) {
+              | ArgumentGiven(expr) => expr
+              | ArgumentHole(_) =>
+                raise(
+                  SyntaxError(
+                    loc,
+                    "To use partial application, prefix the function call with `partial`.",
```
alex-snezhko, line 337: This was kind of a bandaid fix around me being a noob with LR(1) parsers and not being able to figure out how to create separate rules for partial and total applications without avoiding reduce/reduce conflicts. If anyone has insight on this could be done please let me know

ospencer:
Marked this as breaking because of the new keywords.

compiler/src/parsing/ast_helper.re
```diff
@@ -319,6 +319,31 @@ module Expression = {
...
+        arg =>
+          {
+            paa_loc: arg.ppaa_loc,
+            paa_expr:
+              switch (arg.ppaa_expr) {
+              | ArgumentGiven(expr) => expr
+              | ArgumentHole(_) =>
+                raise(
+                  SyntaxError(
+                    loc,
```
spotandjake, line 336: Should we point to the `arg.ppaa_loc`? I feel like this error might be a little annoying as it isn't immediately clear that the `_` hole is specific to partial application, at least that way the position would be telling you hey this argument is the issue specifically, an alteranatie here would be `Hole patterns are only allowed under partial application, prefix the function call with partial.` which is a bit more specific and searchable.

compiler/src/typed/typecore.re
```diff
@@ -1405,6 +1433,95 @@ and type_expect_ =
...
+              | ArgumentHole(_) =>
+                let name =
+                  switch (label) {
+                  | Labeled(name)
+                  | Default(name) => name
+                  | Unlabeled => mknoloc("$arg" ++ string_of_int(i))
+                  };
+
+                (
+                  Expression.ident(
```
spotandjake, line 1469: Can we give all this stuff the location from the arg.

compiler/src/typed/typecore.re
```diff
@@ -1405,6 +1433,95 @@ and type_expect_ =
...
+              param_pat,
+            );
+          },
+          labeled_args,
+        ),
+      );
+
+    let body =
+      Expression.apply(
+        ~loc=Location.dummy_loc,
```
spotandjake, line 1493: Should this take the location from the `func`?

compiler/src/typed/typecore.re
```diff
@@ -1405,6 +1433,95 @@ and type_expect_ =
...
+          switch (arg.ppaa_expr) {
+          | ArgumentHole(_) => Some(label)
+          | ArgumentGiven(_) => None
+          },
+        labeled_args,
+      );
+
+    let new_func_params_pat =
+      Pattern.tuple(
+        ~loc=Location.dummy_loc,
```
spotandjake, line 1511: Same as above.

compiler/src/typed/typecore.re
```diff
@@ -1880,184 +2006,222 @@ and type_application = (~in_function=?, ~loc, env, funct, sargs) => {
...
-            List.filter(((l, _)) => !is_optional(l), remaining_tyargs);
-          raise(Error(loc, env, Apply_too_few_arguments(missing_args)));
-        }
+    let omitted_args =
+      List.map(
+        ((l, ty)) => {
+          switch (l) {
+          | Default(_) =>
+            // omitted optional argument
+            (l, option_none(env, instance(env, ty), Location.dummy_loc))
```
spotandjake, line 2169: Should this take the location from `ty`?

compiler/test/suites/functions.re
```diff
@@ -385,4 +385,173 @@ truc()|},
...
+    "1\n2\n3\n",
+  );
+  assertRun(
+    "partial6",
+    {|
+      let f = (x, y, z) => { print(x); print(y); print(z) }
+      let pf = partial f(1, _, y=_)
+      pf(3, 2)
+    |},
+    "1\n2\n3\n",
```
spotandjake, line 441: as a sanity check can we just add a `print("0")` between the partial application and the actual application of the function i think it's obvious that it will only call the function afterwards but I feel like we should still double check that.

compiler/test/suites/functions.re
```diff
@@ -385,4 +385,173 @@ truc()|},
     |},
     "which has a default value, but the matching argument does not.",
   );
+
+  assertRun(
```
spotandjake, line 389: Should we have a test where a module provides a partially applied function, also maybe a clear test where a function returns a partially applied function. something like:
```
let f = (x, y) => {
  let t  = (a, b) => x + a + b
  let y = partial (_, y)
  y
}
print(f(1, 2)(3)) // 6
```
this ensures that we don't break closures and ensures that we are not losing the partial information when returning from a scope (The way this is implemented I don't think those are major concerns but they seem like good sanity checks. I think your test with the pipeline operator somewhat tests this but doesn't test closures.

compiler/src/typed/typecore.re
```diff
@@ -1405,6 +1433,95 @@ and type_expect_ =
...
+          (i, (label, arg)) => {
+            let (expr, param_pat) =
+              switch (arg.ppaa_expr) {
+              | ArgumentGiven(expr) => (expr, None)
+              | ArgumentHole(_) =>
+                let name =
+                  switch (label) {
+                  | Labeled(name)
+                  | Default(name) => name
+                  | Unlabeled => mknoloc("$arg" ++ string_of_int(i))
```
spotandjake, line 1465: I was having a little trouble visualizing what is actually happening here are we building a tuple from the arguments and then passing the tuple around and at the end passing the tuple values to the function when its fully applied? or did I misunderstand the abstraction here?

compiler/test/suites/functions.re
```diff
@@ -385,4 +385,173 @@ truc()|},
...
+      let makeFn = (x=0) => { print(x); f }
+      let pf = partial makeFn()(1, y=_, _)
+      pf(3)
+    |},
+    "0\n1\n2\n3\n",
+  );
+  assertRun(
+    "partial13",
+    {|
+      let (|>) = (x, f) => f(x)
```
spotandjake, line 510: Should we add this function to pervasives? I feel like the moment we add partial application we are going to be asked for a pipeline operator soon after we add this.

spotandjake (commented):
I had a few questions mostly on the dummy locations, looks great overall though.

PR #2096, opened 2024-04-21
feat(silo): Initial Silo scaffolding

ospencer:
This PR does some initial project scaffolding for Silo, Grain's new project manager and build system. This isn't quite an initial version, but rather some project setup to provide a foundation for us to work from as we build out functionality. It's not yet hooked up to our release process to allow us to continue Grain releases while we develop Silo in tandem.

Two commands are present, `new` and `build`:

```
$ silo --help
Build and manage Grain projects 🌾

Usage: silo [options] <command>

Options:
    -h, -help, --help  Display help
    -v, --version      Print version and exit

Commands:
    new    Create a new Grain project
    build  Compile the current project

Run silo <command> --help for help with individual commands.
```

Silo will eventually manage build artifacts itself rather than offload that functionality to `grainc`, manage dependencies (including the standard library), run wasm files, and more (though not necessarily all of these before an initial release).

As Silo does not yet handle dependencies/stdlib, `build.stdlib` must be set in your `silo.toml` to build projects.

I put together a Silo README as a basis for our Silo documentation which includes the following `silo.toml` configuration information, but I put it here in the PR description as well for easy access.

```toml
[package]
name = "my-package" # the name of this package

[build]
elide-type-info = false # compiles without Grain type info
include-dirs = ["../dir1", "../dir2"] # add paths to compiler code lookup
import-memory = false # imports the wasm memory instead of exporting one
initial-memory-pages = 64 # the number of wasm memory pages to start with
maximum-memory-pages = 128 # the limit of wasm memory pages to grow to
memory-base = 4096 # the memory address to begin Grain allocations
gc = true # enable or disable Grain's garbage collector
stdlib = "../custom-stdlib" # set the location of the standard library to use
use-start-section = false # use a wasm (start) section to run code on startup
wasi-polyfill = "src/wasi-polyfill.gr" # code to polyfill wasi imports

[build.wasm-features]
bulk-memory = true # enable or disable wasm bulk memory
tail-call = true # enable or disable wasm tail calls

[bin]
name = "my-name.wasm" # the name of the resulting wasm file
path = "src/main.gr" # the entrypoint of compilation
source-map = false # produce a source map file
wat = false # produce a WebAssembly Text (.wat) file
```


spotandjake (approved):
Looks good to me.

marcusroberts (approved):
This is great code to read and very comprehensible.   A great start to Silo!

silo/src/config.re
```diff
@@ -0,0 +1,123 @@
...
+    let wasm_features = build |-- field("wasm-features");
+    let wasm_features = {
+      bulk_memory:
+        get(config, wasm_features |-- key("bulk-memory") |-- bool),
+      tail_call: get(config, wasm_features |-- key("tail-call") |-- bool),
+    };
+
+    let build_config = {
+      elide_type_info:
+        get(config, build |-- key("elide-type-info") |-- bool),
```
alex-snezhko, line 71: I'm curious about the choice of kebab case? The convention within the language itself is camel case so why not use that here?

Afaik there's no key naming convention in TOML and I've seen a mixture of all of
* Rust, Gleam: snake case
* Python: kebab case (odd that it's not snake case...)
* Hugo: camel case

silo/src/cli.re
```diff
@@ -0,0 +1,79 @@
...
+      path
+    );
+  let new_ =
+    Cmd.make(
+      ~name="new",
+      ~doc="Create a new Grain project",
+      ~usage="[options] <path>",
+      ~options=[name_opt],
+      ~args=[path_arg],
+      Cmd.Thunk(() => New.new_(~name=?name^, path^)),
```
alex-snezhko, line 40: What do you think about the path optional at which point the current dir will be used?

alex-snezhko (commented):
Why is silo a separate cli tool rather than being integrated into the `grain` command? E.g. `grain build`, `grain new`?

silo/src/config.re
ospencer: Rust uses kebab case, so that + mostly translating command line flags is what got me here. I'd be open to camel case though. I sort of like the configuration language being different than the source language though.

ospencer:
> Why is silo a separate cli tool rather than being integrated into the `grain` command? E.g. `grain build`, `grain new`?

Silo is meant to completely replace the `grain` command, and the `grain` command will be removed once all of its functionality is built into Silo. It gives us an opportunity to truly integrate all of the tools into one and remove our Node.js dependency.

silo/src/new.re
```diff
@@ -0,0 +1,68 @@
...
+  Printf.fprintf(oc, "[package]\nname = \"%s\"\n", name);
+  close_out(oc);
+
+  let oc =
+    open_out(
+      Fp.toString(Fp.join(abs_path, Fp.relativeExn("src/main.gr"))),
+    );
+  Printf.fprintf(oc, "module Main\n\nprint(\"Hello, world!\")\n");
+  close_out(oc);
+};
```
alex-snezhko, line 50: Maybe can report a cheesy feedback message like "New grain project template created at <file path>. Happy coding! 🌾"

silo/src/config.re
alex-snezhko: Ah yeah rust does use kebab case, my bad. Same question about package names and output binary name conventions: I see in the PR description you have kebab case but in the stdlib the convention so far has been camel case. When package management, etc rolls out are we going with kebab case package names like npm?

silo/src/config.re
```diff
@@ -0,0 +1,123 @@
...
+    let bin_config = {
+      name: get(config, bin |-- key("name") |-- string),
+      path: get(config, bin |-- key("path") |-- string),
+      wat: get(config, bin |-- key("wat") |-- bool),
+      source_map: get(config, bin |-- key("source-map") |-- bool),
+    };
+
+    {package: package_config, build: build_config, bin: bin_config};
+  | `Error(msg, loc) => raise(LoadError(ConfigParseError(msg, loc)))
+  | exception (Sys_error(_)) => raise(LoadError(ConfigNotFound))
```
alex-snezhko, line 98: Should parent directories also be looked through?

silo/src/cli.re
```diff
@@ -0,0 +1,79 @@
...
+      ~doc="Compile using the release profile (production mode)",
+      ~value=profile,
+      ~conv=() =>
+      Build.Release
+    );
+  let build =
+    Cmd.make(
+      ~name="build",
+      ~doc="Compile the current project",
+      ~usage="[options]",
```
alex-snezhko, line 56: What do you think about having an optional argument to specify the directory to build from to avoid having to `cd` into it? So if you had `/a/b/silo.toml` and you're in `/a` you could do `silo build ./b`

silo/src/cli.re
```diff
@@ -0,0 +1,79 @@
...
+      ~usage="[options] <path>",
+      ~options=[name_opt],
+      ~args=[path_arg],
+      Cmd.Thunk(() => New.new_(~name=?name^, path^)),
+    );
+
+  let profile = ref(Build.Dev);
+  let release_opt =
+    Cmd.flag(
+      ~names=["--release"],
```
alex-snezhko, line 46: Why is this a CLI flag rather than something in the config TOML?

alex-snezhko (approved):
Left a few comments but looks great overall!

silo/src/cli.re
spotandjake: I assume this is a cli flag over being in the config because the average workflow would be compiling normally and then `release` is only used when your building for production at the end. 

What if instead of having this as a cli flag though we had build profiles in the config. so you could `silo build dev` or `silo build release`, (We could autogenerate a dev and release build, and have the regular build be the development one)

silo/src/config.re
spotandjake: I feel like it would make sense for us to look through parent directories, (especially if we are using `silo.toml` for the lsp as you might open vscode on a nested library and want your settings to follow over), I think if we do this though it would be beneificial to have a message like `Building <path> project` or just something to indicate which config is being used.

One note when coming back to this, there are definetly some security concerns with looking in parent directories, one thing that I have found useful in esy is being able to set `ESY__SANDBOX` which lets you set the project root throgh an env variable.

silo/src/new.re
```diff
@@ -0,0 +1,68 @@
...
+  let name =
+    switch (name) {
+    | Some(name) => name
+    | None =>
+      switch (Fp.baseName(abs_path)) {
+      | Some(name) => name
+      | None => raise(NewError(MustProvideName(path)))
+      }
+    };
+
```
spotandjake, line 35: Should we check if the new project your trying to make already exists and provide a warning? (Maybe we can allow it to be overridden with a force argument if needed).

Issue #2095, opened 2024-04-21
Decrease initial memory pages to 1


ospencer:
I arbitrarily chose the number 64 way back when. I don't know why.

The Grain runtime reserves 2k of space for itself (and only uses about 1k). A hello world program uses only about 400 bytes of actual allocated memory.

I don't think we need to start programs with 4 megabytes of memory 😛

AjaySDwivedi1:
Can I work on this issue? If I can, what files need to need or do I need to look at?

ospencer:
@AjaySDwivedi1 sure! It's mainly changing this line: https://github.com/grain-lang/grain/blob/main/compiler/src/utils/config.re#L367

There may be some references to this number in the CLI docs, so you'd just need to search code nearby `initial_memory_pages` or `initial-memory-pages`.

AjaySDwivedi1:
@ospencer for the code that has initial_memory_pages or initial-memory-pages, what changes need to be made? I checked them and I didn't know what to change.

PR #2097, opened 2024-04-22
Decrease initial memory pages to 1

AjaySDwivedi1:
Changed initial memory pages value to 1

### DIFF
--- a/compiler/src/utils/config.re
+++ b/compiler/src/utils/config.re
@@ -364,7 +364,7 @@ let initial_memory_pages =
     ~names=["initial-memory-pages"],
     ~conv=Cmdliner.Arg.int,
     ~doc="Initial number of WebAssembly memory pages",
-    64,
+    1,
   );
 
 let maximum_memory_pages =


AjaySDwivedi1:
I was having some issues downloading and running node.js so if I need to fix or change anything let me know.

AjaySDwivedi1:
Can the checkpoints that are failing be explained a little to me? I'm a little lost on what to fix or change.

ospencer:
These are two great test failures—they show a case I hadn't considered. We have a flag, `--memory-base`, which allows you to reserve an amount of memory that the Grain code won't touch. Those two tests request ~1mb of memory be reserved. The error we see is happening because Grain code is trying to write to memory addresses that don't exist.

We'll either need to automatically start with more pages or provide an error to the user that they need to also ask for more pages.

Looking into this I realized there's also a potential compiler bug that could happen from a particularly large program. That's a little involved to fix, so we'll send a PR to handle that and then we can rebase this PR once that fix is in.

Lots of fun from just a two character change 😄

spotandjake:
> These are two great test failures—they show a case I hadn't considered. We have a flag, `--memory-base`, which allows you to reserve an amount of memory that the Grain code won't touch. Those two tests request ~1mb of memory be reserved. The error we see is happening because Grain code is trying to write to memory addresses that don't exist.
> 
> We'll either need to automatically start with more pages or provide an error to the user that they need to also ask for more pages.
> 
> Looking into this I realized there's also a potential compiler bug that could happen from a particularly large program. That's a little involved to fix, so we'll send a PR to handle that and then we can rebase this PR once that fix is in.
> 
> Lots of fun from just a two character change 😄

I think we should warn the user that the minimum memory page count is below the requested starting amount, rather than just automagically fixing it. I think in environments where you really care about memory usage it's important that the flags are doing what you say.

ospencer:
Yup, that's going to be a part of the fix. Either a warning or an error.

ospencer:
Specifically, I wouldn't say we're "fixing" it if the program requires more than one starting page and the user didn't specify a specific number of starting pages. In the cases where they do specify the number of pages, we'll either error/warn.

JeysonFlores:
#2095

PR #2101, opened 2024-04-26
fix(compiler): Stabilize CRC checks

### LINKED ISSUE #1843: Investigate CRC behavior
In the current implementation of CRC there is an inconsistency in the usage of file paths including the .gr file extension and file paths without the extension; when new items are added into the CRC table they are keyed on the path with an extension, but when they are checked the extension is not included. When this discontinuity is attempted to be resolved, it leads to the CRC hashes being mismatched and an `Inconsistency` exception being raised, which causes compilation errors on valid programs

ospencer:
Closes #1843

Ensures we always are keying off of object name & removes cache for out-of-date objects. Also ensures consistent CMIs on the first & subsequent builds.

### DIFF
--- a/compiler/src/typed/subst.re
+++ b/compiler/src/typed/subst.re
@@ -35,14 +35,16 @@ type t = {
   types: PathMap.t(type_replacement),
   modules: PathMap.t(Path.t),
   modtypes: Tbl.t(Ident.t, module_type),
-  for_saving: bool,
+  for_cmi: bool,
+  for_crc: bool,
 };
 
 let identity = {
   types: PathMap.empty,
   modules: PathMap.empty,
   modtypes: Tbl.empty,
-  for_saving: false,
+  for_cmi: false,
+  for_crc: false,
 };
 
 let add_type_path = (id, p, s) => {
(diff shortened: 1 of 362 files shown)


marcusroberts (approved):
I'n not a code owner, but I read all the way through and it LGTM!

compiler/src/typed/subst.re
```diff
@@ -132,16 +150,23 @@ let norm =
...
+
 /* Similar to [Ctype.nondep_type_rec]. */
 let rec typexp = (s, ty) => {
   let ty = repr(ty);
   switch (ty.desc) {
   | (TTyVar(_) | TTyUniVar(_)) as desc =>
     /* Is this okay? This causes serialized type variables to lose their proper IDs. -Oscar */
-    /* if s.for_saving || ty.id < 0 then */
-    if (s.for_saving) {
+    /* if s.for_cmi || ty.id < 0 then */
```
phated, line 166: Still haven't answered this question? 😛

phated (approved):

compiler/src/typed/subst.re
ospencer: No 😛

Theoretically if we ever figured this out we wouldn't need the `Translsig` module. I think? The world may never know 🙈

PR #2104, opened 2024-05-04
feat(compiler)!: Custom Grain object files

### LINKED ISSUE #622: Update contributor docs for static linking

### LINKED ISSUE #842: Linker: Add cycle detection
The linker currently doesn't have cycle detection because the typechecker generally won't allow cycles anyway; however, we could introduce cycles ourselves when we subvert the typesystem or do things in post-typechecking phases of the compiler.

This results in a cryptic error message. We should add a cyclic dependency check that raises in the linker to avoid those possibilities. It seems that the ocamlgraph library we use has cycle detection available, we just need to use it.

See also #841

### LINKED ISSUE #1754: Use non .wasm extension for intermediate build files
We want to convey that the intermediate wasm files users get when they compile something are not runnable, and are just to be used by the compiler. I'd also probably take this opportunity to make the default file output simply `.wasm`—the original reason they were `.gr.wasm` was to distinguish them as wasm files that needed to be run in the Grain runtime, but this is no longer the case.

### LINKED ISSUE #1903: Premature optimization of modules with side effects
If you have module `b` with the contents:
```
module B
print("b")
```
and module `a` with the contents:
```
module A
print("b")
include "./b"
print("c")
```
when you call `grain a.gr` instead of getting the expected
```
a
b
c
```
in the console you get 
```
b
c
```

It seems like we might be prematurely optimizing modules because when I add an export and use that export the program does whats expected. I think we need to consider side effects in tree shaking.

### LINKED ISSUE #1997: Re-enable per-module Binaryen optimizations
In #1330 we removed per-module Binaryen optimizations since the API doesn't yet support querying data segments by name. Once this is implemented, we should re-enable it.

Tracking https://github.com/WebAssembly/binaryen/issues/5984

ospencer:
### Overview

This PR replaces Grain's intermediate `.gr.wasm` files with Grain object files. This is for a variety of reasons, including:

- Debuggability. Our wasm-based linker is unable to reconstruct debugging/sourcemap information from wasm binaries. This approach once again generates proper sourcemaps. There is additional work to be done to make the sourcemaps truly usable, but this is a major step in the right direction.
- Optimizations. This change will allow us to do "whole program" optimizations to better assist Binaryen with its optimizations.
- Performance. Going from our IR to Binaryen IR and serializing is the bulk of compilation time. Since we no longer are serializing/deserializing many files, there is a performance boost. For a hello world program, cold compile time is down 10% and warm compile time is down 25%. When running in JS, cold compile time is down 15% and warm compile time is down 21%. This also gets subsecond warm builds in JS on my machine.

### Technical details

Grain object files use the `.gro` extension. As no wasm files are specific to Grain, the default output filename from the compiler uses the `.wasm` extension rather than `.gr.wasm`.

Object files follow this layout:

|magic|version len|version|cmi length|cmi|mash_code|
|---|---|---|---|---|---|
|0xF0 0x9F 0x8C 0xBE|5|0.6.3|1024| \<cmi> |<mash_code>|

We use OCaml's `Marshal` module as our binary format (and for Whole Grain, we can use our `Marshal` module). js_of_ocaml can't marshal floats into a format that can safely be read back by native code. I wanted object files to be the same between native and js, so we store floats as `int64`s in the mashtree to guarantee consistent behavior.

### Reviewing this PR

The code delta for this PR is fairly small, but snapshot tests are now sexprs of mashtrees. As such, we have all new snapshots. I'd recommend clicking the first commit to review the code diff, or using the GitHub UI to hide changes to snapshots.

Closes #622
Closes #842 
Closes #1754 
Closes #1903 
Closes #1997


stdlib/package.json
```diff
@@ -27,7 +27,7 @@
     "index.js"
   ],
   "scripts": {
-    "clean": "del-cli \"**/*.wasm\" \"**/*.wat\" \"**/*.modsig\""
+    "clean": "del-cli \"**/*.wasm\" \"**/*.wat\" \"**/*.gro\" \"**/*.modsig\""
```
spotandjake, line 30: Should we leave the `*.gr.wasm` in here as well for a version or two so we make sure to clean up the legacy files?

compiler/test/suites/basic_functionality.re
```diff
@@ -377,6 +377,6 @@ describe("basic functionality", ({test, testSkip}) => {
     ~config_fn=smallestFileConfig,
     "smallest_grain_program",
     "",
-    4769,
+    6424,
```
spotandjake, line 380: Why did this go up?

Is it because of names changing and debug info?

stdlib/package.json
ospencer: That's covered by `**/*.wasm`.

compiler/test/suites/basic_functionality.re
ospencer: I meant to leave a comment about this; thanks for pointing it out. The old linker had a concept of hard/soft dependencies. If a file was a dependency but no values/functions from that module were actually uses (just types or an unused import), then the linker would compile in type information but not link in the actual files.

At first thought this seems like good behavior, but if this is done, side effects from importing those modules don't happen, which is bad. For example, a normal Grain program that doesn't use anything in `Pervasives` doesn't link in `Pervasives`, which means the `setupExceptions` side effects don't happen, and throwing exceptions in those programs prints `GrainError` instead of the actual error. In practice, users don't run into this because most programs use something from Pervasives.

This new linker links in all dependencies, which is more correct. As I mentioned, in practice this doesn't make a real difference for most users. Side note: the major reason the original implementation did this was because the compiler wasn't great at removing unused code from the resulting binary and this resulted in large module sizes. The compiler has gotten much better and this isn't really an issue anymore.

I also discovered a few more configurations producing smaller programs while working on this, as small as 25 bytes.

spotandjake: Ah okay then this probably closes #1903 then.

compiler/src/codegen/mashtree.re
```diff
@@ -548,10 +556,14 @@ type mash_program = {
   exports: list(export),
   main_body: block,
   main_body_stack_size: stack_size,
-  globals: list((Ident.t, Types.allocation_type)),
+  globals: list((Ident.t, bool, Types.allocation_type, option(constant))),
```
alex-snezhko, line 559: Can we make this a record now? It's not immediately clear what each of these are

compiler/src/typed/cmi_format.re
```diff
@@ -142,22 +138,32 @@ module CmiBinarySection =
...
-  | exception (Error(e)) =>
-    close_in(ic);
-    raise(Error(e));
   };
+  let version_length = input_binary_int(ic);
+  let read_version = really_input_string(ic, version_length);
+  if (read_version != Config.version) {
+    raise(Error(Wrong_version_interface(filename, read_version)));
+  };
+  let _ = input_binary_int(ic);
```
alex-snezhko, line 151: Why not do the same validation as lines 48-51 of emitmod where you verify the length is correct?

compiler/src/codegen/compcore.re
```diff
@@ -278,56 +92,53 @@ let init_codegen_env = name => {
...
-      true;
+let rec resolve_func = (~env, name) => {
+  switch (Hashtbl.find_opt(env.func_import_resolutions, name)) {
+  | Some(resolution) => resolve_func(~env, resolution)
+  | _ => name
   };
-  Ident.unique_name(name);
+};
+
+let reset = () => {
```
alex-snezhko, line 118: What is the point of this function

ospencer: Many of our modules have a `reset` function that clears any internal state that calls whatever smaller reset functions there are. Did the same thing here.

compiler/src/typed/cmi_format.re
ospencer: If it's wrong, `Marshal.from_channel` will raise, which we catch and raise `Corrupted_interface`. We could check it, but it'd be redundant and also make us read the entire channel.

ospencer:
@alex-snezhko This should be ready for you to look at again.

compiler/src/codegen/compcore.re
```diff
@@ -3458,8 +3300,216 @@ let compile_functions = (wasm_mod, env, {functions} as prog) => {
     } else {
       compile_function(wasm_mod, env, func);
     };
+  };
   ignore @@ List.map(handle_attrs, functions);
-  ignore @@ compile_main(wasm_mod, env, prog);
+};
+
+let construct_type_metadata_table = metas => {
```
spotandjake, line 3307: Do we want to move this to a different file?

compiler/src/codegen/compcore.re
```diff
@@ -3387,66 +3200,95 @@ let compile_globals = (wasm_mod, env, {globals} as prog) => {
...
-        wasm_mod,
-        Expression.Call.make(wasm_mod, grain_main, [], Type.int32),
-      ),
+      Expression.Block.make(wasm_mod, "_start", starts),
     );
+  if (Grain_utils.Config.use_start_section^) {
+    Function.set_start(wasm_mod, start);
+  } else {
+    ignore @@
+    Export.add_function_export(wasm_mod, "_start", Comp_utils.grain_start);
```
alex-snezhko, line 3283: Nit: inconsistent use of `grain_start` constant and "_start" literal (a few other times in the lines above)

alex-snezhko (approved):
Did not see any obvious problems, LGTM!

compiler/src/codegen/emitmod.re
```diff
@@ -1,8 +1,72 @@
...
+};
+
+let load_object = (ic): mash_program => {
+  let read_magic = really_input_string(ic, 4);
+  if (read_magic != Cmi_format.magic) {
+    raise(BadObject(InvalidMagic));
+  };
+  let version_length = input_binary_int(ic);
+  let read_version = really_input_string(ic, version_length);
+  if (read_version != Config.version) {
```
spotandjake, line 44: This current logic treats every version as breaking doesn't it? Which would require a full recompile even for small stdlib bug changes right?

compiler/src/codegen/emitmod.re
```diff
@@ -1,8 +1,72 @@
...
+  let oc = open_out_bin(outfile);
+
+  output_string(oc, Cmi_format.magic);
+  let version_length = String.length(Config.version);
+  output_binary_int(oc, version_length);
+  output_string(oc, Config.version);
+  let cmi = Marshal.to_bytes(mashed.signature, []);
+  output_binary_int(oc, Bytes.length(cmi) + version_length + 12);
+  output_bytes(oc, cmi);
+  Marshal.to_channel(oc, mashed, []);
```
spotandjake, line 32: Would it make sense to remove the `signature` here and just have it point to the cmi? Instead of including the cmi twice? It would still allow us to read the cmi without parsing the mashtree.

compiler/src/codegen/emitmod.re
```diff
@@ -1,8 +1,72 @@
...
+    close_out(oc);
+  };
+
+  let oc = open_out_bin(outfile);
+
+  output_string(oc, Cmi_format.magic);
+  let version_length = String.length(Config.version);
+  output_binary_int(oc, version_length);
+  output_string(oc, Config.version);
+  let cmi = Marshal.to_bytes(mashed.signature, []);
```
spotandjake, line 29: Taking a little hint from the wasm bytecode how would we feel about having this be `sectionID ...sectionContents sectionLength` instead of just storing the data randomly, this would allow us to add more data in here easily in the future without breaking things while still being easy and quick to parse.

compiler/src/codegen/emitmod.re
```diff
@@ -1,8 +1,72 @@
...
 
-let emit_module = ({asm, signature}, outfile) => {
+type bad_object =
+  | InvalidMagic
+  | InvalidVersion
+  | Corrupted;
+
+exception BadObject(bad_object);
+
+let emit_object = (mashed: mash_program, outfile) => {
```
spotandjake, line 12: Could we add a small little description of what the object files look like either here or in the contributer docs. Something like whats in the pr description.

compiler/src/codegen/linkedtree.re
```diff
@@ -0,0 +1,215 @@
...
+        [],
+        tree.imports,
+      );
+
+    List.iter(
+      export => {
+        switch (export) {
+        | WasmFunctionExport({ex_function_name, ex_function_internal_name}) =>
+          let internal_name =
+            Printf.sprintf("%s_%d", ex_function_internal_name, dep_id^);
```
spotandjake, line 147: This pattern is used a few times should we make this a function?

compiler/src/typed/cmi_format.re
```diff
@@ -142,22 +138,32 @@ module CmiBinarySection =
...
-    close_in(ic);
+let read_cmi = (ic, filename): cmi_infos => {
+  let read_magic = really_input_string(ic, 4);
+  if (read_magic != magic) {
     raise(Error(Corrupted_interface(filename)));
-  | exception (Error(e)) =>
-    close_in(ic);
-    raise(Error(e));
   };
+  let version_length = input_binary_int(ic);
```
spotandjake, line 146: Would it make sense to move some of this logic into its own module so we can re-use the checks between this file and emit_mod? This would help ensure our checks are staying the same and our object file parsing is the same.

spotandjake (commented):
Mostly looks great to me, I love the new implementation but I have a few questions.

We should probably also update the contributer docs in this pr so they do not get too far out of date.

compiler/src/codegen/emitmod.re
ospencer: Done.

compiler/src/codegen/emitmod.re
ospencer: Done.

compiler/src/typed/cmi_format.re
ospencer: I wanted to do that, but there's some annoying cyclic dependency issues (Grain_codegen depends on Grain_typed, and this would introduce a dependency on Grain_codegen). I admit that this is one annoying place where the two need to be kept in sync, but we'll very very rarely change it and it'll be obviously broken if we do.

spotandjake: Do we maybe want to open an issue for this?

spotandjake (approved):
LGTM, can we also open an issue to update the linking section in the contributer docs, we should probably include the binary format there, and some stuff from the pr description.

compiler/src/typed/cmi_format.re
ospencer: Not particularly; I meant that in a "I don't really plan to do anything about this because I don't think it's a big deal" way. There's no good reason to refactor the whole compiler just for this. The only thing we could do that would be slightly better is implementing the half that doesn't rely on the mashtree somewhere and importing that into emitmod and implementing the other half, but I don't think that's really all that much better.

spotandjake (commented):
This looks great to me, I had one tiny question and it looks like some of the tests are failing currently.

I love that we are using an `exceptionMod` flag it seems a lot less fragile then file paths.

Before we get this merged could we update the static linking contributor docs I think the pr description covers a lot of the content we would want.

compiler/src/codegen/compcore.re
```diff
@@ -658,8 +656,9 @@ let heap_allocate = (wasm_mod, env, num_words: int) =>
                   runtime_heap_next_ptr_name,
                   Type.int32,
                 ),
-                // fake GC refcount of 1
-                Expression.Const.make(wasm_mod, const_int32(1)),
+                // fake GC refcount of 2
+                // this means objects (should) never drop to zero refcount
+                Expression.Const.make(wasm_mod, const_int32(2)),
```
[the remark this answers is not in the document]
ospencer: Allocations in runtime mode happen in the runtime heap and can't ever be freed, but if those objects are ever passed to regular code, reference counting still happens. The bug I uncovered here is that the refcount of runtime objects can fall to 1 if they go out of scope, but that means they get freed. That's a huge issue. The refcount of 2 means that if it ever goes out of scope, the refcount drops to 1 and prevents a free.

compiler/src/compile.re
```diff
@@ -149,6 +149,7 @@ let next_state = (~is_root_file=false, {cstate_desc, cstate_filename} as cs) =>
       Grain_utils.Config.apply_attribute_flags(
         ~no_pervasives=has_attr("noPervasives"),
         ~runtime_mode=has_attr("runtimeMode"),
+        ~no_exception_mod=has_attr("exceptionMod"),
```
alex-snezhko, line 152: What about "noExceptionMod" to keep with the pattern of "noPervasives"? I assume you did this because currently this attribute would only be used on the exception module so it is taken to mean "this is the exception module", but "noExceptionMod" more accurately reflects what the flag is actually doing I feel.

alex-snezhko (approved):

compiler/src/compile.re
ospencer: I had thought of `noExceptions` but I thought that was too ambiguous, but I've come around on it.

ospencer:
@spotandjake I updated the contributor docs.

compiler/src/compile.re
ospencer: Updated.

stdlib/runtime/exception.gr
```diff
@@ -1,5 +1,5 @@
 @noPervasives
-@exceptionMod
+@noExceptions
```
spotandjake, line 2: One tiny NIT, do we just want to add a comment here something like, `// Exceptions cannot be thrown from the exception module or we would have a circular dependency.`?

spotandjake (approved):
This looks great oscar, im excited to get this into main I had one tiny NIT left but otherwise lgtm (as long as tests pass).

marcusroberts (commented):
LGTM!  

I really like the new approach, and the code read fine - I'll not spot any bugs though but looking forward to this being the new way of generating code.

marcusroberts (approved):

phated (approved):
LGTM

PR #2105, opened 2024-05-04
feat(compiler): Enable single-file compilation

ospencer:
Refactors the `DependenciesCompiled`, `ObjectsLinked`, `Compiled`, and `Assembled` states out of `compile.re` and introduces the `grainc` flag `--single-file` to compile a single dependency. The linker is now completely separate from the process of compilation—compiling a file now only means converting source code into an object file. Linking is the combining a number of object files into a wasm file (and `grainc` will still do both compiling and linking). This allows build tools to handle compilation and linking separately.

The future plan is to make single file mode the default once we have a build tool managing building projects (`silo`). I'm not sure that we should get rid of the dependency compilation _entirely_ since it would be useful for development, but that's a decision for later.

### DIFF
--- a/compiler/graindoc/graindoc.re
+++ b/compiler/graindoc/graindoc.re
@@ -47,14 +47,26 @@ type params = {
   current_version: option(string),
 };
 
-let compile_typed = (input: Fp.t(Fp.absolute)) => {
-  switch (
-    Compile.compile_file(
-      ~is_root_file=true,
-      ~hook=stop_after_typed,
-      Filepath.to_string(input),
-    )
-  ) {
+let compile_typed = file => {
+  Module_resolution.load_dependency_graph(file);
+  let to_compile = Module_resolution.get_out_of_date_dependencies();
+  List.iter(
+    file =>
+      ignore(
+        compile_file(
+          ~hook=stop_after_object_emitted,
+          ~outfile=Compile.default_wasm_filename(file),
+          file,
+        ),
+      ),
+    to_compile,
+  );
+  compile_file(~hook=stop_after_typed, file);
+};
+
+let compile = (input: Fp.t(Fp.absolute)) => {
+  reset_compiler_state();
+  switch (compile_typed(Filepath.to_string(input))) {
   | exception exn =>
     let bt =
       if (Printexc.backtrace_status()) {
@@ -180,9 +192,11 @@ let enumerate_runs = opts =>
   };
 
 let graindoc = (opts, runs) => {
+  Config.set_root_config();
+
   List.iter(
     ({input_path, output_path}) => {
-      let program = compile_typed(input_path);
+      let program = compile(input_path);
       try(
         generate_docs(
           ~current_version=opts.current_version,
--- a/compiler/src/language_server/code_file.re
+++ b/compiler/src/language_server/code_file.re
@@ -47,6 +47,20 @@ let warning_to_diagnostic =
   };
 };
 
+let compile = (file, src) => {
+  Module_resolution.load_dependency_graph_from_string(file, src);
+  let to_compile = Module_resolution.get_out_of_date_dependencies();
+  List.iter(
+    file => {
+      ignore(
+        compile_file(~outfile=Compile.default_object_filename(file), file),
+      )
+    },
+    to_compile,
+  );
+  compile_string(~hook=stop_after_typed_well_formed, ~name=file, src);
+};
+
 let compile_source = (uri, source) => {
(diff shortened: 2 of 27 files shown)


marcusroberts (approved):
I read this all through, it made sense, code was clear, LGTM!

compiler/src/typed/dependency_graph.re
```diff
@@ -106,45 +106,27 @@ module Make = (DV: Dependency_value) => {
...
-        ((dep, state), acc) => {
-          switch (acc) {
-          | (true, _) => acc
-          | (false, Some(_)) => acc
-          | (false, None) =>
-            let stop_found =
-              switch (stop) {
-              | Some(d) when DV.equal(d, dep) => true
-              | _ => false
-              };
```
alex-snezhko, line 121: I assume the code you removed here was due to the `stop_found` not being used but is it not important?

compiler/src/typed/env.re
```diff
@@ -714,27 +714,14 @@ let add_import = s => {
...
-  imported_units := old_imported_units;
-  imported_opaque_units := old_opaque_units;
-  ret;
 };
 
-let clear_imports = () => {
+let clear_persistent_structures = () => {
   Consistbl.clear(crc_units);
   Hashtbl.clear(persistent_structures);
   imported_units := StringSet.empty;
```
alex-snezhko, line 724: Nit: `clear_imports`

compiler/src/typed/env.re
```diff
@@ -714,27 +714,14 @@ let add_import = s => {
   imported_units := StringSet.add(s, imported_units^);
 };
 
-let imported_opaque_units = ref(StringSet.empty);
```
alex-snezhko, line 717: Could you explain why you removed these opaque unit imports?

compiler/src/typed/typemod.re
```diff
@@ -1051,6 +1051,7 @@ let initial_env = () => {
 let type_implementation = (prog: Parsetree.parsed_program) => {
   let sourcefile = prog.prog_loc.loc_start.pos_fname;
   let module_name = prog.module_name.txt;
+  Env.clear_imports();
```
alex-snezhko, line 1054: Why is this needed now?

alex-snezhko (approved):
Was a bit unclear about some of the changes to the type checking but LGTM overall

compiler/graindoc/graindoc.re
```diff
@@ -47,14 +47,26 @@ type params = {
...
 
-let compile_typed = (input: Fp.t(Fp.absolute)) => {
-  switch (
-    Compile.compile_file(
-      ~is_root_file=true,
-      ~hook=stop_after_typed,
-      Filepath.to_string(input),
-    )
-  ) {
+let compile_typed = file => {
```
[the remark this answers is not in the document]
ospencer: They're all slightly different between the hooks (which are parametrizable) and compile functions (which aren't really). We could try to find some commonality, but it's mostly just the `load_dependency_graph` and `get_out_of_date_dependencies`; after that the caller is responsible for compiling those any way they see fit. I think it's fine that they're separate, personally.

compiler/src/language_server/code_file.re
```diff
@@ -56,12 +70,7 @@ let compile_source = (uri, source) => {
     Config.preserve_config(() => {
       // Warnings will be reported in diagnostics so no need to print them
       Config.print_warnings := false;
-      Compile.compile_string(
-        ~is_root_file=true,
-        ~hook=stop_after_typed_well_formed,
-        ~name=filename,
-        source,
-      );
+      compile(filename, source);
```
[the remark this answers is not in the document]
ospencer: Good question. It wasn't before, presumably with the idea that while the LSP is running it's compiling the same stuff. We should probably look into that, but that's a separate issue.

compiler/src/typed/dependency_graph.re
ospencer: It was necessary before because it would iterate to find the next dependency that was out of date so it could be worked on. Now, we just grab all of them at once.

compiler/src/typed/env.re
ospencer: It's just old cruft from back in the day.

compiler/src/typed/typemod.re
ospencer: This basically used to happen as a part of the full compile pipeline, but now that things are moved around, we reset it where it's relevant.

PR #2106, opened 2024-05-04
fix(compiler): Don't trigger recompilation for unnecessary compiler flags

ospencer:


marcusroberts (approved):
Approval from me, I did my best to decide if your digestable/notdigestable was correct, but trust you on it!

compiler/src/utils/config.re
```diff
@@ -353,17 +360,24 @@ let stdlib_dir =
     ~conv=option_conv(Cmdliner.Arg.string),
     ~doc="Path to the standard library (stdlib) directory",
     ~env="GRAIN_STDLIB",
+    ~digestable=Digestable,
```
alex-snezhko, line 363: Could you remind me why the stdlib option would be `Digestable` but include_dirs is not?

ospencer: It doesn't need to be, good catch.

ospencer:
@alex-snezhko Made those changes.

alex-snezhko (approved):

PR #2108, opened 2024-05-22
feat(stdlib): Add Mimetypes module

JeysonFlores:
Hey there! This is my first contribution to the Grain project so I wanted to start with something easy to write, debug and review. This module is just a collection of mimetypes and formats with some entrypoints to access to them with a pretty general default behaviour.

Details:
- The list of MIMETypes and format is extracted from [this project](https://github.com/jshttp/mime-db?tab=readme-ov-file) which itself is composed using the media types from Apache, IANA and NGINX.
- The getters return "text/plain" and "txt" when a format or mimetype is not found. This is a pretty standard behaviour for multiple standard libraries across many programming languages.

I hope the value of this module increases in the future when networking modules are added to the standard library. By the time being I think this is a good start.

Thanks!

Note: I just put an arbitrary version in the source code. Feel free to change it


ospencer:
Very cool! We'll discuss this tomorrow during our community call—it'd be great if you could make it (we have them on Discord—you should be able to see the event there)!

ospencer:
To recap what we discussed on the community call: we don't think this library makes sense to be a part of the standard library, but it would make an awesome community package! If you made this its own repository, it'd be great to feature it in https://github.com/grain-lang/awesome-grain!

PR #2111, opened 2024-05-22
chore(stdlib): Removed redundant for-loop in `Set.intersect`

### LINKED ISSUE #2110: Redundant Operations in stdlib Set.intersect
The method [`Set.intersect`](https://github.com/grain-lang/grain/blob/fc19761e9f7e6c3d1cb54a308de721b17d25e12e/stdlib/set.gr#L441) is implemented as:
```
provide let intersect = (set1, set2) => {
  let set = make()
  forEach(key => {
    if (contains(key, set2)) {
      add(key, set)
    }
  }, set1)
  forEach(key => {
    if (contains(key, set1)) {
      add(key, set)
    }
  }, set2)
  set
}
```
Isn't the second `forEach`-loop redundant? If there is a `key` in `set2` such that `contains(key, set1)==true`, then this key would already have been added to the intersection `set` in the first iteration, because then `contains(key, set2)==true` as well? Or is the equality-relation on keys not necessarily symmetric?

In other words, are there sets `set1, set2` such that the second loop adds elements to `set` that the first loop didn't?

lumi-a:
Fixes #2110. I didn't compile nor test this change.

### DIFF
--- a/stdlib/set.gr
+++ b/stdlib/set.gr
@@ -445,11 +445,6 @@ provide let intersect = (set1, set2) => {
       add(key, set)
     }
   }, set1)
-  forEach(key => {
-    if (contains(key, set1)) {
-      add(key, set)
-    }
-  }, set2)
   set
 }
 


spotandjake (approved):
Looks good to me

ospencer (approved):

PR #2125, opened 2024-07-26
feat(lsp): Explicit type annotation code action

alex-snezhko:
Screenshot from 2024-07-25 21-38-55

Screenshot from 2024-07-25 21-39-01

Screenshot from 2024-07-25 21-39-09

### DIFF
--- a/compiler/src/language_server/code_action.re
+++ b/compiler/src/language_server/code_action.re
@@ -38,16 +38,16 @@ let send_no_result = (~id: Protocol.message_id) => {
   Protocol.response(~id, `Null);
 };
 
-let explicit_type_annotation = (range, uri, version, type_str) => {
+let explicit_type_annotation = (range, uri, type_str) => {
   ResponseResult.{
-    title: "Explicitly annotate type",
-    kind: "annotate-types",
+    title: "Annotate type",
+    kind: "annotate-type",
     edit: {
       document_changes: [
         {
           text_document: {
             uri,
-            version: Some(version),
+            version: None,
           },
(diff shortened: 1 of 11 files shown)


marcusroberts (approved):
Great work Alex!  A really useful feature to add too, thank you!

compiler/src/language_server/code_action.re
```diff
@@ -0,0 +1,111 @@
...
+  type t = option(list(code_action));
+};
+
+let send_no_result = (~id: Protocol.message_id) => {
+  Protocol.response(~id, `Null);
+};
+
+let explicit_type_annotation = (range, uri, version, type_str) => {
+  ResponseResult.{
+    title: "Explicitly annotate type",
```
ospencer, line 43: What do you think of just `Annotate type`? If there's no type annotation, then adding one would just be an annotation.

ospencer (commented):
Two questions for you: what's with the version additions, and what happens if you use this code action on something that already has an annotation?

compiler/src/language_server/code_action.re
alex-snezhko: Changed to this

alex-snezhko:
@ospencer I got rid of the version stuff and it doesn't look like it made a difference actually (at least in VSCode's LSP client); I saw in the LSP documentation that the version was one of the response properties so that's why I implemented it initially.

The code action dialog show no longer show if it's tried on something that's already annotated; I realized that there was actually a flaw in my implementation which sometimes showed the code action dialog on something that was already annotated, but I patched it.

ospencer (approved):
Looks great!

PR #2126, opened 2024-07-28
feat(lsp): Use argument label code action

alex-snezhko:
Screenshot from 2024-07-27 19-24-45

Screenshot from 2024-07-27 19-24-55

Screenshot from 2024-07-27 19-25-12

Screenshot from 2024-07-27 19-25-20


ospencer (commented):
What happens if you try to use this action on an argument that already has a label?

alex-snezhko:
@ospencer the coffee action would only show if there is no label

ospencer (approved):

PR #2130, opened 2024-07-29
chore(stdlib): Optimize `List.join`

spotandjake:
This pr optimizes `List.join` similarly to how `Array.join` was optimized in #1948, this change makes `List.join` use only a single allocation, and should reduce the number of times we need to copy while reallocating. 

Closes: #728

### DIFF
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -848,19 +848,20 @@ provide let sub = (start, length, list) => {
 
 // List.join helpers
 @unsafe
-let rec iterSize = (sepSize: WasmI32, size: WasmI32, lst: List<String>) => {
+let rec computeJoinSize = (sepSize: WasmI32, size: WasmI32, lst: List<String>) => {
   use WasmI32.{ (+) }
   use DataStructures.{ stringSize }
   match (lst) {
     [] => size,
+    [hd] => size + stringSize(WasmI32.fromGrain(hd)),
     [hd, ...tl] => {
       let size = size + stringSize(WasmI32.fromGrain(hd)) + sepSize
-      iterSize(sepSize, size, tl)
+      computeJoinSize(sepSize, size, tl)
     },
   }
 }
 @unsafe
-let rec iterBuildString = (
+let rec buildJoinedString = (
   strPtr: WasmI32,
   sepPtr: WasmI32,
   sepSize: WasmI32,
@@ -871,20 +872,21 @@ let rec iterBuildString = (
   use DataStructures.{ stringSize }
   match (lst) {
     [] => void,
-    // Last Element
+    // Last element
     [hd] => {
       let ptr = WasmI32.fromGrain(hd)
       let size = stringSize(ptr)
       Memory.copy(offset, ptr + 8n, size)
+      ignore(hd)
     },
-    // Regular Path
     [hd, ...tl] => {
       let ptr = WasmI32.fromGrain(hd)
       let size = stringSize(ptr)
       Memory.copy(offset, ptr + 8n, size)
+      ignore(hd)
       let offset = offset + size
       Memory.copy(offset, sepPtr, sepSize)
-      iterBuildString(strPtr, sepPtr, sepSize, offset + sepSize, tl)
+      buildJoinedString(strPtr, sepPtr, sepSize, offset + sepSize, tl)
     },
   }
 }
@@ -904,10 +906,10 @@ provide let join = (separator: String, list: List<String>) => {
   use DataStructures.{ allocateString, stringSize }
   let sepPtr = WasmI32.fromGrain(separator)
   let sepSize = stringSize(sepPtr)
-  let strSize = iterSize(sepSize, 0n - sepSize, list)
-  if (strSize <= 0n) return ""
+  let strSize = computeJoinSize(sepSize, 0n, list)
   let newString = allocateString(strSize)
-  iterBuildString(newString, sepPtr + 8n, sepSize, newString + 8n, list)
+  buildJoinedString(newString, sepPtr + 8n, sepSize, newString + 8n, list)
+  ignore(sepPtr)
   return WasmI32.toGrain(newString): String
 }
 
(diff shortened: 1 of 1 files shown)


stdlib/list.gr
```diff
@@ -842,6 +846,48 @@ provide let sub = (start, length, list) => {
   take(length, drop(start, list))
 }
 
+// List.join helpers
+@unsafe
+let rec iterSize = (sepSize: WasmI32, size: WasmI32, lst: List<String>) => {
```
ospencer, line 851: Could we instead call this `computeJoinSize` or something like that?

stdlib/list.gr
```diff
@@ -842,6 +846,48 @@ provide let sub = (start, length, list) => {
...
+  use WasmI32.{ (+) }
+  use DataStructures.{ stringSize }
+  match (lst) {
+    [] => size,
+    [hd, ...tl] => {
```
ospencer, line 856: If you add a case where there's only one list element, you can avoid having to pass a negative value for `size`. The case with just one element adds only the size of that element and the case with more than one adds the size plus the separator size and recurses.

stdlib/list.gr
```diff
@@ -842,6 +846,48 @@ provide let sub = (start, length, list) => {
...
+      let size = size + stringSize(WasmI32.fromGrain(hd)) + sepSize
+      iterSize(sepSize, size, tl)
+    },
+  }
+}
+@unsafe
+let rec iterBuildString = (
```
ospencer, line 863: Could we call this something like `joinHelp` or `buildJoinedString`?

stdlib/list.gr
```diff
@@ -842,6 +846,48 @@ provide let sub = (start, length, list) => {
...
+  sepPtr: WasmI32,
+  sepSize: WasmI32,
+  offset: WasmI32,
+  lst: List<String>,
+) => {
+  use WasmI32.{ (+) }
+  use DataStructures.{ stringSize }
+  match (lst) {
+    [] => void,
+    // Last Element
```
ospencer, line 874:
```suggestion
    // Last element
```

stdlib/list.gr
```diff
@@ -842,6 +846,48 @@ provide let sub = (start, length, list) => {
...
+    [hd] => {
+      let ptr = WasmI32.fromGrain(hd)
+      let size = stringSize(ptr)
+      Memory.copy(offset, ptr + 8n, size)
+    },
+    // Regular Path
```
ospencer, line 880:
```suggestion
```

stdlib/list.gr
```diff
@@ -842,6 +846,48 @@ provide let sub = (start, length, list) => {
...
+) => {
+  use WasmI32.{ (+) }
+  use DataStructures.{ stringSize }
+  match (lst) {
+    [] => void,
+    // Last Element
+    [hd] => {
+      let ptr = WasmI32.fromGrain(hd)
+      let size = stringSize(ptr)
+      Memory.copy(offset, ptr + 8n, size)
```
ospencer, line 878: The hd of the list could get prematurely collected before the `Memory.copy` happens.
```suggestion
      Memory.copy(offset, ptr + 8n, size)
      ignore(hd)
```

stdlib/list.gr
```diff
@@ -842,6 +846,48 @@ provide let sub = (start, length, list) => {
...
+    },
+    // Regular Path
+    [hd, ...tl] => {
+      let ptr = WasmI32.fromGrain(hd)
+      let size = stringSize(ptr)
+      Memory.copy(offset, ptr + 8n, size)
+      let offset = offset + size
+      Memory.copy(offset, sepPtr, sepSize)
+      iterBuildString(strPtr, sepPtr, sepSize, offset + sepSize, tl)
```
ospencer, line 887:
```suggestion
      iterBuildString(strPtr, sepPtr, sepSize, offset + sepSize, tl)
      ignore(hd)
```

stdlib/list.gr
```diff
@@ -852,25 +898,17 @@ provide let sub = (start, length, list) => {
...
-  // Reverse and reduce to take advantage of TCE
-  match (iter(separator, None, reverse(list))) {
-    None => "",
-    Some(s) => s,
-  }
+  use WasmI32.{ (+), (-), (<=) }
+  use DataStructures.{ allocateString, stringSize }
+  let sepPtr = WasmI32.fromGrain(separator)
+  let sepSize = stringSize(sepPtr)
+  let strSize = iterSize(sepSize, 0n - sepSize, list)
```
ospencer, line 907: With the changes I suggested this would just be `0n`.

stdlib/list.gr
```diff
@@ -852,25 +898,17 @@ provide let sub = (start, length, list) => {
...
+  if (strSize <= 0n) return ""
```
ospencer, line 908: And this check wouldn't be needed.

ospencer (changes requested):
Looks pretty good so far, just needs a few tweaks!

stdlib/list.gr
spotandjake: I should put the `ignore` above the function call right otherwise it wont be a tail call?

spotandjake:
Those changes have been commited.

stdlib/list.gr
```diff
@@ -842,6 +846,50 @@ provide let sub = (start, length, list) => {
...
+// List.join helpers
+@unsafe
+let rec computeJoinSize = (sepSize: WasmI32, size: WasmI32, lst: List<String>) => {
+  use WasmI32.{ (+) }
+  use DataStructures.{ stringSize }
+  match (lst) {
+    [] => size,
+    [hd] => size + stringSize(WasmI32.fromGrain(hd)),
+    [hd, ...tl] => {
+      let size = size + stringSize(WasmI32.fromGrain(hd)) + sepSize
```
ospencer, line 858:
```suggestion
      let size = size + stringSize(WasmI32.fromGrain(hd)) + sepSize
      ignore(hd)
```

ospencer (approved):

Issue #387, opened 2020-10-15
Type annotation shorthands?


ospencer:
Type annotations can get unwieldy pretty fast, with types like `List<Option<Number>>`. I think it would be pretty cool to have shorthands for common types, like `[]` for lists, maybe `?` for options, etc.

With these shorthands, this type could be written as `Number?[]`.

I'm not hard set on these particular shorthands, but I'm curious to hear what people think!

phated:
Would type aliases help here? Since they wouldn't be exported, you could do something like:

```
type MaybeNumber = Option<Number>;

export type NumberList = List<MaybeNumber>;
```

It's definitely more verbose, but I don't like things to become "character soup", as I feel most functional languages start to become.

ospencer:
Type aliases definitely help. You'd have to also export `MaybeNumber` here though since it'd be an opaque type otherwise.

I definitely feel where you're coming from with the character soup. I feel like as long as it's fairly limited it can be okay, though. I really like how writing `int[]` feels in languages like C, TypeScript, and Go.

phated:
Oh, something else that might get confusing is if we add `?` as an "unwrap-or-fail" operator. Then, using `?` in type signatures as the Option type would be weird.

I'm also now thinking that using `?` for option doesn't really fit well with Result being an oft-used return type too. Hmm 🤔

ospencer:
Yeah, I tried thinking about something short for `Result`, but it almost feels impossible since you've always got to give it two type params.

chriskrycho:
I actually don’t *love* `?` for `Option` in TS or Swift or C♯. It tends to lead people down the Dark Path™ of sprinkling it everywhere instead of actually thinking about the right model for their data. I find `[]` for array to be a lot less of a nuisance that way.

phated:
I was spinning on this a lot last night and if `?` represents optional data, I'd want it to both be the type annotation AND an alias for `Some(val)`, like `maybeString: () -> ?string` and `maybeString = () => ?"hello world"`. So prefix would be the convenience for wrapping and postfix would be the unwrap-or-throw helper.

sanathusk:
I feel using shorthand for common operation definitely helps.it makes the code more readable.I think typescript has done this really well. We could use  ` ? ` for optional values, `! `To force unwrap. ?. To unwrap only if values is not null would be helpful

sanathusk:
[...]

PR #2132, opened 2024-08-02
feat(compiler): Add support for `Number[]` and `Number[>]` syntax

spotandjake:
This pr adds shorthand syntax for `list` and `array` types which is discussed in #387, I think if we want the `?` and `!` shorthands we should have some more discussion, but the list and array syntax seem generalized enough that it is worth having. As far as I am aware this pr is non breaking because the old syntax still works. The syntax can be seen below:
* `List<Number>` becomes `Number[]`
* `Array<Number>` becomes `Array[>]`

This pr is not ready to be merged as I still need to add tests, make sure the formatter handles this correctly and make sure types get stringified using the shorthand in the lsp and graindoc. I am putting this pr before doing that as I want to make sure that we are okay with doing handling this in the parser instead of adding a parseTree element, given this is syntax sugar it seemed like the best option.

Currently this pr is not breaking as far as I am aware given we do not remove the `List<Number>` syntax and this would have been completely invalid syntax before.

### DIFF
--- a/compiler/src/parsing/ast_helper.re
+++ b/compiler/src/parsing/ast_helper.re
@@ -89,6 +89,10 @@ module Type = {
   let var = (~loc, a) => mk(~loc, PTyVar(a));
   let arrow = (~loc, a, b) => mk(~loc, PTyArrow(a, b));
   let tuple = (~loc, a) => mk(~loc, PTyTuple(a));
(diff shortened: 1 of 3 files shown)


spotandjake:
Closing this because of the decision made on #387 to not have shorthands.

PR #2136, opened 2024-08-07
feat(compiler)!: Disallow whitespace for negative literals + allow infix operators on new line

alex-snezhko:
```grain
let x = 3
- 2
print(x) // 1

let x = 2
+1
-1 // Will give warning to ask for either a space after `-`, wrap in parens, or semicolon on prev line
print(x) // 3

let (|>) = (x, f) => f(x)
let a = [1, 2, 2, 3]
  |> List.reverse
  |> List.unique
```


alex-snezhko:
@ospencer @phated I implemented the changes we discussed

ospencer (commented):
I'm not a fan of the changes to the parsetree. We should be able to have a fully semantic program without needing to read these extra properties, so that probably means we're not doing enough during parsing. We should be able to produce the error we want in the wrapped lexer.

compiler/test/grainfmt/application_indenting.expected.gr
```diff
@@ -3,22 +3,22 @@ module ApplicationIndenting
...
   let riiiiiiightsInheriting = Module64.load(structPtr, 16n)
   (
-    riiiiiiightsInheriting &
-    1N << Module64.extendI32U(Module32.fromGrain(i) >> 1n)
-  ) >
-    0N
+    riiiiiiightsInheriting
+    & 1N << Module64.extendI32U(Module32.fromGrain(i) >> 1n)
+  )
+    > 0N
```
spotandjake, line 9: I feel like this formatting looks a little weird how would we feel about placing the operator on the top line in cases where a parens come first?

compiler/test/grainfmt/infix.expected.gr
```diff
@@ -10,16 +10,35 @@ let trimString = (str: String, end: Bool) => {
...
     if (
       // Spacing
-      currentChar != '\u{0009}' && // Tab
-      currentChar != '\u{000B}' && // LINE TABULATION
-      currentChar != '\u{000C}' && // FORM FEED (FF)
-      currentChar != '\u{0020}' && // Space
-      currentChar != '\u{00A0}' && // No Break Space
-      currentChar != '\u{FEFF}' && // ZERO WIDTH NO-BREAK SPACE
+      currentChar != '\u{0009}'
+      && // Tab
```
spotandjake, line 14: This comment formating doesn't seem right.

stdlib/hash.gr
```diff
@@ -121,8 +121,8 @@ let rec hashOne = (val, depth) => {
   } else if ((val & Tags._GRAIN_NUMBER_TAG_MASK) != 0n) {
     hash32(val)
   } else if (
-    (val & Tags._GRAIN_GENERIC_TAG_MASK) ==
-    Tags._GRAIN_GENERIC_HEAP_TAG_TYPE
+    (val & Tags._GRAIN_GENERIC_TAG_MASK)
+    == Tags._GRAIN_GENERIC_HEAP_TAG_TYPE
```
spotandjake, line 125: Reading through all the formatted code, I find it looks really wierd with the `==` on a newline mostly because its not often that you chain equals together I wonder if we should preserve the old formatting style for `==`.

stdlib/string.gr
```diff
@@ -2184,7 +2184,8 @@ let trimString = (stringPtr: WasmI32, byteLength: WasmI32, fromEnd: Bool) => {
     // TODO(#661): Use unicode whitespace property and unicode line terminator
     if (!fromEnd) {
       if (
-        byte == 0xEFn && // Check for the first BOM byte
+        byte == 0xEFn
+        && // Check for the first BOM byte
```
spotandjake, line 2188: This comment being on the line below really feels like it changes the meaning of it.  There has been a few cases like this.

spotandjake (commented):
Mostly looks good to me, I had a few formatting nits.

alex-snezhko:
@ospencer wouldn't that basically involve having an entire parser within wrapped_lexer? You'd need to know when you encounter a full expression within wrapped_lexer to know whether or not to give the warning

alex-snezhko:
@ospencer bump on the parser comment; let me know if I'm missing something

ospencer (commented):
I still don't like the changes to the parsetree, but I'd be fine if all of that information was consolidated into a `pexp_meta` field.

compiler/test/grainfmt/application_indenting.expected.gr
alex-snezhko: Hmm why should existence of parens change how the operator is formatted?

spotandjake: It's not the parens its just that the line above only contains `)` it seems silly to break it onto a new line here.

stdlib/hash.gr
alex-snezhko: Hmm not a fan of treating this one operator differently from the rest

stdlib/string.gr
alex-snezhko: How would you prefer this to be handled? Moving the comment to the line above? I don't really like that because it changes the order from how you wrote it

compiler/test/grainfmt/infix.expected.gr
alex-snezhko: Wdym?

alex-snezhko:
Ok added @ospencer

stdlib/string.gr
spotandjake: Is there any way we could prevent the line break? In this case there isn't really a reason we need to break the `&&` to the newline. I 100% agree about moving the comment.

compiler/test/grainfmt/infix.expected.gr
spotandjake: Same issue as in the `string` module.

stdlib/string.gr
alex-snezhko: Idk I'd think it's better to be consistent, if we're moving binary operators to the next line I think it'd be better to always do that rather than sometimes not

spotandjake: I agree on that point, maybe it would be worth just manually correcting the placement in the stdlib, as I think there are a few places where the meaning of the comment has changed quite a bit.

ospencer (approved):
Looks awesome, and I'm so much happier with this implementation. Looks like you've just got a typo in one of the tests.

PR #2140, opened 2024-08-13
feat(stdlib): Add `String.repeat` to `String` module

spotandjake:
This pr adds `String.repeat` to the standard library. 

Should this be called `String.make`? It would align better with the `Array` library and `Array.make` the behaviour is similar.


stdlib/string.gr
```diff
@@ -2338,3 +2338,45 @@ provide let toAsciiUppercase = string => {
...
+provide let repeat = (string, count) => {
+  use WasmI32.{ (+), (*), (<), (&), (!=) }
+  if ((WasmI32.fromGrain(count) & 1n) != 1n) {
+    throw InvalidArgument("Invalid count value")
+  }
+  let mut rawCount = untagSimpleNumber(count)
+  if (WasmI32.eqz(rawCount)) return ""
+  let string = if (rawCount < 0n) {
+    rawCount *= -1n
+    reverse(string)
```
ospencer, line 2367: Do other languages do this? This seems like really surprising behavior—I'd probably just want an error.

spotandjake: So checking a few other languages,
``js
"=".repeat(-5)
```
results in a range error.
```py
"=" * -5
```
results in an empty string


I decided to make it flip the string so we handled the behaviour instead of throwing an exception, but I think I am going to make it throw an exception following the Javascript behaviour.

stdlib/string.gr
```diff
@@ -38,6 +38,8 @@ provide enum Encoding {
 
 exception MalformedUnicode
 
+exception RangeError
```
ospencer, line 41: I don't think we should have a new exception. This should just throw `InvalidArgument`. I'm also fine with the empty string behavior, but I think throwing is a little better.

spotandjake: I had already implemented it using `InvalidArgument` I had just forgot to remove the `RangeError` exception I added before making that decision. 

Made that fix now though.

ospencer:
> Should this be called `String.make`? It would align better with the `Array` library and `Array.make` the behaviour is similar.

I personally think that `String.make` would be a little confusing. I would expect a `String.make` to be composed of characters instead of substrings. I think `String.repeat` is very descriptive. I'd welcome a second opinion, though.

stdlib/string.gr
```diff
@@ -2338,3 +2338,42 @@ provide let toAsciiUppercase = string => {
...
+ * @since v0.6.7
+ */
+@unsafe
+provide let repeat = (string, count) => {
+  use WasmI32.{ (+), (*), (<), (&), (!=) }
+  if ((WasmI32.fromGrain(count) & 1n) != 1n) {
+    throw InvalidArgument("Invalid count value")
+  }
+  let mut rawCount = untagSimpleNumber(count)
+  if (WasmI32.eqz(rawCount)) return ""
```
ospencer, line 2364: I personally wouldn't have these `eqz` checks. They make the implementation more complex with little benefit (people are likely calling this function on non-empty strings with count > 0 a vast majority of the time). The code still produces the same result without them.

stdlib/string.gr
```diff
@@ -2338,3 +2338,42 @@ provide let toAsciiUppercase = string => {
...
+  if (rawCount < 0n) {
+    throw InvalidArgument("Invalid count must be a positive integer or zero")
+  }
+  let mut stringPtr = WasmI32.fromGrain(string)
+  let byteLength = WasmI32.load(stringPtr, 4n)
+  if (WasmI32.eqz(byteLength)) return ""
+  stringPtr += 8n
```
ospencer, line 2371: Since we don't ever mutate this again, it's probably better to just make a new binding, `let stringPtr = stringPtr + 8n`.

stdlib/string.gr
```diff
@@ -2338,3 +2338,42 @@ provide let toAsciiUppercase = string => {
...
+  let strPtr = allocateString(byteLength * rawCount)
```
ospencer, line 2372: We should rename this to something more clear, this is too similar to `stringPtr`

spotandjake:
I applied those changes

stdlib/string.gr
```diff
@@ -2338,3 +2338,40 @@ provide let toAsciiUppercase = string => {
   }
   implode(chars)
 }
+
+/**
+ * Produces a new string by repeating a given substring a given number of times.
```
ospencer, line 2343:
```suggestion
 * Produces a new string by repeating a substring a given number of times.
```

stdlib/string.gr
```diff
@@ -2338,3 +2338,40 @@ provide let toAsciiUppercase = string => {
...
+  }
+  let stringPtr = WasmI32.fromGrain(string)
+  let byteLength = WasmI32.load(stringPtr, 4n)
+  let stringPtr = stringPtr + 8n
+  let newStringPtr = allocateString(byteLength * rawCount)
+  let strContentPtr = newStringPtr + 8n
+  for (let mut i = 0n; i < rawCount; i += 1n) {
+    Memory.copy(strContentPtr + byteLength * i, stringPtr, byteLength)
+  }
+  ignore(stringPtr)
```
ospencer, line 2375:
```suggestion
  ignore(string)
```

ospencer:
@spotandjake Can you regen the docs?

spotandjake:
I pushed the updated graindoc

ospencer (approved):

PR #2142, opened 2024-08-14
feat(compiler): Ignore compiler warnings

### LINKED ISSUE #163: Allow setting compiler warnings from flags
We should be able to configure which warnings we want to be emitted via compiler flags. See #158

alex-snezhko:
```grain
match (true) {
  true => void
}
let a = [> 1, 2]
a[1.5]
```
```bash
# will warn
grain compile main.gr
# will not warn
grain compile main.gr --ignore-warnings=all
grain compile main.gr --ignore-warnings=partialMatch,arrayIndexNonInteger
```

Closes #163

### DIFF
--- a/cli/bin/grain.js
+++ b/cli/bin/grain.js
@@ -153,6 +153,12 @@ class GrainCommand extends commander.Command {
       "--verbose",
       "print critical information at various stages of compilation"
     );
+    cmd.forwardOption(
+      "--ignore-warnings <warnings>",
+      "compiler warnings to ignore",
+      list,
+      []
+    );
     return cmd;
   }
 }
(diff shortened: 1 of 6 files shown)


ospencer:
We discussed this on Discord and decided that for now, we don't want the command line flag. For what's needed for Silo, it just needs to be programmatically configurable, but ideally we'd allow disabling a warning using a comment.

alex-snezhko:
@ospencer do we want comments or attributes for this?

ospencer:
Attributes would be a little nicer, but we don't currently support attributes on arbitrary nodes, and that's actually kind of a hard problem. If you want to pair on trying to make that happen we can do that in another PR.

Issue #1334, opened 2022-06-14
`Exception.print()`


ospencer:
Printers can be registered for exceptions, but these values are only shown when an error is thrown. `Exception.print()` should return the string value of an exception.

phated:
I'd probably want this to be `Exception.toString()` because `print` sounds like it'll be printed on stdout or stderr.

spotandjake:
Closing as this was completed with: #2143

PR #2143, opened 2024-08-15
feat(stdlib): Add `Exception.toString`

spotandjake:
This pr adds `Exception.toString` to the `Exception` module closing #1334 

Implementation notes, I decided to rework the way we handle exception printers, moving `panic` into its own module in `runtime/unsafe` and removing exception as a dependecy of `gc` and `malloc`. This allows us to use `Exception` in a regular mode and avoid a lot of the memory leaks we had before. This also allows us to address the `AssertionError` comment in the `Exception` file because we can allocate strings safely now. 

If we are not happy with this implementation I am happy to go back to how we handled things before and implement `Exception.toString` the way we did it before.


ospencer:
I'm not against reworking the libraries in this way, but it borders on being a breaking change. It's only not technically breaking because there isn't a way to catch exceptions. I'd be more comfortable releasing this in a breaking release.

stdlib/runtime/unsafe/panic.gr
```diff
@@ -0,0 +1,28 @@
+@runtimeMode
+module Panic
+
+from "runtime/unsafe/wasmi32" include WasmI32
+use WasmI32.{ (==), (+), (-) }
```
ospencer, line 5:
```suggestion
use WasmI32.{ (+) }
```

stdlib/exception.gr
```diff
@@ -41,11 +41,16 @@ from "runtime/exception" include Exception
...
-  void
+  Exception.registerPrinter(printer)
 }
+
+/**
+ * Gets the string representation of the given exception.
+ *
+ * @param e: The exception to stringify
+ *
+ * @returns The string representation
```
ospencer, line 52:
```suggestion
 * @returns The string representation of the exception
```

stdlib/exception.gr
```diff
@@ -41,11 +41,16 @@ from "runtime/exception" include Exception
...
+ *
+ * @since v0.6.7
+ */
+provide let toString = e => Exception.toString(e)
```
ospencer, line 56:
```suggestion
provide let toString = Exception.toString
```

stdlib/exception.gr
```diff
@@ -41,11 +41,16 @@ from "runtime/exception" include Exception
  */
 @disableGC
```
ospencer, line 42:
```suggestion
```
Honestly, this could just be a re-export like `toString` is.

stdlib/runtime/exception.gr
```diff
@@ -1,96 +1,72 @@
...
-  WasmI32.store(iov, 1n, 12n)
-  fd_write(2n, iov, 2n, written)
-  unreachable()
+  exceptionToString(e, printers)
 }
 
+/**
+ * Throws an exception.
+ *
+ * @param e: The exception to throw
```
ospencer, line 62:
```suggestion
 * Throws an uncatchable exception and traps.
 *
 * @param e: The exception to throw
```

stdlib/runtime/gc.gr
```diff
@@ -21,21 +21,19 @@ module GC
...
 primitive (&&) = "@and"
 primitive (||) = "@or"
-primitive throw = "@throw"
 primitive ignore = "@ignore"
 primitive box = "@box"
 primitive unbox = "@unbox"
 
-exception DecRefError
+let throwDecRefError = () => {
+  Panic.panic("DecRefError: Attempted to decrement a zero reference count")
```
ospencer, line 35:
```suggestion
  Panic.panic("DecRefError: Reference count of zero")
```

spotandjake:
Made those changes

ospencer:
Looks like you need to update the smallest program test.

spotandjake:
Fixed

ospencer (approved):

PR #2146, opened 2024-08-16
fix(lsp): Use correct origin selection range

alex-snezhko:
Holding control highlights the correct origin

Before:
Screenshot from Screencast from 2024-08-15 21-15-36 webm

After:
Screenshot from Screencast from 2024-08-15 21-14-46 webm


ospencer:
I don't quite understand what this is. Can you explain?

alex-snezhko:
@ospencer I updated the description with visuals including the mouse position if that clarifies things. When you hold down the control key, VS Code surfaces this bug by showing the origin selection range as the thing you are targeting instead

ospencer (approved):

PR #2148, opened 2024-08-16
feat(runtime): Simplify `copy` and `fill` polyfills in `Memory` module

spotandjake:
This pr simplifies the `Memory.copy` and `Memory.fill` polyfills in `Memory.gr`, I implemented them using a `for` loop as it means less mutable variables compared to the while loop approach and makes the code a lot easier to follow.

Additionally this removes a few operations per cycle on the functions as we are not modifying as many values every iteration.

### DIFF
--- a/stdlib/runtime/unsafe/memory.gr
+++ b/stdlib/runtime/unsafe/memory.gr
@@ -4,10 +4,17 @@ module Memory
 from "runtime/gc" include GC
 use GC.{ malloc, free, incRef, decRef }
 from "runtime/unsafe/wasmi32" include WasmI32
-use WasmI32.{ (+), (-), (<<), (==), (!=), ltU as (<), gtU as (>) }
+use WasmI32.{ (+), (-), (!=), ltU as (<), gtU as (>) }
 
 provide { malloc, free, incRef, decRef }
 
+/**
+ * Copies the given memory region into the given destination.
+ *
+ * @param dest: The destination memory region
+ * @param src: The source memory region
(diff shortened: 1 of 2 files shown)


stdlib/runtime/unsafe/memory.gr
```diff
@@ -4,38 +4,42 @@ module Memory
...
+use WasmI32.{ (+), (-), (!=), ltU as (<), gtU as (>) }
 
 provide { malloc, free, incRef, decRef }
 
-provide let copy = (dest, src, n) => {
-  let mut dest = dest
-  let mut src = src
-  let mut n = n
+/**
+ * Copies the given memory region into the given destination.
```
ospencer, line 12:
```suggestion
 * Copies the source memory region to the destination memory region. Regions may overlap.
```

stdlib/runtime/unsafe/memory.gr
```diff
@@ -4,38 +4,42 @@ module Memory
...
 
-provide let fill = (dest, c, n) => {
-  let mut dest = dest
-  let mut n = n
-  while (n != 0n) {
-    WasmI32.store8(dest, c, 0n)
-    dest += 1n
-    n -= 1n
+/**
+ * Fills the given memory region with the given value.
```
ospencer, line 34:
```suggestion
 * Fills the given memory region with the given 1-byte value. Values larger than 1 byte will be truncated.
```

ospencer (approved):

PR #2150, opened 2024-08-19
feat(stdlib): Json value access utils

### LINKED ISSUE #1877: stdlib: add a searchable json api
We are keeping the initial json pr small so we can get an mvp, we should come back and make a searchable api later.

Possible Propsal here: https://github.com/grain-lang/grain/pull/1133#discussion_r1156418330

A seperate idea i had was we break this into a seperate module with the a `fromJson` and `toJson` this could allow for reuse, if we ever add a toml or yaml parser to the stdlib.

alex-snezhko:
Closes #1877

### DIFF
--- a/stdlib/json.gr
+++ b/stdlib/json.gr
@@ -2105,7 +2105,7 @@ provide module Lenses {
    * Reads the value focused on by the given lens from the input data.
    * 
    * @param lens: The lens to apply to the subject data
-   * @param subject: The data which will have the lens applied onto it
+   * @param subject: The data which will have the lens applied to it
    * @returns `Some(data)` containing the data read by the lens if the lens matches the given data, or `None` if the data cannot be matched to the lens
    * 
    * @example assert get(number, JsonNumber(123)) == Some(123)
@@ -2120,7 +2120,7 @@ provide module Lenses {
    * 
    * @param lens: The lens to apply to the subject data
    * @param newValue: The new value to set at the focus of the lens
-   * @param subject: The data which will have the lens applied onto it
+   * @param subject: The data which will have the lens applied to it
    * @returns `Some(data)` containing the new data after the lens substitution if the lens matches the given data, or `None` if the data cannot be matched to the lens
    * 
    * @example assert set(number, 123, JsonBoolean(true)) == Some(JsonNumber(123))
@@ -2135,7 +2135,7 @@ provide module Lenses {
    * 
    * @param lens: The lens to apply to the subject data
    * @param fn: The function to apply to the matched data at the lens if matched
-   * @param subject: The data which will have the lens applied onto it
+   * @param subject: The data which will have the lens applied to it
    * @returns `Some(data)` containing the new data after the lens mapping has been applied if the lens matches the given data, or `None` if the data cannot be matched to the lens
    * 
    * @example assert map(number, x => x * 2, JsonNumber(5)) == Some(JsonNumber(10))
--- a/stdlib/json.gr
+++ b/stdlib/json.gr
@@ -451,6 +451,20 @@ record JsonWriter {
   emit: Json => Option<JsonToStringError>,
 }
 
+provide enum ValueError {
+  IncorrectType,
+}
+
+provide enum SearchError {
+  PropertyNotFound,
+  NotJsonObject,
+}
+
+provide record NestedSearchError<err> {
+  propertyAccess: String,
+  error: err,
+}
+
 let emitUTF16EscapeSequence = (codePoint: Number, buffer: Buffer.Buffer) => {
   // Emit the "\u" followed by hexadecimal representation of the codepoint
   // with fixed length of 4 hexadecimal digits corresponding to the two byte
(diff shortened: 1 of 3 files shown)


stdlib/json.gr
```diff
@@ -2069,3 +2083,242 @@ provide let parse: (str: String) => Result<Json, JsonParseError> = (str: String)
...
+ * 
+ * @param json: The JSON value
+ * @returns `Ok(value)` containing the JSON boolean value or `Err(err)` if the JSON value is not a boolean
+ * 
+ * @example assert Json.boolean(Json.JsonBoolean(true)) == Ok(true)
+ * @example assert Json.boolean(Json.JsonNumber(123)) == Err(Json.IncorrectType)
+ * 
+ * @since v0.6.7
+ */
+provide let boolean = json => {
```
spotandjake, line 2099: I feel like we should put these in a submodule named search like you have done with the tests.

stdlib/json.gr
```diff
@@ -2069,3 +2083,242 @@ provide let parse: (str: String) => Result<Json, JsonParseError> = (str: String)
...
+  }
+}
+
+/**
+ * Retrieves the value of a nested JSON object property, recursively accessing
+ * the given property names in order.
+ * 
+ * @param propertyNames: The property names to traverse on the JSON value
+ * @param json: The JSON value
+ * @returns `Ok(value)` containing the property value at the given property access path, `Err({ propertyAccess, error: Json.PropertyNotFound })` if it is not found at some point in the path, or `Err({ propertyAccess, error: Json.NotJsonObject })` if the JSON value is not an object somewhere in the path
```
spotandjake, line 2289: We don't normally document the `Err` contents liket his we normally reference the type on the error.

stdlib/json.gr
```diff
@@ -451,6 +451,20 @@ record JsonWriter {
   emit: Json => Option<JsonToStringError>,
 }
 
+provide enum ValueError {
```
spotandjake, line 454: We should probably document these types.

stdlib/json.gr
```diff
@@ -2069,3 +2083,242 @@ provide let parse: (str: String) => Result<Json, JsonParseError> = (str: String)
     }
   }
 }
+
+/**
+ * Extract the value of a JSON boolean value, or `Err` if the JSON value is not
```
spotandjake, line 2088: I feel like this feels a bit wordy what about something like `Extracts the value of a JSON boolean.`, we already explain the `Err` part in the returns.

This feedback is bassically for all the functions.

stdlib/json.gr
```diff
@@ -2069,3 +2083,242 @@ provide let parse: (str: String) => Result<Json, JsonParseError> = (str: String)
...
+ * a boolean.
+ * 
+ * @param json: The JSON value
+ * @returns `Ok(value)` containing the JSON boolean value or `Err(err)` if the JSON value is not a boolean
```
spotandjake, line 2092:
```suggestion
 * @returns `Ok(value)` containing the JSON boolean or `Err(err)` if the value is not a JSON boolean
```

spotandjake (commented):
The implementation itself mostly looks great to me with a few minor notes.

I personally feel like as we are likely to add other libraries like `Toml`, `Yaml` and maybe more the searching would be better provided as a seperate library and we could have a `toSearchable` or something in the json library that converts the format. The searching semanitcs between bassically every config language is the same and most of the datatypes are the same.

ospencer:
I'd love for us to have combinators to search JSON, somewhat similar to what this TOML library does: https://github.com/ocaml-toml/To.ml?tab=readme-ov-file#lenses

Ours wouldn't need to be quite that complex, and I think you'd only need some minor tweaks.

alex-snezhko:
@ospencer switched over to using lenses

stdlib/json.gr
```diff
@@ -2069,3 +2069,298 @@ provide let parse: (str: String) => Result<Json, JsonParseError> = (str: String)
...
+   * 
+   * @example assert get(prop("x"), JsonObject([("x", JsonNumber(123))])) == Some(JsonNumber(123))
+   * @example
+   * assert set(prop("x"), JsonString("new"), JsonObject([("x", JsonNumber(123))])) ==
+   *   Some(JsonObject([("x", JsonString("new"))]))
+   * 
+   * @since v0.7.0
+   */
+  provide let prop = propertyName =>
+    { get: json => match (json) {
```
alex-snezhko, line 2251: Unrelated to this PR but I don't really like how the formatter handles these records here

stdlib/json.gr
```diff
@@ -2069,3 +2069,298 @@ provide let parse: (str: String) => Result<Json, JsonParseError> = (str: String)
...
+  provide let get = (lens, subject) => lens.get(subject)
+
+  /**
+   * Sets the value focused on by the given lens from the input data to the
+   * desired new value.
+   * 
+   * @param lens: The lens to apply to the subject data
+   * @param newValue: The new value to set at the focus of the lens
+   * @param subject: The data which will have the lens applied onto it
+   * @returns The new data after the lens substitution has been applied wrapped in `Some` if the lens matches the given data, or `None` if the data cannot be matched to the lens
```
spotandjake, line 2110:
```suggestion
   * @returns `Some(data)` containing the new data after the lens substitution if the lens matches the given data, `None` if the data cannot be matched to the lens
```

stdlib/json.gr
```diff
@@ -2069,3 +2069,298 @@ provide let parse: (str: String) => Result<Json, JsonParseError> = (str: String)
...
+  provide let set = (lens, newValue, subject) => lens.set(newValue, subject)
+
+  /**
+   * Updates the value focused on by the given lens from the input data by
+   * applying a function to it and setting the focus to the result of the function
+   * 
+   * @param lens: The lens to apply to the subject data
+   * @param fn: The function to apply to the matched data at the lens if matched
+   * @param subject: The data which will have the lens applied onto it
+   * @returns The new data after the lens mapping has been applied wrapped in `Some` if the lens matches the given data, or `None` if the data cannot be matched to the lens
```
spotandjake, line 2125:
```suggestion
   * @returns `Some(data)` containing the new data after the lens mapping has been applied, if the lens matches the given data, `None` if the data cannot be matched to the lens
```

stdlib/json.gr
```diff
@@ -2069,3 +2069,298 @@ provide let parse: (str: String) => Result<Json, JsonParseError> = (str: String)
...
+ */
+provide module Lenses {
+  /**
+   * A structure which provides functionality for accessing and setting JSON
+   * data.
+   *
+   * @since v0.7.0
+   */
+  provide record Lens<a, b> {
+    get: a => Option<b>,
```
spotandjake, line 2086: can we add graindoc to these fields and change the types from `a => Option<b>` to `(json: a) => Option<b>` so we keep the parameter names in the outputs.

stdlib/json.gr
```diff
@@ -2069,3 +2069,298 @@ provide let parse: (str: String) => Result<Json, JsonParseError> = (str: String)
...
+  }
+
+  /**
+   * A lens whose focus is the properties of a JSON object.
+   * 
+   * @example assert get(objectProps, JsonObject([("a", JsonNumber(123))])) == Some([("a", JsonNumber(123))])
+   * 
+   * @since v0.7.0
+   */
+  provide let objectProps = {
```
spotandjake, line 2227: I feel like `objectProperties` might be a bit verbose but we have been pretty against shorthands so far thoughts? same as for `prop`

stdlib/json.gr
```diff
@@ -2069,3 +2069,298 @@ provide let parse: (str: String) => Result<Json, JsonParseError> = (str: String)
...
+    { get: json => match (json) {
+      JsonObject(props) => {
+        match (List.find(((k, _)) => k == propertyName, props)) {
+          Some((_, v)) => Some(v),
+          None => None,
+        }
+      },
+      _ => None,
+    }, set: (newValue, json) => {
+      let rec replaceFirst = (acc, list) => {
```
spotandjake, line 2260: should we move this to the top level to avoid the closure, it might make the code more readable too.

stdlib/json.gr
```diff
@@ -2069,3 +2069,298 @@ provide let parse: (str: String) => Result<Json, JsonParseError> = (str: String)
...
+   * @param propertyNames: The property path of the JSON object to create a focus on
+   * @returns A lens whose focus is the given property path of a JSON object
+   * 
+   * @example
+   * let nestedObj = JsonObject([("a", JsonObject([("b", JsonNumber(123))]))])
+   * assert get(propPath(["a", "b"]), nestedObj) == Some(JsonNumber(123))
+   * 
+   * @since v0.7.0
+   */
+  provide let propPath = propertyNames => {
```
spotandjake, line 2359: can we move this with the other property stuff, I feel like putting it all the way down here isn't going to make it easy to find.

stdlib/json.gr
```diff
@@ -2069,3 +2069,298 @@ provide let parse: (str: String) => Result<Json, JsonParseError> = (str: String)
...
+ * @since v0.7.0
+ */
+provide module Lenses {
+  /**
+   * A structure which provides functionality for accessing and setting JSON
+   * data.
+   *
+   * @since v0.7.0
+   */
+  provide record Lens<a, b> {
```
spotandjake, line 2085: Should this be `abstract` I didnt see any tests for `Lens.set` or `Lens.get` off a premade Lens.

spotandjake (commented):
This looks sick.

stdlib/json.gr
alex-snezhko: I think it shouldn't be abstract so that users can write their own lenses. Alternatively we could just provide a `lens` function with getter and setter arguments and make this abstract

alex-snezhko: Also not sure what the point of testing custom lenses is, the tests are effectively testing that the lenses provided by the library work

stdlib/json.gr
alex-snezhko: Yeah initially did this so that I could use the `||>` function but I'll just use the alternative provide syntax to change the ordering in the docs

ospencer (approved):
This is beautiful! I'm excited to use this. Just a few minor doc changes and this is good to go.

alex-snezhko:
@ospencer done

ospencer (approved):

PR #2154, opened 2024-08-29
fix(compiler): Correct type approximation on recursive functions

### LINKED ISSUE #2151: Default argument in recursive functions crashes compiler
When I set a default value for an argument in recursive functions, the compiler crashes with

Failure("Impossible: optional argument with non-option type")

```module Main2

let rec parenthesizedExpression = () => {
  let expression1 = expression(0)
  expression1
}

and expression = (prec=0) => {  // setting the default value here causes the compiler crash
  0
}

provide let parse = (source: String) => {
  expression(0)
}

spotandjake:
This pr fixes #2151, we seem to have made an invalid assumption when creating the type here which was causing the crashes. 

For anyone interested the root cause of the issue was in `type_approx` which is used when we are guessing types for recursive functions, we were producing the a `TTyvar` instead of an `option(TTyvar)` on default arguments.

### DIFF
--- a/compiler/src/typed/typecore.re
+++ b/compiler/src/typed/typecore.re
@@ -635,7 +635,14 @@ let rec type_approx = (env, sexp: Parsetree.expression) =>
   | PExpLambda(args, e) =>
     newty(
       TTyArrow(
-        List.map(x => (x.pla_label, newvar()), args),
+        List.map(
+          x =>
+            switch (x.pla_label) {
+            | Default(_) => (x.pla_label, type_option(newvar()))
+            | _ => (x.pla_label, newvar())
+            },
+          args,
+        ),
         type_approx(env, e),
         TComOk,
       ),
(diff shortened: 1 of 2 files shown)


spotandjake:
I noticed the line in the image below as well which wasn't causing the problem but uses the same pattern, I'm not sure if this behaviour is incorrect at the moment but It would probably be a good idea to look into if we are handling things correctly there as well. 
image

ospencer:
@spotandjake I looked into this and I don't love the solution here. Yes, default values are `Option` types, but I think this will just hide bugs. The `Impossible` case we're hitting is in fact possible, in this exact case (where the type of the function is not known beforehand).

Do you want to update that in this PR or close it and have me make the changes?

ospencer:
@spotandjake and I chatted offline and decided that it's better to just handle it this way rather than handle this case in each of the other places it comes up.

ospencer (approved):

spotandjake:
I just added an additional test that tests for the invalid case in the original issue where the labels are missing.

PR #2158, opened 2024-09-17
feat(stdlib): Reimplement `Number.sin`, `Number.cos`, `Number.tan`

spotandjake:
This pr reintroduces `Number.sin` after it was removed in #2046, this uses the implementation found in [libc](https://git.musl-libc.org/cgit/musl/tree/src/math), with some adaptations made for wasm and grain. It uses a small kernal function that accuratly can compute `sin` for values between `-pi/4` to `pi/4` and a range reduction algorithm to map any values outside the range into the range efficently.

Notes:
* I made a new math folder in the stdlib, as these functions can be used for multiple number types it makes sense to seperate them out so they are easier to understand.
* I am going to make a follow up pr for `gamma` and `factorial`
* I am going to make another follow up pr for `Float32` and `Float64`, `sin`, `cos` and `tan`
Some Notes:
* When it comes to bigints we lose a lot of accuracy do to how `f64`s store numbers.

work for #1478

### DIFF
--- a/stdlib/runtime/math/trig.gr
+++ b/stdlib/runtime/math/trig.gr
@@ -12,6 +12,8 @@ from "runtime/math/kernel/sin" include KernelSin
 use KernelSin.{ sin as kern_sin }
 from "runtime/math/kernel/cos" include KernelCos
 use KernelCos.{ cos as kern_cos }
+from "runtime/math/kernel/tan" include KernelTan
+use KernelTan.{ tan as kern_tan }
 from "runtime/math/rempio2" include Rempio2
 use Rempio2.{ rempio2 }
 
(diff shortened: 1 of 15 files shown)


ospencer (commented):
This looks really good! The only major feedback I have is that I'd name the kernel modules just the actual names, like instead of `KernelCos` I'd call it just `Cos` or `Cosine`.

spotandjake:
Made that change

stdlib/runtime/math/trig.gr
```diff
@@ -0,0 +1,128 @@
...
+ */
+module Trig
+
+from "runtime/unsafe/wasmi32" include WasmI32
+from "runtime/unsafe/wasmi64" include WasmI64
+from "runtime/unsafe/wasmf64" include WasmF64
+from "runtime/unsafe/conv" include Conv
+use Conv.{ fromInt32, fromFloat64 }
+from "runtime/math/kernel/sin" include Sine
+use Sine.{ sin as kern_sin }
```
ospencer, line 12: Can we do these in camel case?

spotandjake: done

ospencer (approved):

Issue #2156, opened 2024-09-16
Implement round functions on Float64 and Float32


spotandjake:
Currently we do not expose `Float64.round`, `Float64.floor`, `Float64.ceil` in the float libraries as these are common instructions we should probably add them.

This should be a pretty easy feature to implement as wasm has most of these lower level functions.

spotandjake:
This was closed with #2162

PR #2161, opened 2024-09-18
feat(stdlib): Add `copySign`, `sqrt`, `min`, `max`, `round`, `trunc`, `floor`, `ceil` to `Float32`

spotandjake:
This pr adds some utilities to `Float32` which map easily to `wasmF32` such as `copySign`, `sqrt`, `min`, `max`, `round`, `trunc`, `floor`, `ceil`

Work Towards #2156

### DIFF
--- a/stdlib/float32.gr
+++ b/stdlib/float32.gr
@@ -426,11 +426,11 @@ provide let round = (x: Float32) => {
   let ptr = newFloat32(WasmF32.nearest(xv))
   WasmI32.toGrain(ptr): Float32
 }
-// let rint = () => {}
+
 /**
  * Computes the square root of its operand.
  *
- * @param x: The float32 to square root
+ * @param x: The operand to square root
  * @returns The square root of the operand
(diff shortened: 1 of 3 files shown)


stdlib/float32.gr
```diff
@@ -315,3 +353,114 @@ provide let neg = (x: Float32) => {
...
+  let xv = WasmF32.load(WasmI32.fromGrain(x), _VALUE_OFFSET)
+  let ptr = newFloat32(WasmF32.sqrt(xv))
+  WasmI32.toGrain(ptr): Float32
+}
+
+/**
+ * Copys the sign of the second operand to the first operand.
+ *
+ * @param x: The operand to be copied
+ * @param y: The operand containing the sign
```
ospencer, line 451:
```suggestion
 * @param x: The operand to modify
 * @param y: The operand to copy the sign from
```

ospencer (approved):

PR #2162, opened 2024-09-19
feat(stdlib): Add `copySign`, `sqrt`, `min`, `max`, `round`, `trunc`, `floor`, `ceil` to `Float64`

spotandjake:
This pr adds some utilities to `Float64` which map easily to `wasmF64` such as `copySign`, `sqrt`, `min`, `max`, `round`, `trunc`, `floor`, `ceil`

Work Towards #2156

### DIFF
--- a/stdlib/float64.gr
+++ b/stdlib/float64.gr
@@ -447,8 +447,8 @@ provide let sqrt = (x: Float64) => {
 /**
  * Copys the sign of the second operand to the first operand.
  *
- * @param x: The operand to be copied
- * @param y: The operand containing the sign
+ * @param x: The operand to be modify
+ * @param y: The operand to copy the sign from
  * @returns The first operand with the sign of the second operand
  *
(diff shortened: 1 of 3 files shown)


stdlib/float64.gr
```diff
@@ -315,3 +353,114 @@ provide let neg = (x: Float64) => {
...
+  let xv = WasmF64.load(WasmI32.fromGrain(x), _VALUE_OFFSET)
+  let ptr = newFloat64(WasmF64.sqrt(xv))
+  WasmI32.toGrain(ptr): Float64
+}
+
+/**
+ * Copys the sign of the second operand to the first operand.
+ *
+ * @param x: The operand to be copied
+ * @param y: The operand containing the sign
```
ospencer, line 451:
```suggestion
 * @param x: The operand to be modify
 * @param y: The operand to copy the sign from
```

ospencer (approved):

PR #2170, opened 2024-10-04
feat(stdlib)!: Seeded hashing

spotandjake:
This pr refactors the hashing function so we do not keep global state, it also implements a new `Hash.makeSeeded` and `Hash.make` function which returns a `HashInstance`. If we want a different api I am happy to change things around.

API:
```
let globalHash = Hash.make()
assert Hash.hash(globalhash, 1) == Hash.hash(globalHash, 1)
assert Hash.hash(globalhash, "a") == Hash.hash(globalHash, "a")
```
```
let seeded1 = Hash.makeSeeded(1)
assert Hash.hash(seeded1, 1) == Hash.hash(seeded1, 1)
assert Hash.hash(seeded1, "a") == Hash.hash(seeded1, "a")
let seeded2 = Hash.makeSeeded(2)
assert Hash.hash(seeded1, 1) != Hash.hash(seeded2, 1)
```

Questions:
* Should we generate a seperate seed on `Hash.make` every time or do we want to use the globalSeed? Maybe if we want a globalSeed we should provide it a `makeGlobal`.
* Should we use `makeSeeded` with a set seed on `Map` and `Set` so we can get rid of the possible exception.

My original implementation had the signature of `Hash.make: (seed: Number) => (anything: a) => Number` but this causes our inference engine to lock the hashing function a particular data type when you do something like: 
```rs
let hash = Hash.make()
assert hash("H") == hash("H")
assert hash(1) == hash(1) // fails to typecheck because `a` is now `String` so you would need to make a separate instance for each dataType.
```
The alternative approach suggested was to use a submodule for seeded but as the api's would differ this did not seem ideal.

One benefit to this approach is because we call `Random.random()` on `make` we will not throw a possible exception on `Hash.hash` which means all of our `Set` and `Map` functions besides `make` wont throw.

Succeeds: #1952 

This is breaking as we changed the type of `Hash.hash` from `(a) => Number` to `(HashInstance, a) => Number`.

### DIFF
--- a/stdlib/hash.gr
+++ b/stdlib/hash.gr
@@ -3,8 +3,12 @@
  *
  * @example from "hash" include Hash
  *
- * @example Hash.hash(1)
- * @example Hash.hash("Hello World")
+ * @example
+ * let hashInstance = Hash.make()
+ * assert Hash.hash(hashInstance, "Hello World") == Hash.hash(hashInstance, "Hello World")
+ * @example
+ * let hashInstance = Hash.makeSeeded(10)
+ * assert Hash.hash(hashInstance, "Hello World") == Hash.hash(hashInstance, "Hello World")
  *
  * @since v0.1.0
  */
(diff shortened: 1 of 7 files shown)


stdlib/hash.gr
```diff
@@ -130,141 +120,207 @@ let rec hashOne = (val, depth) => {
...
+}
+
+/**
+ * Represents a particular hashing instance.
+ *
+ * @since v0.7.0
+ */
+abstract record HashInstance {
+  seed: Number,
+}
```
ospencer, line 261: I think this would be better as just `abstract type HashInstance = Number`. You'll need more annotations in this file, but it lets us avoid an extra layer of indirection.

spotandjake:
I applied that change.


Another small question on this pr, should we fix the seed used for `Map` and `Set` to keep the order consistent between runs on operations like `toList` this helps to preserve same input `==` same output semantics.

ospencer (approved):

PR #2171, opened 2024-10-04
chore(runtime): Avoid branches in `164abs` and `164neg` in `runtime/numbers`

spotandjake:
This is a small pr that changes the way we do `abs` and `neg` in the runtime for some of our helpers, normally this would not matter but given we call `gcd` a lot it makes sense to avoid branching.

### DIFF
--- a/stdlib/runtime/numbers.gr
+++ b/stdlib/runtime/numbers.gr
@@ -1602,8 +1602,9 @@ let numberDivide = (x, y) => {
 
 @unsafe
 let i64abs = x => {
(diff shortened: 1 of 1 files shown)


ospencer (commented):
This is the sort of change that without some benchmarking I'd be reluctant to merge the change. It makes the code much harder to read, and this is the exact type of optimization that binaryen/the wasm runtimes would make for us.

spotandjake:
I'm going to close this issue, I think this might just reveal that we need todo some more tuning with our binaryen passes and pass ordering.

Issue #1185, opened 2022-04-15
Optimize pattern matching on constants


ospencer:
When matching on constants, we essentially compile something like `match (...) { 'a' => ..., 'b' => ..., ... }` into an `if` chain of `if (val == 'a') { goto branch 1 } else if (val == 'b') { goto branch 2 } else ...`. We can instead compile them as though they were variants, and just do a `br_table`.

spotandjake:
Investigating this a bit, it does not seem as though using a br_table is always the most efficent, if you have sparse datasets the table is not the best optionally additionally we need to map the data to a 0 based range as such it probably makes sense to use a few heuristics here notably:
* Primarily continuous dataset (We can have a few caps and just direct them to the default branch manually)
* More then a few elements as br_table is only really more efficent then branching in the case that checking all the options would take longer.
* for void constants it would make sense to just eliminate the match all together
* for bool constants it would make sense to build a select statement instead of a conditonal

We can however do an additional optimization when implementing this that is mentioned in the todo commment and even if the dataset is sparse we can load the number data from the heap and untag and do a direct match for some of our number types to speed things up.

PR #2173, opened 2024-10-09
feat(compiler): Improve constant pattern matching

spotandjake:
This improves the code generation for our pattern matching engine, the goal is to make it so binaryen can build `br_tables` in more cases.

So far I have switched some of our static constants for short types away from the condtional logic to the same switch logic we use for `constructors`, this should make it easier to optimize the codegen for both cases in the future, this already has benefits in the codegen as can be seen from the test we are generating more select operations and less branching operations.

Draft as I want to look into some more optimizations mostly related to the jump table.

Work for: #1185

### DIFF
--- a/compiler/src/middle_end/matchcomp.re
+++ b/compiler/src/middle_end/matchcomp.re
@@ -57,6 +57,7 @@ type decision_tree =
       with an optional default branch. */
     Switch(
       switch_type,
+      option(Ident.t),
       list((int, decision_tree)),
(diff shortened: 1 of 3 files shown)


alex-snezhko (commented):
This is awesome! Just had a question about the unoptimized case

spotandjake:
I'm converting this to a draft as I would like to wait until the gc changes are all made to redo this, I think it can be done in a neater way.

PR #2177, opened 2024-10-14
feat!: Remove `->` token in favor of `=>`

spotandjake:
This removes the `->` support in the compiler as we said we were only going to keep it through `v0.6`, This is my first change to the parser that changes the way tokens are evaluated, I went through and manually updated the `parsers.messages` file I dont know if I was meant todo it another way but menhir seemed fine with it :).

One thing I was wondering is maybe instead of fully removing support for `->` do we want to still handle lexing it and throw an error in the lexer about it being deprecated for a few versions? I think this behaviour is desirable over the formatter approach as it signifies we dont support the syntax but makes it clear why things broke.

Breaking as we are removing support for `->`


alex-snezhko:
I think doing a lexer in the error instead makes sense to be helpful to the user

spotandjake:
Alright I switched this to throw a `DeprecatedToken` error in the lexer, I think this behaviour makes sense to keep around for a few versions and we can remove the error if we decide to use `->` for something different later.

ospencer:
I don't think the error is necessary. We already had a transition period, and it could have made sense to have a warning for it before dropping support, but as we're at the point that the support is over, we should remove it from the language entirely.

spotandjake:
> I don't think the error is necessary. We already had a transition period, and it could have made sense to have a warning for it before dropping support, but as we're at the point that the support is over, we should remove it from the language entirely.

Reverted the error.

ospencer (approved):

PR #2178, opened 2024-10-21
feat(stdlib): `isAscii`,`isAsciiControl`,`isAsciiWhitespace` to `Char`

spotandjake:
I've been working in rust a little bit and found the variety of ascii functions super helpful notably `isAscii` and `isAsciiWhitespace`, this pr just adds a few more little utility functions for use in the char library.

Question: Would it make sense to move all of the ascii related functions both in this pr and the existing ones to an `Ascii` submodule. (If we do this I will probably also move over a few of the other common ascii group methods like isPunctuation).

### DIFF
--- a/stdlib/char.gr
+++ b/stdlib/char.gr
@@ -315,11 +315,25 @@ provide let (>=) = (x: Char, y: Char) => {
 }
 
 /**
+ * Checks if the character is an ASCII character.
+ *
+ * @param char: The character to check
+ * @returns `true` if the character is an ASCII character or `false` otherwise
+ *
+ * @example assert Char.isAscii('1')
+ * @example assert Char.isAscii('a')
+ * @example assert !Char.isAscii('🌾')
+ *
+ * @since v0.7.0
+ */
+provide let isAscii = char => char <= '\u{007F}' // usv <= 0x7F
+
+/**
  * Checks if the character is an ASCII digit.
  *
  * @param char: The character to check
  * @returns `true` if the character is an ASCII digit or `false` otherwise
- * 
+ *
  * @example assert Char.isAsciiDigit('1')
  * @example assert !Char.isAsciiDigit('a')
  *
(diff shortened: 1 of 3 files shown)


ospencer:
I'm in favor of an `Ascii` submodule.

spotandjake:
> I'm in favor of an `Ascii` submodule.

Done

stdlib/char.gr
```diff
@@ -315,64 +315,170 @@ provide let (>=) = (x: Char, y: Char) => {
 }
 
 /**
- * Checks if the character is an ASCII digit.
+ * Sub module for working with ASCII characters.
```
ospencer, line 318:
```suggestion
 * Utilities for working with ASCII characters.
```

stdlib/char.gr
```diff
@@ -315,64 +315,170 @@ provide let (>=) = (x: Char, y: Char) => {
...
  *
- * @param char: The character to check
- * @returns `true` if the character is an ASCII digit or `false` otherwise
- * 
- * @example assert Char.isAsciiDigit('1')
- * @example assert !Char.isAsciiDigit('a')
+ * @example Char.Ascii.isAscii('1')
```
ospencer, line 320: I think it's fine to do the examples as just `Ascii.` without the leading `Char.`.

stdlib/char.gr
```diff
@@ -315,64 +315,170 @@ provide let (>=) = (x: Char, y: Char) => {
...
+ * @since v0.7.0
  */
-provide let isAsciiDigit = char => char >= '0' && char <= '9'
+provide module Ascii {
+  /**
+   * The minimum valid ASCII character code.
+   *
+   * @since v0.7.0
+   */
+  provide let min = 0x00
```
ospencer, line 330: Shouldn't these be chars? If someone really needs the code then they can do a `Char.code` on it.

stdlib/char.gr
```diff
@@ -315,64 +315,170 @@ provide let (>=) = (x: Char, y: Char) => {
...
+   * @param char: The character to check
+   * @returns `true` if the character is an ASCII character or `false` otherwise
+   *
+   * @example assert Char.Ascii.isValid('1')
+   * @example assert Char.Ascii.isValid('a')
+   * @example assert !Char.Ascii.isValid('🌾')
+   *
+   * @since v0.7.0
+   */
+  provide let isValid = char => char <= '\u{007F}' // usv <= 0x7F
```
ospencer, line 351:
```suggestion
  provide let isValid = char => char <= '\u{007F}'
```

ospencer (commented):
This is awesome. Really looking forward to this.

stdlib/char.gr
spotandjake: We currently use the code on the top level module `Char.min`. I think if we really want to change that we should do it in a seperate pr after this.

stdlib/char.gr
spotandjake: I think we should keep the `Char.Ascii.isAscii`, I feel like it's not very long and it makes the examples runnable, in other modules such as array I was using the pattern:
```
use Array.{ module Immutable }
assert Immutable.isEmpty(Immutable.empty) == true
```
but I felt that the module name ascii was small enough here to just directly inline it.

stdlib/char.gr
ospencer: That makes sense. We can leave it as-is.

ospencer (approved):

PR #2180, opened 2024-10-24
chore(stdlib): Clean up runtime printing library

spotandjake:
Small cleanup pr, I noticed we had an extra printing utility in `runtime/utils/printing` this pr removes it as we already provide `debugPrint` for this use case.

I also changed `debugPrint` to use `@noPervasives` which makes `debugPrint` usable in libraries without pervasive without any issues, this pr also just adds an ignore to the string to better help with the garbage collector.


stdlib/runtime/bigint.gr
```diff
@@ -136,26 +136,26 @@ let init = (limbs: WasmI32) => {
 provide let debugDumpNumber = (num: WasmI32) => {
```
ospencer, line 136: Let's just remove this whole debug function. I really don't like the debug import.

spotandjake: Done.

ospencer (approved):

PR #2200, opened 2024-11-18
chore(stdlib): Add examples to `Queue` module

spotandjake:
This pr adds examples to the `Queue` module. I also noticed a few of the history items were missing so this corrects them and I also moved the `toList` and `fromList` closer to the `toArray` and `fromArray`.

### DIFF
--- a/stdlib/queue.gr
+++ b/stdlib/queue.gr
@@ -7,6 +7,13 @@
  *
  * @example from "queue" include Queue
  *
+ * @example
+ * let queue = Queue.fromList([0, 1])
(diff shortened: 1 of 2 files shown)


stdlib/queue.gr
```diff
@@ -125,12 +159,60 @@ provide let pop = queue => {
   }
 }
 
+/**
+ * Clears the queue by removing all of its elements
```
ospencer, line 163:
```suggestion
 * Clears the queue by removing all of its elements.
```

ospencer (approved):

PR #2202, opened 2024-11-18
feat(stdlib): `List.Associative` Submodule

spotandjake:
This pr adds a `List.Assocative` submodule which contains functions for working with key-value pairs in lists.

```
let data = [
  ("name", "Alice"),
  ("name", "Bob"),
  ("age", "30"),
]
assert List.Associative.get("name", data) == Some("Alice")
```

### Use case
This is similar to the associative list methods in ocaml https://ocaml.org/manual/5.2/api/List.html#1_Associationlists

I wasn't quite sure if we wanted this in the standard library or not but I think the pattern occurs enough with record information that it makes sense. A few examples being `Map.toList`, `JSONObject`.
Currently duplicates are ignored and the first item is always treated as the one your referring too.

### Alternative proposal 
An alternative proposal to this pr is we implement a new `Collection` module, where collections are of type `abstract type Collection<key, value> = List<(key, value)>` and then in that case we can handle the duplication semantics when you do `List.toCollection`, I think this might be overkill though.

### DIFF
--- a/stdlib/list.gr
+++ b/stdlib/list.gr
@@ -972,3 +972,129 @@ provide let sort = (compare=compare, list) => {
 
   mergesort(list)
 }
+
+/**
+ * Utilities for working with lists of key-key value pairs.
+ *
+ * @example
+ * let data = [
+ *  ("name", "Alice"),
+ *  ("age", "30"),
+ * ]
+ * assert List.Associative.get("name", data) == Some("Alice")
+ *
+ * @since v0.7.0
+ */
+provide module Associative {
+  /**
+   * Checks if the given key is present in the list of key-value pairs.
+   *
+   * @param key: The key to search for
+   * @param list: The list of key-value pairs
+   *
+   * @returns `true` if the key is found or `false` otherwise
+   *
+   * @example
+   * let data = [
+   *   ("name", "Alice"),
+   *   ("age", "30"),
+   * ]
+   * assert List.Associative.has("name", data) == true
+   * @example List.Associative.has("age", []) == false
+   *
+   * @since v0.7.0
+   */
+  provide let rec has = (key, list) => {
+    match (list) {
+      [] => false,
+      [(k, _), ...rest] when key == k => true,
+      [_, ...rest] => has(key, rest),
+    }
+  }
+  /**
+   * Provides the first value in the list of key-value pairs that matches the given key.
+   *
+   * @param key: The key to search for
+   * @param list: The list of key-value pairs
+   *
+   * @returns `Some(value)` if the key is found or `None` otherwise
+   *
+   * @example
+   * let data = [
+   *  ("name", "Alice"),
+   *  ("name", "Bob"),
+   *  ("age", "30"),
+   * ]
+   * assert List.Associative.get("name", data) == Some("Alice")
+   *
+   * @example List.Associative.get("age", []) == None
+   *
+   * @since v0.7.0
+   */
+  provide let rec get = (key, list) => {
+    match (list) {
+      [] => None,
+      [(k, v), ...rest] when key == k => Some(v),
+      [_, ...rest] => get(key, rest),
+    }
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


stdlib/list.gr
```diff
@@ -972,3 +972,129 @@ provide let sort = (compare=compare, list) => {
...
+
+  /**
+   * Provides a new list with the first value in the list of key-value pairs that matches the given key replaced.
+   * If the key is not found the item is appended to the list.
+   *
+   * @param key: The key to replace
+   * @param value: The new value to set
+   * @param list: The list of key-value pairs
+   *
+   * @returns The new list with the key-value pair replaced
```
ospencer, line 1050:
```suggestion
   * @returns A new list with the key-value pair replaced
```

stdlib/list.gr
```diff
@@ -972,3 +972,129 @@ provide let sort = (compare=compare, list) => {
...
+   */
+  provide let rec has = (key, list) => {
+    match (list) {
+      [] => false,
+      [(k, _), ...rest] when key == k => true,
+      [_, ...rest] => has(key, rest),
+    }
+  }
+  /**
+   * Provides the first value in the list of key-value pairs that matches the given key.
```
ospencer, line 1015:
```suggestion
   * Retrieves the first value in the list of key-value pairs that matches the given key.
```

stdlib/list.gr
```diff
@@ -972,3 +972,129 @@ provide let sort = (compare=compare, list) => {
...
+  provide let rec get = (key, list) => {
+    match (list) {
+      [] => None,
+      [(k, v), ...rest] when key == k => Some(v),
+      [_, ...rest] => get(key, rest),
+    }
+  }
+
+  /**
+   * Provides a new list with the first value in the list of key-value pairs that matches the given key replaced.
```
ospencer, line 1043:
```suggestion
   * Creates a new list with the first value in the list of key-value pairs that matches the key replaced.
```

stdlib/list.gr
```diff
@@ -972,3 +972,129 @@ provide let sort = (compare=compare, list) => {
...
+  provide let rec set = (key, value, list) => {
+    match (list) {
+      [] => [(key, value)],
+      [(k, _), ...rest] when key == k => [(k, value), ...rest],
+      [first, ...rest] => [first, ...set(key, value, rest)],
+    }
+  }
+
+  /**
+   * Provides a new list with the first value in the list of key-value pairs that matches the given key removed.
```
ospencer, line 1073:
```suggestion
   * Creates a new list with the first value in the list of key-value pairs that matches the key removed.
```

ospencer (commented):
This PR is beautiful, but the spacing in your PR description is a crime.

ospencer (approved):
I love this but would like to see additional input from @alex-snezhko or @phated.

alex-snezhko:
Yeah I'm totally in favor of this, I've basically had to implement this functionality myself several times so I think it makes sense to include in the stdlib

stdlib/list.gr
```diff
@@ -972,3 +972,129 @@ provide let sort = (compare=compare, list) => {
...
+   *  ("name", "Bob"),
+   *  ("age", "30"),
+   * ]
+   * assert List.Associative.get("name", data) == Some("Alice")
+   *
+   * @example List.Associative.get("age", []) == None
+   *
+   * @since v0.7.0
+   */
+  provide let rec get = (key, list) => {
```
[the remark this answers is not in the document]
spotandjake: What if we just had an `all=false` defaulted argument.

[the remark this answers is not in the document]
alex-snezhko: You can't really do that because the return type would change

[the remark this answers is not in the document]
spotandjake: That's fair enough, in that case I think we should have the `getAll`, `removeAll` and we might as well add `setAll` though I feel like that one isn't quite as useful.

PR #2206, opened 2024-11-19
chore(compiler): No longer rely on joo_global_object

hhugo:
It's deprecated and will result in a warning with the next release of jsoo


ospencer (approved):
Thanks @hhugo!

PR #2207, opened 2024-11-21
feat(cli): Support node v22 and v23

spotandjake:
I don't know if this is the best title for this but before if you tried to run the grain cli with node `v22` or above it would error because the `--experimental-wasm-return-call` flag was used, this just checks if we are on those versions and doesn't apply the flag as the behaviour is the default.

I think this pr might be controversial the goal here isn't to change the version of js we use with the pkg binaries just to allow newer versions of node to be used when you've built the compiler from source.

### DIFF
--- a/cli/bin/grainrun.js
+++ b/cli/bin/grainrun.js
@@ -10,7 +10,11 @@ const v8 = require("v8");
  *
  * This seems to work for our needs with Node 18, but we should be cautious when updating.
  */
-v8.setFlagsFromString("--experimental-wasm-return-call");
+if (
+  !process.versions.node.startsWith("22.") &&
+  !process.versions.node.startsWith("23.")
+)
+  v8.setFlagsFromString("--experimental-wasm-return-call");
 
 const { readFile } = require("fs/promises");
 const { WASI } = require("wasi");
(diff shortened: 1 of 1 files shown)


ospencer:
I would reverse the logic and only apply the flag for versions of Node that support it.

spotandjake:
> I would reverse the logic and only apply the flag for versions of Node that support it.

I flipped that logic, I just did a manual check for `18, 19, 20 and 21` as We only support down to `18`

cli/bin/grainrun.js
```diff
@@ -10,7 +10,13 @@ const v8 = require("v8");
  *
  * This seems to work for our needs with Node 18, but we should be cautious when updating.
  */
-v8.setFlagsFromString("--experimental-wasm-return-call");
+if (
+  process.versions.node.startsWith("18.") ||
+  process.versions.node.startsWith("19.") ||
+  process.versions.node.startsWith("20.") ||
+  process.versions.node.startsWith("21.")
+)
```
ospencer, line 18: Here's something a bit more compact:
```suggestion
if (process.version.match(/^v(18|19|20|21)\./))
```

cli/bin/grainrun.js
```diff
@@ -10,7 +10,8 @@ const v8 = require("v8");
  *
  * This seems to work for our needs with Node 18, but we should be cautious when updating.
```
ospencer, line 11:
```suggestion
 * This is valid in Node 18, 19, 20, and 21.
```

ospencer (approved):

PR #2209, opened 2024-11-26
feat(stdlib): Add `Array.tryInit`

spotandjake:
This pr adds an `Array.tryInit` function to make this a bit easier. `Array.tryInit: (length: Number, fn: (index: Number) => Result<a, b>) => Result<Array<A>, b>`. This makes building arrays from result functions quite a bit easier as there was no good alternative before. I do not love the `tryInit` name but I couldn't think of anything better. 

Notes:
* I do not love the `tryInit` name, if anyone has a better one let me know
* If we like the feature and name I am happy to add it to `List` and `Array.Immutable`
  * happy todo that in this pr or a separate one
  * `Array.tryInit` is the most important as with something like a list you can just build it manually.
* I also noticed we didn't have tests for `Array.make` and `Array.init` so I added those in this pr but I am happy to move those to a separate pr if wanted.

### DIFF
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -130,10 +130,10 @@ provide let init = (length: Number, fn: Number => a) => {
 }
 
 /**
- * Creates a new array of the specified length where each element is
- * initialized with the result of an initializer function. The initializer
- * is called with the index of each array element. If the initializer function
- * returns `Err(_)`, the array creation is stopped and the error is returned.
+ * Creates a new array where each element is initialized with the `Ok`
+ * value result of an initializer function. The initializer is called with the
+ * index of each element. Returns the new array if all calls to the
+ * initializer succeed or the first error otherwise.
  *
  * @param length: The length of the new array
  * @param fn: The initializer function to call with each index, where the value returned will be used to initialize the element
@@ -168,7 +168,8 @@ provide let tryInit = (length: Number, fn: Number => Result<a, b>) => {
       },
       Err(e) => {
         for (let mut j = 0n; j < i; j += 4n) {
-          ignore(Memory.decRef(WasmI32.load(array, _ARRAY_START_OFFSET + j)))
+          Memory.decRef(WasmI32.load(array, _ARRAY_START_OFFSET + j))
+          Memory.free(array)
         }
         return Err(e)
       },
--- a/stdlib/array.gr
+++ b/stdlib/array.gr
@@ -120,13 +120,61 @@ provide let init = (length: Number, fn: Number => a) => {
   let mut index = 0n
   for (let mut i = 0n; i < byteLength; i += 4n) {
     WasmI32.store(
-      array + i,
+      array,
       Memory.incRef(WasmI32.fromGrain(fn(tagSimpleNumber(index)))),
-      _ARRAY_DATA_OFFSET
+      _ARRAY_DATA_OFFSET + i
     )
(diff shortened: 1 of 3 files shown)


stdlib/array.gr
```diff
@@ -121,15 +121,63 @@ provide let init = (length: Number, fn: Number => a) => {
...
+      Ok(value) => {
+        WasmI32.store(
+          array,
+          Memory.incRef(WasmI32.fromGrain(value)),
+          _ARRAY_START_OFFSET + i
+        )
+        index += 1n
+      },
+      Err(e) => {
+        for (let mut j = 0n; j < i; j += 4n) {
```
spotandjake, line 171: We `incRef` when we add the items so we should probably decRef all the items we `incRef`'d. I would appreciate if someone could verify that I am doing this right.

I also think we might need to decRef the allocated array if i'm not mistaken?

spotandjake:
> This feels a little niche. Can you give some examples of use cases of where this would be useful?

The scenario I needed this for was parsing wasm, I already know how many types will be in the section so it makes sense to initialize the array ahead of time, but I need to parse each type separately like:
```
let typeCount = Stream.GetAndAdvance.uLeb128(stream)
Array.init((_) => {
  parseWasmType(stream) // This could fail so we want to escape
}, typeCount)
```
The only other ideal way I could think todo this was either use a list or create an array of options (which has other issues). I know the instruction itself seems niche and my use case was niche but I feel like the operation makes sense especially during streaming operations where you know the count but not if the data is correct.

ospencer:
I see. Let me sit with it for a bit.

stdlib/array.gr
```diff
@@ -121,15 +121,63 @@ provide let init = (length: Number, fn: Number => a) => {
...
     index += 1n
   }
   WasmI32.toGrain(array): Array<a>
 }
 
+/**
+ * Creates a new array of the specified length where each element is
+ * initialized with the result of an initializer function. The initializer
+ * is called with the index of each array element. If the initializer function
+ * returns `Err(_)`, the array creation is stopped and the error is returned.
```
ospencer, line 137:
```suggestion
 * Creates a new array where each element is initialized with the `Ok`
 * value result of an initializer function. The initializer is called with the
 * index of each element. Returns the new array if all calls to the
 * initializer succeed or the first error otherwise.
```

stdlib/array.gr
ospencer: That's correct.

stdlib/array.gr
```diff
@@ -121,15 +121,63 @@ provide let init = (length: Number, fn: Number => a) => {
...
+        WasmI32.store(
+          array,
+          Memory.incRef(WasmI32.fromGrain(value)),
+          _ARRAY_START_OFFSET + i
+        )
+        index += 1n
+      },
+      Err(e) => {
+        for (let mut j = 0n; j < i; j += 4n) {
+          ignore(Memory.decRef(WasmI32.load(array, _ARRAY_START_OFFSET + j)))
```
ospencer, line 172:
```suggestion
          Memory.decRef(WasmI32.load(array, _ARRAY_START_OFFSET + j))
          Memory.free(array)
```

ospencer (commented):
After thinking for awhile, I think this is worth it to have.

ospencer (approved):

PR #2214, opened 2024-12-13
fix(compiler): Reduce stack usage when allocating lists

### LINKED ISSUE #2109: Improve compilation of large list/array literals
When the compiler is compiled to JavaScript, large lists/arrays can experience stack overflows in various parts of the compiler.

ospencer:
Fixes #2109 

This allows list literals with up to ~3000 elements when the compiler is compiled to JS and running on Node.

Tested native, 50k elements without issue. If we want a test for this, I'd keep it to just a few hundred elements.


phated (approved):
Slick

spotandjake (approved):
This is great, I do feel like we should have a regression test for this either testing some sane list sizes or that only runs on release or something (I imagine we will go with the former since we don't have any infra for release only tests at the moment).

ospencer:
Added a test.

Issue #684, opened 2021-06-01
Better UTF-8 code sharing between String and Char


peblair:
With the changes in #683, there is now some duplicated UTF-8 logic between `Char` and `String` (specifically, `String.getCodePoint` and `String.writeUtf8CodePoint`). While these functions are not identical, they do largely the same thing. It would be preferable if these were combined into a single implementation that both modules used, but we would like them to not depend on one another, if possible. That means that we are probably looking at some sort of third module which both can depend on.

cc @ospencer in case there's anything else to add here.

ospencer:
We'll still need to figure out how we want to handle shared helper libraries in the stdlib, but otherwise 👍

PR #2215, opened 2024-12-16
feat(stdlib): Add `Bytes.setChar` and `Bytes.getChar`

spotandjake:
This pr was originally only meant to add `Bytes.setChar` and `Bytes.getChar` but I wanted it to share the utf8 impl with the string library, One thing led to another and I added a new `Utf8` library into the runtime.

The `Utf8` module houses all of the shared low level utf8 logic for encoding and decoding I also refactored the runtime and stdlib to make use of these sharedHelpers which removed a lot of duplicate code and I think should be enough to close out #684 

As this pr got a little off track I am happy to make this pr just a refactor and move the `setChar` and `getChar` changes to a separate pr. A small note on the function themselves, compared to the other set and get methods `chars` are slightly different as we need to account for the range of byte sizes this is going to make `Buffer.setChar` slightly more annoying when I get around to that.


stdlib/bytes.gr
```diff
@@ -396,6 +405,69 @@ provide let clear = (bytes: Bytes) => {
...
+  let byte = WasmI32.load8U(ptr + offset, _VALUE_OFFSET)
+  let charSize = getUtf8ByteCount(byte)
+  checkIndexIsInBounds(offset, charSize, size)
+  let codePoint = getCodePoint(ptr + offset + _VALUE_OFFSET)
+  ignore(bytes)
+  tagChar(codePoint)
+}
+
+/**
+ * Sets a character starting at the given byte index.
```
ospencer, line 442:
```suggestion
 * UTF-8 encodes a character starting at the given byte index.
```

ospencer (commented):
Overall looks great!

ospencer (approved):

Issue #2217, opened 2024-12-23
Improve pattern matching errors for lists


spotandjake:
Currently we treat lists internally as regular adts and whenever we print them we give them the pretty list syntax it seems as though we do not handle this for pattern matching warnings related to exhaustiveness.

```gr
module Main

match ([]) {
  [1] => void,
}
```

results in the almost unreadable warning:
```
Warning 8: this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
([...](1, [...](_, _))|[...](0, _)|[])
```
I think this would be done in `parmatch.re` under the `Conv.conv` function for Constructors.

PR #2220, opened 2024-12-23
feat(compiler): Improve exhaustive warning for lists

spotandjake:
This pr improves the way we print exhaustiveness warnings for list constructors an example can be seen below:
Screenshot 2024-12-22 at 8 22 40 PM

As a side note, and without looking into this I wonder if it would make sense or if we should at least open an issue to move from a custom printer here to somehow using the formatter, we do have a function to convert from the typedtree nodes back into the parsetree nodes which we should be able to pipe into the formatter this would maintain a consistent

Closes: #2217


ospencer (approved):

ospencer:
> I wonder if it would make sense or if we should at least open an issue to move from a custom printer here to somehow using the formatter

Maybe? I really hate the untypeast stuff, but it would be nice to use the formatter.

PR #2222, opened 2024-12-29
feat(lsp): Add or remove block braces code action

alex-snezhko:
```
// Before
let f = () => print("Hi")
// After code action
let f = () => { print("Hi") }

// Before
let f = () => { print("Hi") }
// After code action
let f = () => print("Hi")
```

Implemented only for function bodies as I felt this would be where this would get the most use, and allowing it _everywhere_ would add a lot of noise as this code action would basically show everywhere in code. I'm open to also adding it to a few other specific places like ifs, loops, etc. though

### DIFF
--- a/compiler/src/language_server/code_action.re
+++ b/compiler/src/language_server/code_action.re
@@ -111,6 +111,81 @@ let rec process_named_arg_label = (uri, results: list(Sourcetree.node)) => {
   };
 };
 
+let rec process_add_or_remove_braces = (uri, results: list(Sourcetree.node)) => {
+  Typedtree.(
+    switch (results) {
+    | [Value({exp: {exp_desc: TExpLambda([mb], _)}}), ...rest] =>
+      switch (mb.mb_body.exp_desc) {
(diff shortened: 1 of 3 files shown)


spotandjake (approved):
This looks good to me, though I do wonder if we should be adding the line breaks as well, given I imagine you would use this to start adding more statements given if you just add the braces inline the formatter will format them away.

alex-snezhko:
@spotandjake yeah I considered adding line breaks as well though my initial thought was that formatting the generated code properly should be the formatter's job and not the LSP's: furthermore the LSP would have to take an opinionated stance on how many spaces to insert after the line break, which the user might not prefer if they don't like the formatter's prescriptive styling e.g. if they are using tabs or 4 space indent, etc.

I can definitely see the pragmatic argument of adding a line break though, since realistically that's what the user will always do after using this code action: of course you could just run the formatter right after running the code action but that adds another step I suppose. I will pass this question to @ospencer @phated to arrive on a consensus if we want the LSP's output to be opinionated in line with the formatter.

ospencer (approved):

PR #2226, opened 2025-01-04
feat(stdlib): add `Path.removeExtension`

spotandjake:
This adds `Path.removeExtension` to the `Path` library as I noticed there wasn't really a great way todo it before


alex-snezhko:
Would it be better to have an `updateExtension`/`withExtension` function that you could just pass in `None` to remove

spotandjake:
> Would it be better to have an `updateExtension`/`withExtension` function that you could just pass in `None` to remove

I was debating an `updateExtension` but I honestly think we should still have `removeExtension` it's more intuitive and clear what it does and then `updateExtension` can just take a string.

ospencer (commented):
I think this is reasonable to have. I could additionally see an `updateExtension` to change it, but I like them being separate because it's easy to search for.

alex-snezhko (approved):

PR #2228, opened 2025-01-09
chore: fix 404 binaryen.ml URL

zhoufanjin:
fix 404 status URL


ospencer (approved):
We should probably get that main branch name updated, but this is much quicker. Thanks for catching this!

Issue #2229, opened 2025-01-12
formatter replaces tuple argument types


lue-bird:
The compiling code
```grain
let tuple_second: ((a, b)) => b = ((_, y)) => y
```
gets formatted to
```grain
let tuple_second: (a, b) => b = ((_, y)) => y
```
which doesn't compile as multiple arguments and tuples are different constructs.

A temporary workaround for users can be switching to
```grain
type Tuple2<a, b> = (a, b)
let tuple_second: Tuple2<a, b> => b = ((_, y)) => y
```

Info:
- ```
  grain --version
  0.6.6
  ```
  built from source a few days ago

PR #2230, opened 2025-01-13
fix(grainfmt): Fix formatting of function types with a single tuple arg

spotandjake:
This fixes formatting of single argument tuple function types such as `((a, b) => Void` before they were being formatted to `(a, b) => Void`, this is because we drop parens around single argument lambda's but we need to keep them in the case of a pattern.

Closes: #2229

### DIFF
--- a/compiler/src/formatting/fmt.re
+++ b/compiler/src/formatting/fmt.re
@@ -2702,8 +2702,7 @@ let print_type = (fmt, {ptyp_desc, ptyp_loc}) => {
           ptyp_arg_label: Unlabeled,
           ptyp_arg_type: {
             ptyp_desc:
-              PTyAny | PTyVar(_) | PTyArrow(_, _) | PTyConstr(_, _) |
-              PTyPoly(_, _),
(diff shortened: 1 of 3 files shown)


compiler/src/formatting/fmt.re
```diff
@@ -2696,7 +2696,19 @@ let print_type = (fmt, {ptyp_desc, ptyp_loc}) => {
...
     )
-  | PTyArrow([{ptyp_arg_label: Unlabeled} as param], return) =>
+  | PTyArrow(
+      [
+        {
+          ptyp_arg_label: Unlabeled,
+          ptyp_arg_type: {
+            ptyp_desc:
+              PTyAny | PTyVar(_) | PTyArrow(_, _) | PTyConstr(_, _) |
+              PTyPoly(_, _),
```
ospencer, line 2706: I would probably just keep this to `PTyAny`, `PTyVar`, and `PTyArrow`.

spotandjake: I removed the `PtyPoly` and limited `PTyConstr` to not have any data attached, we need to allow `PtyConstr` still though because something like `Number` is a constructor.

ospencer (approved):

Issue #2152, opened 2024-08-28
Duplicate imports


lann:
Grain produces Wasm modules with duplicate imports. This is permitted by the Wasm core spec, but with caveats:

https://webassembly.github.io/spec/core/syntax/modules.html#imports

> It is possible to import the same module/name pair multiple times [...] However, embedders are not required to support such overloading

The [Wasm Component Model](https://github.com/WebAssembly/component-model) in particular does not support duplicate imports, which means that WASI preview1 modules compiled with Grain cannot be converted to components with [the preview1 adapter](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi-preview1-component-adapter/README.md).

Example:

```console
$ grain --version
0.6.6
$ cat hello.gr
module Main
print("Hello, World")
$ grain compile hello.gr
$ wasm-tools dump hello.gr.wasm | grep Import
   0x16b | 16 77 61 73 | import [func 0] Import { module: "wasi_snapshot_preview1", name: "fd_write", ty: Func(2) }
   0x18d | 16 77 61 73 | import [func 1] Import { module: "wasi_snapshot_preview1", name: "fd_write", ty: Func(2) }
   0x1af | 16 77 61 73 | import [func 2] Import { module: "wasi_snapshot_preview1", name: "fd_write", ty: Func(2) }
   0x1d1 | 16 77 61 73 | import [func 3] Import { module: "wasi_snapshot_preview1", name: "fd_write", ty: Func(2) }
```

spotandjake:
If you run `grain compile <file> --release` it will produce a module without multiple imports.

ospencer:
To add to what @spotandjake mentioned, we have successfully used the adapter to make components. Let me know if you want some pointers!

We've been aware of this issue but have been slow to fix it because of the available workaround. This is something we can resolve in the next major release.

lann:
Aha thanks! Feel free to resolve this unless you'd like to keep it for tracking.

PR #2233, opened 2025-01-14
feat(compiler)!: Deduplicate imports

spotandjake:
This enables the binaryen pass to de-duplicate imports, if were against using the binaryen pass I think we can also do this manually in `compcore` and just filter through the imports in `compile_imports` but we'll have to manually handle re-patching the import identifiers through the program.

This makes developing for runtimes such as `wasmtime` a little nicer.

Closes: #2152


ospencer:
I would lean towards trying to prevent the issue rather than rely on the Binaryen pass. I think you're right that this would be annoying to resolve during codegen, but I don't think this is bad to fix as a part of linking at all, because it's already keeping track of names to rewrite. You could dedupe the imports and just keep the needed name in the resolver.

spotandjake:
I refactored this change to handle things in compcore but I still need to figure out why our smallest program got larger, it seems to be that the wrapper isn't being inlined anymore but I am unsure why.

compiler/test/suites/includes.re
```diff
@@ -207,4 +207,38 @@ describe("includes", ({test, testSkip}) => {
...
+    let name = "dedupe_includes";
+    let outfile = wasmfile(name);
+    ignore @@
+    compile(
+      ~hook=Grain.Compile.stop_after_assembled,
+      name,
+      {|
+      module DeDupeIncludes
+      // Ensures fd_write is only included once
+      print("test")
```
ospencer, line 221: Can you make this test more obvious by literally importing an external function twice?

compiler/src/codegen/compcore.re
```diff
@@ -22,6 +24,7 @@ type codegen_env = {
   /* Allocated closures which need backpatching */
   backpatches: ref(list((Expression.t, closure_data))),
   required_imports: list(import),
+  raw_resolutions: ref(StringSet.t),
```
ospencer, line 27: Let's call this something like `foreign_import_resolutions`.

spotandjake:
Both those changes have been made.

compiler/src/codegen/compcore.re
```diff
@@ -3090,37 +3101,53 @@ let compile_imports = (wasm_mod, env, {imports}) => {
...
-      let typ = wasm_type(typ);
-      Import.add_global_import(
-        wasm_mod,
-        internal_name,
-        module_name,
-        item_name,
-        typ,
-        mut,
-      );
+    let imp_key = (module_name, item_name, mimp_kind, mimp_type);
```
ospencer, line 3104: `import_key`?

ospencer:
Why is this marked as breaking?

spotandjake:
> Why is this marked as breaking?

I wasn't sure if we considered this optimization a breaking change or not, I'll mark it as non breaking though.

spotandjake:
As a note if we merge this before a fix is made for #2243 we will need to make sure to handle the `foreign_import_resolutions` logic for globals as well.

ospencer (approved):

PR #2234, opened 2025-01-16
fix(compiler): Write universal exports on linked module

ospencer:
In #2104, the `write_universal_exports` function was previously called after linking, but I forgot to reinstate it. This restores that behavior.

I deleted some tests because they were testing old behavior that was valid for the intermediate (object) wasm files but are no longer valid in the output of final wasm modules.


compiler/test/suites/modules.re
```diff
@@ -126,44 +126,4 @@ describe("modules", ({test, testSkip}) => {
     "nestedModules",
     "hello from foo\nhello from bar\n[2, 3, 4]\n9\n[> 2, 3, 4]\nfalse\nfoo\n",
   );
-  test("reprovided_module", ({expect}) => {
```
spotandjake, line 129: Would it make sense to change this to a snapshot test of the object file to ensure we are representing these correctly there.

compiler/test/suites/provides.re
```diff
@@ -220,32 +220,12 @@ describe("provides", ({test, testSkip}) => {
...
-    "module Test; module M { provide let x = 1; provide let y = 2 }; use M.*; provide { x, y }",
+    "module Test; module M { provide let x = () => 1; provide let y = () => 2 }; use M.*; provide { x, y }",
     [
-      ("GRAIN$EXPORT$x", Binaryen.Export.external_global),
-      ("GRAIN$EXPORT$y", Binaryen.Export.external_global),
+      ("x", Binaryen.Export.external_function),
+      ("y", Binaryen.Export.external_function),
     ],
   );
-  assertHasWasmExport(
```
spotandjake, line 229: If I understood the changes correctly we no longer allow you to export values right, which is why these tests are being removed ?

spotandjake (approved):
Mostly looks good just two small questions.

compiler/test/suites/provides.re
ospencer: There is no change to the final wasm output. The old wasm object files had these exports but the final linked wasm files did not.

compiler/test/suites/modules.re
ospencer: Done.

ospencer:
@spotandjake Good catch. I restored the old behavior. We should maybe consider a warning for reserved export names, but that's not for this PR.

spotandjake:
I just tried this again the `_start` stuff is fixed but I found a new bug if you have a program like:
`b.gr`:
```
module B

provide let x = (y) => 1
```
`a.gr`
```
module A
from "./b.gr" include B
provide let x = B.x
```
we seem to hit a `Not_found` exception.

ospencer:
This is good to go. @spotandjake I added tests for that behavior and another test caught a different bug.

compiler/src/codegen/comp_utils.re
```diff
@@ -195,152 +165,147 @@ let type_of_repr = repr => {
...
-              arg_types,
-              result_types,
-              [||],
+                ),
+                Expression.Null.make(),
+              ),
               function_body,
-            );
-            // Remove existing Grain export (if any)
-            Export.remove_export(wasm_mod, name);
```
alex-snezhko, line 333: Why is the remove_export no longer necessary

compiler/src/codegen/comp_utils.re
```diff
@@ -195,152 +165,147 @@ let type_of_repr = repr => {
...
-          switch (item) {
-          | TSigValue(
-              id,
-              {
-                val_repr: ReprFunction(args, rets, direct),
-                val_fullpath: path,
-              },
-            ) =>
-            let name = Ident.name(id);
-            let exported_name = "GRAIN$EXPORT$" ++ name;
```
alex-snezhko, line 212: Why is the `GRAIN$EXPORT$` prefix gone?

compiler/src/codegen/comp_utils.re
ospencer: When we used wasm files as object files, we wrote special Grain export names in the Wasm binaries to facilitate linking later. We don't do this anymore before linking happens before we generate wasm (so there was no Grain export in the first place to worry about removing).

compiler/src/codegen/comp_utils.re
ospencer: We don't write special Grain exports anymore; now we only write the universal exports (see my other comment).

alex-snezhko (approved):

compiler/src/codegen/compcore.re
```diff
@@ -3311,8 +3267,7 @@ let compile_main = (wasm_mod, env, prog) => {
   if (Grain_utils.Config.use_start_section^) {
     Function.set_start(wasm_mod, start);
   } else {
-    ignore @@
-    Export.add_function_export(wasm_mod, grain_start, Comp_utils.grain_start);
+    ignore @@ Export.add_function_export(wasm_mod, start_name, grain_start);
```
spotandjake, line 3270: Should we be doing a check here and warn and override if a user `_start` exists?

compiler/src/codegen/compcore.re
```diff
@@ -3506,7 +3461,6 @@ let compile_wasm_module =
     ignore @@ compile_imports(wasm_mod, env, prog, import_map);
     ignore @@ compile_globals(wasm_mod, env, prog);
     ignore @@ compile_functions(wasm_mod, env, prog);
-    ignore @@ compile_exports(wasm_mod, env, prog);
```
spotandjake, line 3509: I feel like this should be a breaking change, given were changing the module signature.

spotandjake (approved):
Looks good to me I had one small question, but otherwise lets get this merged.

compiler/src/codegen/compcore.re
ospencer: Maybe, but that's not for this PR.

compiler/src/codegen/compcore.re
ospencer: It's breaking in the sense that the module signature will be different than it currently is on the `main` branch, but the behavior on main is a regression and this just restores the signature to the way it worked before custom object files were merged.

Issue #703, opened 2021-06-06
Remove all offset magic constants in stdlib


peblair:
Code such as this:
```grain
// get the boxed number tag from the pointer x
let tag = WasmI32.load(x, 4n)
```
is neither maintainable nor simple to read. Instead, we should have constants for all magic numbers like this. For example:
```grain
// get the boxed number tag from the pointer x
let tag = WasmI32.load(x, _BOXED_NUMBER_TAG_OFFSET)
```

PR #2237, opened 2025-01-24
chore(runtime): Cleanup offset constants in `String` module

spotandjake:
This pr works towards #703 I started with the `string` module, I also changed a few other modules to use the common `offsets` module for their offsets, if we are good with the approach I am taking to this I will do the same to the other modules but I figure that might be best todo in a separate pr.

Notes:
* I found some other utf-8 cleanup that I threw into this pr as it wasn't much.
* I also made the other modules with offsets already defined in variables point to the `Offsets` module for consistency sake.

### DIFF
--- a/stdlib/runtime/unsafe/offsets.gr
+++ b/stdlib/runtime/unsafe/offsets.gr
@@ -1,0 +1,18 @@
+/**
+ * Various common offsets for data structures throughout the runtime.
+ */
+@runtimeMode
+module Offsets
+
+// String
+provide let _STR_LEN_OFFSET = 4n
+provide let _STR_DATA_OFFSET = 8n
+
+// Array
+provide let _ARRAY_LEN_OFFSET = 4n
+provide let _ARRAY_DATA_OFFSET = 8n
+provide let _Array_ITEM_SIZE = 4n
+
+// Bytes
+provide let _BYTES_LEN_OFFSET = 4n
+provide let _BYTES_DATA_OFFSET = 8n
--- a/stdlib/runtime/unsafe/offsets.gr
+++ b/stdlib/runtime/unsafe/offsets.gr
@@ -11,7 +11,7 @@ provide let _STR_DATA_OFFSET = 8n
 // Array
 provide let _ARRAY_LEN_OFFSET = 4n
 provide let _ARRAY_DATA_OFFSET = 8n
-provide let _Array_ITEM_SIZE = 4n
+provide let _ARRAY_ITEM_SIZE = 4n
 
 // Bytes
 provide let _BYTES_LEN_OFFSET = 4n
--- a/stdlib/runtime/utf8.gr
+++ b/stdlib/runtime/utf8.gr
@@ -16,6 +16,19 @@ from "runtime/unsafe/wasmi32" include WasmI32
 provide exception MalformedUnicode
 
 /**
+ * Determines if the given byte is a leading byte in a UTF-8 sequence.
+ *
+ * @param byte: The byte to check
+ *
+ * @returns `true` if the byte is a leading byte, `false` otherwise
+ */
+@unsafe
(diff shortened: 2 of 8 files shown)


stdlib/runtime/unsafe/offsets.gr
```diff
@@ -0,0 +1,18 @@
...
+module Offsets
+
+// String
+provide let _STR_LEN_OFFSET = 4n
+provide let _STR_DATA_OFFSET = 8n
+
+// Array
+provide let _ARRAY_LEN_OFFSET = 4n
+provide let _ARRAY_DATA_OFFSET = 8n
+provide let _Array_ITEM_SIZE = 4n
```
alex-snezhko, line 14: Unused; if you want to keep it change to all uppercase

stdlib/runtime/utf8.gr
```diff
@@ -15,6 +15,19 @@ from "runtime/unsafe/wasmi32" include WasmI32
...
 
+/**
+ * Determines if the given byte is a leading byte in a UTF-8 sequence.
+ *
+ * @param byte: The byte to check
+ *
+ * @returns `true` if the byte is a leading byte, `false` otherwise
+ */
+@unsafe
+provide let isLeadingByte = byte => {
```
alex-snezhko, line 26: Was this intended to be used somewhere?

alex-snezhko (approved):

stdlib/runtime/utf8.gr
spotandjake: This is used in `string.gr` so its just a tiny bit of cleanup I noticed while looking for constants and decided to address in this pr.

stdlib/string.gr
```diff
@@ -454,9 +476,9 @@ provide let split = (separator: String, string: String) => {
       if (Memory.compare(ptr, pptr, psize) == 0n) {
         let strSize = ptr - last
         let str = allocateString(strSize)
-        Memory.copy(str + 8n, last, strSize)
-        WasmI32.store(arr + arrIdx, str, 8n)
-        arrIdx += 4n
+        Memory.copy(str + _STR_DATA_OFFSET, last, strSize)
+        WasmI32.store(arr + _ARRAY_DATA_OFFSET, str, arrIdx)
```
alex-snezhko, line 480: Why did you flip these

stdlib/string.gr
```diff
@@ -966,30 +978,24 @@ let _OFFSET_NAME = "offset"
...
-  if (WasmI32.eqz(num & 1n)) {
-    let str = " argument must be an integer"
-    throw Exception.InvalidArgument(name ++ str)
-  }
-  let num = num >> 1n
-  if (num < 0n) {
-    let str = " argument must be non-negative"
-    throw Exception.InvalidArgument(name ++ str)
+  use WasmI32.{ (<) }
+  let _num = convSimpleNumber(num, name ++ " argument must be an integer")
```
alex-snezhko, line 982: A note is that in the nominal case this is now less efficient due to needing to do this string allocation + concat every time rather than only in the bad case

alex-snezhko (approved):

stdlib/string.gr
spotandjake: I flipped them as I figured we should be offsetting into the array for the item rather than moving the pointer, it doesn't really make a difference though.

stdlib/runtime/utf8.gr
ospencer: Was this already its own function or did you separate out this logic into a function?

spotandjake: > Was this already its own function or did you separate out this logic into a function?

I separated it out, happy to move it to a separate pr but I didn't feel like it was worth it, especially as its just work on generally cleaning up more of our `utf8` logic.

PR #2238, opened 2025-01-26
feat(stdlib): Add `Char.encodedLength`

spotandjake:
This adds `Char.byteCount` which can be used with a specified encoding to get the encoded length of a char.

Example use case for this is stepping through a byte array when you do `Byte.getChar` you have no clue how far to advance for the next char without this.

### DIFF
--- a/stdlib/char.gr
+++ b/stdlib/char.gr
@@ -190,14 +190,14 @@ provide enum Encoding {
  * @param char: The character
  * @returns The byte count of the character in the given encoding
  *
- * @example Char.byteCount(Char.UTF8, 'a') == 1
- * @example Char.byteCount(Char.UTF8, '🌾') == 4
- * @example Char.byteCount(Char.UTF16, '©') == 1
+ * @example Char.encodedLength(Char.UTF8, 'a') == 1
+ * @example Char.encodedLength(Char.UTF8, '🌾') == 4
+ * @example Char.encodedLength(Char.UTF16, '©') == 1
  *
  * @since v0.7.0
  */
 @unsafe
-provide let byteCount = (encoding, char: Char) => {
+provide let encodedLength = (encoding, char: Char) => {
   let usv = untagChar(char)
   let utf8ByteCount = usvEncodeLength(usv)
   let utf8ByteCount = tagSimpleNumber(utf8ByteCount)
--- a/stdlib/char.gr
+++ b/stdlib/char.gr
@@ -184,9 +184,9 @@ provide enum Encoding {
 }
 
 /**
- * Returns the byte count of the given character in the given encoding.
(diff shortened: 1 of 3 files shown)


ospencer:
This would need to be called `utf8EncodedLength` or something like that to be specific about what byte length it's referring to.

spotandjake:
> This would need to be called `utf8EncodedLength` or something like that to be specific about what byte length it's referring to.

The model takes an `encoding` parameter and will return the byte count in the specified encoding.

stdlib/char.gr
```diff
@@ -172,6 +172,42 @@ provide let toString = (char: Char) => {
...
+ *
+ * @since v0.7.0
+ */
+@unsafe
+provide let byteCount = (encoding, char: Char) => {
+  let usv = untagChar(char)
+  let utf8ByteCount = usvEncodeLength(usv)
+  let utf8ByteCount = tagSimpleNumber(utf8ByteCount)
+  match (encoding) {
+    UTF32 => 4,
```
alex-snezhko, line 205: Can check this at the beginning since it'll always be 4 but nbd

alex-snezhko (approved):

stdlib/char.gr
spotandjake: I don't think it's worth it as it makes the match non exhaustive and adds an extra branch, `UTF32` probably isn't going to be used very often anyways so optimizing for it doesn't seem worth it.

ospencer:
@spotandjake I think it's still not clear that this gives you some kind of encoded length. To me, `byteCount` sounds like it's going to tell me how many bytes this character is. That's not really meaningful. Total number of bytes that Grain considers it? Is it referring to Unicode characters that consist of multiple code points and telling me how many? Calling this something like `encodedLength` tells me that it's going to tell me how long it is when encoded (some kind of way, which I can specify with a parameter), which is clear.

spotandjake:
That changes have been made.

stdlib/char.gr
```diff
@@ -172,6 +172,42 @@ provide let toString = (char: Char) => {
...
+ * @since v0.7.0
+ */
+provide enum Encoding {
+  UTF8,
+  UTF16,
+  UTF32,
+}
+
+/**
+ * Returns the byte count of the given character in the given encoding.
```
ospencer, line 187:
```suggestion
 * Returns the byte count of a character if encoded in the given encoding.
```

stdlib/char.gr
```diff
@@ -172,6 +172,42 @@ provide let toString = (char: Char) => {
...
+ *
+ * @param encoding: The encoding to use
```
ospencer, line 189:
```suggestion
 * @param encoding: The encoding to check
```

ospencer (approved):

PR #2239, opened 2025-01-28
chore(docs): Add architectural flowchart

spotandjake:
I did up a flowchart for the grain compiler walkthrough and realized it would make a good addition to the compiler reference.


alex-snezhko:
Feels weird to have the OCaml typechecker and "thresher" simultaneously in the diagram, let's just include the current state and then change it once the typechecker is changed? That won't really even change the diagram at all right? Also the title "Pre-0.7" isn't really useful.

spotandjake:
Addressed that feedback

spotandjake:
I think i'm going to close this my initial goal was to add the graph as I made it for other purposes, but I agree with oscar's point above.

PR #2242, opened 2025-02-07
fix: typos in documentation files

maximevtush:


spotandjake:
This change actually needs to be made in the doc comment in `path.gr` and then you can run `npm run stdlib doc` to regenerate the markdown.

Issue #2244, opened 2025-02-13
GrainFmt removes parens around type casting causing a parse error.


spotandjake:
If you format the below program:
```gr
module Main
type X = Int32

(1l: X): Int32
```
the formatter will output:
```gr
module Main
type X = Int32

1l: X: Int32
```
which fails to parse.

PR #2245, opened 2025-02-13
fix(grainfmt): Correct formatting of nested constraints

spotandjake:
This fixes a small bug where `(1: Number): Number` was being formatted into `1: Number: Number` which is a parsing issue.

Closes: #2244


ospencer (commented):
I tested this on your branch and it works, but can you add a test for `((1: Number): Number): Number`?

spotandjake:
I added that test, I didn't have it initially because its still just `(expr: type)` at each level but it makes sense to test the nested behaviour.

ospencer (approved):

Issue #2246, opened 2025-02-16
Stack overflow on equality comparsions of large lists


spotandjake:
With the following test program:
```gr
module ArrayTest

from "list" include List

let i = 6122

let x = List.init(i, (i) => i)
let y = List.init(i, (i) => i)

assert x == y
```

Testing off commit `2f16ba5` It seems we can compare lists up to `723` elements and up to `6121` elements in release mode before we overflow the stack.

The source of the issue is that adt comparisons within `heapEqualHelp` are not being tco'd which I don't think we can solve easily in the general case, however in the case of lists I think we can easily handle them seperatly in a tail call optimized away.

ospencer:
I wouldn't call this a bug.

ospencer:
(It can be better, but the behavior isn't incorrect.)

PR #2247, opened 2025-02-17
feat(runtime): Optimize `==` for lists 

spotandjake:
This optimizes `==` for specifically for lists and also refactors it to use early return. Instead of the recursive approach used in the general case for adt's because we know lists will always have two elements (item, nextItemPtr) we handle it in an iterative manner avoiding the stack overflow.

**Notes**:
* The smallest program size increased because `==` is bundled to every grain program.
* `==` would now lock with a cyclic list, however from my understanding such a list would be impossible to build in grain. 
  * If we want to handle cycles we can but it adds overhead as we need to go back through the scanned items and do the unmarking after.
  * This will all need to get refactored with the wasm gc changes we plan to make so I think it's fine to take the perf improvement for now.
* We should be able to support any list size now however it seems we start running into issues with `List.init` around `5005` elements.

Closes: #2246

### DIFF
--- a/compiler/test/stdlib/pervasives.test.gr
+++ b/compiler/test/stdlib/pervasives.test.gr
@@ -75,3 +75,9 @@ record Comparable2 {
 }
 assert compare({ a: 1, b: true, c: void }, { a: 1, b: true, c: void }) == 0
 assert compare({ a: 1, b: true, c: void }, { a: 1, b: false, c: void }) > 0
+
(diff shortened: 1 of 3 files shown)


ospencer (commented):
Should we add a test with a very very long list?

spotandjake:
I added that test and this should be ready to merge now.

ospencer (approved):

PR #2248, opened 2025-02-17
feat(runtime): Prevent stack overflow while collecting large lists

spotandjake:
### Summary
This pr prevents stack overflows during the collection of large and extremely large lists.

## Notes
* Smallest program size decreased because of increased code sharing in `decRefChildren`
* All data structures now support extremely deep nested objects in the final position.
  
----

### Problem
While working on #2247 I noticed that `decRef` was overflowing the stack with lists that were only around `1350` elements.

### Solution (For those interested)
The reason for this is that `decRef` and `decRefChildren` were not tco optimized, so the two options were to either handle lists separately or find a way to optimize `decRef` for tail calls. I decided on the second approach as it is more generic and provides improvements for other data structures as well.

The way I went about this was with a slightly cursed but valid modification where we `free` the parent before the children, this only works because we immediately `decRef` the children and when we `free` we don't zero the memory but instead  just merge the blocks which will leave our child pointers unaffected. When freeing children we free most of the children in a loop and then free the final child outside of the loop in the tail position of `decRefChildren` allowing for the tco optimization. Initially this bloated our smallest program size by 100 bytes but I noticed that we are doing effectively the same operations for all dataTypes so I combined everything into a shared helper function that actually reduced our bundle size by `47` bytes.

## Why not handle lists separately
Handling lists separately would have certainly been easier and if we are not up for the changes I made here it is still a viable approach however, it would add overhead to every ADT decRef and likely result in a hundred or so additional bytes being added to our smallest program size.

### DIFF
--- a/stdlib/runtime/gc.gr
+++ b/stdlib/runtime/gc.gr
@@ -94,9 +94,9 @@ let rec decRef = (userPtr: WasmI32, ignoreZeros: Bool) => {
 
       if (WasmI32.eqz(refCount)) {
         /*
-         * Note: We dangerously call free before decRefChildren to allow for
-         *       tco optimization. This works because `free` only merges the
-         *       empty blocks and does not actually zero out the memory.
+         * Note: We call free before decRefChildren to allow for a tail call.
+         * This is okay because no allocations occur while we traverse the
+         * structure and free does not mangle the data.
          */
         free(userPtr)
         decRefChildren(userPtr)
@@ -137,7 +137,7 @@ and decRefChildren = (userPtr: WasmI32) => {
     t when t == Tags._GRAIN_ARRAY_HEAP_TAG || t == Tags._GRAIN_TUPLE_HEAP_TAG => {
       decRefChildrenHelp(userPtr, 4n, 8n)
     },
-    // No travelsal necessary for other tags
+    // No traversal necessary for other tags
     _ => void,
   }
 }
--- a/stdlib/runtime/gc.gr
+++ b/stdlib/runtime/gc.gr
@@ -25,6 +25,7 @@ from "runtime/unsafe/panic" include Panic
 from "runtime/unsafe/wasmi32" include WasmI32
 use WasmI32.{ (+), (-), (*), (&), (==), (!=) }
 
+primitive (!) = "@not"
 primitive (&&) = "@and"
 primitive (||) = "@or"
 primitive ignore = "@ignore"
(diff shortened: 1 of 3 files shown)


ospencer (commented):
Can you explain what `parentPtr` is for? It's not obvious to me reading through the code.

spotandjake:
> Can you explain what `parentPtr` is for? It's not obvious to me reading through the code.

Essentially `Memory.decRef` has the behaviour that it returned the pointer passed in, I needed to find a way to preserve this behaviour through the tco optimization which required passing the `parentPtr` through `decRef` and `decRefChildren`, otherwise we wouldn't be able to preserve the behaviour while putting `decRefChildren` in the tail position.

ospencer:
That behavior isn't used by `decRef` or `decRefChildren` at all; you could just return it in the function that's exported at the end, e.g. `let decRef = userPtr => { decRef(userPtr, false); userPtr }`.

ospencer:
The internal `decRef` can just return void.

spotandjake:
> The internal `decRef` can just return void.

I made that change, that simplifies the pointer handling semantics quite a bit.

stdlib/runtime/gc.gr
```diff
@@ -84,24 +85,37 @@ let rec decRef = (userPtr: WasmI32, ignoreZeros: Bool) => {
...
     } else {
       let refCount = refCount - 1n
       setRefCount(userPtr, refCount)
 
       if (WasmI32.eqz(refCount)) {
-        decRefChildren(userPtr)
+        /*
+         * Note: We dangerously call free before decRefChildren to allow for
+         *       tco optimization. This works because `free` only merges the
+         *       empty blocks and does not actually zero out the memory.
```
ospencer, line 99: It's not dangerous if we know what we're doing 😉
```suggestion
         * Note: We call free before decRefChildren to allow for a tail call.
         * This is okay because no allocations occur while we traverse the
         * structure and free does not mangle the data.
```

stdlib/runtime/gc.gr
```diff
@@ -110,43 +124,25 @@ and decRefChildren = (userPtr: WasmI32) => {
...
-      for (let mut i = 0n; WasmI32.ltU(i, maxOffset); i += 4n) {
-        ignore(decRef(WasmI32.load(userPtr + i, 16n), false))
-      }
-    },
-    _ => {
-      // No travelsal necessary for other tags
-      void
+      decRefChildrenHelp(userPtr, 4n, 8n)
     },
+    // No travelsal necessary for other tags
```
ospencer, line 140:
```suggestion
    // No traversal necessary for other tags
```

ospencer (commented):
Nice.

ospencer (approved):

PR #2249, opened 2025-02-17
feat(stdlib): Optimize `List.init` to support large lists

spotandjake:
## Notes
* `2500` is a rather arbitrary choice to switch over, We start having stack overflows with the grain runner around 5k elements and I figured other runtimes might have smaller stacks so it probably makes sense to switch over before that.
  * If we choose to low a number we leave performance on the table because of the `List.reverse` overhead
  * If we choose to high a number we could overflow the wasm call stack.
* If #2247 gets merged first then I can use `==` instead of the the helper function.
* Draft as without #2248 I need to use a bind for the list to prevent an immediate free and stack overflow


stdlib/list.gr
```diff
@@ -27,17 +34,18 @@ from "runtime/dataStructures" include DataStructures
...
 provide let init = (length, fn) => {
-  // This method can be further improved by checking the length against a specific size
-  // and determining if it can be made tail-recursive, then use List.reverse on it.
-  // Which would be the same as OCaml does it in https://github.com/ocaml/ocaml/blob/03839754f46319aa36d9dad56940a6f3c3bcb48a/stdlib/list.ml#L79
-  let rec iter = (idx, length) => {
-    if (idx >= length) {
-      []
-    } else {
-      [fn(idx), ...iter(idx + 1, length)]
+  if (length < 2500) {
```
ospencer, line 37: Let's go with 2048 because powers of two warm my heart.

stdlib/list.gr
```diff
@@ -13,6 +13,13 @@ from "runtime/unsafe/memory" include Memory
 from "runtime/unsafe/wasmi32" include WasmI32
 from "runtime/dataStructures" include DataStructures
 
+let rec revHelp = (list, acc) => {
+  match (list) {
+    [] => acc,
+    [first, ...rest] => revHelp(rest, [first, ...acc]),
+  }
+}
+
```
ospencer, line 22: Why not just move the whole reverse function up here?

ospencer (approved):

PR #2251, opened 2025-03-02
fix(grainfmt): correct fmt of operator funcs applied with labeled args

spotandjake:
Corrects formatting of both infix and prefix functions called with labeled arguments.

ex:
```
(+)(x=1, y=2)
(!)(rhs=true)
```
previously it formatted the bad output of:
```
x=1 + y=2
!rhs=true
```

### DIFF
--- a/compiler/test/grainfmt/application.expected.gr
+++ b/compiler/test/grainfmt/application.expected.gr
@@ -85,3 +85,5 @@ let preExistingObjectsWithRefCountMismatch2 = Map.make():
   >
 
 Int32.(-)(1l, 2l)
+Int32.(+)(x=1l, y=2l)
+Pervasives.(!)(rhs=true)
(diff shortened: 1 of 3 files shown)


compiler/test/grainfmt/application.expected.gr
```diff
@@ -85,3 +85,5 @@ let preExistingObjectsWithRefCountMismatch2 = Map.make():
   >
 
 Int32.(-)(1l, 2l)
+Int32.(+)(x=1l, y=2l)
```
alex-snezhko, line 88: In this example wouldn't the code path you edited not get applied? For example the line directly above this is formatted as `Int32.(-)(1l, 2l)`. Would you want to `use Int32.{ (+) }` first?

spotandjake: 100%, I added the correct tests, I kept these ones around for completeness sake.

alex-snezhko (approved):

ospencer (approved):

PR #2252, opened 2025-03-02
feat(grainfmt): Add inline flag to grainfmt

spotandjake:
This adds the `--inline` flag to the `grain format` command specifically useful with tools like [Treefmt](https://treefmt.com/latest/reference/formatter-spec/). Usage of the command is as such, `grain format input.gr --inline` which will write the formatted code back to `input.gr` this also works on directories.


ospencer:
Prettier names this flag `--write`. I don't know what other formatters call this.

spotandjake:
Closing as after a conversation in discord it we leaned towards the current `-o` semantics.

PR #2253, opened 2025-03-09
fix(compiler): Fix 'inconsistent assumptions over interface' bug

alex-snezhko:
Fix 'The files A and B make inconsistent assumptions over interface C' error

### DIFF
--- a/compiler/src/typed/env.re
+++ b/compiler/src/typed/env.re
@@ -793,14 +793,12 @@ module Persistent_signature = {
   };
 
   let load =
-    ref((~loc=Location.dummy_loc, unit_name) => {
-      switch (Module_resolution.locate_object_file(~loc, unit_name)) {
-      | filename =>
-        let ret = {filename, cmi: Module_resolution.read_file_cmi(filename)};
-        Some(ret);
-      | exception Not_found => None
+    ref(obj_filename =>
+      {
+        filename: obj_filename,
+        cmi: Module_resolution.read_file_cmi(obj_filename),
       }
-    });
+    );
 };
 
 let acknowledge_pers_struct = (check, {Persistent_signature.filename, cmi}) => {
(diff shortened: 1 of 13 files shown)


compiler/src/typed/env.re
```diff
@@ -2245,15 +2246,10 @@ let crc_of_unit = filename => {
...
-  let resolved_units =
-    List.map(
-      unit => Module_resolution.locate_unit_object_file(unit),
-      imported_units,
-    );
-  List.map2(
-    (unit, resolved_unit) => (unit, crc_of_unit(resolved_unit)),
+
+  List.map(
+    (unit) => (unit, crc_of_unit(unit)),
```
alex-snezhko, line 2251: Root cause appeared to be that `crc_of_unit` which calls `find_pers_struct` operates as if the filename passed in was to the `.gr` file rather than the `.gro` file

compiler/src/typed/env.re
```diff
@@ -848,8 +848,10 @@ let acknowledge_pers_struct = (check, {Persistent_signature.filename, cmi}) => {
 
 let find_pers_struct = (~loc, check, filepath) => {
   switch (Hashtbl.find(persistent_structures, filepath)) {
-  | Some(ps) => ps
-  | None => raise(Not_found)
+  | Some(ps) =>
+    failwith("find_pers_struct unexpectedly found persistent_structures for .gr filepath " ++ filepath ++ ". Please open an issue.");
```
alex-snezhko, line 852: Idea here is that `persistent_structures` should _never_ have a `filepath` entry in it because it is a `.gr` file whereas this map should only have `.gro` files in it.

To be honest I was not confident enough that some code path does not trigger this so rather than deleting this code entirely I just included these safeguards which would warrant additional investigation.

ospencer (commented):
I think you're right about the what the issue is here, but this isn't quite the solution. `persistent_structures` is basically a cache of information about compiled objects, and your changes bypass the issue by just avoiding the cache entirely and always loading the object from disk. `Persistent_signature.load` does the right thing because it resolves the filepath to the object location.

This code is all kinds of busted and needs some real attention.

alex-snezhko:
Updated @ospencer

compiler/src/typed/env.re
```diff
@@ -2278,30 +2210,33 @@ let build_signature_with_imports =
     cmi_type_metadata: type_metadata,
     cmi_config_sum: Cmi_format.config_sum(),
   };
+};
 
+let add_cmi_to_persistent_structures = (modname, filename, cmi) => {
```
ospencer, line 2215: It looks like the `modname` here isn't used anywhere, so we should be able to remove it from here, and the anftree, right?

alex-snezhko: Yep good point, this was from a previous iteration of this PR that I ended up restructuring

ospencer (approved):
Looks great!

PR #2254, opened 2025-03-20
Upgrade jsoo

hhugo:


ospencer:
It's usually not so bad. I don't know what's going on here, though.

hhugo:
The CI is happy on ubuntu https://github.com/hhugo/grain/actions/runs/13989268415/job/39169369435?pr=2

Failing on macos because of cmake
Failing on windows because of ocp-indent


- I had to add `(data_only_dirs node_modules)` in `compiler/dune`, otherwise dune would complain about dune files in `compiler/node_modules/ocaml`
- pin binaryen to include jsoo 6 fixes
- pin dune because it has a `with-dev-setup` dependency` on a recent version of ppx_expect that is not available on 4.14.  I don't understand why esy want to install `with-dev-setup` for dune.
- the hacks.js files are no longer necessary because already part of the latest release.

The diff in `esy.json` can probably be cleaned up and reduced. It's the result of many experiment.

spotandjake:
On my branch #2233 I got past the ocp-indent issue by vendoring and disabling the test suite, the issue is related to a symlink in the tests [censored version here](https://github.com/spotandjake/ocp-indent), I think we might need to make a pr handling to the esy opam patches repo handling this behavior correctly, still ran into issues with conf-cmake though.

hhugo:
In #2255, I left the binaryen dep unchanged and didn't see any issue with dune

Issue #1816, opened 2023-04-18
Remove --experimental-wasm-return-call flag when wasm-return-call is no longer experimental


phated:
We set the v8 flag in the runner currently.

PR #2256, opened 2025-03-25
chore(cli)!: Require node `>=22.13`

spotandjake:
This pr switches us from node `v18` to `22` (lts). As [pkg](https://github.com/vercel/pkg) was deprecated, we are now using a maintained fork [@yao-pkg/pkg](https://github.com/yao-pkg/pkg) however after a discussion in discord and some concerns with the security of the binary I decided to fork the repos so we have upstream control.

Closes: #1816


.github/workflows/build-js.yml
```diff
@@ -42,9 +42,9 @@ jobs:
           ref: ${{ inputs.ref }}
 
       - name: Setup node.js
-        uses: actions/setup-node@v3.6.0
+        uses: actions/setup-node@v4.3.0
```
ospencer, line 45: Can we use hashes instead?

spotandjake: Done

cli/package.json
```diff
@@ -38,7 +38,7 @@
     "commander": "^8.1.0"
   },
   "devDependencies": {
-    "@phated/pkg": "5.9.1-grain",
+    "@spotandjake/pkg": "6.4.0",
```
ospencer, line 41: Could we instead fork this in the Grain org?

spotandjake: Below should be a working workflow to fork it into the grain org (obviously using the pkg-fetch versions that match with upstream):
1) Fork pkg-fetch
2) Run build_all with no release
3) Update shas
4) make commit
5) Publish the release
6) publish to npm with npm run publish, publish as v3.6 not v3.6.0 in the tag

----

7) Fork pkg
8) Switch to use @grain/pkg-fetch
9) Regen yarn config
10) Test
11) Publish

From what I remember we need a `github PAT` and `npm PAT` for the release process. I'm happy to fork it over and everything but I don't think I currently have permissions todo that.

spotandjake: This should be ready to go now.

ospencer (approved):
Excellent!

Issue #2250, opened 2025-02-24
Add `Float64.reinterpretInt64`, `Int64.reinterpretFloat64`


spotandjake:
While working on a `miniFloat` implementation I noticed that while we have good support for converting back and forth between number types we do not have anything equivalent to `WasmF64.reinterpretI64` which allows for raw conversations more useful when working with the bits themselves. I think implementation of these would essentially be `untag`, `call wasm primitive`, `retag` we should probably add these to all numeric data types.

PR #2257, opened 2025-03-31
feat(stdlib): Implement `reinterpret` between float and int values

spotandjake:
This implements the following functions:
- `Int32.reinterpretFloat32`
- `UInt32.reinterpretFloat32`
- `Int64.reinterpretFloat64`
- `UInt64.reinterpretFloat64`
- `Float32.reinterpretInt32`
- `Float32.reinterpretUint32`
- `Float64.reinterpretInt64`
- `Float64.reinterpretUint64`
These act pretty similarly to their lower level wasm counterparts. 

Note: While I was working I also did some cleanup related to #703 though I'm happy to undo that if we decide it doesn't fit into this pr.

Closes: #2250

### DIFF
--- a/compiler/test/stdlib/float32.test.gr
+++ b/compiler/test/stdlib/float32.test.gr
@@ -26,6 +26,18 @@ assert fromNumber(0) == 0.0f
 assert toNumber(555.0f) == 555
 assert toNumber(0.0f) == 0
 
+// Float32.reinterpretInt32
+assert Float32.reinterpretInt32(0l) == 0.0f
+assert Float32.reinterpretInt32(1065353216l) == 1.0f
+assert Float32.reinterpretInt32(-1082130432l) == -1.0f
+assert Float32.reinterpretInt32(1109917696l) == 42.0f
+
+// Float32.reinterpretUint32
+assert Float32.reinterpretUint32(0ul) == 0.0f
+assert Float32.reinterpretUint32(1065353216ul) == 1.0f
+assert Float32.reinterpretUint32(3212836864ul) == -1.0f
+assert Float32.reinterpretUint32(1109917696ul) == 42.0f
+
 // Float32.pow tests are based on test cases from libc-test: http://nsz.repo.hu/git/?p=libc-test
 /*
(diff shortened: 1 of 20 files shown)


spotandjake:
I undid the offset constant changes in this pr and only the new functions make use of the constants for now.

alex-snezhko:
Not necessarily against this PR but I was curious when people would actually use these? My sense is that if people are looking to do low-level bit manipulation algorithms they would most likely be using the low level unsafe types rather than the high level versions that have allocation overhead

spotandjake:
> Not necessarily against this PR but I was curious when people would actually use these? My sense is that if people are looking to do low-level bit manipulation algorithms they would most likely be using the low level unsafe types rather than the high level versions that have allocation overhead

I think that goes directly against our philosophy that unsafe is not to be used by users and only for runtime use, these are certainly a bit more low level then what we currently have but if our goal is to encourage users to stay out of unsafe grain then these seem essential for certain algorithms. 

I generally agree with your argument on dropping down for perf reasons but I don't think we should be encouraging or restricting our stdlib api's under that assumption.

ospencer:
Pretty much what Jake said—we don't want users to do the unsafe thing, so when they _do_ need to reach for something like this it's available. OCaml gives you this as `Int32.bits_of_float` and `Int32.float_of_bits`. I actually kind of like those names better because they're more obvious than the wasm instruction imo, but I think using the wasm instruction name is fine.

spotandjake:
> Pretty much what Jake said—we don't want users to do the unsafe thing, so when they _do_ need to reach for something like this it's available. OCaml gives you this as `Int32.bits_of_float` and `Int32.float_of_bits`. I actually kind of like those names better because they're more obvious than the wasm instruction imo, but I think using the wasm instruction name is fine.

I feel like with the ocaml names I would expect them to return a `Bytes` object. You can already do the raw interpret using `setFloat64` and `setInt64` with a `Bytes` object but that obviously incurs overhead and isn't a very great workflow.

ospencer (approved):

Issue #2258, opened 2025-03-31
Implement `Buffer.grow` or `Buffer.resize`


spotandjake:
I noticed while working with the `Buffer` library that we only have capabilities to quickly shrink a buffer but not grow one. It might be useful to implement a `Buffer.resize` or `Buffer.grow`

PR #2260, opened 2025-03-31
feat(stdlib): Add `Buffer.resize(len, buf)`

spotandjake:
This pr implements `Buffer.resize`, it either resizes the buffer to the given length or truncates it to the set length. This is mostly useful for scenarios where buffer is being used in combo with `get` and `set` functions.

Closes: #2258

### DIFF
--- a/compiler/test/stdlib/buffer.test.gr
+++ b/compiler/test/stdlib/buffer.test.gr
@@ -185,6 +185,19 @@ assert Buffer.getInt8(0, buf) == 1s
 let bytes = Buffer.toBytes(buf)
 assert Bytes.length(bytes) == 1
 
+// Buffer.resize
+let buf = Buffer.make(0)
+assert Buffer.length(buf) == 0
+Buffer.resize(0, buf)
+assert Buffer.length(buf) == 0
+Buffer.resize(8, buf)
+assert Buffer.length(buf) == 8
+Buffer.setInt8(7, 2s, buf)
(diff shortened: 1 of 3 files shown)


ospencer:
What's the use case here? Generally users shouldn't have the ability to mess with the underlying `Bytes`; the whole point of `Buffer` is to not really have to think about it. I didn't really follow what you meant about this being useful with `get` and `set` functions.

spotandjake:
The main use I had here was in an interpreter using `buffer` as a linear memory, if I just use `Buffer.getInt8` or `Buffer.setInt8` I run into bounds errors and the only workaround at the moment is something along the lines of `Buffer.addBytes(Bytes.make(Buffer.length(buf)-desiredSize), buf)` see [here](https://github.com/spotandjake/Grain-Brainf/blob/cbc5d9f5d2c0f5b0099ab6644c754d8b5ac22597/brainf.gr#L25).

Maybe a `Buffer.grow` would be better then resize here though.

ospencer:
For your brainfuck interpreter? You definitely don't want to use a buffer for that; buffers are for appending data, not for random access. It's also going to be slower. I know it's handy to not have to worry about resizing since you don't know the length of the tape up front, but it's better to just allocate a big `Bytes` and then use `Bytes.resize` if you learn you need more space. Using a buffer doesn't get you much because you need to check that the read/write is in bounds anyway.

PR #2262, opened 2025-04-01
feat(stdlib): Add `Buffer.getChar`

spotandjake:
This adds `Buffer.setChar` and `Buffer.getChar` to complement their bytes counterparts.

Note: We have to read the first byte before `getChar` so we can correctly do the bounds check similar to how its done in `Bytes.getChar`.

Note: I realized we missed a piece of documentation on `Bytes.getChar` so I added it here.

### DIFF
--- a/stdlib/buffer.gr
+++ b/stdlib/buffer.gr
@@ -384,7 +384,7 @@ provide let addString = (string, buffer) => {
  * @returns A character starting at the given index
  *
  * @throws IndexOutOfBounds: When `index` is negative
- * @throws IndexOutOfBounds: When `index + 1` is greater than the bytes size
+ * @throws IndexOutOfBounds: When `index + 1` is greater than the buffer size
  * @throws MalformedUnicode: When the bytes at the index are not a valid UTF-8 sequence
  *
  * @example
--- a/stdlib/buffer.gr
+++ b/stdlib/buffer.gr
@@ -24,7 +24,7 @@ from "char" include Char
 from "runtime/numbers" include Numbers
 use Numbers.{ coerceNumberToWasmI32 }
 from "runtime/utf8" include Utf8
-use Utf8.{ usvEncodeLength }
+use Utf8.{ usvEncodeLength, utf8ByteCount, exception MalformedUnicode }
 from "runtime/unsafe/offsets" include Offsets
 use Offsets.{ _BYTES_LEN_OFFSET, _BYTES_DATA_OFFSET }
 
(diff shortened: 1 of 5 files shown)


stdlib/buffer.gr
```diff
@@ -376,6 +376,68 @@ provide let addString = (string, buffer) => {
...
+ * @returns A character starting at the given index
+ *
+ * @throws IndexOutOfBounds: When `index` is negative
+ * @throws IndexOutOfBounds: When `index + 1` is greater than the bytes size
+ * @throws MalformedUnicode: When the requested character is not a valid UTF-8 sequence
+ *
+ * @example
+ * let buf = Buffer.make(32)
+ * Buffer.addString("Hello World 🌾", buf)
+ * assert Buffer.getChar(12, buf) == 'H'
```
alex-snezhko, line 393: Is this a mistake?

stdlib/buffer.gr
```diff
@@ -376,6 +376,68 @@ provide let addString = (string, buffer) => {
...
+ *
+ * @since v0.7.0
+ */
+@unsafe
+provide let getChar = (index, buffer) => {
+  use WasmI32.{ (+), (&), (+), (==), (>) }
+  checkIsIndexInBounds(index, 1, buffer)
```
alex-snezhko, line 400: Why is this 1 if it's UTF-8?

stdlib/buffer.gr
```diff
@@ -376,6 +376,68 @@ provide let addString = (string, buffer) => {
   buffer.len += bytelen
 }
 
+/**
+ * Gets the UTF-8 encoded character at the given byte index.
```
alex-snezhko, line 380: If the operation is intended to get a character starting at a byte index then if you point your index in the middle of a UTF-8 character then what would the expectation be?

spotandjake: I'm pretty sure it would either return a character in the case that the rest was a valid char (I can't recall if thats ever the case), but more likely `MalformedUnicode`.

stdlib/buffer.gr
spotandjake: Good catch I'll correct that in a second.

stdlib/buffer.gr
spotandjake: Characters can be between `1` and `4` bytes, we need to ensure the first byte exists so we can check the char size on line 406 and we do an additional length check on line 408 with the actual length.

stdlib/buffer.gr
```diff
@@ -376,6 +376,68 @@ provide let addString = (string, buffer) => {
...
+ *
+ * @param index: The byte index to access
+ * @param buffer: The buffer to access
+ * @returns A character starting at the given index
+ *
+ * @throws IndexOutOfBounds: When `index` is negative
+ * @throws IndexOutOfBounds: When `index + 1` is greater than the bytes size
+ * @throws MalformedUnicode: When the requested character is not a valid UTF-8 sequence
```
ospencer, line 388:
```suggestion
 * @throws MalformedUnicode: When the bytes at the index are not a valid UTF-8 sequence
```

stdlib/buffer.gr
```diff
@@ -376,6 +376,68 @@ provide let addString = (string, buffer) => {
...
 
+/**
+ * Gets the UTF-8 encoded character at the given byte index.
+ *
+ * @param index: The byte index to access
+ * @param buffer: The buffer to access
+ * @returns A character starting at the given index
+ *
+ * @throws IndexOutOfBounds: When `index` is negative
+ * @throws IndexOutOfBounds: When `index + 1` is greater than the bytes size
```
ospencer, line 387: It's not just + 1, right? If you ask for a character that's 4 bytes long then that whole range needs to be valid.

ospencer (commented):
I'm somewhat against these get/set APIs (not just the ones introduced in this PR, the existing ones too). I think it really encourages the wrong use of Buffers; they're really just meant for appending data, and they add a bunch of mental burden of "am I supposed to use `add` or `set`?". I'm more in favor of removing the existing ones rather than adding more. What are the use cases here?

stdlib/buffer.gr
spotandjake: So its `index + 1` to get the usv and if the length doesn't match up then its just an invalid usv.

spotandjake:
> I'm somewhat against these get/set APIs (not just the ones introduced in this PR, the existing ones too). I think it really encourages the wrong use of Buffers; they're really just meant for appending data, and they add a bunch of mental burden of "am I supposed to use `add` or `set`?". I'm more in favor of removing the existing ones rather than adding more. What are the use cases here?

If we have any of the `get` and `set` commands I think these should exist to match. If we want to talk about removing them though I think its fair, most applications would possibly be better served by `Bytes`, and `Bytes.resize`.

ospencer:
We only have them for integers (so they're missing for everything else), and they landed in the initial implementation... I think we were just kinda caught up in that implementation and didn't audit the full API.

spotandjake:
I removed `Buffer.setChar` as discussed at the community meeting and will follow up with a pr removing the other `Buffer.set` apis in a pr closer to 0.8.

stdlib/buffer.gr
ospencer:
```suggestion
 * @throws IndexOutOfBounds: When `index + 1` is greater than the buffer size
```

Issue #1983, opened 2024-01-26
GrainDoc: Improve Errors


spotandjake:
Currently, if you mess up the label on a `@param` you get an error such as: `Unable to find a matching function parameter for str. Make sure a parameter exists with this label or use `@param <param_index> str` for unlabeled parameters.`, it would be beneficial if this included the file and function that this is related to as when generating docs for a directory its unclear where the error is coming from.

After looking into this more, it seems that none of our errors contain locations or function names.

PR #2263, opened 2025-04-01
feat(graindoc): Add bind information to errors

spotandjake:
This just adds the bind that is causing the issue to help with debugging. I think it would be beneficial if we had the full error location as well but I don't quite understand how our error_reporting works enough to make that change yet. 

Work towards: #1983

### DIFF
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -95,10 +95,22 @@ exception
     attr: string,
   });
 
-exception MissingLabeledParamType({name: string});
-exception MissingUnlabeledParamType({idx: int});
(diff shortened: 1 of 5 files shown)


ospencer:
I think it'd be better to just provide the location. We do lex the comments, so theoretically the information is there.

spotandjake:
> I think it'd be better to just provide the location. We do lex the comments, so theoretically the information is there.

That would probably be better, I didn't know how to properly hook into our error handler with the nice locations from the compiler.

spotandjake:
Closing as #2291 implements the desired solution.

Issue #2049, opened 2024-03-01
Formatter: Missing trailing comma on inline record fields


phated:
I noticed in the JSON PR that the new formatter removed a bunch of trailing commas on the last inline record field.

PR #2266, opened 2025-04-02
feat(fmt): Add trailing comma on inline records

spotandjake:
This adds a trailing comma on inline record variants

Closes: #2049


ospencer (commented):
Why is this breaking?

spotandjake:
> Why is this breaking?

I didn't know if we count formatting changes as breaking or not, but I've changed it to non breaking.

ospencer (approved):

Issue #2235, opened 2025-01-19
Includes test suites fails when run indiviudally


spotandjake:
It seems as though we are capturing some external state from the rest of the test suites when running the `includes` suite that causes the snapshots to change when its run individually. We should do some investigation to this and correct this behaviour.

PR #2272, opened 2025-04-12
chore(compiler): Properly handle disable_stamps state

ospencer:
I noticed that we weren't properly resetting the `disable_stamps` state in a handful of tests, causing a good chunk of the test suite to not have stamps in their snapshots. For some test suites, this would cause them to fail if run individually.

There's somewhat of a discussion to be had for disabling stamps in a bunch of the snapshots, but I find the stamps really helpful for debugging and think we should keep them. (They really only change when we add new global types.)


spotandjake:
Closes: #2235

spotandjake (approved):
Looks good to me.

marcusroberts (approved):
LGTM!

phated (approved):

Issue #2221, opened 2024-12-29
Pattern matching exhaustiveness for records


alex-snezhko:
The following program does not complain at compile-time but will fail at runtime

```
module Main
record T { a: Option<String> }
match ({ a: Some("b") }) {
  { a: None } => void,
}
```

spotandjake:
This might be related to #2219

Issue #2219, opened 2024-12-23
Inline Records On Adt's Ignore exhaustiveness checking


spotandjake:
I was working on #2217 and noticed that we do not currently have any printing logic for inline records, when I went to test what happens I noticed that somehow once inline records are introduced we don't emit an exhaustiveness error at all which is rather interesting behaviour, see below for an example: 
```gr
module Test

enum Test {
  Case1{ v: Void, },
  Case2{ v: Void, },
}
let n = Case2{ v: void }
match (n) {
  Case1{ _ } => void,
}
```
The code above compiles just fine and doesn't cause an exhaustiveness error even though it is clearly non exhaustive, this is extremely interesting as this isn't just a case of the record matches not being exhaustive but the actual constructor check not being exhuastive.

To the original issue as well we do not implement printing logic in `Printpat.re` for inline records.

spotandjake:
Closing as this was fixed with #2274

PR #2274, opened 2025-04-15
fix(compiler): Handle non-exhaustive record patterns

spotandjake:
This implements exhaustiveness checking both for regular records and inline records. I split the two into seperate commits as we need the record pattern logic for inline adt's but some additional debugging was needed to get inlines working.

### Record Exhaustiveness

I mostly ported the logic over from [parmatch.ml in the ocaml compiler](https://github.com/ocaml/ocaml/blob/8c52855f31804a86384ba1ea8981b14b7a5f1c02/typing/parmatch.ml), though they are not 1 for 1 as our implementation has diverged a decent bit, the foundational logic is the same.

### Inline ADT Records

It was a little weird that previously including a record variant would cause you to get no exhaustiveness info even on non record variants in your enum, it turned out this was because when converting back to our parsetree nodes we were just treating everything as a tuple causing a `misused_inline_record` error internally.

### Changes to `queue.gr`, `mixedPatternMatching.gr`

This actually revealed some cases we missed, so I implemented them, the case in `queue.gr` would never actually be possible because of how we build the queues.

Closes: #2221, #2219

### DIFF
--- a/compiler/src/typed/parmatch.re
+++ b/compiler/src/typed/parmatch.re
@@ -438,7 +438,7 @@ let record_arg = ph => {
   switch (ph.pat_desc) {
   | TPatAny => []
   | TPatRecord(args, _) => args
-  | _ => fatal_error("Parmatch.as_record")
+  | _ => fatal_error("Parmatch.record_arg")
   };
 };
 
--- a/compiler/src/typed/parmatch.re
+++ b/compiler/src/typed/parmatch.re
@@ -177,6 +177,12 @@ let all_coherent = column => {
         false
       }
     | (TPatTuple(l1), TPatTuple(l2)) => List.length(l1) == List.length(l2)
+    | (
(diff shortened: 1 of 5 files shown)


compiler/src/typed/parmatch.re
```diff
@@ -390,23 +426,45 @@ let simple_match = (p1, p2) =>
...
   | (_, TPatAny | TPatVar(_)) => true
   | (_, _) => false
   };
 
+/* extract record fields as a whole */
+let record_arg = ph => {
+  switch (ph.pat_desc) {
+  | TPatAny => []
+  | TPatRecord(args, _) => args
+  | _ => fatal_error("Parmatch.as_record")
```
ospencer, line 441:
```suggestion
  | _ => fatal_error("Parmatch.record_arg")
```

PR #2275, opened 2025-04-15
fix(lsp): Prevent crash on alias module hover

### LINKED ISSUE #2179: LSP: Crash when aliasing a module
The following code causes the lsp to crash with a `Not_Found`:
```gr
from "./host.gr" include Host
use Host.{ module Bindings as A }
print(A.add(1ul, 2ul))
```
the issue seems to come from aliasing in the `module Bindings as A` the exception seems to be thrown from [Sourcetree](https://github.com/grain-lang/grain/blob/584c4d29a883fca7d774efb91102bad35854afa6/compiler/src/language_server/sourcetree.re#L308) when we call `find_module`, we could fix this by just adding a try catch and not including the item in sourcetree but that just hides the bug and I think this is indicative of a deeper problem. This exception happens when the `didOpen` event fires.

spotandjake:
This pr prevents the lsp crash caused by #2179 by using the value path over looking up the module, I think this works well in this case however I noticed that using an aliased module with a `use` statement causes a deeper compiler crash which makes me think we should correct the behaviour in `find_module`. I don't quite know the best way about adding that but wanted to get a pr up preventing the crashes until we can investigate deeper. 

I'm not sure if we want to close #2179 with this if we do then we should probably open a seperate issue for the `use` crash.

### DIFF
--- a/compiler/src/language_server/sourcetree.re
+++ b/compiler/src/language_server/sourcetree.re
@@ -303,35 +303,48 @@ module Sourcetree: Sourcetree = {
             Path.(
               switch (exp.exp_desc) {
               | TExpIdent(
-                  PExternal(path, _),
+                  PExternal(_, _),
                   {txt: IdentExternal(IdentName({loc}), _)},
-                  desc,
+                  {val_type, val_loc, val_internalpath: PExternal(path, _)},
                 ) =>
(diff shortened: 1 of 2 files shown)


ospencer:
Let's not try to work around the bug; let's fix the bug.

spotandjake:
Closing in favour of #2285

PR #2278, opened 2025-04-19
chore(compiler): Clean up duplicate `Warnings.is_active` checks

spotandjake:
Just some tiny cleanup, `Location.prerr_warning` checks `Warnings.is_active` [source here](https://github.com/grain-lang/grain/blob/3cf6021b7c7a228542fd300e9bf33c5b63bcc42d/compiler/src/parsing/location.re#L416) before actually emitting the error so doing it in these locations just makes things slower and adds code noise.


ospencer (approved):
Note that we typically use the `compiler` scope for compiler changes (no one outside of Grain core really thinks about `grainc`).

Issue #2211, opened 2024-11-26
Grain seems to treat a type as a different type when the module "as" functionality is used when importing its module


godotdot:
The following code (also implemented in this attachment: 
[bad_module_example.zip](https://github.com/user-attachments/files/17926231/bad_module_example.zip)) exhibits the following behavior:
```
grain main.gr 
File "main.gr", line 9, characters 10-20:
Error: This expression has type DifferentName.Type
       but an expression was expected of type TypeProvider.Type
TypeProvider::Type is abstract because no corresponding cmi file was found in path.
```

I would expect DifferentName.Type and TypeProvider.Type would be treated as the same type, as DifferentName.Type is just an alias/renaming of TypeProvider.Type.

There are similarities between this issue and https://github.com/grain-lang/grain/issues/1758.

main.gr
```
module Main    
    
from "./type_provider.gr" include TypeProvider as DifferentName    
use DifferentName.{ type Type }    
                                  
from "./function_provider.gr" include FunctionProvider    
use FunctionProvider.{ function }    
    
function( {field: 5} )
 ```

type_provider.gr
```
module TypeProvider

provide record Type {
    field: Number
}
```

function_provider.gr
```
module FunctionProvider    
    
from "./type_provider.gr" include TypeProvider     
use TypeProvider.{ type Type }     
    
provide let function = (t: Type) => {    
    void    
}    

```

PR #2285, opened 2025-04-23
fix(compiler): Resolve Not_found for use statements with module aliases

### LINKED ISSUE #2179: LSP: Crash when aliasing a module

ospencer:
Fixes #2179
Closes #2275

### DIFF
--- a/compiler/test/suites/provides.re
+++ b/compiler/test/suites/provides.re
@@ -255,4 +255,20 @@ describe("provides", ({test, testSkip}) => {
     "reprovideTypeUnifyA",
     "true\n",
   );
+  assertRun(
+    "use_multilevel_alias",
+    {|
+      module A {
+        provide module B {
+          provide let b = 1
+        }
+      }
+
+      use A.{ module B as C }
+      use C.{ b }
+
+      print(b)
+    |},
+    "1\n",
+  );
 });
--- a/compiler/src/typed/env.re
+++ b/compiler/src/typed/env.re
@@ -1925,10 +1925,10 @@ let rec add_signature = (sg, env) =>
 
 /* Open a signature path */
 
-let add_components = (slot, root, env0, ~type_aliases=Tbl.empty, comps) => {
+let add_components = (slot, root, env0, ~aliases=Tbl.empty, comps) => {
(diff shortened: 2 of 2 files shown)


compiler/test/suites/provides.re
spotandjake, line 1: Do you think it would make sense to add a regression test specifically for the lsp as well, something like:
```
module Main

module A {
  provide module B {
    provide let b = 1
  }
}

use A.{ module B as C }
print(C.b)
```
testing with a hover request.  This would cause the lsp to crash but the compiler didn't have any issues because it's a value lookup.

spotandjake (commented):
This looks great, one concern I still have is related to #2211 we still treat `from "x" include Mod as Alias` completely separately, but we could probably leave that as future work, the lsp working as expected now.

compiler/test/suites/provides.re
ospencer: I can, if you want, but I think the regular test covers this case pretty well since this is core language functionality. It doesn't really have anything to do with the LSP (the LSP was crashing because the compiler was broken, not because the LSP was broken).

ospencer:
> This looks great, one concern I still have is related to #2211 we still treat `from "x" include Mod as Alias` completely separately, but we could probably leave that as future work, the lsp working as expected now.

They're kind of different things! There isn't really a way to make them completely use the same code path because there would need to be a module that the import already existed in (but we're just now importing it).

spotandjake (approved):
Both fair points, let's get this merged, thanks for fixing this.

alex-snezhko (approved):

Issue #2287, opened 2025-05-17
Grainfmt: Expression with line comment before end of block formats with additional newline


ospencer:
Formatting

```grain
module Foo

module Bar {
  1 // foo
}
```

will add an additional newline at the end of the module, and formatting again will do the same. It should add no additional lines.

PR #2288, opened 2025-05-17
fix(grainfmt): Correct formatting of submodules ending with comments

spotandjake:
This corrects the formatting of cases like:
```gr
module Foo {
  foo // foo
}
```
Previously we were adding a new hardline at the end each time the formatter was run, it seems we were not correctly handling the linebreak calculation as were not printing the comment range as if it were at the end of a block. 

Closes: #2287


ospencer (approved):
Thanks @spotandjake!

PR #2291, opened 2025-05-18
feat(graindoc): Add locations to Graindoc errors

ospencer:


marcusroberts (approved):
Very useful!

spotandjake (approved):
LGTM.

Closes: #1983

compiler/src/diagnostics/comment_attributes.re
```diff
@@ -1,11 +1,15 @@
-exception InvalidAttribute(string);
-exception MalformedAttribute(string, string);
+open Grain_parsing;
 
 type param_id =
-  | LabeledParam(string)
-  | PositionalParam(int);
+  | LabeledParam(string, Location.t)
```
alex-snezhko, line 4: These locs looks unused, unless I missed it?

compiler/graindoc/docblock.re
```diff
@@ -7,28 +8,41 @@ type param = {
   param_id: string,
   param_type: string,
   param_msg: string,
+  param_loc: Location.t,
```
alex-snezhko, line 11: Nit could use `Location.loc('a)` for these? May make some of the pattern matches a bit more consistent rather than having to keep track of the location field of each type :shrug:

compiler/graindoc/docblock.re
```diff
@@ -89,61 +103,60 @@ and provided = {
...
-        Printf.sprintf("Attribute @%s is only allowed to appear once.", attr);
-      Some(msg);
-    | InvalidAttribute({name, attr}) =>
-      let msg = Printf.sprintf("Invalid attribute @%s on %s", attr, name);
-      Some(msg);
-    | _ => None
-    }
-  });
+  Location.register_error_of_exn(
+    fun
```
alex-snezhko, line 155: Nit I thought we generally avoid using `fun`

alex-snezhko (approved):

compiler/graindoc/docblock.re
ospencer: The `loc` type is good for when you don't need to carry the location information forward. I feel like it's too easy to do a `.txt` and then pass these record types along, and when you realize you don't have the location information you have to go back and thread it through.

That said, I actually don't like having these types here. They're only used in this file and I think named arguments would be just fine. I didn't want to make that opinionated change in this PR though.

compiler/graindoc/docblock.re
ospencer: No fun allowed! This is how we do all of the `register_error_of_exn`s so I just copied it, but I'll change it if you want.

compiler/src/diagnostics/comment_attributes.re
ospencer: Not used. I was mostly setting up the opportunity to warn/error specifically on the name/position.

PR #2292, opened 2025-05-19
feat(graindoc): Pretty markdown tables

ospencer:
Prints Graindoc markdown tables in the pretty format, with all rows of equal length. Is slightly faster (up to 10-15% across the stdlib) than before because it does less concatenation to create the tables.


spotandjake (approved):

ospencer:
Realized I ran the wrong command locally.

PR #2293, opened 2025-05-22
fix(graindoc): Add all function parameters to the table, even if missing documentation

### LINKED ISSUE #2281: Graindoc: Consider warning or adding undocumented params to the param table.
I noticed if you document a program like:
```gr
module Main
/**
 * @param a: This is the first
 */
provide let test = (a, b) => a + b
```
The output parameter table only has an entry for `a`, while this isn't the end of the world because the output signature is still correct I think it would still be beneficial to output the `b` param to the table with type info and a blank description. Alternatively we might want to consider emitting a warning (maybe we allow a `--lint` flag on graindoc? this is a bigger discussion). The reason the current behaviour is problematic is it's easy to miss undocumented parameters and then the parameter table (which as a user reading docs I would consider a source of truth) isn't complete and doesn't match the function signature.

marcusroberts:
Partially Fixes #2281 

Add all function parameters to the documentation table in the order they are defined.   

Add an empty documentation field for any that are missing user supplied documentation in the comment above.

(New and Updated tests to follow)

### DIFF
--- a/compiler/graindoc/docblock.re
+++ b/compiler/graindoc/docblock.re
@@ -4,6 +4,8 @@ open Grain_typed;
 open Grain_utils;
 open Grain_diagnostics;
 
+module StringSet = Set.Make(String);
+
 type param = {
   param_id: string,
   param_type: string,
@@ -110,6 +112,7 @@ type error =
     })
   | MissingLabeledParamType({name: string})
   | MissingUnlabeledParamType({idx: int})
+  | DuplicateLabelParamType({name: string})
   | MissingReturnType
   | AttributeAppearsMultipleTimes({attr: string})
   | InvalidAttribute({
@@ -144,6 +147,12 @@ let report_error = (ppf, err) => {
       ppf,
       "Unable to find a type for parameter at index %d. Make sure a parameter exists at this index in the parameter list.",
       idx,
+    )
+  | DuplicateLabelParamType({name}) =>
+    Format.fprintf(
+      ppf,
+      "Multiple `@param` attributes found for parameter %s. Each parameter may only have one associated `@param` attribute.",
+      name,
     )
   | MissingReturnType =>
     Format.fprintf(ppf, "Unable to find a return type. Please file an issue!")
@@ -489,19 +498,32 @@ let for_value_description =
     };
 
   // Validate that all parameter attributes were used (i.e. no invalid labels, no duplicates)
-  List.iter(
-    ((label, (_, loc, used))) =>
+  ignore @@
+  List.fold_left(
+    (acc, (label, (_, loc, used))) => {
+      let label_name =
+        switch (label) {
+        | LabeledArg(name) => name
+        | UnlabeledArg(idx) => string_of_int(idx)
+        };
       if (used^ == false) {
         raise(
           Error(
             loc,
-            switch (label) {
-            | LabeledArg(name) => MissingLabeledParamType({name: name})
-            | UnlabeledArg(idx) => MissingUnlabeledParamType({idx: idx})
+            if (StringSet.mem(label_name, acc)) {
+              DuplicateLabelParamType({name: label_name});
+            } else {
+              switch (label) {
+              | LabeledArg(name) => MissingLabeledParamType({name: name})
+              | UnlabeledArg(idx) => MissingUnlabeledParamType({idx: idx})
+              };
             },
           ),
         );
-      },
+      };
+      StringSet.add(label_name, acc);
+    },
+    StringSet.empty,
     param_attributes,
   );
 
--- a/compiler/test/suites/graindoc.re
+++ b/compiler/test/suites/graindoc.re
@@ -59,7 +59,7 @@ describe("graindoc", ({test, testSkip}) => {
   assertGrainDocError(
     "paramDuplicate",
     "paramDuplicate",
-    "Error: Unable to find a matching function parameter for a. Make sure a parameter exists with this label or use `@param <param_index> a` for unlabeled parameters.",
+    "Error: Multiple `@param` attributes found for parameter a. Each parameter may only have one associated `@param` attribute.",
     [|"--current-version=v0.2.0"|],
   );
   assertGrainDocError(
--- a/compiler/test/runner.re
+++ b/compiler/test/runner.re
@@ -644,7 +644,6 @@ let makeGrainDocRunner = (test, name, filename, arguments, stderr) => {
       let (result, err, _) = doc(infile, arguments);
 
       // we need do a binary content comparison to ensure the EOL is correct
-      let outfile = graindoc_out_file(name);
       expect.ext.binaryFile(result).toBinaryMatch(graindoc_out_file(name));
       expect.string(err).toMatch(stderr);
     },


compiler/graindoc/docblock.re
```diff
@@ -376,6 +376,68 @@ let for_value_description =
...
+          | PositionalParam(idx) => Some((string_of_int(idx), attr))
+          | LabeledParam(name) => Some((name, attr))
+          }
+        | _ => None
+        },
+      attributes,
+    );
+
+  let match_label_to_arg = arg => {
+    List.find_opt(((name, _)) => name == arg, param_attributes);
```
spotandjake: I think `List.assoc_opt` or `List.assq_opt` might be what you want to use here https://ocaml.org/manual/5.3/api/List.html#1_Associationlists

compiler/graindoc/docblock.re
```diff
@@ -376,6 +376,68 @@ let for_value_description =
...
+            | Typedtree.Labeled(l)
+            | Default(l) => l.txt
+            | _ => string_of_int(index)
+            };
+
+          // First try to match by name, then by position
+          let matched =
+            switch (match_label_to_arg(param_id)) {
+            | Some(_) as result => result
+            | None => match_label_to_arg(string_of_int(index))
```
spotandjake: Is the behaviour here that if you documented a param as `x` but the function only has position params we would map `x` to `0` for instance?

compiler/graindoc/docblock.re
```diff
@@ -376,6 +376,68 @@ let for_value_description =
...
+            };
+
+          // Extract documentation from the matched attribute
+          let documentation =
+            switch (matched) {
+            | Some((_, Comment_attributes.Param({attr_desc}))) => attr_desc
+            | _ => ""
```
spotandjake: Should we be throwing some sort of exception here?

compiler/graindoc/docblock.re
```diff
@@ -376,6 +376,68 @@ let for_value_description =
...
+            };
+
+          // Get parameter type
+          let param_type =
+            switch (lookup_arg_by_label(param_id, args)) {
+            | Some((Labeled(_), typ)) => Printtyp.string_of_type_sch(typ)
+            | Some((Default(_), {desc: TTyConstr(_, [typ], _)})) =>
+              Printtyp.string_of_type_sch(typ)
+            | Some((_, typ)) => Printtyp.string_of_type_sch(typ)
+            | None => ""
```
spotandjake: Should we be throwing some sort of exception here? And do you know if there is any real case this would occur?

compiler/graindoc/docblock.re
```diff
@@ -376,6 +376,68 @@ let for_value_description =
...
+          let documentation =
+            switch (matched) {
+            | Some((_, Comment_attributes.Param({attr_desc}))) => attr_desc
+            | _ => ""
+            };
+
+          // Get parameter type
+          let param_type =
+            switch (lookup_arg_by_label(param_id, args)) {
+            | Some((Labeled(_), typ)) => Printtyp.string_of_type_sch(typ)
```
spotandjake: Nit: This could probably use an `or` pattern, instead of handling each branch with the same logic.

marcusroberts:
Take care when merging #2292 which makes changes to Graindoc.

compiler/test/graindoc/params.input.gr
spotandjake, line 1: Should we also test a function without the params documented at all, and an only positional argument.

We might also want to test the failing cases, i.e a param doesn't exist.

compiler/graindoc/docblock.re
```diff
@@ -376,6 +376,72 @@ let for_value_description =
...
+            let documentation =
+              switch (matched) {
+              | Some((_, Comment_attributes.Param({attr_desc}))) => attr_desc
+              | _ => ""
+              };
+
+            // Get parameter type
+            let param_type =
+              switch (lookup_arg_by_label(param_id, args)) {
+              | Some((Labeled(_), typ)) => Printtyp.string_of_type_sch(typ)
```
spotandjake: nit: We could use an `or` pattern here to reduce the number of branches.

compiler/graindoc/docblock.re
```diff
@@ -414,41 +480,9 @@ let for_value_description =
...
-        | Param({attr_id: param_id, attr_desc: param_msg}) =>
-          let (param_id, param_type) =
-            switch (param_id) {
-            | PositionalParam(idx) =>
-              switch (lookup_type_expr(~idx, args)) {
-              | Some((_, typ)) => (
-                  string_of_int(idx),
-                  Printtyp.string_of_type_sch(typ),
-                )
-              | None => raise(MissingUnlabeledParamType({idx: idx}))
```
spotandjake: Why do we never throw these errors anymore?

* Do we now allow labeling a positional arg with a label and the other way around? i.e:
```
/*
 * @param x: test
 */
 let test = (_ as x) => void
 ```
 * What happens now if you document a non existing param, or too many positional params? 
 
 I see we removed it here and decided to handle the cases but are we sure we want to from a linting and correctness standpoint?

File: compiler/graindoc/docblock.re [hunk 8]
```diff
@@ -376,6 +376,72 @@ let for_value_description =
...
+          (index, (lab, exp)) => {
+            let param_id =
+              switch (lab) {
+              | Typedtree.Labeled(l)
+              | Default(l) => l.txt
+              | _ => string_of_int(index)
+              };
+
+            // First try to match by name, then by position
+            let matched =
```
spotandjake: Shouldn't the logic match specifially by name or position? i.e if we see an `int` its a position and a name its a position? This way we can continue to give the well formedness errors from before.

compiler/test/graindoc/params.input.gr
marcusroberts: You're right.  Every other test covered this case as they all have undocumented params so I use those as my test case, but an explicit one would be better

compiler/test/suites/graindoc.re
```diff
@@ -43,4 +43,17 @@ describe("graindoc", ({test, testSkip}) => {
     "line 5",
     [|"--current-version=v0.2.0"|],
   );
+  assertGrainDocOutput("params", "params", [|"--current-version=v0.2.0"|]);
+  assertGrainDocError(
+    "paramDuplicate",
+    "paramDuplicate",
+    "Error: Unable to find a matching function parameter for a. Make sure a parameter exists with this label or use `@param <param_index> a` for unlabeled parameters.",
```
spotandjake: I just use the generic error here because of how I check for this; we could produce a more specific error, but it just wouldn't be efficient as we would need to check that any unused label attribute is unique in the label attribute list. I think for the purpose of the error, this is good enough for now.

compiler/test/graindoc/params.expected.md
```diff
@@ -0,0 +1,95 @@
...
+
+Parameters:
+
+| param | type     | description   |
+| ----- | -------- | ------------- |
+| `a`   | `Number` | This is the a |
+| `b`   | `Number` |               |
+| `c`   | `Number` | This is the c |
+
+File "/Users/spotandjake/Documents/github/grain/compiler/test/graindoc/params.input.gr", line 47, characters 12-17:
```
spotandjake: Because the test runner checks on stdout, the warning appears in the file for this.

compiler/src/utils/warnings.re
```diff
@@ -42,9 +42,10 @@ type t =
   | UselessRecordSpread
   | PrintUnsafe(string)
   | ToStringUnsafe(string)
-  | ArrayIndexNonInteger(string);
+  | ArrayIndexNonInteger(string)
+  | ParameterDocumentationMissing(string);
```
ospencer, line 46: I don't think we want to add this as a warning to the compiler proper. We should start a set of Graindoc warnings.

spotandjake: As a noted in discord until #2343, I don't think there's an easy way todo this while using the same handler and printing logic we use in the compiler. I think it's ideal we use the same printing logic so any printing improvements there get reflected back in the grain doc warnings. 

Maybe for #2342 the warning handler could be a functor that takes the error types and printing handler allowing us to use it in all of our tooling.

spotandjake: As a noted in discord until #2343, I don't think there's an easy way todo this while using the same handler and printing logic we use in the compiler. I think it's ideal we use the same printing logic so any printing improvements there get reflected back in the grain doc warnings. 

Maybe for #2342 the warning handler could be a functor that takes the error types and printing handler allowing us to use it in all of our tooling.


Unless you have a better idea I think we should keep the error handling here for now and move with #2342, we can renumber our errors then which shouldn't be a problem as we don't reference the numbers anywhere currently.

PR #2294, opened 2025-05-25
SIMD

ospencer:


compiler/esy.json
```diff
@@ -2,9 +2,7 @@
   "name": "@grain/compiler",
   "version": "0.7.0",
   "esy": {
-    "build": [
-      "dune build @native --no-buffer"
-    ],
+    "build": ["dune build @native --no-buffer"],
```
spotandjake, line 5: Why did the formatting change here?

compiler/test/stdlib/wasmv128.test.gr
```diff
@@ -0,0 +1,1040 @@
...
+
+from "runtime/unsafe/wasmi32" include WasmI32
+from "runtime/unsafe/wasmi64" include WasmI64
+from "runtime/unsafe/wasmf32" include WasmF32
+from "runtime/unsafe/wasmf64" include WasmF64
+
+from "runtime/unsafe/wasmv128" include WasmV128
+use WasmV128.*
+
+let (==) = (x, y) => {
```
spotandjake, line 13: Why not provide this in the runtime module?

compiler/test/stdlib/wasmv128.test.gr
spotandjake, line 1: `compiler/test/runtime` might be a better place for this. I know the other unsafe tests exist here but we should honestly move those too. Especially as we would like to eventually split the runtime and stdlib.

compiler/src/typed/typecore.re
```diff
@@ -391,6 +448,124 @@ let primn_type =
...
+        ("align", Builtin_types.type_wasmi32),
+        ("lane", Builtin_types.type_wasmi32),
+        ("vec", Builtin_types.type_wasmv128),
+      ],
+      grain_type_of_wasm_prim_type(ret_type),
+    )
+  | WasmSimdConstI8x16 =>
+    prim_type(
+      [
+        ("lane0", Builtin_types.type_wasmi32),
```
spotandjake, line 517: I wonder if it would have been worth using a helper like `List.init((index) => ("lane" ++ toString(index),Builtin_types.type_wasmi32), 15)` for all of these.

compiler/src/codegen/compcore.re
```diff
@@ -3434,6 +3516,12 @@ let compile_wasm_module =
     } else {
       default_features;
     };
+  let features =
+    if (Config.simd^) {
+      [Module.Feature.simd128, ...features];
```
spotandjake, line 3521: Binaryen's default behaviour is to error if a feature isn't enabled but the code is used right?

spotandjake (commented):
This looks awesome overall, I've just done an initial review so far, but want to go back over against the wasm api.

compiler/src/codegen/comp_utils.re
```diff
@@ -10,20 +10,30 @@ let grain_global_function_table = "tbl";
 // TODO(#): Use a more descriptive name once we get fixes into Binaryen
 let grain_memory = "0";
 
+let gensym_counter = ref(0);
```
spotandjake, line 13: Don't these already exist?

compiler/src/codegen/comp_wasm_prim.re
```diff
@@ -178,3 +559,372 @@ let compile_wasm_prim2 = (wasm_mod, env, instr, ret_type, arg1, arg2) => {
...
+      wasm_mod,
+      const_vec128(
+        Bytes.get_int32_le(bytes, 0),
+        Bytes.get_int32_le(bytes, 4),
+        Bytes.get_int32_le(bytes, 8),
+        Bytes.get_int32_le(bytes, 12),
+      ),
+    );
+  | None =>
+    // The wrapper function generated for this can never provide immediate
```
spotandjake, line 601: The problem here is we have no way to say something needs to be a proper constant for `align` and such right??

Similar to the issue we run into with `load` and `store`?

PR #2297, opened 2025-06-14
chore(docs): Clean up pre 0.6 examples

spotandjake:
Some of our stdlib examples were still using the pre 0.6 syntax this corrects them.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,10 +35,10 @@ It can sometimes be helpful to run small Grain programs directly to test some fu
 
 ### Standard library
 
-It's usually easiest to create a small Grain program that imports your library to try it out, like so:
+It's usually easiest to create a small Grain program that includes your library to try it out, like so:
 
 ```grain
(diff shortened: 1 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -47,7 +47,7 @@ print(appended)
 The tests for the standard library are located in `compiler/test/stdlib`. Since the standard library tests are written in Grain, rather than running the whole test suite, you can just run them directly:
 
 ```bash
-grain compiler/test/stdlib/array.test.gr
+grain compiler/test/stdlib/array.test.gr --stdlib stdlib
```
ospencer, line 50: Is the `--stdlib` necessary? That should be the default.

spotandjake: It makes a difference if you are developing with the prebuilt binaries / didn't build the stdlib from source in which case it would default to the prepackaged ones.

ospencer: Right. Good catch.

PR #2298, opened 2025-06-14
feat(lsp): Implement document symbol outline

spotandjake:
This adds the lsp symbol outline which makes it easier to navigate large files.

Screenshot 2025-06-14 at 6 20 23 PM
Screenshot 2025-06-14 at 6 23 05 PM

### DIFF
--- a/compiler/src/language_server/symbol.re
+++ b/compiler/src/language_server/symbol.re
@@ -1,0 +1,246 @@
+open Grain_typed;
+
+// https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentSymbolParams
+module RequestParams = {
+  [@deriving yojson({strict: false})]
+  type t = {
+    [@key "textDocument"]
+    text_document: Protocol.text_document_identifier,
+  };
+};
+
+// https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentSymbol
+module ResponseResult = {
+  /** A symbol kind. */
+  [@deriving (enum, yojson)]
+  type symbol_kind =
+    | [@value 1] File
+    | [@value 2] Module
+    | [@value 3] Namespace
+    | [@value 4] Package
+    | [@value 5] Class
+    | [@value 6] Method
+    | [@value 7] Property
+    | [@value 8] Field
+    | [@value 9] Constructor
+    | [@value 10] Enum
+    | [@value 11] Interface
+    | [@value 12] Function
+    | [@value 13] Variable
+    | [@value 14] Constant
+    | [@value 15] String
+    | [@value 16] Number
+    | [@value 17] Boolean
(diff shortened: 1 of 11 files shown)


compiler/src/language_server/symbol.re
```diff
@@ -0,0 +1,246 @@
...
+        },
+        data_decls,
+      ),
+    )
+  | TTopLet(_, _, binds) =>
+    Some(
+      List.filter_map(
+        (bind: Typedtree.value_binding) => {
+          switch (bind.vb_pat.pat_desc) {
+          | TPatVar(ident, _) =>
```
ospencer, line 151: Shouldn't we at least have `TPatConstraint`? I might argue that we'd want to process every `TPatVar` that appears in a toplevel binding.

ospencer (commented):
This looks really clean overall.

compiler/src/language_server/symbol.re
spotandjake: Doesn't `TPatConstraint` only appear under `pat_extra`?.

As per the others I would generally agree though I was thinking of leaving these for a small follow pr:
* cases like `let [x,y] = [() => void, () => void]` in order to determine that `x` and `y` should have the kinds of functions I would have to start inspecting and breaking down the `exp_type` which seems a little complicated. Alternatively we could just say destructured items are of kind `Variable`.
* I am not exactly sure what the range and selection ranges should be for destructured items.

ospencer: `TPatConstraint` is just when the user writes a type annotation. I don't think looking at the type is too bad, really. If it's an arrow then you know it's a function, and it can be a variable otherwise. As for range, for the big one I'd just do the whole binding, and for the small one I'd do just the name.

spotandjake: `TPatConstraint` is under `pat_extra` though which means we don't need to look at right? 
Screenshot 2025-06-14 at 8 34 40 PM


As per the other patterns what I really meant was two fold, patterns are recursive so I need to go down both the pattern and the type to figure out whats what i.e `let [x,y] = [() => void, () => void]` I need to start extracting the `a` from the `list<a>`, this isn't insanely difficult but it felt like a lot of complexity for the initial feature, this gets more complex when you consider you can have patterns like `let [(x, y), (a, b)]  = ...` and other nested patterns, I am happy todo it though, I also don't know how well those selection ranges and ranges would work given vscodes symbol headers
Screenshot 2025-06-14 at 8 42 45 PM


For reference reasons lsp doesn't handle the nested patterns
Screenshot 2025-06-14 at 8 43 33 PM

I'm happy to add it if we really want it though.

ospencer: I think it's fine to start here.

ospencer (approved):
Thanks for implementing this @spotandjake!

PR #2313, opened 2025-10-15
fix(stdlib): Correct include examples in `Bigint` and `PriorityQueue`

ghost:
Updated `include` name in the source file that creates the Docs for BigInt: from `Bigint` to `BigInt` to prevent compiler error if cut and pasted into a grain code file.

### DIFF
--- a/stdlib/priorityqueue.gr
+++ b/stdlib/priorityqueue.gr
@@ -3,7 +3,7 @@
  *
  * An immutable priority queue implementation is available in the `Immutable` submodule.
  *
- * @example from "priorityqueue" include Priorityqueue
+ * @example from "priorityqueue" include PriorityQueue
  *
  * @since v0.5.3
  */


stdlib/bigint.gr
```diff
@@ -13,7 +13,7 @@ module BigInt
 
 from "runtime/unsafe/wasmi32" include WasmI32
 from "runtime/unsafe/memory" include Memory
-from "runtime/bigint" include Bigint as BI
+from "runtime/bigint" include BigInt as BI
```
spotandjake, line 16: This is changing an actual include you need to change the example in the docs.  line 4

spotandjake (commented):
You'll need to run `npm run stdlib doc` to regen the docs.

stdlib/bigint.gr
ghost: ok,  why don't you decline/delete this pull request and I will resubmit and put each change in a separate branch too.

spotandjake: Your good to just do a second commit, we use squash commits and both of these changes are bassically the same so the scope is small enough they can be combined.

PR #2319, opened 2025-10-21
chore(stdlib): Organize runtime and stdlib tests

spotandjake:
Small pr moving some of our runtime tests into the proper test suite (This will better help when we split the runtime), it also organizes some of our stdlib tests with subfolders as the runtime and stdlib grow it seems helpful for us to organize them a little.

I also noticed while making this pr that we weren't running a couple test suites so I made sure to enable those.


ospencer (approved):
Great organization!

spotandjake:
I switched the order of our `assertions` for the runner results as well. I noticed we would fail do to the error number not matching first which meant you never got the error information from asserts from the result check. By doing the result check first we get more info on failures and how to fix them without running the test suites manually.

I also opened #2320 as I was running into issues when enabling the test suites related to artifacts and timeouts.

PR #2322, opened 2025-10-22
chore: Pin ci to `macos-14`

### LINKED ISSUE #2321: Grain fails to build on latest version of macos
Grain fails to build on the latest version of macos due to clang being upgraded by xtools to clang 17, which isn't compatible with our cmake fork.

A working branch is available here: https://github.com/spotandjake/grain/tree/spotandjake-clang-17 

We need to merge a bunch of downstream dependencies before we can merge these changes into main and fix this issue however.

spotandjake:
Attempting to pin our current ci to `macos-14` so we can merge changes, until we fix #2321

### DIFF
--- a/.github/workflows/build-js.yml
+++ b/.github/workflows/build-js.yml
@@ -13,7 +13,7 @@ on:
         default: "ubuntu-latest"
         options:
           - ubuntu-latest
-          - macos-latest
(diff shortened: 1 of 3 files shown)


ospencer:
@spotandjake So this isn't needed anymore, right?

spotandjake:
Nope

PR #2324, opened 2025-10-26
feat(cli): Use esm modules over commonjs

spotandjake:
Small pr switching from `require` to `import` in the cli, I figure we might as well switch to esm given we are on a version of node that supports it.

Draft as I would still like to test the js build, but it takes forever locally.


cli/bin/graindoc.js
```diff
@@ -1,3 +1,3 @@
 // A wrapper around grainc.bc.js that prepares the `pkg` env
-require("./pkg");
-require("./graindoc.bc.js");
+import ".pkg.js";
+import "./graindoc.bc.js";
```
ospencer: Looks like we lost a line ending.

cli/bin/grainlsp.js
```diff
@@ -1,3 +1,3 @@
 // A wrapper around grainlsp.bc.js that prepares the `pkg` env
-require("./pkg");
-require("./grainlsp.bc.js");
+import ".pkg.js";
+import "./grainlsp.bc.js";
```
ospencer: And one here.

spotandjake:
Closing as it doesn't look like we can make this change with pkg just yet sadly.

PR #2325, opened 2025-10-27
feat(stdlib): Standardize `path` module examples

spotandjake:
I noticed in #2282 that the examples in the path module didn't have the `Path` prefix, this pr adds it to the path functions to ensure our examples are runnable. I also noticed while making this change that some of our `@since` tags were missing so I added those in while I was here.

This does make the examples slightly more verbose but I think it's worth it to be able to just copy them into your editor and run them especially as we don't really encourage the use of `use Path.*`


ospencer (approved):
Looks great! This one is a `feat`.

PR #2326, opened 2025-10-28
chore(stdlib): Simplify path types in `path` module

spotandjake:
When the `path` module was originally implemented, we didn't have `abstract` types, as it was before the module system overhaul. This removes the unneeded records and simplifies them to regular abstract types.

I'm not sure if this is considered a `feat` or `chore` given the types are abstract, it's really just cleanup.


ospencer (approved):
Sweet. I think this is a `chore`.

PR #2333, opened 2025-11-12
feat(compiler): Upgrade to Binaryen v124

spotandjake:
This upgrades to binaryen.ml v124.

### DIFF
--- a/compiler/esy.json
+++ b/compiler/esy.json
@@ -25,7 +25,7 @@
     "copy:exe": "cp -f #{$GRAINC_BIN_PATH} #{$GRAINDOC_BIN_PATH} #{$GRAINFORMAT_BIN_PATH} #{$GRAINLSP_BIN_PATH} #{$CLI_BIN_DIR}",
     "copy:js": "cp -f #{$GRAINC_JS_PATH} #{$GRAINDOC_JS_PATH} #{$GRAINFORMAT_JS_PATH} #{$GRAINLSP_JS_PATH}  #{$CLI_BIN_DIR}",
     "clean": "rm -rf #{self.root}/_esy",
-    "build:js": "esy b dune build @js --no-buffer #{os == 'windows' ? '-j 2' : ''}",
+    "build:js": "esy b dune build @js --no-buffer -j 2",
     "test": "#{$TEST_EXEC_PATH}",
     "test:js": "node #{$TEST_JS_PATH}",
     "format": "dune build @fmt --auto-promote",


compiler/src/codegen/compcore.re
```diff
@@ -3478,6 +3482,13 @@ let compile_wasm_module =
       Option.is_none(Grain_utils.Config.memory_base^),
     );
 
+  // Our compiler generates code that should never trap, so enabling this allows
+  // Binaryen to perform more aggressive optimizations
+  let _ = Settings.set_traps_never_happen(true);
```
spotandjake, line 3487: Should we only enable this under our `Release` profile, just to help us catch any codegen errors?

compiler/src/codegen/compcore.re
```diff
@@ -3478,6 +3482,13 @@ let compile_wasm_module =
...
+  // Grain does not permit external edits to GC objects or function references
+  // not directly exposed via imports or exports
+  let _ = Binaryen.Settings.set_closed_world(true);
```
spotandjake: I'm not sure if we should enable this in this pr or when we merge the gc work, these optimizations bassically all relate to wasm-gc so It might just slow our optimization pipeline down slightly at the moment.

compiler/src/codegen/compcore.re
ospencer: I think it's fine to keep it on.

compiler/src/codegen/compcore.re
ospencer: Yeah, let's do that in the GC PR.

compiler/esy.json
```diff
@@ -25,14 +25,14 @@
     "copy:exe": "cp -f #{$GRAINC_BIN_PATH} #{$GRAINDOC_BIN_PATH} #{$GRAINFORMAT_BIN_PATH} #{$GRAINLSP_BIN_PATH} #{$CLI_BIN_DIR}",
     "copy:js": "cp -f #{$GRAINC_JS_PATH} #{$GRAINDOC_JS_PATH} #{$GRAINFORMAT_JS_PATH} #{$GRAINLSP_JS_PATH}  #{$CLI_BIN_DIR}",
     "clean": "rm -rf #{self.root}/_esy",
-    "build:js": "esy b dune build @js --no-buffer #{os == 'windows' ? '-j 2' : ''}",
+    "build:js": "esy b dune build @js --no-buffer -j 2",
```
spotandjake, line 28: I lowered the number of builders as we were running into out-of-memory issues when building on Ubuntu, likely because binaryen has grown.

I decided to lower on all platforms because I noticed that we were actually building faster with fewer workers, even on macOS where there was enough RAM available, likely do to less memory pressure, causing less swap usage. I noticed similar behaviour when building locally previously. I saw macos build times go from ~24minutes to ~15minutes


I went with 2 as it's what we were using on Windows previously; however, it may make sense to play with this number a little more in a separate pr for optimum performance.

Issue #2330, opened 2025-11-06
Replace bulkMemory polyfill with `llvm-memory-copy-fill-lowering` binaryen pass


spotandjake:
It seems a pass named `llvm-memory-copy-fill-lowering` was added to binaryen which polyfills the bulk memory operations this seems like a better solution then our internal polyfills. 

Blocked: This is blocked until we upgrade binaryen

PR #2334, opened 2025-11-12
feat(compiler): Use Binaryen's bulk memory polyfill

spotandjake:
This replaces our custom bulk memory polyfills to use the binaryen pass implementation.

Notes:
* We no longer test the polyfill behaviour as we are not trying to test binaryen and we don't have a good way of inspecting the wasm output.
* The feature flag must be left on, as the instructions get optimized away, and if it was off, adding the instructions would throw exceptions in binaryen.
* Bulk memory is now supported across the board, so this polyfill is far less important than it was initially. This reduces our maintenance complexity, shifting it over to binaryen.

Closes: #2330


compiler/src/typed/typecore.re
```diff
@@ -368,16 +368,16 @@ let primn_type =
   | WasmMemoryCopy =>
     prim_type(
       [
-        ("source", Builtin_types.type_wasmi32),
-        ("destination", Builtin_types.type_wasmi32),
+        ("dest", Builtin_types.type_wasmi32),
```
spotandjake, line 371: I noticed these signatures were slightly wrong previously, we never saw this behaviour because the names ended up coming from the polyfills rather than from the primitives however without correcting this mistake this pr would be breaking if you tried to use named arguments.

compiler/src/codegen/compcore.re
```diff
@@ -3456,23 +3456,15 @@ let compile_wasm_module =
     );
   let wasm_mod = Module.create();
 
-  let default_features = [
+  let features = [
+    Module.Feature.bulk_memory,
+    Module.Feature.bulk_memory_opt,
     Module.Feature.mvp,
```
ospencer: Feels weird to not have mvp at the top:
```suggestion
    Module.Feature.mvp,
    Module.Feature.bulk_memory,
    Module.Feature.bulk_memory_opt,
```

compiler/src/codegen/compcore.re
```diff
@@ -3538,6 +3530,10 @@ let compile_wasm_module =
 
   validate_module(~name?, wasm_mod);
 
+  if (! Config.bulk_memory^) {
```
ospencer, line 3536: Can we add a comment here? This doesn't seem too obvious and reads like it requires LLVM/does something special with LLVM.

spotandjake:
Made those changes

compiler/test/suites/flags.re
```diff
@@ -0,0 +1,85 @@
...
+    let wasm_mod = load_module(outfile);
+    Module.set_features(wasm_mod, Compcore.features);
+    expect.bool(
+      List.mem(Module.Feature.bulk_memory, Module.get_features(wasm_mod)),
+    ).
+      toBe(
+      true,
+    );
+    // Validate the module uses the bulk memory feature
+    expect.int(Module.validate(wasm_mod)).toBe(1);
```
ospencer: This is just checking that Binaryen can get and set features. I don't think we need this test.

compiler/test/suites/flags.re
```diff
@@ -0,0 +1,85 @@
...
+      ~config_fn=() => {Grain_utils.Config.bulk_memory := false},
+      name,
+      {|
+      module NoBulkMemoryCalls
+
+      from "runtime/unsafe/memory" include Memory
+      @disableGC
+      provide let foo = () => {
+        Memory.fill(0n, 0n, 0n)
+        Memory.copy(0n, 0n, 0n)
```
ospencer: Could we make these actually fill and copy values? A clever optimizer could remove these because they don't do anything (and thus this test wouldn't test what we want it to).

compiler/test/suites/flags.re
```diff
@@ -0,0 +1,85 @@
...
+        Compcore.features,
+      ),
+    );
+    expect.bool(
+      List.mem(Module.Feature.bulk_memory, Module.get_features(wasm_mod)),
+    ).
+      toBe(
+      false,
+    );
+    // Validate the module doesn't use the bulk memory feature
```
ospencer:
```suggestion
    // The module will fail validation if bulk memory opcodes are still present
```

compiler/test/suites/flags.re
spotandjake: Maybe I should be testing that this case fails when the bulk memory flags are removed to validate they exist in the first place?

ospencer: That's better, but I still don't think we need to test that here.

ospencer (approved):
Thanks, looks great!

PR #2350, opened 2026-01-24
fix(compiler)!: Correct constructors incorrectly being brought in scope

### LINKED ISSUE #1968: Compiler: module values incorrectly brought into scope
With the following example module:
```
module A

provide enum Variant {
  VariantA,
  VariantB
}
```

A consuming module will experience strange behavior with regards to accessing the definitions defined in the module
```
module B

include "./a"

// compiles as expected
print(A.VariantA)
// fails to compile as expected, since `VariantB` hasn't been directly extracted from the module
print(VariantB)

// compiles even though `VariantB` hasn't been brought into scope from the module
print(A.VariantA == VariantB)
```

At first glance it appears that the entire module is brought into scope for the expression hence why `A.VariantA == VariantB` compiles

spotandjake:
This pr corrects a bug where module values were incorrectly being brought into scope.

If we had a function such as `x: (a, a) => void` then in a case like `x(A.VariantA, VariantB)` the typechecker would lookup the second value through the of `a` when disambiguating causing constructors to be brought into scope incorrectly. This also fixes a very simialr issue with pattern matching.

The fix I chose is minimally invasive; we just skip the scope-based lookup when disambiguating in certain places. We still want the scope and type logic to exist for other cases like inference, which is why it's left alone. 

NOTE: I'm not sure if we consider this a breaking fix or not; some invalid code that previously would compile now won't. 

Closes: #1968

### DIFF
--- a/compiler/src/typed/disambiguation.re
+++ b/compiler/src/typed/disambiguation.re
@@ -136,6 +136,7 @@ module NameChoice =
         ~warn=Location.prerr_warning,
         ~check_lk=(_, _) => (),
         ~scope=?,
+        ~scoped=false,
         lid,
         env,
         opath,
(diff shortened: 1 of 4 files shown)


ospencer (commented):
We'll consider this a breaking fix.

ospencer (approved):

Issue #2273, opened 2025-04-12
Warn about `_start` override for user exports.


spotandjake:
Should we be doing a check here and warn and override if a user `_start` exists?

_Originally posted by @spotandjake in https://github.com/grain-lang/grain/pull/2234#discussion_r2040554569_

Currently if compile:
```gr
module Main

provide let _start = () => void
```
you will get an error propagated from binaryen that `_start` already exists, unless you compile with `--use-start-section` we should probably just wrap this with a nicer error. I think we want todo this directly in `write_universal_exports` or `compcore` as we only want to apply this on the emitted module. The only slightly annoying / tricky part is in order to use our regular error handler we would need to depend on `Location` from `Grain_Parsing`.

PR #2351, opened 2026-02-03
feat(compiler): Provide a user friendly error on `_start` override

spotandjake:
This pr provides a nice error in cases where a user overrides `_start` without compiling using `--use-start-section`. Previously binaryen would throw an error because we try to export `_start` twice once for the user's export and once for the actual module start, this error however wasn't very informative and didn't explain to the user how to fix the problem.

Note:
* I'm not 100% sure the language I used in this error is the best I tried to avoid mentioning how wasi uses `_start` which is the reason for the problem as I think that raises more questions than answers.
* Despite the error not fully describing why this is a problem, I think it's still a major improvement on what we had before as it tells them where the problematic function is and how to fix it.

Closes: #2273


ospencer (approved):
I like it!

Issue #1834, opened 2023-04-26
Find way to consistently test size of type metadata


ospencer:
See #1833 

We had a test that checked the default amount of type metadata, but as changes are made to the runtime and Pervasives, this number changes, which means the test would constantly need to be updated. We could maybe test a range of acceptable values, or craft an example that shouldn't change with respect to the runtime/Pervasives.

spotandjake:
Another approach is some sort of snapshot where we allow it to grow but only within a range of the old value given a range would still need to be updated if we pushed too far. but I think if it grows in one pr more then a couple of bytes we might have a problem.

PR #2352, opened 2026-02-03
feat(compiler): Test typemetadata size

spotandjake:
This pr tests the base typemetadata size in runtimeMode avoiding changes from modifications to the runtime or pervasives.

Note:
* I'm pretty sure this is a feat because we are adding a test though it may be a chore.

Closes: #1834


ospencer (commented):
I'd love to also see a test that creates a couple of data structures and then tests the size of the metadata. I don't think it would change between test runs, so it should be a pretty stable test.

Issue #2353, opened 2026-02-03
LSP: Support displaying multiple compiler errors


spotandjake:
In #2348 we added support for displaying multiple errors for the first time through the CLI. We should adjust the way we handle errors, either throughout the compiler or how we consume them in the lsp to support displaying multiple errors there as well.

Issue #2342, opened 2026-01-13
Compiler: Consider refactoring error handling, warnings and locations


spotandjake:
Currently our error handling, warning and location logic comes from ocaml's compiler, and is contained in `location.re` in the parser. 

Locations are generic and not specific to the parser we should consider moving these to `Grain_utils`, we should also consider splitting out the error and warning handling to their own modules allowing us to make error handling more consistent throughout the compiler.

PR #2354, opened 2026-02-05
feat(lsp): Support displaying multiple errors in the lsp

spotandjake:
This pr adds support for displaying multiple well formedness errors in the lsp, this is a follow up pr to #2348. The way I implemented this was to keep a list of all reported errors and then we can go through the list when one is raised. I think whenever we complete #2342 we may want to streamline the error tracking api a little to handle this better but this should work as a good solution for now.

Demo Screenshot:
Screenshot 2026-02-05 at 2 38 58 PM

Closes: #2353

### DIFF
--- a/compiler/src/parsing/location.re
+++ b/compiler/src/parsing/location.re
@@ -594,11 +594,11 @@ let () =
 
 external reraise: exn => 'a = "%reraise";
 
-let reported_exns = ref([]);
-let reset_exns = () => reported_exns := [];
+let reported_exceptions = ref([]);
+let reset_exceptions = () => reported_exceptions := [];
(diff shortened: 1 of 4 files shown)


compiler/src/parsing/location.re
```diff
@@ -593,7 +594,11 @@ let () =
 
 external reraise: exn => 'a = "%reraise";
 
+let reported_exns = ref([]);
+let reset_exns = () => reported_exns := [];
```
ospencer: Can we call these `reported_exceptions` and `reset_exceptions`?

ospencer (approved):

Issue #2311, opened 2025-09-11
Exhaustiveness failure on record variants combined with `when` patterns.


spotandjake:
It seems that our exhaustiveness check incorrectly reports the below pattern match as exhaustiveness, this seems to occur when record variants are combined with `when` patterns.

```gr
module Test

enum T {
  A{ n: Number, },
  B{ n: Number, },
}
match (B{ n: 1 }) {
  A{ n } => void,
  B{ n } when false => void,
}
```

PR #2355, opened 2026-02-05
fix(compiler): Correct exhaustive check on record adt with when pattern

spotandjake:
This pr corrects exhaustive checking when inlined records are mixed with when clauses, like in the example below:

```gr
enum T {
  A{ n: Number, },
  B{ n: Number, },
}
match (B{ n: 1 }) {
  A{ n } => void,
  B{ n: 0 } when false => void,
}
```

Previously, we were not providing a warning, which was due to a small bug in `Conv`, where we didn't propagate any `Any` pattern. This handles the edge case correctly, allowing a warning to show up. However, I think we may want to take a deeper look into pattern matching semantics eventually, as the `Conv` logic feels fragile in general. 

Closes: #2311

### DIFF
--- a/compiler/src/typed/parmatch.re
+++ b/compiler/src/typed/parmatch.re
@@ -357,8 +357,34 @@ module Compat =
     | (TPatOr(p1, p2), _) => compat(p1, q) || compat(p2, q)
     | (_, TPatOr(q1, q2)) => compat(p, q1) || compat(p, q2)
     /* Constructors, with special case for extension */
-    | (TPatConstruct(_, c1, ps1), TPatConstruct(_, c2, ps2)) =>
+    | (
+        TPatConstruct(
+          _,
+          c1,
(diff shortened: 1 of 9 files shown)


compiler/src/typed/parmatch.re
```diff
@@ -2153,6 +2153,11 @@ module Conv = {
         };
         Hashtbl.add(constrs, id, cstr);
         switch (lst) {
```
ospencer: Let's change this to match on both the values as well as `cstr_inlined`, explicitly matching on the `Some(_)` and `None` cases.

In the `Some` case, for the constructor values, if the list is empty, we should fail. If the list contains more than one item, we should fail. We then need to handle _all_ of the cases that are valid. Is it just `Any`? Feels like `Or` and `Alias` are valid here too.

spotandjake: Instead of just handling all the casees I refactored this a tiny bit so that it's more explicit what type a constructor is.

Issue #2301, opened 2025-06-22
lsp: Add graindoc code action.


spotandjake:
It would be useful if we had a code action in the lsp where if you hovered over something like:
```gr
let test = (x,y) => {
  print(x)
  return 1
}
```
it would offer to insert a template graindoc header like:
```gr
/**
 * 
 *
 * @param x: 
 * @param y:
 * @returns:
 *
 * @example
 *
 * @since
 */
let test = (x,y) => {
  print(x)
  return 1
}
```

We would probably only want to target provided values and module headers with this, and we may want to discuss what attributes should be in the template, the params and returns seem obvious, but the example and since might not make sense to include.

This should be fairly easy to implement the only thing to consider is we don't currently expose binds through sourcetree, so we would need to modify that slightly to support this, we would also need to mark if the binds are exported or not.

PR #2356, opened 2026-02-06
feat(lsp): Implement an add graindoc code action

spotandjake:
This pr implements an add graindoc code action that lets you add a graindoc template to `let expressions` and `modules`.

by default for a value or module we add:
```js
/**
 *
 *
 * @example
 *
 * @since
 */
```
and for a function we add:
```js
/**
 *
 *
 *
 * @param a:
 * @param b:
 * @returns 
 * @example
 *
 * @since
 */
provide let x = (a, b) => a + 1
```

This pr isn't quite ready to merge as it still needs tests but I wanted to get some feedback on this before writing those. 

Closes: #2301

### DIFF
--- a/compiler/test/suites/grainlsp.re
+++ b/compiler/test/suites/grainlsp.re
@@ -219,6 +219,25 @@ let abc = { x: 1 }
           ),
         ),
       ]),
+      `Assoc([
+        ("title", `String("Add Graindoc")),
+        ("kind", `String("add-graindoc")),
+        (
+          "edit",
+          lsp_text_document_edit(
+            "file:///a.gr",
+            [
+              (
+                lsp_range(
+                  (1, 4611686018427387871),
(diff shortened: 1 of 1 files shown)


compiler/test/suites/grainlsp.re
```diff
@@ -219,6 +219,25 @@ let abc = { x: 1 }
...
+        ("title", `String("Add Graindoc")),
+        ("kind", `String("add-graindoc")),
+        (
+          "edit",
+          lsp_text_document_edit(
+            "file:///a.gr",
+            [
+              (
+                lsp_range(
+                  (1, 4611686018427387871),
```
spotandjake, line 232: I wasn't quite sure if this is worthy of a code commnt, but this number should be `Int.max_int` however it seems that `yojson` treats these internally as `IEEE-754` and we are getting some rounding.

PR #2364, opened 2026-03-01
Spotandjake/gc stdlib

spotandjake:


compiler/test/stdlib/bytes.test.gr
spotandjake, line 1: I added a few more tests here, because I noticed we were never really validating the encoding itself just the operation. These tests validate things like endianness.

compiler/test/stdlib/buffer.test.gr
```diff
@@ -239,6 +239,10 @@ let bytes = Bytes.fromString(str)
 let buf = Buffer.make(0)
 Buffer.addBytes(bytes, buf)
 assert Buffer.toBytes(buf) == bytes
+let buffer = Buffer.make(0)
```
spotandjake, line 242: I ran into a bug elsewhere for this in the stdlib and figured it should have a dedicated test.

stdlib/runtime/atof/parse.gr
spotandjake, line 1: There still seems to be a bug in `atof` somewhere causing overreads and slightly invalid numbers which is causing the `Number` and `Json` modules to fail their test.

I was having trouble tracking down the source though. I noticed that there are some cases where we were a bit laxed on our boundary checks before here, which may be part of the problem.

stdlib/runtime/atof/parse.gr
```diff
@@ -185,27 +185,29 @@ let parseFloatToParts = (string: String) => {
...
-            let digit = char - _CHAR_CODE_0
-            if (digit >= 0n && digit < 10n) {
-              mantissa = addWasmI64(WasmI64.extendI32U(digit), mantissa * 10N)
-              i += 1n
-              numDigitsAfterDot += 1n
-            } else if (char == _CHAR_CODE_UNDERSCORE) {
-              i += 1n
-            } else {
-              break
+      if (i < len) {
```
spotandjake, line 195: This is one such case where we were being a bit laxed before,  the loop could go all the way to `len` and then when this tries to read it fails. I tried to correct the cases I could find, but I couldn't track them all down.

stdlib/runtime/unsafe/conv.gr
```diff
@@ -99,20 +97,19 @@ provide let fromFloat64 = (n: Float64) => {
...
-  // call to tagSimpleNumber and instead doing it inline. For the case of
-  // numbers not fitting into the range, the cost of these simple ALU
-  // instructions are negligible compared to heap allocations.
+  // WasmI32.leS by about 20%. For the case of numbers not fitting into the
+  // range, the cost of these simple ALU instructions are negligible compared
+  // to heap allocations.
 
   let simple = tagSimpleNumber(n)
 
-  if (untagSimpleNumber(simple) is n) {
```
spotandjake, line 115: The `is` check here was causing an invalid module error.

stdlib/runtime/unsafe/tags.gr
```diff
@@ -4,7 +4,6 @@ module Tags
 provide let _GRAIN_NUMBER_TAG_TYPE = 0b0001n
 provide let _GRAIN_SHORTVAL_TAG_TYPE = 0b0010n
 provide let _GRAIN_CONST_TAG_TYPE = 0b0110n
-provide let _GRAIN_GENERIC_HEAP_TAG_TYPE = 0b0000n
```
spotandjake, line 7: This is no longer a concept

stdlib/runtime/equal.gr
```diff
@@ -39,26 +38,26 @@ let _LIST_ID = WasmRef.fromGrain(builtinId("List"))
...
       }
 
-      let xarray = DataStructures.getCompoundValueArrayRef(xptr)
-      let yarray = DataStructures.getCompoundValueArrayRef(yptr)
+      let xarray = DataStructures.getCompoundValueArrayRef(xRef)
+      let yarray = DataStructures.getCompoundValueArrayRef(yRef)
 
       // Handle lists separately to avoid stack overflow
-      if (DataStructures.loadVariantTypeId(xptr) is _LIST_ID) {
-        if (xVariantTag >> 1n == 1n) return true // End of list
```
spotandjake, line 61: I think you introduced this bug in one of your prior commits, but we already untagged it. So this was causing equality issues.

stdlib/runtime/string.gr
```diff
@@ -63,6 +62,7 @@ primitive typeMetadata = "@heap.type_metadata"
 let findTypeMetadata = typeHash => {
   let typeMetadata = typeMetadata()
```
spotandjake, line 64: I think we may want to make `typeMetadata()` return `-1` or `0` if type metadata is elided. I need the `eqz` check on the number of buckets or we get a divide by zero error.

stdlib/map.gr
```diff
@@ -436,26 +435,17 @@ provide let fromList = list => {
...
-    // We store them directly to prevent GC on uninitialized array data.
-    let arrayPtr = WasmI32.fromGrain(array)
-    let item = (key, value)
-    WasmI32.store(
-      addWasmI32(arrayPtr, untagSimpleNumber(i) * 4n),
-      Memory.incRef(WasmI32.fromGrain(item)),
-      8n
-    )
-    ignore(array)
+    arr[i] = (key, value)
```
spotandjake, line 444: We may want to consider doing this in more of the places we are doing `Array.setAnyValue`

compiler/test/runtime/unsafe/wasmi32.test.gr
```diff
@@ -85,9 +85,6 @@ let test = () => {
   assert WasmI32.extendS16(0x8000n) == 0xffff8000n
 
   // Grain conversion tests
-  assert WasmI32.fromGrain(true) == 0xfffffffen
-  assert WasmI32.toGrain(0xfffffffen): Bool
-
```
ospencer, line 90: I think it'd be good to keep this test.

docs/contributor/data_representations.md
```diff
@@ -28,11 +28,11 @@ Since heap-allocated values (`pointers` in the chart above) are "tagged" with `0
...
 
-31 bits allow us to represent more than 2 billion numbers (from -1073741824 to 1073741823), which is a few more than the 500 million we'd get from 29 bits. If larger numbers (or non-integer numbers) are needed, then Grain will fall back onto one of the other (heap-allocated) number types. The heap-allocated numbers are described in detail later.
+30 bits allow us to represent more than 2 billion numbers (from -536,870,912 to 536,870,911), which is a few more than the 500 million we'd get from 29 bits. If larger numbers (or non-integer numbers) are needed, then Grain will fall back onto one of the other (heap-allocated) number types. The heap-allocated numbers are described in detail later.
```
ospencer:
```suggestion
30 bits allow us to represent more than 1 billion numbers (from -536,870,912 to 536,870,911), which is four times as many as the 250 million we'd get from 28 bits. If larger numbers (or non-integer numbers) are needed, then Grain will fall back onto one of the other (heap-allocated) number types. The heap-allocated numbers are described in detail later.
```

stdlib/bytes.gr
```diff
@@ -701,14 +688,15 @@ provide let getUint16 = (index: Number, bytes: Bytes) => {
...
+  let srcSize = getBytesSize(src)
+  let srcData = getBytesArrayRef(src)
   let offset = coerceNumberToWasmI32(index)
-  checkIndexIsInBounds(offset, _INT16_BYTE_SIZE, size)
+  checkIndexIsInBounds(offset, _INT16_BYTE_SIZE, srcSize)
   let v = untagUint16(value)
-  WasmI32.store16(ptr + offset, v, _BYTES_DATA_OFFSET)
-  ignore(bytes)
+  WasmArrayRef.setI8(srcData, offset + 0n, v & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 1n, v >> 8n & 0xffn)
```
ospencer: Does this error without the mask? I imagine it doesn't, and we could save a couple of instructions here (though I imagine an optimizer would catch this).

stdlib/bytes.gr
```diff
@@ -759,14 +751,17 @@ provide let getInt32 = (index: Number, bytes: Bytes) => {
...
   let offset = coerceNumberToWasmI32(index)
-  checkIndexIsInBounds(offset, _INT32_BYTE_SIZE, size)
+  checkIndexIsInBounds(offset, _INT32_BYTE_SIZE, srcSize)
   let v = Conv.fromInt32(value)
-  WasmI32.store(ptr + offset, v, _BYTES_DATA_OFFSET)
-  ignore(bytes)
+  WasmArrayRef.setI8(srcData, offset + 0n, v & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 1n, v >> 8n & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 2n, v >> 16n & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 3n, v >> 24n & 0xffn)
```
ospencer: Same question about the masks.

stdlib/bytes.gr
```diff
@@ -817,14 +816,17 @@ provide let getUint32 = (index: Number, bytes: Bytes) => {
...
   let offset = coerceNumberToWasmI32(index)
-  checkIndexIsInBounds(offset, _INT32_BYTE_SIZE, size)
+  checkIndexIsInBounds(offset, _INT32_BYTE_SIZE, srcSize)
   let v = Conv.fromUint32(value)
-  WasmI32.store(ptr + offset, v, _BYTES_DATA_OFFSET)
-  ignore(bytes)
+  WasmArrayRef.setI8(srcData, offset + 0n, v & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 1n, v >> 8n & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 2n, v >> 16n & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 3n, v >> 24n & 0xffn)
```
ospencer: Same question.

stdlib/bytes.gr
```diff
@@ -875,14 +881,17 @@ provide let getFloat32 = (index: Number, bytes: Bytes) => {
...
-  checkIndexIsInBounds(offset, _INT32_BYTE_SIZE, size)
-  let v = Conv.fromFloat32(value)
-  WasmF32.store(ptr + offset, v, _BYTES_DATA_OFFSET)
-  ignore(bytes)
+  checkIndexIsInBounds(offset, _FLOAT32_BYTE_SIZE, srcSize)
+  let v = WasmI32.reinterpretF32(Conv.fromFloat32(value))
+  WasmArrayRef.setI8(srcData, offset + 0n, v & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 1n, v >> 8n & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 2n, v >> 16n & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 3n, v >> 24n & 0xffn)
```
ospencer: Same question.

stdlib/bytes.gr
```diff
@@ -933,14 +953,24 @@ provide let getInt64 = (index: Number, bytes: Bytes) => {
...
+  let hi = WasmI32.wrapI64(v >> 32N)
+  use WasmI32.{ (+), (&), (>>) }
+  WasmArrayRef.setI8(srcData, offset + 0n, lo & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 1n, lo >> 8n & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 2n, lo >> 16n & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 3n, lo >> 24n & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 4n, hi & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 5n, hi >> 8n & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 6n, hi >> 16n & 0xffn)
+  WasmArrayRef.setI8(srcData, offset + 7n, hi >> 24n & 0xffn)
```
ospencer: Same question.

stdlib/bytes.gr
```diff
@@ -991,14 +1032,24 @@ provide let getUint64 = (index: Number, bytes: Bytes) => {
...
```
ospencer: Same question.

stdlib/bytes.gr
```diff
@@ -1050,11 +1112,22 @@ provide let getFloat64 = (index: Number, bytes: Bytes) => {
...
```
ospencer: Same question.

stdlib/int16.gr
```diff
@@ -144,11 +123,9 @@ provide let (+) = (x: Int16, y: Int16) => {
  */
 @unsafe
 provide let (-) = (x: Int16, y: Int16) => {
-  let x = WasmI32.fromGrain(x)
-  let y = WasmI32.fromGrain(y)
-  let val = x - y
-  let tagged = val | _TAG_BYTE
-  WasmI32.toGrain(signExtend(tagged)): Int16
+  let x = untagInt16(x)
+  let y = untagInt16(y)
```
ospencer, line 127: The full untagging is a little rough but it's more obvious.

stdlib/map.gr
ospencer: Maybe, but `setAny` avoids the bounds checks.

stdlib/runtime/string.gr
ospencer: That should be simple enough to do.

stdlib/runtime/unsafe/conv.gr
ospencer: It's because `is` only works on references now. (So it'll work for Grain values, but not unsafe values anymore.)

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,294 @@
+/**
+ * Utilities for working with WebAssembly reference types in grain.
```
ospencer:
```suggestion
 * Utilities for working with WebAssembly reference types in Grain.
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,294 @@
...
+ * Utilities for working with WebAssembly array reference types in grain.
+ */
+provide module WasmArrayRef {
+  /** Represents a reference to an unsafe WebAssembly array. */
+  abstract type WasmArrayRef = WasmRef
+
+  /**
+   * Performs an unchecked cast from a `WasmRef` to a `WasmArrayRef`.
+   * The caller must ensure that the provided reference is indeed a valid WebAssembly array reference,
+   * as using an invalid reference may lead to runtime errors.
```
ospencer: You say "may" everywhere in this file, but it definitely will. We should update that.

stdlib/wasi/process.gr
```diff
@@ -193,53 +207,57 @@ provide let argv = () => {
...
   }
-  let arr = allocateArray(envc)
+  let arrRef = allocateArray(envc, WasmRef.fromGrain(void))
+  let arrData = DataStructures.getCompoundValueArrayRef(arrRef)
 
   let envsLength = envc * 4n
-  for (let mut i = 0n; i < envsLength; i += 4n) {
-    let strPtr = WasmI32.load(envvPtr + i, 0n)
+  let mut i = 0n
+  for (let mut ptr = 0n; ptr < envsLength; ptr += 4n) {
```
ospencer, line 241: Feels weird not to just combine these; should be able to do `let mut ptr = 0n, i = 0n` and do the increment in here too.

stdlib/int16.gr
spotandjake: I imagine optimizers could possibly optimize away some of this stuff. 

I also think it's well worth it for how readable it is.

compiler/test/runtime/unsafe/wasmi32.test.gr
spotandjake: Do we want to just write a test suite for `wasmref` that tests all of the functions there?

docs/contributor/data_representations.md
spotandjake: I was more just making small changes here and going to come back for docs before we get the main gc pr through. Though I'll just rewrite the contributor docs in this pr.

stdlib/wasi/process.gr
spotandjake: How would I chain the expressions in the increment? I certainly can do `let mut ptr = 0n and i = 0n` I just wasn't aware that there was a way in grain to chain the expressions for the increment.

stdlib/map.gr
spotandjake: Thats true. I think it makes sense to leave this one at least, I feel like `Map.toArray` is unlikely to be the bottleneck in user code and it greatly simplifies this function.

stdlib/bytes.gr
spotandjake: It doesn't seem to, so I'll remove the mask.

docs/contributor/low_level_programming.md
```diff
@@ -4,16 +4,14 @@ Grain is overwhelmingly a high-level programming language, but it provides a few
 
 ## Unsafe Types
 
-Grain has four unsafe types, `WasmI32`, `WasmI64`, `WasmF32`, and `WasmF64`. Unlike all other Grain types which are always represented as an `i32`, these special types are represented by their corresponding WebAssembly types. This makes these types **incompatible** with regular Grain values, and they cannot be used with generic functions (like `print`) or appear on Grain's heap (like in a list, array, or tuple). While the typechecker will often prevent you from using them incorrectly, some uses may not be caught (#1153).
+Grain has four unsafe types, `WasmI32`, `WasmI64`, `WasmF32`, and `WasmF64`. Unlike all other Grain types which are always represented as an `wasmRef`, these special types are represented by their corresponding WebAssembly types. This makes these types **incompatible** with regular Grain values, and they cannot be used with generic functions (like `print`) or appear in Grain's heap values (like in a `list`, `array`, `tuple` or any other high level dataStructure). While the typechecker will often prevent you from using them incorrectly, some uses may not be caught (#1153).
```
ospencer:
```suggestion
Grain has four unsafe types, `WasmI32`, `WasmI64`, `WasmF32`, and `WasmF64`. Unlike all other Grain types which are always represented as a `WasmRef`, these special types are represented by their corresponding WebAssembly types. This makes these types **incompatible** with regular Grain values, and they cannot be used with generic functions (like `print`) or appear in Grain's heap values (like in a `list`, `array`, `tuple` or any other high level dataStructure). While the typechecker will often prevent you from using them incorrectly, some uses may not be caught (#1153).
```

docs/contributor/low_level_programming.md
```diff
@@ -4,16 +4,14 @@ Grain is overwhelmingly a high-level programming language, but it provides a few
...
+While the number of gotchas has been minimized compared to past versions of Grain, there are still a few things to keep in mind when working with unsafe types. As unsafe values are not a subtype of `WasmRef` they cannot be used in any context where a regular Grain value is expected. This means that, they cannot be stored in any high level data structures such as `list`, `array`, `tuple`, `records` or any other high level dataStructure, they cannot be passed to any generic functions (like `print`, `==`, `marshal` or `compare`) as they are incompatible with regular Grain values, and they cannot be captured by closures as closures are implemented as heap values and are also subject to the same limitations as other heap values. To work around these limitations, convert them to native Grain `Int32`/`Int64`/`Float32`/`Float64` values first using functions in `runtime/unsafe/conv`.
```
ospencer:
```suggestion
As unsafe values are not a subtype of `WasmRef` they cannot be used in any context where a regular Grain value is expected. This means that, they cannot be stored in any high level data structures such as `list`, `array`, `tuple`, `records` or any other high level dataStructure, they cannot be passed to any generic functions (like `print`, `==`, `marshal` or `compare`) as they are incompatible with regular Grain values, and they cannot be captured by closures as closures are implemented as heap values and are also subject to the same limitations as other heap values. To work around these limitations, convert them to native Grain `Int32`/`Int64`/`Float32`/`Float64` values first using functions in `runtime/unsafe/conv`.
```

stdlib/runtime/atof/decimal.gr
```diff
@@ -203,160 +202,47 @@ provide let round = d => {
...
-    WasmI32.store16(_TABLE, 0x949Fn, 116n)
-    WasmI32.store16(_TABLE, 0x94C8n, 118n)
-    WasmI32.store16(_TABLE, 0x9CF2n, 120n)
-    WasmI32.store16(_TABLE, 0x051Cn, 122n)
-    WasmI32.store16(_TABLE, 0x051Cn, 124n)
-    WasmI32.store16(_TABLE, 0x051Cn, 126n)
-    WasmI32.store16(_TABLE, 0x051Cn, 128n)
-  }
-  _TABLE
+provide let get_TABLE = (index: WasmI32) => {
```
ospencer, line 206: This implementation doesn't cache the initialization, which is fairly important here. It's fine to just move the reference out of the function; that's good enough. Same deal with the other tables.

stdlib/runtime/atof/slow.gr
```diff
@@ -55,42 +55,21 @@ use Decimal.{ parseDecimal, _DECIMAL_POINT_RANGE }
...
-    WasmI32.store8(_POWERS, 39n, 12n)
-    WasmI32.store8(_POWERS, 43n, 13n)
-    WasmI32.store8(_POWERS, 46n, 14n)
-    WasmI32.store8(_POWERS, 49n, 15n)
-    WasmI32.store8(_POWERS, 53n, 16n)
-    WasmI32.store8(_POWERS, 56n, 17n)
-    WasmI32.store8(_POWERS, 59n, 18n)
-  }
-  _POWERS
+let get_POWERS = (index: WasmI32) => {
```
ospencer, line 61: Same thing with a table.

stdlib/runtime/atof/table.gr
```diff
@@ -1,2017 +1,49 @@
...
-    WasmF64.store(_F64_POWERS10_FAST_PATH, 0.0W, 200n)
-    WasmF64.store(_F64_POWERS10_FAST_PATH, 0.0W, 208n)
-    WasmF64.store(_F64_POWERS10_FAST_PATH, 0.0W, 216n)
-    WasmF64.store(_F64_POWERS10_FAST_PATH, 0.0W, 224n)
-    WasmF64.store(_F64_POWERS10_FAST_PATH, 0.0W, 232n)
-    WasmF64.store(_F64_POWERS10_FAST_PATH, 0.0W, 240n)
-    WasmF64.store(_F64_POWERS10_FAST_PATH, 0.0W, 248n)
-  }
-  _F64_POWERS10_FAST_PATH
+provide let get_F64_POWERS10_FAST_PATH = (index: WasmI32) => {
```
ospencer, line 13: Same thing with a table.

stdlib/runtime/math/rempio2.gr
```diff
@@ -50,7 +53,24 @@ let pio2_right = (q0: WasmI64, q1: WasmI64) => { // see: jdh8/metallic/blob/mast
 }
 
 @unsafe
-let mut _PIO2_TABLE = -1n
+let get_PIO2_TABLE = (index: WasmI32) => {
```
ospencer, line 57: Same thing with a table.

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
+/**
+ * Utilities for working with WebAssembly reference types in grain.
```
ospencer:
```suggestion
 * Utilities for working with WebAssembly reference types in Grain.
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ *
+ * This modules provides unsafe functions and should be used with extreme caution,
```
ospencer:
```suggestion
 * This modules provides unsafe functions and should be used with extreme caution;
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ * incorrect usage will lead to runtime errors or undefined behavior.
+ */
 @runtimeMode
 module WasmRef
 
+from "runtime/unsafe/wasmi32" include WasmI32
+
+/**
+ * Casts a grain value to a WebAssembly reference type.
```
ospencer:
```suggestion
 * Casts a Grain value to a WebAssembly reference type.
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ *
+ * @param value: The value to cast
+ * @returns A `WasmRef` corresponding to the provided grain value
```
ospencer:
```suggestion
 * @returns A `WasmRef` corresponding to the provided Grain value
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ */
 provide primitive fromGrain = "@wasm.fromGrain"
+/**
+ * Casts a WebAssembly reference type to a grain value.
```
ospencer:
```suggestion
 * Casts a WebAssembly reference type to a Grain value.
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ *
+ * NOTE:
+ * This function does not tell the typechecker the exact type of the resulting grain value,
+ * it is the caller's responsibility to ensure that the provided `WasmRef` corresponds to a
+ * valid grain value, and the caller must ensure that the resulting grain value is used in a
```
ospencer:
```suggestion
 * valid Grain value, and the caller must ensure that the resulting Grain value is used in a
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ * type-safe manner, as using an invalid reference will lead to runtime errors or undefined behavior.
+ *
+ * @param value: The `WasmRef` to cast
+ * @returns A grain value corresponding to the provided `WasmRef`
+ */
 provide primitive toGrain = "@wasm.toGrain"
 
+/**
+ * Checks if the provided `WasmRef` represents a WasmI31 Reference.
```
ospencer:
```suggestion
 * Checks if the provided `WasmRef` is a `ref i31`.
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ *
+ * @param ref: The `WasmRef` to inspect
+ * @returns `true` if the reference is a WasmI31 Reference, `false` otherwise
+ */
 provide primitive isRefI31 = "@is_refi31"
+
+/**
+ * Checks if the provided `WasmRef` represents a grain heap value.
```
ospencer:
```suggestion
 * Checks if the provided `WasmRef` is a Grain heap value.
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ * Casts a WebAssembly reference type to a grain value.
+ *
+ * NOTE:
+ * This function does not tell the typechecker the exact type of the resulting grain value,
+ * it is the caller's responsibility to ensure that the provided `WasmRef` corresponds to a
+ * valid grain value, and the caller must ensure that the resulting grain value is used in a
+ * type-safe manner, as using an invalid reference will lead to runtime errors or undefined behavior.
+ *
+ * @param value: The `WasmRef` to cast
+ * @returns A grain value corresponding to the provided `WasmRef`
```
ospencer:
```suggestion
 * @returns A Grain value corresponding to the provided `WasmRef`
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ */
 provide primitive toGrain = "@wasm.toGrain"
 
+/**
+ * Checks if the provided `WasmRef` represents a WasmI31 Reference.
+ *
+ * @param ref: The `WasmRef` to inspect
+ * @returns `true` if the reference is a WasmI31 Reference, `false` otherwise
```
ospencer:
```suggestion
 * @returns `true` if the reference is a `ref i31`, `false` otherwise
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ */
 provide primitive isRefI31 = "@is_refi31"
+
+/**
+ * Checks if the provided `WasmRef` represents a grain heap value.
+ *
+ * @param ref: The `WasmRef` to inspect
+ * @returns `true` if the reference is a grain heap value, `false` otherwise
```
ospencer:
```suggestion
 * @returns `true` if the reference is a Grain heap value, `false` otherwise
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ */
 provide primitive isGrainHeapValue = "@is_grain_heap_value"
 
+/**
+ * Retrieves the signed 31-bit integer value from the provided `WasmRef` if it represents a WasmI31 Reference.
```
ospencer:
```suggestion
 * Retrieves the signed 31-bit integer value from the provided `WasmRef`.
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ *
+ * NOTE: This function will throw a runtime error if the provided reference is not a WasmI31 Reference.
```
ospencer:
```suggestion
 * NOTE: This function will trap if the provided reference is not a `ref i31`.
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ *
+ * @param ref: The `WasmRef` to inspect
+ * @returns The signed 31-bit integer value if the reference is a WasmI31 Reference
```
ospencer:
```suggestion
 * @returns The signed 31-bit integer value
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ */
 provide primitive i31GetS = "@i31.get_s"
+
+/**
+ * Retrieves the unsigned 31-bit integer value from the provided `WasmRef` if it represents a WasmI31 Reference.
```
ospencer:
```suggestion
 * Retrieves the unsigned 31-bit integer value from the provided `WasmRef`.
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ *
+ * NOTE: This function will throw a runtime error if the provided reference is not a WasmI31 Reference.
```
ospencer:
```suggestion
 * NOTE: This function will trap if the provided reference is not a `ref i31`.
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
+ *
+ * @param ref: The `WasmRef` to inspect
+ * @returns The unsigned 31-bit integer value if the reference is a WasmI31 Reference
```
ospencer:
```suggestion
 * @returns The unsigned 31-bit integer value
```

stdlib/runtime/unsafe/wasmref.gr
```diff
@@ -1,20 +1,331 @@
...
-provide primitive arrayLen = "@wasm.ref_array_len"
-provide primitive arrayI8GetS = "@wasm.ref_array_i8_get_s"
-provide primitive arrayI8GetU = "@wasm.ref_array_i8_get_u"
-provide primitive arrayI8Set = "@wasm.ref_array_i8_set"
-provide primitive arrayI64Get = "@wasm.ref_array_i64_get"
-provide primitive arrayI64Set = "@wasm.ref_array_i64_set"
-provide primitive arrayAnyGet = "@wasm.ref_array_any_get"
-provide primitive arrayAnySet = "@wasm.ref_array_any_set"
+/**
+ * Utilities for working with WebAssembly array reference types in grain.
```
ospencer:
```suggestion
 * Utilities for working with WebAssembly array reference types in Grain.
```

stdlib/wasi/process.gr
ospencer: Talked about this offline.

PR #2366, opened 2026-03-10
fix(stdlib): correct the behavior of float pow opperations

Kara-Zor-El:
Previously, `Numbers.powf(0.0W, 0.0W)` and various other cases of numbers being put to the power of `0.0` would result in the value `NaN` being produced when instead it should result in `1.0` being produced.
https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=8766229 (page 63)


ospencer (approved):
Thanks so much for tracking this down and submitting the fix, much appreciated!

PR #2368, opened 2026-03-16
fix(stdlib): Correct bugs in regex parsing

Kara-Zor-El:
If wanted can also add tests cases for these

Changes:
- fix(stdlib): Change posix word parsing to parse word instead of hex
- fix(stdlib): Make word validation only match A-Z, a-z, _

Previously due to the <=, any character with a code of >= 95 was treated as part of a word's characters. This broke `\b` and `\B` matching for backticks alongside {, |, }, ~. You can confirm this behavior by doing 
```
Regex.isMatch(Result.unwrap(Regex.make("\\bhello\\b")), "`hello`")
``` 
which will return false when they should be true (you can do 
```
console.log(/\bhello\b/.test("`hello`"))
``` 
in node to see intended behavior).
- fix(stdlib): Correct asciiCharToLower range
- fix(stdlib): correct findAll termination and advance

zero-width matches would cause some incorrect results:
ex.
```
let matches = Regex.findAll(Result.unwrap(Regex.make("a*")), "aab")
List.forEach(m => {
  let { group, _ } = m
  print(group(0))
}, matches)
```
would result in:
```
Some("aa")
Some("a")
Some("")
```
when it should be
```
Some("aa")
Some("")
Some("")
```
js version for validation:
```
console.log("aab".match(/a*/g))
// Result: [ 'aa', '', '' ]
```

### DIFF
--- a/stdlib/regex.gr
+++ b/stdlib/regex.gr
@@ -741,9 +741,13 @@ and parsePosixCharClass = (buf: RegExBuf) => {
                 Ok(
                   rangeAdd(
                     rangeAddSpan(
-                      rangeAddSpan([], Char.code('a'), Char.code('f')),
-                      Char.code('A'),
-                      Char.code('F')
+                      rangeAddSpan(
+                        rangeAddSpan([], Char.code('a'), Char.code('z')),
+                        Char.code('A'),
+                        Char.code('Z')
+                      ),
+                      Char.code('0'),
+                      Char.code('9')
                     ),
                     Char.code('_')
                   ),
--- a/stdlib/regex.gr
+++ b/stdlib/regex.gr
@@ -2946,7 +2946,6 @@ let makeReferenceMatcher = eq =>
 
 let referenceMatcher = makeReferenceMatcher(((a, b)) => a == b)
 
-
 let referenceMatcherCaseInsensitive = makeReferenceMatcher(
   ((a, b)) => Char.Ascii.toLowercase(a) == Char.Ascii.toLowercase(b)
 )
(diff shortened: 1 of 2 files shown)


stdlib/regex.gr
```diff
@@ -2943,7 +2947,7 @@ let makeReferenceMatcher = eq =>
 let referenceMatcher = makeReferenceMatcher(((a, b)) => a == b)
 
 let asciiCharToLower = c => {
```
spotandjake: We may actually just want to replace this entire section with `Char.Ascii.toLowercase`

stdlib/regex.gr
```diff
@@ -3624,7 +3628,7 @@ let rec fastDriveRegexMatchAll = (rx, string, startOffset, endOffset) => {
           ...fastDriveRegexMatchAll(
             rx,
             string,
-            startPos + startOffset + 1,
+            (if (endPos > startPos) endPos else startPos + 1) + startOffset,
```
spotandjake, line 3623: Under what cases is `endPos > startPos`, and maybe it would make more sense to sort it in `searchMatch`, so we don't need any similar logic elsewhere?

spotandjake (commented):
Thanks for finding this, looks good to me overall, just one question.

We should probably add the regression tests.

stdlib/regex.gr
Kara-Zor-El: fixed

stdlib/regex.gr
Kara-Zor-El: `endPos > startPos` whenever the match consumes a character. so like if `ab` matches `ab` startPos=0, endPos=2 but `endPos == startPos` when the match consumes no characters. So if `a*` is matching ``, at a position with no a's it would be the case as an example

Kara-Zor-El: also, in regards to the `searchMatch` I think it is only supposed to affect the next search position

ospencer:
I agree with Jake! Looks like it just needs regression tests and this is good to go.

ospencer (approved):
Thanks so much!

PR #2376, opened 2026-03-28
feat(runtime): Move `Memory.compare` from compiler to runtime

### LINKED ISSUE #2375: Runtime: Consider moving `Memory.compare` primitive into the runtime.
Currently we have a primitive `@wasm.memory_compare` in the compiler however, there is no matching wasm instruction, and as such we bassically do this manually by generating the instructions needed. It may make more sense to move this into the runtime in the `runtime/memory` module, which is more inline with how we tend to handle this sort of thing especially as we move away from linear memory and towards wasm gc. 

The advantages here:
* Less compiler complexity
* Easier to maintain

The dissadvantange of this change is there may be some slight overhead from the grain code vs raw wasm instructions but it should be negligable especially considering we don't use `Memory.compare` much in the runtime or stdlib anymore with the wasm gc change.

spotandjake:
This pr removes our `@wasm.memory_compare` primitive in favour of implementing it in the runtime.

The reason I am making this change, is as we move away from linear memory and towards wasm gc it isn't as important that this function is in the compiler. Removing it from compcore lowers our maintance burden and moving it to grain makes it easier to manage.  The bigger motivator is that our primitives tend to represent either low level things that aren't practical todo in the language itself or things with a direct wasm instruction and this fits neither of those, making the runtime a better home. Similar to how we handle all of the gc array comparisons once that pr lands.

I'm not sure if we actually want to go forward with this change or not so i'm really just putting this pr in the case we do.

Notes:
* Instead of returning `-1` or `1` we now return the magnitude of the difference, this fits the compare api however it is a slight change to the logic we had before.
* I added tests for all the functions in `memory.gr` I can move those to a seperate pr but I figured if i'm testing one I might as well test the things it relies on.  

Closes: #2375


ospencer:
Want to rebase this off of the GC branch?

spotandjake:
> Want to rebase this off of the GC branch?

Made that change

ospencer (approved):

PR #2378, opened 2026-04-01
feat!: Switch to Wasm GC

### LINKED ISSUE #703: Remove all offset magic constants in stdlib

### LINKED ISSUE #813: Verify reference management in `@throw` compilation
In `compcore.re`, we currently don't do any special reference management for `throw`. We should determine whether the current semantics are correct, as this will become actually relevant once we have exception handling.

### LINKED ISSUE #841: Compiler: Support memory allocation in the wasi polyfill

### LINKED ISSUE #989: Cyclic GC

### LINKED ISSUE #2095: Decrease initial memory pages to 1

### LINKED ISSUE #2123: Memory leak on unused function calls
In
```
let f = () => (1, 2)
f()
f()
f()
f()
```
the memory allocated by the calls to `f` is not reclaimed. This is not the case if `f` is inlined (which is probably why this was not discovered).

### LINKED ISSUE #2197: DecRefError related to closures for non heap based mutable values.
I ran into a wierd decRef error with the below code:
```
let test = () => {
  let mut dependencyExists = false
  (() => {
    print(dependencyExists)
  })()
  (() => {
    print(dependencyExists)
  })()
  return dependencyExists
}
print(test())
```
This also happens if dependencyExists is a char or really any stack type. However if it is a Number, or heap type this error does not occur. Some cases that do work are below:
```
let test = () => {
  let dependencyExists = box(void)
  (() => {
    print(dependencyExists)
  })()
  (() => {
    print(dependencyExists)
  })()
  return dependencyExists
}
print("Test1")
print(test())

let test = () => {
  let mut dependencyExists = 1
  (() => {
    print(dependencyExists)
  })()
  (() => {
    print(dependencyExists)
  })()
  return dependencyExists
}
print("Test2")
print(test())
```

### LINKED ISSUE #2335: Does the current version support compiling to Wasm-GC?
I tried compiling the following simple program, as you presented on WasmIO:
```
module WasmIO

provide let apply = (f, x) => f(x)
```
```
grain compile --wat apply1.gr
```
However, the compilation result is still a wasm file that doesn't contain wasm-gc instructions.

ospencer:
# Wasm GC

At last, this PR switches Grain from full use of linear memory to Wasm GC objects! Details about the layouts of these objects can be found in `docs/contributor/data_representations.md`.

## Highlights

* No more compiler-caused GC bugs. GC code is completely removed from the compiler.
* The smallest Grain program is now 329 bytes, down from 3091.
* Grain programs now only use 1 page of linear memory.
* Performance of Grain programs is either similar or greatly increased.

## Breaking changes

* Simple numbers are now 30 bits, down from 31. This is because stack values in Wasm GC are 31 bits, for the same reason our numbers were 31 bits; however we don't have control over that bit anymore and need to distinguish from other stack tags.
* `@disableGC` has been removed.
* Closures can no longer be marshaled.

Closes #2335 
Closes #2197 
Closes #2123 
Closes #989 
Closes #841 
Closes #813 
Closes #703 
Closes #2095
Closes #984 
Closes #1330 
Closes #2097


compiler/src/codegen/comp_utils.re
```diff
@@ -75,39 +115,128 @@ let rec compile_const = (c): Literal.t => {
...
+let build_func_type = (args, rets) => {
+  let builder = Type_builder.make(1);
+  Type_builder.set_signature_type(builder, 0, Type.create(args), rets);
+  switch (Type_builder.build_and_dispose(builder)) {
+  | Ok([ty]) => ty
+  | _ => assert(false)
+  };
+};
+
+let build_basic_func_type = arity => {
```
spotandjake, line 219: Think we could either move the `+ 1` for the closure arg to the callee or rename this to something like `build_basic_grain_func_type`.

compiler/src/codegen/comp_utils.re
```diff
@@ -4,42 +4,88 @@ open Grain_typed;
...
 
 let grain_main = "_gmain";
 let grain_start = "_start";
-let grain_env_name = "_genv";
-let grain_global_function_table = "tbl";
+
 // TODO(#): Use a more descriptive name once we get fixes into Binaryen
 let grain_memory = "0";
 
+let page_size = 65536;
```
spotandjake, line 62: NOTE: Once this is merged, we should open an issue to move this to `Grain_utils.Config` and have this read from that along with the page_size definition in `runtime/memory.gr`. The two reasons for this are, we shouldn't store the same information in two places and it gets us set up to easily support the custom page sizes proposal.

compiler/src/codegen/compcore.re
```diff
@@ -478,11 +339,21 @@ let compile_bind =
   };
 };
 
-let get_swap = (~ty as typ=Types.Unmanaged(WasmI32), wasm_mod, env, idx) =>
+let get_swap = (~ty as typ=Types.GrainValue, wasm_mod, env, idx) =>
```
spotandjake, line 342: NOTE: Once this is merged we should probably open an issue to refactor `get_swap`, `set_swap` and `tee_swap` to a single `swap_help` that takes an action as thats the only part that differs. 

Then we can just do:
```
let get_swap = (~ty as typ=Types.GrainValue(GrainAny), wasm_mod, env, idx) =>
  swap_help(BindGet, ~ty, wasm_mod, env, idx);
let set_swap = (~ty as typ=Types.GrainValue(GrainAny), wasm_mod, env, idx) =>
  swap_help(BindSet({value: value}), ~ty, wasm_mod, env, idx);
let tee_swap = (~ty as typ=Types.GrainValue(GrainAny), wasm_mod, env, idx) =>
  swap_help(BindTee({value: value}), ~ty, wasm_mod, env, idx);
```

compiler/src/codegen/compcore.re
```diff
@@ -870,45 +586,43 @@ let error_if_true = (wasm_mod, env, cond, err) =>
...
-    /* Note that we're assuming the type-checker has done its
-       job and this access is not out of bounds. */
-    call_incref(
+    Expression.Array.get(
       wasm_mod,
-      env,
-      load(~offset=4 * (idx_int + 2), wasm_mod, tup()),
+      Expression.Struct.get(
+        wasm_mod,
+        2,
```
spotandjake, line 598: NOTE: We may want to open an issue after this is merged to build a little api to allow referencing struct fields by name rather than index. 

This would make it easier to read through the compiler and help make the compiler more resilient to data layout changes.

compiler/src/codegen/compcore.re
```diff
@@ -2447,6 +2162,131 @@ let compile_primn = (wasm_mod, env: codegen_env, p, args): Expression.t => {
...
+      gensym_label("ref_array_fill"),
+      [
+        Expression.Array.fill(
+          wasm_mod,
+          Expression.Ref.cast(
+            wasm_mod,
+            compile_imm(wasm_mod, env, List.nth(args, 0)),
+            array_type,
+          ),
+          compile_imm(wasm_mod, env, List.nth(args, 1)), // offset
```
spotandjake, line 2222: ~I really like how we use comments to document what arguments we're passing it would have been nice if we did this throughout. Might be something to come back to in a future pr though.~~~

Actually thinking about this just a little bit more, I think the case I would be making is that we should expose labelled arguments in `Binaryen.ml`  and then use the labels in compcore to make it easier to tell what the arguments were working with are.

compiler/src/codegen/compcore.re
```diff
@@ -3502,32 +3249,64 @@ let compile_wasm_module =
...
+  let (initial_memory, maximum_memory) =
+    switch (Config.initial_memory_pages^, Config.maximum_memory_pages^) {
+    | (initial_memory, Some(maximum_memory)) => (
+        initial_memory,
+        maximum_memory,
+      )
+    | (initial_memory, None) => (initial_memory, Memory.unlimited)
+    };
+
+  let initial_memory = max(initial_memory, minimum_memory_pages^);
```
spotandjake, line 3270: NOTE: We need to open an issue to properly handle intial_memory especially in comparson to maximum_memory.

compiler/src/codegen/compcore.re
```diff
@@ -3390,67 +3164,52 @@ let compile_type_metadata = (wasm_mod, env, prog) => {
...
 
+let default_features = [
+  Module.Feature.mvp,
+  Module.Feature.multivalue,
+  Module.Feature.tail_call,
+  Module.Feature.sign_ext,
+  Module.Feature.mutable_globals,
+  Module.Feature.reference_types,
+  Module.Feature.gc,
+  Module.Feature.bulk_memory,
```
spotandjake, line 3200: I think we removed the part where we disabled the `bulk_memory` feature when bulk memory is not disabled. 

Not sure if we should be doing the same thing for `tail_call`?

compiler/src/codegen/data_representations.re
```diff
@@ -0,0 +1,336 @@
...
+    | Some(supertype) => Type_builder.set_sub_type(builder, 0, supertype)
+    | None => ()
+    };
+    if (open_) {
+      Type_builder.set_open(builder, 0);
+    };
+    switch (Type_builder.build_and_dispose(builder)) {
+    | Ok([ty]) =>
+      switch (name) {
+      | Some(name) when Config.profile^ != Some(Release) =>
```
spotandjake, line 36: When I added the names I figured we only wanted them when compiling in non release mode.

I just want to make sure this is acutally the case. 

I think the way we handle other names is we just include them no matter what and then when binaryen is optimizing if your emitting `wat` it keeps them if your emitting `wasm` it strips them.

compiler/src/codegen/data_representations.re
spotandjake: Just a fun little note I thought of while reviewing this. It probably wouldn't be overly complex for us to generate the `data_representations` contributor docs from this file in the future to ensure their in sync. 

Unrelated to this pr just found it interesting.

compiler/src/middle_end/analyze_purity.re
spotandjake: Unrelated to this pr but I was here anyway, shouldn't we also be checking that the condition is pure, not just the branches? (Also out of curiousty if this were to be a partial match that could trap and all the branches were pure shouldn't we still count this as impure because of the default trapping behaviour?)

spotandjake: This probably shouldn't just be false, we should probably do a lookup if we know the function. I think we need to take into consideration that `return` is only an impure statement, though it doesn't make the function impure, similar to what I mentioned about the for loop.

compiler/esy.json
```diff
@@ -32,11 +32,11 @@
     "check-format": "dune build @fmt"
   },
   "dependencies": {
-    "@grain/binaryen.ml": ">= 0.34.0 < 0.35.0",
+    "@grain/binaryen.ml": "*",
```
spotandjake, line 35: NOTE: Once binaryen.ml is out with binaryen v126 we need to change this line

compiler/esy.json
```diff
@@ -67,10 +67,10 @@
     "@opam/file-context-printer": "reasonml/reason-native:file-context-printer.opam#20b1997b6451d9715dfdbeec86a9d274c7430ed8",
     "@opam/pastel": "reasonml/reason-native:pastel.opam#20b1997b6451d9715dfdbeec86a9d274c7430ed8",
     "@opam/rely": "reasonml/reason-native:rely.opam#20b1997b6451d9715dfdbeec86a9d274c7430ed8",
-    "@opam/ocamlfind": "1.9.8",
-    "@opam/num": "1.5"
+    "@opam/num": "1.5",
+    "@grain/binaryen.ml": "grain-lang/binaryen.ml:package.json#451cb0df06f7254fffc5eec02ab31ec495d83340"
```
spotandjake, line 71: NOTE: Once binaryen.ml is out with binaryen v126 we need to change this line

stdlib/runtime/atof/table.gr
```diff
@@ -1,2017 +1,55 @@
...
+    )
   }
-  _F64_POWERS10_FAST_PATH
+  use WasmI32.{ (+), (<<), (|) }
+  let data = DataStructures.getBytesArrayRef(tableRef)
+  let lo = WasmArrayRef.getI8U(data, index + 3n) << 24n
```
spotandjake, line 23: NOTE: I think once this pr is merged, we should open an issue related to adding API's for getting larger integers from the `int8` array, I think the api needs workshopping though.

spotandjake (commented):
Final review done, I didn't find anything critical, I just had a few nits and wanted to keep track of a few issues we should open after this is merged.

I didn't put an approval as we still have the `binaryen.ml` resolution that needs to be switched.

PR #2381, opened 2026-04-02
Remove unused MArityOp and MTagOp from mashtree

deepnav4:
Removed unused `MArityOp` and `MTagOp` from the mashtree.

* Deleted corresponding variants from `instr_desc`
* Removed related unused types
* Cleaned up references in garbage collection logic

These operations were not implemented (NYI) and appear to be unused.


phated:
Thanks for submitting this but we're not accepting the contribution at this time.

PR #2382, opened 2026-04-03
feat(compiler): Stub panic with unreachable to prevent bad codegen

### LINKED ISSUE #2379: Compiler: Have `call_panic_handler` produce `unreachable` in `@runtimeMode`
Currently, we have `call_panic_handler` which calls `runtime/unsafe/panic` however, when working in `@runtimeMode`, we don't have `panic` available to us. This isn't a problem if you write your code in a low-level manner however, if you try to use pattern matching or array indexing, we generate bad WASM code. A simple solution to this would be to emit an unreachable instruction in its place.

spotandjake:
This pr stubs panic with unreachable when we are compiling in `@runtimeMode`.

Because we use `@runtimeMode` itself to compile the panic module, we can't allow a circular dependency, so we don't import panic when compiling in runtime mode.

This had the side effect that if someone tried to use a higher-level expression such as `arr[x]`, we would generate invalid WASM. In order to avoid this, I figure we can just fail with `unreachable` instead of properly panicking. The downside is we don't get a nice error, but I figure that is a perfectly reasonable tradeoff in the context of the runtime. 

Closes: #2379

### DIFF
--- a/compiler/src/codegen/compcore.re
+++ b/compiler/src/codegen/compcore.re
@@ -143,20 +143,33 @@ let get_metadata_ptr = wasm_mod =>
 let get_grain_imported_name = (mod_, name) => Ident.unique_name(name);
 
 let call_panic_handler = (wasm_mod, env, args) => {
-  let args = [
-    Expression.Global_get.make(
+  switch (env.compilation_mode) {
+  | Runtime =>
+    Expression.Block.make(
       wasm_mod,
-      resolve_global(~env, panic_with_exception_name),
+      gensym_label("runtime_panic_stub"),
+      List.fold_left(
+        (acc, arg) => [Expression.Drop.make(wasm_mod, arg), ...acc],
+        [Expression.Unreachable.make(wasm_mod)],
+        args,
+      ),
+    )
+  | _ =>
+    let args = [
+      Expression.Global_get.make(
+        wasm_mod,
+        resolve_global(~env, panic_with_exception_name),
+        Type.int32,
+      ),
+      ...args,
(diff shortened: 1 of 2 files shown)


compiler/src/codegen/compcore.re
```diff
@@ -143,20 +143,33 @@ let get_metadata_ptr = wasm_mod =>
...
-  let args = [
-    Expression.Global_get.make(
+  switch (env.compilation_mode) {
+  | Runtime =>
+    Expression.Block.make(
       wasm_mod,
-      resolve_global(~env, panic_with_exception_name),
+      gensym_label("runtime_panic_stub"),
+      List.fold_left(
+        (acc, arg) => [Expression.Drop.make(wasm_mod, arg), ...acc],
```
ospencer: This is a little weird to me. This is only the exception being passed to the panic handler, so there's no sense in keeping it around. We should be able to just emit `unreachable` and have it be a little cleaner.

spotandjake: That's fair. I made the change. 

The reason I chose to drop the arguments initially was suppose that any of the arguments had side effects like `local.tee`, we might be setting a value and using it later, in which case not emitting the wasm would cause issues. Reviewing the implementation of `compile_imm`, this won't happen right now.

ospencer (approved):
The very next instruction is `unreachable`, so side effects like a local tee wouldn't matter. The thing to look out here for would be if we didn't do SSA, since an argument could be something like a function call. But not something for us to worry about!

PR #2384, opened 2026-04-13
feat(compiler)!: Out of source builds

### LINKED ISSUE #468: Out of source builds
Currently, Grain builds programs in-source. This makes things cumbersome for cleaning up artifacts, running the compiler in multiple parallel processes, etc. There are probably even more benefits with this strategy that I'm not thinking of, which I say because it seems most languages (Rust, OCaml) use this strategy.

ospencer:
This PR moves all produced build artifacts to a target directory. By default, it's called `target`.

The folder structure is as follows:
```
target/
└── [profile]/
    ├── build/
    │   └── <mirror of source dir>
    ├── deps/
    │   ├── stdlib
    │   └── [library]*
    └── artifact.wasm
```

The `build` directory contains all artifacts for the current project and is anchored by a project root, which can optionally be configured via `--project-root`. The `deps` folder contains all build artifacts for external libraries. Besides the stdlib, additional libraries can be included with `-L`. This differs from `-I`, which is now only for including directories of object files (relevant for build systems). By default, the wasm artifact is now placed at `target/[profile]/artifact.wasm`.

Compilation behavior remains the same, but there is now much better support for build systems that want to compile libraries separately and include them later with `-I`.

Typed filepaths from the `Fp` library were used as much as possible through the refactor of `module_resolution`.

Closes #468


cli/bin/grain.js
```diff
@@ -180,7 +204,7 @@ program
   .action(function (file, options, program) {
     const success = exec.grainc(file, options, program);
     if (success) {
-      const outFile = options.o ?? file.replace(/\.gr$/, ".wasm");
+      const outFile = options.o ?? defaultWasmLocation(file, options);
```
spotandjake, line 204: How would we feel about setting `options.o` by default in the cli to the root directory so the wasm file is still copied out by our cli.

I think this provides a nicer user experience with the js api having the wasm file just be there.

cli/bin/grain.js
spotandjake, line 98:
```suggestion
      "add additional included object file directories",
```

Is it useful for us to actually expose this through our cli, should we just remove the flag here and if someone needs that they can invoke `grainc` directly?

My main worry is that people are going to get confused about what this is / does

compiler/grainc/grainc.re
```diff
@@ -90,21 +85,20 @@ let grainc = (single_file_mode, name, outfile) => {
...
       let dependencies = Module_resolution.get_dependencies();
 
       error_wrapped(() => Link.link(~main_object, ~outfile, dependencies));
     };
   };
-
-  `Ok();
 };
 
-/** Converter which checks that the given output filename is valid */
```
spotandjake, line 103: Why did we remove this comment? The function still looks todo the same thing?

compiler/src/typed/env.re
```diff
@@ -732,6 +732,7 @@ type pers_struct = {
   ps_crcs: list((string, Digest.t)),
   ps_crc: Digest.t,
   ps_filename: string,
+  ps_base_dir: string,
```
spotandjake, line 735: Related to #216 should this be `Fp.t` and then we push the `toString` conversion to the where it's being consumed.

All i'm thinking here is that if we start making things `FP.T` at their definitions we are more likely to prefer working with file paths as we write more code.

compiler/src/typed/module_resolution.re
```diff
@@ -3,15 +3,62 @@ open Grain_utils;
...
+  | No_module_file(Location.t, string, option(string))
+  | Source_outside_roots(Fp.t(Fp.absolute));
 
 exception Error(error);
 
 let error = err => raise(Error(err));
 
+type origin =
+  | Project
+  | Stdlib
```
spotandjake, line 15: Is there a reason we treat the `stdlib` separately here?  Why not consider it a `Project`?

compiler/src/typed/module_resolution.re
```diff
@@ -79,44 +169,19 @@ module PathTbl = {
   type t('a) = Hashtbl.t(string, 'a);
   let create: int => t('a) = Hashtbl.create;
 
-  let add: (t('a), (string, string), 'a) => unit =
-    (tbl, (dir, unit_name), v) => {
-      let dir = Filepath.String.realpath_quick(dir);
-      Hashtbl.add(tbl, Filepath.String.smart_cat(dir, unit_name), v);
-    };
+  let key = (dir: Fp.t(Fp.absolute), unit_name) =>
```
spotandjake, line 172: Would it be a good idea to replace the `/` or `\` after doing `Fp.toString` with a separator like `:` or something that isn't path related so keys are the same across platforms?

compiler/src/typed/module_resolution.re
```diff
@@ -126,169 +191,238 @@ let current_located_module_cache = () => {
...
+      | Some(result) => result
+      | None => raise(Not_found)
       }
+    };
+  };
+
+let try_locate_module = (base_dir: Fp.t(Fp.absolute), name: string, loc) =>
+  try(locate_object(base_dir, name)) {
+  | Not_found =>
+    if (Filename.check_suffix(name, ".gr")) {
```
spotandjake, line 328: Think we should have constants somewhere for  the filepaths`"gr"` and `"gro"`?

compiler/src/typed/module_resolution.re
```diff
@@ -511,24 +657,25 @@ let () = {
...
   ));
-  Fs_access.register_cache_flusher((
-    Hashtbl.remove(resolutions),
-    () => Hashtbl.clear(resolutions),
-  ));
 };
 
 let dump_dependency_graph = Dependency_graph.dump;
 
-/* Error report */
```
spotandjake, line 522: Why did we drop this comment?

compiler/src/utils/config.re
```diff
@@ -363,13 +365,43 @@ let include_dirs =
...
+  opt(
+    ~names=["target-dir"],
+    ~conv=Filepath.Args.MaybeExistingDirectory.cmdliner_converter,
+    ~doc="Directory where build artifacts are written.",
+    ~docv="DIR",
+    ~digestible=NotDigestible,
+    Fp.At.(Filepath.get_cwd() / "target"),
+  );
+
+let libraries =
```
spotandjake, line 395: Should we move this up with `-I`?

compiler/src/utils/filepath.re
```diff
@@ -101,6 +124,16 @@ module String = {
   // TODO(#216): We should consider switching to type safe Fp.t where ever filepaths are used
   let chop_suffix = Filename.chop_suffix;
 
+  let first_segment = path =>
+    switch (String.index_opt(path, '/')) {
```
spotandjake, line 128: shouldn't we be checking for both `/` and `\` or normalizing first?

compiler/test/runner.re
```diff
@@ -19,16 +19,27 @@ let customMatchers = createMatcher => {
...
+    Filepath.String.concat("input", name ++ ".gr"),
+  );
 let stdlibfile = name =>
   Filepath.to_string(Fp.At.(test_stdlib_dir / (name ++ ".gr")));
 let runtimefile = name =>
   Filepath.to_string(Fp.At.(test_runtime_dir / (name ++ ".gr")));
 let objectfile = name =>
-  Filepath.to_string(Fp.At.(test_input_dir / (name ++ ".gro")));
+  Filepath.to_string(
+    Fp.At.(test_target_dir / "debug" / "build" / "input" / (name ++ ".gro")),
```
spotandjake, line 33: Having this default to `debug` might become confusing in the future.

compiler/test/runner.re
```diff
@@ -19,16 +19,27 @@ let customMatchers = createMatcher => {
...
+  );
 let wasmfile = name =>
   Filepath.to_string(Fp.At.(test_output_dir / (name ++ ".wasm")));
 let mashfile = name =>
-  Filepath.to_string(Fp.At.(test_input_dir / (name ++ ".mashtree")));
+  Filepath.to_string(
+    Fp.At.(
+      test_target_dir / "debug" / "build" / "input" / (name ++ ".mashtree")
```
spotandjake, line 40: same as above.

compiler/test/runner.re
```diff
@@ -283,15 +308,34 @@ let format = file => {
...
 let doc = (file, arguments) => {
-  let cmd = Array.concat([[|"grain", "doc", file|], arguments]);
+  let cmd =
+    Array.concat([
+      [|
+        "grain",
+        "doc",
+        "--target-dir",
+        Filepath.to_string(test_target_dir),
+        "-L",
```
spotandjake: Why does this need the `test-libs` stuff included?

compiler/test/runner.re
```diff
@@ -283,15 +308,34 @@ let format = file => {
...
 
 let lsp = stdin_input => {
-  let cmd = [|"grain", "lsp"|];
+  let cmd = [|
+    "grain",
+    "lsp",
+    "--target-dir",
+    Filepath.to_string(test_target_dir),
+    "-L",
+    Format.sprintf("test-libs=%s", Filepath.to_string(test_libs_dir)),
```
spotandjake: Same question as above?

spotandjake (commented):
This looks great to me, I did an initial review and has a few thoughts, I'll get you a more in depth one tomorrow or later this week.

cli/bin/grain.js
ospencer: I think it's better to not have the file be in the source directory by default. `grain file.gr` will still run the file for you, and it's not hard to find. Plus, if you want it somewhere else you can always pass `-o`.

cli/bin/grain.js
ospencer: I could see an argument for that.

compiler/src/typed/env.re
ospencer: I don't think so; that starts getting into this being a much bigger refactor. I already did a bunch of work towards #216 in this PR though.

compiler/src/typed/module_resolution.re
ospencer: `Project` is not _a_ project, it's _this_ project. (A better argument would be why the stdlib isn't just a `Library`, but that's because the module resolution happens differently because we don't prefix all the stdlib imports with `stdlib`.)

compiler/src/typed/module_resolution.re
ospencer: I don't think that really matters much, does it?

compiler/test/runner.re
ospencer: You mean if we were ever to change the tests to run in release mode?

spotandjake: > You mean if we were ever to change the tests to run in release mode?

Or if we were to ever add a new test that runs in release mode.

ospencer: If we were ever to do that we'd get a pretty obvious error.

cli/bin/grain.js
spotandjake, line 1: Does this pr allow us to drop the `pkg.js` file in the cli? 

It only exists to map the build outputs to a target directory but we are doing that inside the compiler now?


(Also just remembered that we are going to need to update the playground to handle the out of source stuff before we release this version if we merge this).

Issue #2087, opened 2024-04-05
Rework: `toString` to use a buffer


spotandjake:
Currently `toString` in the `runtime/string` module rely's on `concat` for building the output string, this often incurs a performance penality from the many allocations done especially with more complex data structures like nested records. We implemented the bytes literal and a `stdlib/buffer` module into the language since `toString` was written. It would make sense to implement a basic `buffer` implementation in the runtime so we can make use of buffers throughout the `stdlib` and `runtime`.

Issue #2127, opened 2024-07-28
Improve `String.map` and `String.mapi`


ospencer:
We currently implement these via `String.explode` and `String.implode`, but should instead iterate and build a new string. We probably don't want to make `String` depend on `Buffer`, but we could guess the next string size and realloc/copy if necessary.

spotandjake:
It may make sense to implement an internal simplified version of the buffer library for this and #2087

spotandjake:
> We currently implement these via `String.explode` and `String.implode`, but should instead iterate and build a new string. We probably don't want to make `String` depend on `Buffer`, but we could guess the next string size and realloc/copy if necessary.

Instead of using a buffer for this would we be able to allocate a string of size `4*charCount` which is the max bytesize the output string could be and, do the mapping and then shrink the strings bytelength to the real size without doing a realloc? This wouldn't hurt memory allocation in any way right but it would mean strings use more memory then they need beause we cant free half a chunk right?

Issue #1794, opened 2023-04-01
Stdlib: Add `toString` functions


spotandjake:
Currently grain has a global `toString` function that works on all datatypes in pervasvies. We should also add `toString` functions to libraries that work on specific datatypes such as `Char.toString` a few advantages of this is, these functions will be smaller and faster, and more type safe.

Future Question: after this issue is completed a future question is if we should remove the pervasvies `toString` to enforce better typesafety.

Issue #2374, opened 2026-03-28
Stdlib: Dynamic Array library


spotandjake:
Grain currently provides two primary iterable types: `List` and `Array`. These cover many use cases, but there’s a clear gap between them:
* `Array` offers constant-time indexing but has a fixed size.
* `List` supports growth but sacrifices indexing performance.

What’s missing is a structure that provides both dynamic growth and constant-time lookup—a dynamic array (commonly called `Vector `or `ArrayList`).

While many ML-family languages avoid including this abstraction, that tradeoff makes less sense in Grain. Since Grain leans more into imperative patterns, the lack of a dynamic array forces users into suboptimal choices that can carry real performance costs.

**Prior Art:**
OCaml [Dynarray](https://ocaml.org/manual/5.2/api/Dynarray.html) module provides a similar abstraction:
```
record Vector<a> {
  mut data: Array<Option<a>>,
  mut size: Number
}
```

This approach works but introduces overhead due to repeated Option wrapping/unwrapping, ocaml has measured this to be around 25% in the worst case.

Further reading can be found on the [wiki](https://en.wikipedia.org/wiki/Dynamic_array)

**Proposed Direction:**
We already have a need for a dynamic array in the runtime, in fact, we implement one in `runtime/string.gr` for tracking cycles. My proposal is we implement a rather unsafe low-level vector that stores elements as raw `wasmRefs` we can use void as the default value, given you can never read past `size`. The runtime implementation would be unsafe and avoid any unnecessary bounds checks. The stdlib module would wrap the functions with bounds checking properly to provide a nice user interface. This avoids the extra boxing entirely. This implementation is similar to what I have built [here](https://github.com/spotandjake/Grain-toString/blob/5cc0bcb720dcc867ec07da2c9172d27119f6d826/libs/vector.gr) for the runtime side anyway.

**Why This Matters:**
Without a dynamic array:
* Users either overuse the Array and manually manage resizing
  * They implement it themselves, which results in duplicate non-standard code.
* Or fall back to List and pay for poor indexing performance
Neither is great, especially in performance-sensitive or imperative-heavy code.

**Notes:**
One thing to consider here is that there are already questions as to when to use an array vs a list, and adding another array type probably doesn't help. I think we're still in the clear, as this won't be a first-class citizen, so we don't have direct syntax for initializing an array, making it more likely people will only use this when they actually need it.

**Questions:**
[...]

PR #2385, opened 2026-04-29
feat(stdlib): Rewrite `toString` to use a layout tree

spotandjake:
This pr refactors `toString` to improve the formatting of printed items. Previously, we were naively concatenating strings while stringifying. While this approach was relatively performant and very compact, it meant we weren't able to support high-quality formatting akin to how we format grain code itself. In order to solve this problem, I ported the `doc printing engine` used in the formatter into Grain and rewrote our implementation of it. 
Screenshot 2026-04-29 at 5 09 02 PM

More information on the implementation itself and design decisions can be found here: https://github.com/spotandjake/Grain-toString/tree/spotandjake/gc, along with some more in-depth notes on both the size and performance changes.

A few more notable things:
* When testing as a library, I was only seeing about a `40%` increase in the size of Hello World.
  * Our smallest grain module increased in size by about 100 bytes, looking into the wasm itself this seems to be related to the builders in `doc.gr`, these are capable of being optimized away however I don't think our current passes handle them very well.
* Performance for small prints has stayed the same; however, it seems that when printing very long lists with around 750k items, performance degrades a little bit compared to before.
  * I tested this in wasmtime, and most of the time seemed to have been spent in gc. I think this is much less of an issue in v8 and will improve in wasmtime in the future. 
  * Most of the time, users are going to be printing small items or singular large dumps when debugging. The performance I was seeing seemed very acceptable under these workloads.
* This refactor adds a few very useful libraries to the runtime, `doc`, `miniBuffer` and `vector`, all of which are generic enough to be used elsewhere. 
  * The vector library can be used in `marshal` getting rid of the second implementation there, it could also back a higher level vector library fairly easily.
  * The `MiniBuffer` library would be rather useful through the stdlib, notably things like `List.join`
* While I did optimize the `doc` library for grain, I didn't change the api from the formatter; this means that we can use it in other places, such as improving JSON stringification in the future. 
* Each datatype exposes a separate printer, which will allow us to provide more specific implementations or, in the future, maybe have the compiler monomorphize things like `print("Hello World")`.
* Now instead of just printing `unknown` when we don't have type metadata for a variant or record we print the data inside and leave the fields or variant itself as unknown. 

This work is based on: #2378

Closes: #2087
Work towards: #1794, #2127


stdlib/runtime/toString.gr
```diff
@@ -0,0 +1,956 @@
...
+  /**
+   * Converts an Int64 value to its string representation.
+   *
+   * @param val: The Int64 value to convert
+   *
+   * @returns A document layout representing the stringified Int64 value
+   */
+  @unsafe
+  provide let int64 = (val: Int64) => {
+    let num = Conv.fromInt64(magic(val): Int64)
```
Kara-Zor-El: magic is unnecessary

stdlib/runtime/toString.gr
```diff
@@ -0,0 +1,956 @@
...
+  /**
+   * Converts a Grain value to its string representation.
+   *
+   * @param value: The grain value to convert
+   * @param state: The current stringification state
+   *
+   * @returns A document layout representing the stringified value
+   */
+  and value = (value: a, state) => {
+    let ref = WasmRef.fromGrain(value)
```
Kara-Zor-El, line 922: most of the helper functions convert ref inside them and are only used here. will make note of each but seems like you're converting to `WasmRef` only to do so again in most of the helpers

stdlib/runtime/toString.gr
```diff
@@ -0,0 +1,956 @@
...
+    if (Helpers.isStackValue(ref)) {
```
Kara-Zor-El, line 923: passing in `WasmRef` when function already does `WasmRef.fromGrain`

stdlib/runtime/toString.gr
```diff
@@ -0,0 +1,956 @@
...
+      if (Helpers.isSimpleNumberValue(value)) {
+        number(magic(value): Number)
+      } else if (Helpers.isShortValue(ref)) {
```
Kara-Zor-El, line 926: same as above

stdlib/runtime/toString.gr
```diff
@@ -0,0 +1,956 @@
...
+        shortValueToString(ref, state)
+      } else if (Helpers.isConstantValue(ref)) {
```
Kara-Zor-El, line 928: same as above

stdlib/runtime/toString.gr
```diff
@@ -0,0 +1,956 @@
...
+        constantValueToString(ref)
+      } else {
+        Doc.Builder.asciiString("<unknown value>")
+      }
+    } else if (Helpers.isHeapValue(ref)) {
+      heapValueToString(ref, state)
```
Kara-Zor-El, line 934: same as above

stdlib/runtime/toString.gr
```diff
@@ -0,0 +1,956 @@
...
+  provide let isStackValue = val => {
+    let ref = WasmRef.fromGrain(val)
+    WasmRef.isRefI31(ref)
+  }
+
+  /**
+ * Checks if the given grain value is a simple number.
+ *
+ * NOTE:
+ * This throws a wasm trap if the value is a not a (ref i31),
```
Kara-Zor-El:
```suggestion
 * This throws a wasm trap if the value is not a (ref i31),
```

stdlib/runtime/toString.gr
```diff
@@ -0,0 +1,956 @@
...
+    let ref = WasmRef.fromGrain(val)
+    (WasmRef.i31GetU(ref) & Tags._GRAIN_NUMBER_TAG_MASK)
+      == Tags._GRAIN_NUMBER_TAG_TYPE
+  }
+
+  /**
+ * Checks if the given grain value is a constant value.
+ *
+ * NOTE:
+ * This throws a wasm trap if the value is a not a (ref i31),
```
Kara-Zor-El:
```suggestion
 * This throws a wasm trap if the value is not a (ref i31),
```

stdlib/runtime/toString.gr
```diff
@@ -0,0 +1,956 @@
...
+    let ref = WasmRef.fromGrain(val)
+    (WasmRef.i31GetU(ref) & Tags._GRAIN_GENERIC_TAG_MASK)
+      == Tags._GRAIN_CONST_TAG_TYPE
+  }
+
+  /**
+ * Checks if the given grain value is a short value.
+ *
+ * NOTE:
+ * This throws a wasm trap if the value is a not a (ref i31),
```
Kara-Zor-El:
```suggestion
 * This throws a wasm trap if the value is not a (ref i31),
```

stdlib/runtime/vector.gr
Kara-Zor-El, line 1: nit: personally i prefer the name vector be used when referring to math. ultimately the name doesn't matter too much but i think something else could be a bit better

stdlib/runtime/vector.gr
```diff
@@ -0,0 +1,98 @@
...
+ * @param vector: The Vector to append to
+ * @param value: The value to append to the Vector
+ */
+@unsafe
+provide let push = (vector: Vector<a>, value: a) => {
+  use WasmI32.{ (+), (<<), (==) }
+  let size = untagSimpleNumber(vector.size)
+  let capacity = WasmArrayRef.length(vector.data)
+  if (size == capacity) {
+    let capacity = capacity << 2n
```
Kara-Zor-El, line 53: this grows it by 4x. was there a reason to do so vs 2x?

stdlib/runtime/vector.gr
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * This function does not do a bounds check, it is the callers responsibility
+ * to ensure the vector is not empty before calling this function.
+ *
+ * @param vector: The Vector to pop from
+ *
+ * @returns The value that was removed from the vector
+ */
+@unsafe
+provide let pop = (vector: Vector<a>) => {
```
Kara-Zor-El, line 73: correct me if wrong but in theory we could have a empty vector popped creating a negative tagSimpleNumber

stdlib/runtime/vector.gr
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * @param vector: The Vector to inspect
+ *
+ * @returns The length of the Vector
+ */
+@unsafe
+provide let length = (vector: Vector<a>) => untagSimpleNumber(vector.size)
+
+/**
+ * Appends a value to the end of the Vec, resizing if necessary.
```
Kara-Zor-El:
```suggestion
 * Appends a value to the end of the Vector, resizing if necessary.
```

stdlib/runtime/miniBuffer.gr
```diff
@@ -0,0 +1,90 @@
...
+@unsafe
+let autogrow = (buffer: MiniBuffer, addedSize: WasmI32) => {
+  use WasmI32.{ (+), (*), (>=), (<) }
+  let bufferRef = WasmRef.fromGrain(buffer.data)
+  let bufferData = getBytesArrayRef(bufferRef)
+  let mut bufferSize = getBytesSize(bufferRef)
+  let bufferPosition = untagSimpleNumber(buffer.pos)
+  let desiredSize = bufferPosition + addedSize
+  if (bufferSize >= desiredSize) return
+  while (bufferSize < desiredSize) bufferSize *= 2n
```
Kara-Zor-El, line 46: if initial buffer size is 0n, infinite loop when trying to grow

stdlib/runtime/doc.gr
```diff
@@ -0,0 +1,791 @@
...
+      breakingWidth: breakingWidth(breaking),
+    }
+  /**
+   * Constructs a node that introduces indentation equal to the number of spaces
+   * specified when the enclosing group is broken. When newline characters are emitted,
+   * they are followed by space characters equal to the amount of indentation that has
+   * been applied by all groups, unless this would lead to trailing whitespace.
+   * Note that if the enclosing group has not been broken, the
+   * indentation will not apply. For example, in this document,
+   * ```grain
```
Kara-Zor-El, line 382: make sure to close this codeblock

Kara-Zor-El (approved):
Looks great. some small nits

stdlib/runtime/toString.gr
spotandjake: This is kind of by design and done for type safety. Currently, we only expose `Stringify.value` but the idea is in the future we might offer `Stringify.int64` to provide a more narrow `Int64.toString`.

Converting is a `no-op` as all it does is change the type, so it's not a performance issue. We need to convert them to get the tag information.

stdlib/runtime/toString.gr
spotandjake: This is correct behaviour `value` isn't always a `wasmRef` its often a grain value itself, but we don't know the type.

stdlib/runtime/vector.gr
spotandjake: I think `vector` is fine in the runtime given thats normally the low level name, this may be something we revisit when #2374 gets done.

stdlib/runtime/vector.gr
spotandjake: Through testing, I found that there was negligible performance or space loss from this on smaller and medium test cases, but it greatly helped with the performance of massive cases.

stdlib/runtime/vector.gr
spotandjake: we absolutely could but this isn't a user library its a runtime library, and doing a bounds check here would incurr quite a cost given its in the hot path of printing.

The `vector` library is a small no checks runtime only library that must be used with extreme care.

compiler/test/runtime/toString.test.gr
```diff
@@ -0,0 +1,412 @@
...
+      value2: 2,
+      value3: 3,
+      value4: 4,
+      value5: 5,
+      value6: 6,
+      value7: 7,
+      value8: 8,
+    }
+  )
+    == "{ \n  value: 1,\n  value2: 2,\n  value3: 3,\n  value4: 4,\n  value5: 5,\n  value6: 6,\n  value7: 7,\n  value8: 8, \n}"
```
ospencer: Looks like you've got trailing whitespace here. Should be fixable with a `breakableSpace`.

ospencer (commented):
Overall this looks like it's going to be a big improvement! Direction looks good. Noting a couple of small things I found.

I mentioned this on the community call, but it looks like arrays lost their space in the beginning (`[> 1, 2, 3]`). That's how the formatter does it.

I think you've got a bug in how `FitFlat` and `FitBreaking` work versus the original Doc engine, but I haven't verified. Worth adding tests since `toString` doesn't use those.

spotandjake:
Fixed all those concerns and added a test for the different engine behaviours.

PR #2389, opened 2026-04-30
feat: Cleanup to shrink smallest program size

spotandjake:
I wasn't sure if this was a `feat` or `chore`. I was looking through the binary of our smallest program, and noticed that a few uneccessairy functions were being left behind. This is a rather small pr that corrects that.

Notably, `bigIntToString` was using `_SIZES`, which created a closure in every program. I did some additional optimizations while looking at the function.
* Remove the unused `_SIZES` array and closure dependency.
* Pull out `getDigit`, which removes a secondary closure and allows us to lazily initialize `_DIGITS` saving some bytes.
* Track length with `result` so we don't need to iterate over the result buffer again. (This also removed a helper function.

The second change was much smaller; we were creating a function in pervasives called `setupExceptions`. This was being done because previously the exception handling interacted with some unsafe code; however, now it's completely safe, so we can drop the function wrapper, which saves a few bytes.

Just as a note from what I could tell in the binary, our smallest program is pretty well optimized, with most of the remaining code being related to the exception handling setup we do in pervasives. I don't really see much room left to optimize this away, however, beyond determining if exceptions can be thrown in the program and eliminating them at the compiler level.

This work is based on: #2378

### DIFF
--- a/compiler/test/suites/basic_functionality.re
+++ b/compiler/test/suites/basic_functionality.re
@@ -485,6 +485,6 @@ describe("basic functionality", ({test, testSkip}) => {
     ~config_fn=smallestFileConfig,
     "smallest_grain_program",
     "",
-    329,
+    284,
   );
 });
--- a/stdlib/pervasives.gr
+++ b/stdlib/pervasives.gr
@@ -230,21 +230,17 @@ provide primitive unbox = "@unbox"
 
 // Setup exception printing
 primitive elideTypeInfo = "@meta.elide_type_info"
-@unsafe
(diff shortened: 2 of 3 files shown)


ospencer (commented):
It seems like we should have a new test for that small bug fix in the bigint printing.

spotandjake:
> It seems like we should have a new test for that small bug fix in the bigint printing.

There wasn't really a bug just closures causing the code to get included.

ospencer (approved):

PR #2392, opened 2026-05-02
fix(lsp): make spec complient JSON-RPC response and properly implement definition link suport

Kara-Zor-El:
Howdy,

I'm working on making a Neovim plugin for the LSP and I got some errors, when looking into them I discovered there was some issues with the LSP:
- definitionProvider must be a boolean (or [DefinitionOptions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#definitionOptions)) see here: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#serverCapabilities
- linkSupport was moved into [definition.clientCapabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#definitionClientCapabilities) and [typeDefinition.clientCapabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#typeDefinitionClientCapabilities)
- locationLink was previously assumed always an option but some clients only support location
- json-rpc result sucess messages should've not included the null error field (technically doesn't explicitly say in the spec field but specified in json-rpc (https://www.jsonrpc.org/specification#response_object) and breaks neovim)
- json-rpc error messages should've not included the null result field (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#responseMessage)

sorry for how link dense the links are 🥲

### DIFF
--- a/compiler/src/language_server/goto.re
+++ b/compiler/src/language_server/goto.re
@@ -45,6 +45,17 @@ let send_location_link =
   );
 };
 
+let send_location =
+    (~id: Protocol.message_id, ~uri: Protocol.uri, ~range: Protocol.range) => {
+  Protocol.response(
+    ~id,
+    Protocol.location_to_yojson({
+      uri,
+      range,
+    }),
+  );
+};
+
 type check_position =
   | Forward
   | Backward;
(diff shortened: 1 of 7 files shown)


compiler/src/language_server/goto.re
```diff
@@ -148,15 +159,28 @@ let process =
...
     switch (result) {
     | None => send_no_result(~id)
     | Some((origin_loc, target_loc, target_uri)) =>
-      send_location_link(
-        ~id,
-        ~origin_range=Utils.loc_to_range(origin_loc),
-        ~target_uri,
-        ~target_range=Utils.loc_to_range(target_loc),
-      )
+      if (use_link) {
```
spotandjake: Nit: If this is a support thing think this logic would just make sense to implement in send_location_link itself, so it works everywhere?

compiler/src/language_server/initialize.re
```diff
@@ -22,9 +47,27 @@ module RequestParams = {
...
 
+let take_definition_link_support = (params: RequestParams.t) =>
+  switch (params.capabilities) {
+  | None => false
+  | Some({text_document: None}) => false
+  | Some({text_document: Some({definition: None})}) => false
+  | Some({text_document: Some({definition: Some({link_support})})}) => link_support
+  };
+
+let take_type_definition_link_support = (params: RequestParams.t) =>
```
spotandjake: Nit: Would it make sense to pull these out as a tuple in a single function seems like your always going to be looking for both?

spotandjake (commented):
Looks great to me just two nits

spotandjake (approved):
This looks good to me.

ospencer (approved):
Thank you!

PR #2395, opened 2026-07-10
fix(stdlib): Correct path bounds checks in `wasi/File` module

Kara-Zor-El:
It was noticed that `Fs.exists` was throwing an uncaught error and while looking into that error it was also noticed that doing `--dir=.` instead of `--dir=./` caused another similar uncaught error.

The problem relating to `Fs.exists` stemmed from a couple of out of bounds reads that were solved by checking if we are past the boundary of the string before rather than after, this bug existed before the gc work however as we are now using gc arrays which are bounds checked the error surfaced.

The problem with the preopens stemmed from `stripPrefix` we were mistaking checking for `./` before `.` which led to an overread similar to before this error never surfaced before the gc work as linear memory isn't bounds checked.

### DIFF
--- a/stdlib/wasi/file.gr
+++ b/stdlib/wasi/file.gr
@@ -1913,16 +1913,16 @@ let stripPrefixes = (path: String) => {
     if (WasmArrayRef.getI8U(origPathData, ptr) == _CHAR_SLASH) {
       ptr += 1n
       len -= 1n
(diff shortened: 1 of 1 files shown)


spotandjake (approved):

ospencer:
Are there any tests we can add here?

spotandjake:
> Are there any tests we can add here?

I'm pretty sure this is covered by our issue to enable `fs.test.gr`

PR #2396, opened 2026-07-10
feat(lsp): Add `textDocument/didClose` to lsp

Kara-Zor-El:
Added `textDocument/didClose` functionality to the LSP. ~While not a huge problem on small codebases, if you are working on a large codebase with a lot of files, it can cause the LSP to eat up a lot more memory by still keeping the typedTree and sourceTree~

Edit: the ocaml runtime seems to not free the heap to the OS immediately so even if its removed memory will stay the same. Can be fixed with `Gc.compact()` (see: https://ocaml.org/manual/5.5/api/Gc.html#VALcompact) but its (probably) not worth it

### DIFF
--- a/compiler/src/language_server/code_file.re
+++ b/compiler/src/language_server/code_file.re
@@ -332,3 +332,26 @@ module DidChange = {
     };
   };
 };
+
+module DidClose = {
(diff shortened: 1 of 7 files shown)


spotandjake (approved):
Looks good to me.

PR #2397, opened 2026-07-10
chore(lsp): Remove manual line breaks from hover

Kara-Zor-El:
Currently, only VS Code (and maybe other editors I haven't tested with) support using HTML in their markdown for hovering. The `<br><br>` is currently redundant so removing it help other clients look the same

Previously:
| VS Code | Zed | Neovim |
| :---: | :---: | :---: |
| image | image | image | 

Now:
| VS Code (no change) | Zed | Neovim |
| :---: | :---: | :---: |
| image | image | image |

### DIFF
--- a/compiler/src/language_server/hover.re
+++ b/compiler/src/language_server/hover.re
@@ -50,7 +50,7 @@ let markdown_join = (a, b) => {
   // Horizonal rules between code blocks render a little funky
   // so we manually add linebreaks
   Printf.sprintf(
-    "%s\n---\n<br><br>\n%s",
+    "%s\n---\n\n%s",
     a,
     b,
   );
--- a/compiler/test/suites/grainlsp.re
+++ b/compiler/test/suites/grainlsp.re
@@ -602,7 +602,7 @@ let a = { x: 1 }
           (
             "value",
             `String(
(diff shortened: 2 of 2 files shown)


spotandjake (approved):

spotandjake:
Do you know if the lsp spec mentions anything about the content type supported. I.e if it's markdown, html, or raw text? Not a blocker to this just curious.

Kara-Zor-El:
It supports `MarkedString`, `MarkedString[]` or `MarkupContent` (and `MarkupContent.MarkupKind` can be plaintext to markdown) https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#hoverClientCapabilities

It should be noted it does explicitly specify html may be removed by clients though (so in general it is best to avoid it)
> * *Please Note* that clients might sanitize the return markdown. A client could
> * decide to remove HTML from the markdown to avoid script execution.

 from: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markupContent

Kara-Zor-El:
Also, a client can specify what html tags they support: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markdownClientCapabilities
though, i think its more work than its worth unless doing fancy LSPstuff with svgs (normally done by extension on clients for like rending images based on paths or urls)

PR #2398, opened 2026-07-11
feat(lsp): Add `textDocument/completion` to LSP

Kara-Zor-El:
Hey, been working on getting completions working in the LSP. So far its going well but hopefully much more cleanup and work to come.

This pr adds context-aware `textDocument/completion` support to the LSP.

Completions are driven by Tree-sitter for syntax context (so they still work on incomplete / broken code while editing) and supplements it with the typed environment when we have a successful compilation. Depending on where the cursor is, we can offer keywords, in-scope values/types, import paths, module names, graindoc attributes, etc. 

Huge credit to @spotandjake for helping out and answering questions throughout as well as his immense help with the tree-sitter work. Work here is in part based on his earlier work in https://github.com/grain-lang/grain/pull/1900

Currently I am using: 
- https://github.com/Kara-Zor-El/tree-sitter-grain/tree/Kara-Zor-El/completions-testing for the tree-sitter impl
- https://github.com/Kara-Zor-El/grain-tree-sitter.ml/ for the bindings

Currently tree-sitter-grain is licensed as MIT. I think currently Jake and Oscar mentioned we might be able to change the license for the parser code to MIT. Will wait for confirmation but can change to LGPL-3 for the time being.

I do still have a couples areas I am trying to smoothen out and squash a few bugs so will leave this as a draft for now and would love some feedback even if not in a complete state


compiler/grainlsp/dune
Kara-Zor-El, line 1: Going through and adding some clarifying comments throughout. Had to expand this separately since the implementation for native vs js backends for grain-tree-sitter.ml use different underlying libraries for each target (ocaml tree_sitter bindings vs wasm tree-sitter-web bindings). Not sure if there's a better way to do it here or in the layer I wrote but open to feedback

compiler/src/parsing/keywords_gen.re
Kara-Zor-El, line 1: you're going to notice that currently the keyword generation here is not actively being used. Originally was using it and then was working on context specific completions and will switch back to integrating this into the current way of doing things soon

compiler/test/suites/grainlsp_completions.re
Kara-Zor-El, line 1: @spotandjake mentioned wanted to have a discussion about how is best to test the lsp and completions going forwards. Obviously completions need to be thoroughly tested but the current manual positioning method could probably use some work. Will leave it up to Jake to explain more but my current thoughts are to have a helper function and a position and/or range character
for example
```grain
let x: §
```
where the section sign (`§`) specifies where the cursor is. And for ranges to use `«` and `»` for ranges

compiler/dune-project
```diff
@@ -1,4 +1,4 @@
-(lang dune 3.0)
+(lang dune 3.7)
```
Kara-Zor-El, line 1: This had to be changed to allow for `(include_subdirs qualified)` (https://dune.readthedocs.io/en/stable/reference/dune/include_subdirs.html#qualified) to make it part of a module hierarchy. This was mainly done for organizational reasons. In my testing I haven't observed anything breaking due to this change. Happy to discuss further

compiler/esy.json
```diff
@@ -33,6 +34,7 @@
   },
   "dependencies": {
     "@grain/binaryen.ml": ">= 0.34.0 < 0.35.0",
+    "@grain/grain-tree-sitter.ml": "*",
```
Kara-Zor-El, line 36: the intent here would be to ideally have grain-tree-sitter.ml and tree-sitter-grain (which I should probably change the name of one of them, lol) to become part of the grain project (not the repo but rather than org and under the grain umbrella)

compiler/esy.json
```diff
@@ -17,13 +17,14 @@
       "GRAINLSP_BIN_PATH": "#{self.target_dir / 'default' / 'grainlsp' / 'grainlsp.exe'}",
       "GRAINLSP_JS_PATH": "#{self.target_dir / 'default' / 'grainlsp' / 'grainlsp.bc.js'}",
       "CLI_BIN_DIR": "#{self.root / '..' / 'cli' / 'bin'}",
+      "TREE_SITTER_GRAIN_WASM_PATH": "#{self.root / '..' / '..' / 'grain-tree-sitter.ml' / 'vendor' / 'tree-sitter-grain' / 'tree-sitter-grain.wasm'}",
```
Kara-Zor-El: For the JS build we end up using tree-sitter-web which has wasm bindings, hence the necessary inclusion of the wasm file here

compiler/src/parsing/parser.mly
```diff
@@ -19,27 +19,27 @@ module Grain_parsing = struct end
...
 %token <string> INFIX_ASSIGNMENT_10
 
-%token ENUM RECORD TYPE MODULE INCLUDE USE PROVIDE ABSTRACT FOREIGN WASM PRIMITIVE
-%token AND
-%token EXCEPT FROM STAR
+%token ENUM [@keyword] RECORD [@keyword] TYPE [@keyword] MODULE [@keyword] INCLUDE [@keyword] USE [@keyword] PROVIDE [@keyword] ABSTRACT [@keyword] FOREIGN [@keyword] WASM [@keyword] PRIMITIVE [@keyword]
+%token AND [@keyword]
+%token EXCEPT [@keyword] FROM [@keyword] STAR
 %token SLASH DASH PIPE
 %token EOL EOF
```
Kara-Zor-El, line 45: on L46 there's the reserved keywords not used yet. Whenever these are added, they'd need a attribute applied as well

cli/package.json
```diff
@@ -14,7 +14,10 @@
     "build-pkg": "pkg ./package.json --no-bytecode --compress Brotli --options=experimental-wasi-unstable-preview1,no-warnings --output ../pkg/grain"
   },
   "pkg": {
-    "assets": "bin/*.js"
+    "assets": [
+      "bin/*.js",
+      "bin/*.wasm"
```
Kara-Zor-El, line 19: wasm file needed for js side of the completions

Kara-Zor-El:
A couple of other things I want to bring up related to tree-sitter. I think tree-sitter gives us great opportunity elsewhere within the language server.

1. [`textDocument/hover`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_hover) (which already exists) and [`completionItem/resolve`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem_resolve) can have docblock support added to them. @spotandjake  mentioned it might make sense (once we refine docblocks in the tree-sitter impl) to potentially rewrite the docblock parsing to take advantage of tree-sitter potentially when doing this and to have a better way of caching docblocks (maybe as json)

2. Another one of Jake's ideas which I personally think is great is using tree-sitter to replace our current formatting code. [Topiary](https://topiary.tweag.io/) currently is an example of this being done. It would make it much easier for [`textDocument/rangeFormatting`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_rangeFormatting) to be supported. I think something part of the CLI instead of an external tool would be the way to go here but it does seem like it could potentially be helpful here

3. Potentially can be used as a fallback for goto when dirty. Also rename and references

compiler/src/language_server/completions/sources/keywords.re
```diff
@@ -0,0 +1,362 @@
...
+         switch (entry.snippet) {
+         | Some(_) => Some(SnippetFormat)
+         | None => None
+         };
+       let command =
+         switch (entry.snippet) {
+         | Some(_) =>
+           let command: Protocol.command = {
+             title: "",
+             command: "editor.action.triggerSuggest",
```
Kara-Zor-El: when doing snippets, unfortunately we can't force the LSP to automatically suggest options. This is the closest I was able to figure out but unfortunately `editor.action.triggerSuggest` is vscode (and derivative) specific.

compiler/src/language_server/initialize.re
```diff
@@ -109,9 +113,14 @@ module ResponseResult = {
   type t = {capabilities: lsp_capabilities};
 
   let capabilities = {
+    position_encoding: "utf-8",
```
Kara-Zor-El: **NOTE:** This is used because grain assume UTF-8. It should be noted that the LSP protocol requires UTF-16 support so this is not [spec compliant](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#positionEncodingKind).

Issue #2399, opened 2026-08-15
Compiler: `Not_found` compiler exception raised from nested `let` inside `let mut`


Kara-Zor-El:
Apologies for the bad explanation. Was doing some testing for the completions work I have been doing and stumbled across this weird case where having the `let y = 1` inside the mut binding, for some reason the compiler crashes

Code will probably explain it better:

```grain
let mut x = if (true) {
  let y = 1
  y
} else {
  2
}
x = 5
print(x)
```

produces the following (tested on both main and gc):
```
/Users/karawilson/Dev/grain/grain/cli/bin/grainc.exe: internal error, uncaught exception:
Not_found
Raised at Grain_typed__Ident.find_same in file "src/typed/ident.re", line 303, characters 13-29
Called from Grain_codegen__Transl_anf.find_id in file "src/codegen/transl_anf.re", line 161, characters 16-50
Called from Grain_codegen__Transl_anf.compile_comp in file "src/codegen/transl_anf.re", line 483, characters 8-26
Called from Grain_codegen__Transl_anf.compile_anf_expr.(fun) in file "src/codegen/transl_anf.re", line 695, characters 42-69
Called from Grain_codegen__Transl_anf.compile_anf_expr in file "src/codegen/transl_anf.re", lines 691-703, characters 8-9
Called from Grain_codegen__Transl_anf.compile_function_body in file "src/codegen/transl_anf.re", line 726, characters 13-40
Called from Grain_codegen__Transl_anf.transl_anf_program in file "src/codegen/transl_anf.re", line 1143, characters 4-45
Called from Grain__Compile.next_state in file "src/compile.re", line 149, characters 19-59
Called from Grain__Compile.compile_resume in file "src/compile.re", line 172, characters 19-32
Called from Grain_utils__Config.preserve_config in file "src/utils/config.re", line 305, characters 12-19
Re-raised at Grain_utils__Config.preserve_config in file "src/utils/config.re", line 311, characters 4-15
Called from Grain_utils__Config.preserve_root_config in file "src/utils/config.re", line 273, characters 12-19
Re-raised at Grain_utils__Config.preserve_root_config in file "src/utils/config.re", line 281, characters 4-15
Called from Dune__exe__Grainc.compile_file in file "grainc/grainc.re", line 58, characters 9-49
Called from Dune__exe__Grainc.error_wrapped in file "grainc/grainc.re", line 31, characters 5-10
[...]

PR #2400, opened 2026-08-16
fix(compiler): Correct mutation bug in dead branch optimization

spotandjake:
This pr corrects a small bug related to our dead branch optimization.

Programs that contained a `let` bind inside of constant `if` statement, inside of another `let mut` bind would lose the mut_flag of the outer binding.

This pr corrects that by properly propagating the mut_flag.

Issue: #2399

Thanks to @Kara-Zor-El for the help.


compiler/src/middle_end/optimize_dead_branches.re
```diff
@@ -59,7 +58,7 @@ module BranchArg: Anf_mapper.MapArgument = {
             recursive,
             mutable_,
             binds,
-            relinearize(id, global, body, cont),
+            relinearize(id, global, body, mut_flag, cont),
           ),
```
Kara-Zor-El: Tested and it seems to fix the issue reported. However, turns out that there's another where:
```grain
let mut x = if (true) {
  let mut y = 1
  y
} else {
  2
}

let z = () => x

x = 3
print(z())
```

produces 1 instead of 3. 

This should fix be:
```patch
-    | AELet(global, recursive, mutable_, binds, body) => {
+    | AELet(inner_global, recursive, mutable_, binds, body) => {
        ...a,
        anf_desc:
          AELet(
-            global,
+            inner_global,
```

Kara-Zor-El (changes requested):

ospencer (approved):

PR #2401, opened 2026-09-03
feat: Upgrade to binaryen v0.42

### LINKED ISSUE #2358: Improve testing of `type_metadata` to check if the segment exists
In https://github.com/grain-lang/grain/pull/2352 I had to use the statement `Binaryen.Memory.get_num_segments(asm) > 0` to check if the `type_metadata` segment exists as calling `Binaryen.Memory.get_segment_data(asm, "type_metadata")` throws a fatal exception that cannot be caught if the segment doesn't exist.

Once https://github.com/grain-lang/binaryen.ml/issues/271 is figured out we should correct this as the current check could be fragile if something like #1330 we're ever to be merged.

spotandjake:
This pr upgrades our binaryen.ml version from `0.35` to `0.42`.

In binaryen.ml `0.36` we upgraded jsoo from `6.3` to `6.4` which forces us to update our cmdliner version to `2.0`. This change broke the `ppx_deriving_cmdliner` module as it hasn't been updated in 4 years.

As such I dropped the library and instead of using the macros to generate our cmdliner terms we just write them manually, it's slightly more verbose but not to bad.

This pr also updates our optimization passes to follow binaryenes default ordering until we decide to diverge.

Closes: #2358


compiler/src/codegen/optimize_mod.re
```diff
@@ -56,13 +56,23 @@ let default_global_optimization_pre_passes =
...
+                Passes.cfp;
+              },
+              Passes.gsi,
+            ],
+            if (Settings.get_closed_world()) {
+              [Passes.abstract_type_refining, Passes.unsubtyping];
+            } else {
+              [];
+            },
+          ]);
```
Kara-Zor-El: Seems like optimize_level should be >= 3 here and gsi should be outside the closed world branch

compiler/src/utils/filepath.re
```diff
@@ -182,23 +182,23 @@ module Args = {
...
+        Error(Format.sprintf("Invalid path: %s", fname))
       };
     };
 
     let prntr = (formatter, value) => {
       Format.fprintf(formatter, "File: %s", to_string(value));
     };
 
-    let cmdliner_converter = (prsr, prntr);
+    let cmdliner_converter = Cmdliner.Arg.Conv.make(~parser=prsr, ~pp=prntr);
```
Kara-Zor-El: doesn't this need a `~docv` and `()`?

Kara-Zor-El (changes requested):

compiler/src/codegen/optimize_mod.re
spotandjake: Good catch

Kara-Zor-El (approved):

ospencer (approved):

PR #2403, opened 2026-09-04
feat: Allow `@elideTypeInfo` as an attribute

### LINKED ISSUE #2365: Allow eliding type info of an individual type or module
Currently we ship all type info no matter where it exists in a program, it would be nice if we could use an attribute like `@elideTypeInfo` on a specific type or entire module to tell the compiler that we don't care about this types info. This would is useful in both the stdlib and runtime where we are currently shipping type info that is never exposed to the user bloating their program with no real benefit.

I don't believe that this would acutally be that hard todo we would just need to add support for the property on the places we want it, carry the property on the type itself to the cmi and then omit the metadata when we compile.

spotandjake:
This PR adds a new `@elideTypeInfo` attribute. It is similar to eliding type information at the CLI level, but is intended more for use within the runtime or library internals where we only want to elide type information from specific parts of a program.

The attribute can be applied in the following places:

**Above a record definition**:
* Elides the type information for that record specifically.

**Above an enum definition**
* Elides the type information for that enum specifically.

**Above a submodule definition**:
* Elides the type information for everything within the submodule.

**Above a module definition**:
* This might initially seem equivalent to eliding type information for the entire program, but it only affects the module itself.
* For example, if you have a Main module with type information elided, you can still print records from other modules where type information has not been elided.

Currently the `@elideTypeInfo` flag can't be applied to exceptions and exceptiosn will not be elided if the parent module is. One reason for this is that we do not allow the attribute on exceptions in the parser and second reason is that because exceptions are a `gadt` eliding one won't remove the actual core type itself which breaks our type metadata section with the way i've choosen to implement things. This is something that we can address in the future.

While working on #2385, I noticed that a significant portion of the added program size came from the use of higher-level types such as records and enums. Some of that additional size comes from the type information associated with those types.

The intention of introducing this attribute is to give us a more selective way to strip type information from specific modules, submodules, records or enums. This is particularly useful for runtime and internal library code where we know that the type information will never be needed, and want to reduce the size of the resulting program or in the case of the runtime want to ship slim libraries.

Closes: #2365


compiler/src/parsing/well_formedness.re
```diff
@@ -338,6 +342,22 @@ let disallowed_attributes = (errs, super) => {
...
+      )
+    ) {
+    | Some({Asttypes.attr_name: {txt, loc}}) =>
+      errs :=
+        [
+          AttributeDisallowed(
+            "`elideTypeInfo` is only allowed on module, record, and variant declarations.",
+            loc,
+          ),
+        ]
```
Kara-Zor-El, line 358: is there a reason these aren't prepended to the error list and instead replaces it?

compiler/src/parsing/well_formedness.re
```diff
@@ -395,6 +415,27 @@ let disallowed_attributes = (errs, super) => {
...
+      | PTopData(_) => ()
+      | _ =>
+        errs :=
+          [
+            AttributeDisallowed(
+              "`elideTypeInfo` is only allowed on module, record, and variant declarations.",
+              loc,
+            ),
+          ]
+      }
```
Kara-Zor-El, line 436: same as above

compiler/test/suites/basic_functionality.re
Kara-Zor-El, line 1: you should probably add tests for elideTypeInfo at a file level and maybe testing calling toString on a type in one file from another

stdlib/path.gr
```diff
@@ -53,27 +53,28 @@ from "char" include Char
...
-
+@elideTypeInfo
 enum FileType {
   File,
   Dir,
 }
 
 // hack to be able to concretely distinguish TypedPath from PathInfo and
 // enforce TypedPath's type parameters
+@elideTypeInfo
```
Kara-Zor-El, line 77: these two cause toString/print to change
```grain
module Main

from "path" include Path

let p = Path.fromString("/file.txt")
print(p)
```
goes from 
```
AbsoluteFile(({
  base: Abs(Root)
}, {
  fileType: File
}, ["file.txt"]))
```
to
```
AbsoluteFile(({
  base: Abs(Root)
}, <record value>, ["file.txt"]))
```

Kara-Zor-El (dismissed):
took a quick look tonight. Also, side note but in general i think it might make more sense to change elide to omit in this and in general. I do think its more common in everyday language and would increase clarity.

compiler/src/parsing/well_formedness.re
spotandjake: We already had `AttributeDisallowed` so I just used it. I think if we want to talk about changing that then we should do it in a seperate pr. (maybe one cleaning up the entire error system).

stdlib/path.gr
Kara-Zor-El: it still resolves to:
```grain
AbsoluteFile(({
  base: Abs(Root)
}, {
  fileType: <enum value>
}, ["file.txt"]))
```
instead of
```grain
AbsoluteFile(({
  base: Abs(Root)
}, {
  fileType: File
}, ["file.txt"]))
```

compiler/src/parsing/well_formedness.re
```diff
@@ -338,6 +342,22 @@ let disallowed_attributes = (errs, super) => {
...
     | None => ()
     };
+    switch (
+      List.find_opt(
+        ({Asttypes.attr_name: {txt}}) => txt == "elideTypeInfo",
+        attrs,
+      )
+    ) {
+    | Some({Asttypes.attr_name: {txt, loc}}) =>
+      errs :=
```
ospencer, line 353: This should be pushing onto errs, not replacing it. If we've got other places in the file doing this, we can fix them in this PR.

stdlib/path.gr
```diff
@@ -53,20 +53,19 @@ from "char" include Char
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
-
+@elideTypeInfo
```
ospencer, line 57: Looks like the spacing got messed up.

compiler/src/parsing/well_formedness.re
```diff
@@ -338,6 +342,22 @@ let disallowed_attributes = (errs, super) => {
...
+        [
+          AttributeDisallowed(
+            "`elideTypeInfo` is only allowed on module, record, and variant declarations.",
```
ospencer: We should call them enums, not variants.

compiler/src/parsing/well_formedness.re
```diff
@@ -395,6 +415,27 @@ let disallowed_attributes = (errs, super) => {
...
+    ) {
+    | Some({Asttypes.attr_name: {txt, loc}}) =>
+      switch (desc) {
+      | PTopModule(_)
+      | PTopData(_) => ()
+      | _ =>
+        errs :=
+          [
+            AttributeDisallowed(
+              "`elideTypeInfo` is only allowed on module, record, and variant declarations.",
```
ospencer: Here too.

ospencer (commented):
Do we have a test for a module-scoped `@elideTypeInfo`?

ospencer:
@Kara-Zor-El Not sure I 100% agree with that name change, but changing it would make this a breaking change. That'd have to be separate from this (or this would have to wait to land).

spotandjake:
> @Kara-Zor-El Not sure I 100% agree with that name change, but changing it would make this a breaking change. That'd have to be separate from this (or this would have to wait to land).

Maybe `stripTypeInfo` would fit. Because we are stripping it from the file. I do agree that `elide` feels like a weird choice coming from other languages, though i think you get used to it quickly. 

Either way it seems like a discussion for a seperate pr.

spotandjake:
> Do we have a test for a module-scoped `@elideTypeInfo`?

yup in `compiler/test/input/typeMetadataElidedModule.gr` I test if we elide the type info for the main module. Making sure that we are not affecting the types in another module.

ospencer (approved):

PR #2404, opened 2026-09-10
feat(graindoc): Complete refactor

### LINKED ISSUE #2281: Graindoc: Consider warning or adding undocumented params to the param table.

spotandjake:
This PR is the polished version of an experiment I had been working on locally.

This PR refactors `graindoc` from one monolithic module into multiple smaller files.

Previously, `docblock.re` was responsible for both building the `graindoc` AST and emitting Markdown. This made it difficult to integrate `graindoc` with the language server and made it harder to modify or add new backends.

The new approach splits the module into the following components:

### Frontend

* `graindoc.docblock` — Responsible for building the `graindoc` AST.
* `graindoc.docir` — Defines the `graindoc` AST itself.
* `graindoc.errors` — Provides error handling and utilities for `graindoc`.

### Backend

* `graindoc.Markdown_backend` — Converts a `graindoc` AST into a `Grain_format` doctree.

### Decisions

* I decided to use `doctree` because it allows us to produce consistently formatted documents in real time. Compared to writing directly to a buffer, this also makes output generation simpler and more customizable.
* The `graindoc` AST is designed so that, in the future, we can support raw `sexp` output. This would allow external tools to consume the AST and generate custom outputs.

### Long-Term Vision

Shortly after this PR, I plan to follow up with a PR that integrates `graindoc` into the LSP.

Longer term, I would like to replace the logic in `Markdown_backend` with a templating library similar to EJS or Jinja. This would make it easier to support multiple backends and allow users to customize how documentation is generated.

### Performance note

While performance of this pr seems to be about the same if not slightly faster then the previous edition, I've noticed that graindoc can be a tiny bit slow when documenting the entire stdlib, while working on this refactor I looked into it a bit and it seems to be that when we `enumerate_runs` over the stdlib directory we call `compile` which calls `reset_compiler_state` this bassically means that despite compiling something like `runtime/compare` when we document `stdlib/pervasvies` we need to recompile it again because we throw out all the cached info. I did some performance testing with a soft reset that just resets the following:
```
Location.reset_exceptions();
Ident.setup();
Ctype.reset_levels();
Grain_utils.Warnings.reset_warnings();
 ``` 
and generation times speed up by a rather noticable amount, I didn't apply this change in this pr as i'm not sure it's 100% safe but we should probably open an issue to look into that and maybe make a `soft_reset_compiler_state`.

Closes: #2281 (not sure if we want to add a warning here)

### CODE REVIEW

File: stdlib/uri.md
Comment (Kara-Zor-El) [inline stdlib/uri.md:1]: we should probably also add a description to the `encodeQuery` parameter list

Comment (spotandjake) [reply stdlib/uri.md:1]: That should be handled in a seperate pr

File: compiler/graindoc/errors.re [hunk 1]
```diff
@@ -0,0 +1,170 @@
...
+      param_name: string,
+    })
+  | /** Indicates that an attribute that should appear on a function is not on a function. */
+    AttributeAppearsOnNonFunction({
+      attr: string,
+    })
+  | /** Indicates that an attribute that can only appear once appears multiple times. */
+    AttributeAppearsMultipleTimes({
+      attr: string,
+    })
```
Kara-Zor-El: Currently `ParameterAttributeAppearsMultipleTimes` is unused and instead `AttributeAppearsMultipleTimes` is raised

compiler/graindoc/errors.re
```diff
@@ -0,0 +1,170 @@
...
+let invalid_attribute = (~name, attribute) => {
+  let {attr, attr_loc}: Comment_attributes.t = attribute;
+  Error(
+    attr_loc,
+    InvalidAttribute({
+      name,
+      attr: get_attr_name(attr),
+    }),
+  );
+};
```
Kara-Zor-El: this currently is left unused

compiler/graindoc/docblock.re
```diff
@@ -1,1206 +1,533 @@
...
+let saved_comments = Hashtbl.create(64);
 
-let output_for_since = (~current_version, {since_version, since_loc}) => {
-  let current_version =
-    switch (current_version) {
-    | Some(version) => version
+/**
+ * Retrieves the docblock comment associated with the given location.
+ *
+ * @param including_attributes Weather to scan past grain attributes to find the comment. (should be false inside of type content)
```
Kara-Zor-El: sorry to nit but whether

compiler/graindoc/docblock.re
```diff
@@ -1,1206 +1,533 @@
...
-    List.fold_left(
-      (
-        {provided_types, provided_values, provided_modules},
-        sig_item: Types.signature_item,
-      ) => {
+/**
+ * Generates a docir from a module signature.
+ *
+ * @param namespace The root namespace of the module
+ * @param ident The identifier of the module
```
Kara-Zor-El: should be name instead of ident

compiler/graindoc/docblock.rei
```diff
@@ -0,0 +1,61 @@
...
+ */
+let from_value_description:
+  (~namespace: option(string), ~ident: Ident.t, Types.value_description) =>
+  value_info;
+
+/**
+ * Generates a docir from a module signature.
+ *
+ * @param namespace The root namespace of the module
+ * @param ident The identifier of the module
```
Kara-Zor-El: should be name instead of ident

compiler/graindoc/markdown_backend.re
```diff
@@ -0,0 +1,598 @@
...
+    ++ hardline;
+  };
+
+  /**
+   * Generates a heading for a markdown document.
+   * @param level The level of the heading (1-6)
+   * @param doc The content of the heading
+   * @return A doctree containing the heading
+   */
+  let heading = (~level=1, doc) =>
```
Kara-Zor-El, line 41: heading alongside paragraph, alongside paragraph, code_block, bold, and frontmatter seem to be very similar to their Grain_utils.Markdown counterparts. Is there any reason why this is also being done but only (from a quick glance) just stylistically different here?

spotandjake: Bassically because I switched from our output being completely text based to being doctree based I wrote some markdown counterparts that are doctree based, we could use the `Markdown` string based ones but they don't compose as neatly with the doctree.

Issue #1674, opened 2023-02-17
support Mac M1 cpu


xunfeng1980:

spotandjake:
Grain does support m1

phated:
When building Grain from source, everything works great on aarch64 Mac computers. We also supply aarch64 Linux docker images: https://github.com/grain-lang/grain/pkgs/container/grain/71393126?tag=main

No CI service provides aarch64 Mac runners and the `pkg` tool doesn't allow x86 -> aarch64 cross-compiles, so we don't have the ability to build binaries for it.

I'll change your title to "Provide aarch64 binaries" so it is accurately tracking what *doesn't* currently exist.

adsouza:
Would it be possible to have the Homebrew cask build it from source on ARM instead of defaulting to just downloading the prebuilt x64 binary?

phated:
@adsouza no, the build tools aren't easily available via brew.

Phault:
@phated Things have changed and the `macos-14`, `macos-14-xlarge`, `macos-13-xlarge` and `macos-latest-xlarge` runners are now available for arm64. The first one is available for everyone regardless of any GitHub plan.

mortax:
@phated @Phault where does this stand? are `arm64` binaries available somewhere? yesterday's release of macOS Golden Gate marks the final version that will support Rosetta and i'd **really** prefer not to install it at all, which means i can't run Grain unless i build it myself.

PR #2409, opened 2026-09-16
feat(cli)!: Switch to node-sea and move to node 24

spotandjake:
This pr switches our compiler from `yao-pkg/pkg`s classic build mode to their `node-sea` based mode.

Benefits:
* The security concern of malicious binaries has been negated as we are now using the offical node binaries
* Upgrading node versions is now a simple process that doesn't require waiting on pkg-fetch builds

Currently this pr is using my fork of `pkg` as I needed to expose the internal vfs in order to implement our write hack for the stdlib. I am going to try to upstream this or we can move to using grain's pkg fork.

Breaking:
I'm not sure if we actually consider this pr breaking or not, because of node-sea requirements we are dropping support for node 22 completely.

Seperate note:
I noticed that the a large majority of the time spent during packaging was during `Brotli` compression specifically of our `.bc.js` artifacts. I switched our compression to `Zstd` which speeds up build times and produces a binary of bassically the same size (within 100kb). Though of note, we would proabbly be fine not compression at all, as we don't seem to be getting a ton from it.

Longer term:
* This pr works with both #2378 and #2384 
* I need todo a bit more research but I think this helps us solve #1674 through cross compilation.


cli/bin/pkg.js
```diff
@@ -2,26 +2,104 @@
...
+    const originalFunc = fs[name];
+    if (typeof originalFunc !== "function") continue;
+    fs[name] = function (arg0, ...args) {
+      if (typeof arg0 == "string" && isInStdlib(arg0)) {
+        if (isGrainFile(arg0)) {
+          const result = vfs[name](arg0, ...args);
+          if (name === "openSync" || name === "open")
+            virtualFileDescriptors.add(result);
+          return result;
+        } else {
```
spotandjake, line 86: This part isn't needed once #2384 gets merged but it wouldn't hurt to keep it.

cli/package.json
```diff
@@ -38,7 +48,7 @@
     "commander": "^8.1.0"
   },
   "devDependencies": {
-    "@grain/pkg": "6.5.0",
+    "@yao-pkg/pkg": "github:spotandjake/pkg#81b9c9fa075629831b6b882db4f1adba7a8c84c5",
```
spotandjake: We can switch to `yao-pkg/pkg`if https://github.com/yao-pkg/pkg/pull/304 gets merged and the `openSync` stuff is patched.

Kara-Zor-El:
can you update the docker file as well to make it also use node 24?

PR #2410, opened 2026-09-18
feat(cli): Refactor the cli

spotandjake:
This pr refactors the entire cli making it more modern and modular.

I noticed previously that having all of the commands handled in one large file made it hard to understand what was going on in the cli this pr switches things so that each command has it's own file and the execution of the actual compiler components are now standardized.

Additionally this pr switches us to esm modules allowing for better treeshaking.

This work is based off my deno cli rewrite experiments.


cli/src/utils.js
```diff
@@ -0,0 +1,132 @@
+import { Option } from "commander";
+import { execSync } from "node:child_process";
+import { env, execPath } from "node:process";
+import { fileURLToPath } from "node:url";
+// import path from "node:path";
```
Kara-Zor-El, line 5: when you get a second, maybe just remove this commented out import

cli/src/commands/compile.js
```diff
@@ -0,0 +1,29 @@
...
+
+export const compile = async (file, opts, program) => {
+  return await runCommand("grainc", optionsToFlags(program, opts, [file]));
+};
+
+export default (cli, unprocessedArgs) => {
+  cli
+    .command("compile <file:string>")
+    .description("compile a grain program into wasm")
+    .addOption(new ForwardOption("-o <filename:string>", "output filename"))
```
Kara-Zor-El: issue with ForwardOption but will flag here, doing `grain compile filename.gr -o outputfile.wasm` will produce a file with the name `=outputfile.wasm`

.github/workflows/build-native.yml
```diff
@@ -79,7 +79,7 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: native-build-artifacts-${{ runner.os }}-${{ runner.arch }}
-          path: cli/bin/*.exe
+          path: cli/src/*.exe
```
Kara-Zor-El: shouldn't this be `cli/src/artifacts/*.exe`

cli/src/cli.js
```diff
@@ -174,60 +157,24 @@ program
...
-
-  .action(function (file, options, program) {
-    const success = exec.grainc(file, options, program);
+  .action(async (file, options, program) => {
+    const success = await compile(file, options, program);
     if (success) {
-      const outFile = options.o ?? file.replace(/\.gr$/, ".wasm");
-      exec.grainrun(unprocessedArgs, outFile, options, program);
+      const outFile = options.o ?? defaultWasmLocation(file, options);
+      run(outFile, options, unprocessedArgs);
```
Kara-Zor-El: this should probably use await

cli/src/cli.js
```diff
@@ -88,7 +65,13 @@ class GrainCommand extends commander.Command {
     cmd.forwardOption(
       "-I, --include-dirs <dirs>",
       "add additional dependency include directories",
-      list,
+      listParser,
+      [],
+    );
+    cmd.forwardOption(
```
Kara-Zor-El, line 71: library is not in grainc yet

cli/src/cli.js
```diff
@@ -97,15 +80,20 @@ class GrainCommand extends commander.Command {
       null,
       stdlibPath,
     );
+    cmd.forwardOption(
+      "--target-dir <dir>",
```
Kara-Zor-El, line 84: same as above

cli/src/cli.js
```diff
@@ -97,15 +80,20 @@ class GrainCommand extends commander.Command {
...
+      "directory where build artifacts are written",
+    );
+    cmd.forwardOption("--project-root <dir>", "project root directory");
```
Kara-Zor-El, line 87: same as above

cli/src/utils.js
```diff
@@ -0,0 +1,132 @@
...
+  //   : path.resolve("target");
+  // const profile = options.release ? "release" : "debug";
+  // const basename = path.basename(file).replace(/\.gr$/, ".wasm");
+  // return path.join(targetDir, profile, basename);
+  return file.replace(/\.gr$/, ".wasm");
+};
+
+const exec = (command, args, options) => {
+  try {
+    execSync(`${command} ${args.join(" ")}`, options);
```
Kara-Zor-El, line 84: `join(" ")` is probably not the best idea. if your path has a string in it (ex. `--stdlib /Users/karawilson/Dev/grain/my\ new\ stdlib/stdlib`, it would split on the path

cli/src/commands/compile.js
```diff
@@ -0,0 +1,29 @@
+import { ForwardOption, optionsToFlags, runCommand } from "../utils.js";
+
+export const compile = async (file, opts, program) => {
+  return await runCommand("grainc", optionsToFlags(program, opts, [file]));
+};
+
+export default (cli, unprocessedArgs) => {
+  cli
+    .command("compile <file:string>")
```
Kara-Zor-El: Whts up with the `:string`. Is this a byproduct from porting from deno?

Kara-Zor-El (changes requested):
Great work! Just a few areas that need some additional cleaning up

cli/src/cli.js
spotandjake: By the time this gets merged it will be.

cli/src/utils.js
spotandjake: That's around for 0.8 compat

cli/src/cli.js
spotandjake: Again by the time this is merged that'll probably be in.

cli/src/utils.js
spotandjake: It's the same behaviour as before.

Kara-Zor-El: Okay, maybe not in this PR since it'd be a breaking change, but at some point might be good to fix it so that it correctly handles spaces

spotandjake: > `join(" ")` is probably not the best idea. if your path has a string in it (ex. `--stdlib /Users/karawilson/Dev/grain/my\ new\ stdlib/stdlib`, it would split on the path

Keep in mind that this is `join` instead of `split`, commander will handle this properly on the js side and then cmdliner should handle this correctly on the ocaml side of thigns so there shouldn't be any problems really.

PR #2413, opened 2026-09-18
fix(lsp): correct advertisement of CodeLens resolveProvider

Kara-Zor-El:
`codeLens/resolve` is not implemented and we already return a fully populated lense with `textDocument/codeLens`


spotandjake (approved):
Good catch.

ospencer (approved):

